PR #34, opened 2016-12-29
Fix cli of `odoc compile`

dbuenzli:
Before this patch specifying `-pkg $PKG -o $P` would write to `$PKG/$P` if `$P` was relative and `$P` if it was absolute.

After this patch the -o option is always an exact specification of where the output will be written.

As before if `-o` is not specifed the odoc file is written to `$PKG/$(BASE).odoc` where `$(BASE)`
is the basename of the input file.


dbuenzli:
The patch is not as clean as I wanted this `Odoc.Fs` thing is creating more noise than it needs but basically now the API at Odoc.Compile simply takes input and output files and all the path fiddling is done at the driver level in `main.ml`.

dbuenzli:
> As before if -o is not specifed the odoc file is written to $PKG/$(BASE).odoc where $(BASE)
> is the basename of the input file.

Btw. I think this a bit non-sense. We should do it like most tool would do which is to write at the same location as the input file. If I get an ack I'll tweak the patch to make it work that way.

lpw25:
> Btw. I think this a bit non-sense. We should do it like most tool would do which is to write at the same location as the input file. If I get an ack I'll tweak the patch to make it work that way.

I agree. That makes the most sense to me.

dbuenzli:
> I agree. That makes the most sense to me.

I added a commit for this. If the rest is ok I'll let you or @trefis merge this.

PR #85, opened 2017-07-08
Add missing sub-commands

vramana:


vramana:
@trefis May be `to_xml` can be renamed to `to-xml` for the sake of consistency. What do you think?

trefis:
Thanks for the suggestion!

PR #104, opened 2018-02-10
Force <i> font style to italic

thizanne:
`<i>` tags generated by `{i foo}` weren't rendered in italic. This patch simply applies the same treatment as `<b>` tags.

I checked by manually editing `odoc.css` in another project that the issue is fixed.


aantron:
Thanks!

PR #107, opened 2018-03-18
css: fix codeblock overflow

slice:
small css fix that allows you to scroll in codeblocks, preventing text overflow

before:

image

after:

image

### DIFF
--- a/etc/odoc.css
+++ b/etc/odoc.css
@@ -111,7 +111,7 @@ pre
   padding-top: 0.3125rem; padding-bottom: 0.3125rem;
   margin-top: 1.25rem; /*  margin-bottom: 0.625rem; */
   line-height: 1.1875rem;
-  background: #F1F1F1; }
+  background: #F1F1F1; overflow: auto; }
 
 h1 tt, h1 code, h2 tt, h2 code, .h7 tt, .h7 code { font-size: 1.125rem }
 h3 tt, h3 code { font-size: 1rem }


dbuenzli:
I'm rather against this. It's better to stick to the page measure (if only to let it slightly overflow sometimes). Here are a few reasons:

1. "Modern" (sic) browsers like to challenge basic usability rules by hiding scroll bars (as your example shows). If the overflow happens on white space you may miss the fact that the code block is scrollable. 
2. IIRC they are annoying on touch based interfaces as they might prevent further vertical scrolling if your scrolling swipes hit such a region, which can easily happen on a whole screenful if your code fences are large enough.
3. Horizontal scrolling is in general a bit painful.

In your particular case note that you are wasting quite a bit of space on the left. Also you could certainly other ways to present the data (not necessarily in a code fence) so that it fits the constraints.

slice:
The screenshot directly references [a portion of Base.](https://ocaml.janestreet.com/ocaml-core/latest/doc/base/Base/Float/#val-round)

Additionally, the scrollbar hiding behavior is heavily OS/platform dependent. For example, if I have macOS always show scrollbars:

image

Then the problem goes away:

image

On platforms such as Linux, scrollbars are shown by default and therefore aren't a problem:

image

Also, well-behaved touch systems do not prevent vertical scrolling while the cursor is inside of the scrollable region:

kapture 2018-03-18 at 12 03 23

I have found this behavior to be identical under Linux, too.

dbuenzli:
In general it's better if people stick to 80 columns in verbatim and code fences (see the [guidelines](https://ocaml.org/learn/tutorials/guidelines.html#Width-of-the-page)). Currently the stylesheet only accomodates for 73 columns, I'll make sure when I tweak it [again](https://github.com/ocaml-doc/odoc/issues/102#issuecomment-359748669) that it accomodates 80 columns without overflowing. I really think this particular example should simply be typeset differently.

It [seems](https://raw.githubusercontent.com/janestreet/base/master/src/float.mli) base programmers do not know about the 80-columns conventions, maybe they should.

Maybe you can convince another member of the project to merge this. But in general I think it's better to have these constraints rather than scrollbars.

Khady:
There is a lot of code out there which doesn't stick to 80 columns. It would be nice to support this code properly rather than to treat it as a third world citizen.

slice:
@dbuenzli Hm, alright. I was not aware of the 80-column convention until now. However, I do personally agree with @Khady.

Feel free to reopen if necessary.

trefis:
W.r.t. the space wasted on the left: this is (I believe) due to a bug in octavius and happens for all blocks of code. @aantron might have a fix for it though.

Issue #133, opened 2018-04-18
Fix indentation of code blocks


aantron:
This issue is about adjusting the parser slightly, to delete some leading whitespace from code blocks. 

<br/>

## Problem

When the author of docs writes

```
(** Some text

    {[
      let foo = 10
      let () = print_endline (string_of_int foo)
    ]} *)
```

...the code block gets parsed as

```
"
      let foo = 10
      let () = print_endline (string_of_int foo)
"
```

...and rendered in the HTML with all that left whitespace, which shouldn't be there. This forces people to use the ugly workaround:

```
(** Some text

{[
let foo = 10
let () = print_endline (string_of_int foo)
]} *)
```

odoc should do this automatically instead. odoc should find a line with the *least* amount of left whitespace (in this case, both lines have 6 leading spaces), and delete that much leading whitespace from all lines in the code block, in this case resulting in:

```
"
let foo = 10
let () = print_endline (string_of_int foo)
"
```

...which will get displayed nicely.

This was originally reported in #132 by @bobbypriambodo.

<br/>

## Fixing it

The issue should be pretty easy to solve. odoc already postprocesses code blocks immediately after reading them from comments, to remove leading and trailing blank lines. We need to add one more postprocessing step to that code, to remove the leading whitespace.

The rest of this post is a guide on how to do that :)

<br/>

1. **(Setup)** [Clone the repo and install development dependencies][dev-setup]:

    ```
    git clone https://github.com/ocaml/odoc.git
    cd odoc
    opam pin add --no-action odoc .
    opam install --deps-only odoc
    ```

2. **(Sanity check)** Run `make test` to check the setup, and make sure tests start out passing :)

3. **(Coding)** The existing steps for postprocessing code blocks are [here][code-block]:

    https://github.com/ocaml/odoc/blob/af59e44913110f17380bd956f51fd736e726dc56/src/parser/lexer.mll#L256-L259

    and the functions being called are defined [here][steps].

    So, we basically just need to add a third function, that goes through all lines in the code block text, finds which one has the *least* leading whitespace, and then goes through the lines again and removes that much whitespace from each one (you are welcome to use any other method, this is just the obvious suggestion :)).

4. **(Testing)** The parser is pretty heavily tested. So, once you are calling this new function, some of the tests will start failing. This is a good thing :) The first test to fail should be [this one][first-test]:

[...]

PR #134, opened 2018-04-21
Handle leading whitespace on code blocks

bobbypriambodo:
Tries to address #133.

The implementation brings in `Str` module to split strings (since `split_on_chars` is available only since 4.04.0). Let me know if you have concerns.

I also wondered whether we need to specially handle weird spaces and tabs combo, but decided not to since I don't think there's a use case for that.

### DIFF
--- a/src/parser/lexer.mll
+++ b/src/parser/lexer.mll
@@ -70,6 +70,32 @@ let trim_trailing_blank_lines : string -> string = fun s ->
     in
     String.sub s 0 trim_from
 
+let trim_leading_whitespace : string -> string = fun s ->
+  let count_leading_whitespace : string -> int = fun line ->
+    let rec count_leading_whitespace' : int -> int = fun index ->
+      if index >= String.length line then
+        index
+      else
+        match line.[index] with
+        | ' ' | '\t' -> count_leading_whitespace' (index + 1)
+        | _ -> index
+    in
+    count_leading_whitespace' 0
+  in
+  let lines = Str.(split (regexp "\n") s) in
+  let least_amount_of_whitespace =
+    lines
+    |> List.map count_leading_whitespace
+    |> List.fold_left min max_int
+  in
+  let remove_whitespace : string -> string = fun line ->
+    String.sub line least_amount_of_whitespace (String.length line - least_amount_of_whitespace)
+  in
+  lines
+  |> List.map remove_whitespace
+  |> String.concat "\n"
+
+
 
 
 module Location = Model.Location_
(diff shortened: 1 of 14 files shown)


src/parser/lexer.mll
```diff
@@ -70,6 +70,32 @@ let trim_trailing_blank_lines : string -> string = fun s ->
...
+        | ' ' | '\t' -> count_leading_whitespace' (index + 1)
+        | _ -> index
+    in
+    count_leading_whitespace' 0
+  in
+  let lines = Str.(split (regexp "\n") s) in
+  let least_amount_of_whitespace =
+    lines
+    |> List.map count_leading_whitespace
+    |> List.fold_left min max_int
```
aantron: I'd add a comment, that using `max_int` is okay because if the list is empty, we will not use the result of the `fold` (in the `map` below).

bobbypriambodo:
Thanks for the review. I went with the astring approach since it touches less files :D I've also added it on the `depends` field on the opam file.

Thanks for the catching on `max_int`, but it makes me wonder if the fact that it needed a comment means it's a bit of a smell. Unfortunately I can't think of a more straightforward implementation now, so I just added the note anyway.

Do tell if you have more concerns, would be happy to address them.

aantron:
Thanks for all that, it looks good to me :)

The `max_int` is a little bit of a smell, I guess, but I also can't think of something better.

aantron:
@bobbypriambodo There is one last thing remaining. I think we need to filter out empty or all-whitespace lines from computing the minimum:

```
{[
let () =
  Lwt_main.run begin
    let%lwt data = Lwt_io.(read_line stdin) in
    let%lwt () = Lwt_io.printl data in
    Lwt.return ()
  end

(* ocamlfind opt -linkpkg -package lwt.ppx,lwt.unix echo.ml && ./a.out *)
]}
```

In that example, the blank line is completely empty, so even when I indent the first and last line by two spaces, the blank line, which has no characters, forces the overall amount of leading whitespace to be computed as zero. Blank lines could have whitespace characters in general, though, even though that's sloppy coding :)

Of course, after ignoring the blank line, we will get two for the amount of leading space, so we have to be prepared for the blank line having less characters than that to substring out.

Let me know if you'd like me to fix this, but a quick follow-on PR is also welcome :)

bobbypriambodo:
Ah, that is true, I already tried it on several examples but none had that characteristics. Sorry for that. I think I'll be able to fix it tomorrow (it's past 11 PM on this part of the world). If that's not quick enough then please handle it as you see fit. Thanks.

aantron:
Thanks, that is fine :) and sorry for not catching right away.

PR #136, opened 2018-04-25
Restore support for {%...%}

### LINKED ISSUE #112: Restore support for {%...%}, perhaps with a warning

aantron:
Writing `{%html:<b>foo</b>%}` in a comment now puts the raw HTML `<b>foo</b>` into odoc's HTML output, as it does with ocamldoc, and did with older versions of odoc.

There are a couple of notes:

1. In ocamldoc and older odoc, the default target is LaTeX. I made the default target HTML, i.e. you can write just `{%<b>foo</b>%}`. I think this is ok, because when I ran odoc against most opam packages, I found no instances of raw output being used for LaTeX.

    We can restore the default target to LaTeX, however odoc doesn't support LaTeX at all yet. Alternatively, we can remove the notion of a default target completely.

2. The current odoc parser has a notion of inline elements (like emphasized text spans) and block elements (like paragraphs). `{%...%}` is always parsed as an inline element. However, if it is alone in a paragraph, it is emitted as if it is a block element. So,

    ```
    foo

    bar {%<e>baz</e>%}
    ```

    becomes

    ```html
    <p>
      foo
    </p>

    <p>
      bar <e>baz</e>
    </p>
    ```

    ...but

    ```
    foo

    {%
    <blockquote>
      bar
    </blockquote>
    %}
    ```

    becomes

    ```html
    <p>
      foo
    </p>

    <blockquote>
      bar
    </blockquote>
    ```

    and odoc does not generate a `<p>` tag around the `<blockquote>`.

Resolves #112.

cc @dbuenzli @rizo @bobbypriambodo


dbuenzli:
Thanks @aantron. All that seems perfect. 

      and odoc does not generate a <p> tag around the <blockquote>.

Are you sure about the separating blank line though ? (Just asking I don't have these thing in my head).

Regarding: 

     In ocamldoc and older odoc, the default target is LaTeX. I made the default target HTML,

No objection to this, but please make sure these deviations get recorded somewhere (or even you could propose them upstream, at least documentation-wise). So far the reference has always been [this document](https://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#sec349) and balkanisation of the documentation language should be avoided.

dbuenzli:
Also closes https://github.com/dbuenzli/odig/issues/35

aantron:
> Are you sure about the separating blank line though ?

It doesn't get generated, I put it into the issue comment for presentation purposes. Should have known you will catch such things :)

> please make sure these deviations get recorded somewhere

Indeed, I have a file for this, and should probably clean it up and commit it. However, thinking about this specific case more, I think we should disable default target support. Since we are only focusing on HTML, every user would have had to write `{%html:...%}` in ocamldoc, and we can require the same thing for now in odoc.

We can deal with the default target again later, perhaps when looking at LaTeX output (cc @Octachron).

Issue #95, opened 2017-11-12
Documentation: second level includes are not visible


Chris00:
When an included interface includes a second level one, the latter is not visible enough when it is open.
![32697375-45a9a756-c78f-11e7-9d90-eabd517aa7d6](https://user-images.githubusercontent.com/1255665/32702400-e74bbf0c-c7e6-11e7-8862-63d915f4f960.png)
A closed second level also changes the line width, which looks odd.

dbuenzli:
I'm not sure I understand what the problem is here. And where does the line width change ? Could you please rather send a link on an example and also mention your browser.

Chris00:
> And where does the line width change ?

In front of `include Base__.Monad_intf/S_without...`

Chris00:
> what the problem is here.

Did you notice two braces are open?  If not, that's the problem — knowing what is the origin of each value.

PR #139, opened 2018-05-21
HTML/CSS improvements

rizo:
This PR makes small changes to the HTML generation and rewrites the CSS styling rules. I started with the visual CSS changes and tried to minimise actual code changes.

More specifically:

- The page is wrapped into a `div` with the `content` class. This gives more control over styling/responsiveness of the page and prepares the page structure for the inclusion of new elements like sidebar (also in this PR) and search bar (will have a PR for that in the future).
- The `odoc.css` was formatted with the common CSS formatting options. I think the previous style was more compact but also harder to refactor.
- Added `toc` class to the table of contents `nav`. The sidebar is responsive and is inlined on smaller screens.
- A new visual theme is proposed with (IMO of course) improved typography and better visual structure of the code items. I'd like to use custom fonts too but this could be done separately.
- All code items are now annotated with a `spec` class and their particular spec kind (`value`, `module-type`, etc).

Currently the Table of Contents is not rendered in nested modules AFAIK, I can work on that after this PR is merged.

Let me know if something could be improved or if there's something you don't like about this PR. I'm open to ideas and will make any changes you think make sense.

CC @aantron 

---
## Example Packages

- <http://odis.io/odoc-demo/order/Order/index.html>
- <http://odis.io/odoc-lwt/lwt/Lwt/index.html>
- <http://odis.io/odoc-base/base/Base/index.html>


pmetzger:
This is very nice! I could quibble with some of the details (I prefer serif body fonts for example) but those are details. All in all, this is great!

(You might want to squash the git commits down to each functional change for review and merge purposes.)

lpw25:
Pretty good. Could you show some examples of the more awkward elements e.g. functors and includes in your demo?

The style looks nice for your front page, and in other places with lots of text. But I don't think it looks great for pages with little text e.g. http://odis.io/odoc-demo/order/Order/Equality/index.html. I would suggest focusing on those pages as they are both more common and much harder to make look nice than the tutorial style pages with lots of text.

I don't really have any concrete suggestions, but I think perhaps the blue side thing is the main thing that looks unattractive on those pages. Also there seems to be no styling for OCaml keywords on those lines.

Finally, and this may well be an existing bug, it seems that module type declarations are coming out as:
```ocaml
module type Equal0 : sig ... end
```
instead of:
```ocaml
module type Equal0 = sig ... end
```

aantron:
For pages like http://odis.io/odoc-demo/order/Order/Equality/index.html, the previous CSS and layout gave the option of *not* documenting every single value. If there is a group of consecutive signature items of the same kind, they would all get displayed in one "block," and the documentation of the last one of them (if any) would appear as if it was the documentation for all of them. This could work well for groups of related values, like on that page.

lpw25:
(I'm assuming that was a response to my post, rather than a general comment). My point was not specific to groups of related values. Far more pages contain only a single line or no description for each item than contain a paragraph between each item. Styling these nicely is much more challenging.

rizo:
@lpw25 Thank you for your feedback!

The package has some examples of functors. See this pages:

- <http://odis.io/odoc-demo/order/Order/Ordered0/Extend/index.html>
- <http://odis.io/odoc-demo/order/Order/Ordered0/module-type-Extension/index.html>

I will definitely spend more time looking at edge cases and investigate this style with other popular OCaml packages (both with and without extensive documentation). The goal of this PR was to validate my direction so that I could focus on those.

> I think perhaps the blue side thing is the main thing that looks unattractive on those pages.
I'm sad to hear that! 😛 I wanted to make the code items visually distinguishable. I will try some simpler approaches.

Both lack of "styling for OCaml keywords on those lines" and the issue with "module type declarations" are a bit out of scope of this PR. I can address them separately.

Thanks again!

rizo:
@aantron for reference here's an example of that case in action: <http://odis.io/odoc-demo/order/Order/Comparator/index.html#val-descending>. `descending` and `ascending` are groupped. Not sure if that's clear enough in the rendered UI.

In general, I agree that less text-intens declarations are harder to style.

rizo:
Another problem that I noticed is that currently the first paragraph for module items is not included in the generated HTML (like here: <http://odis.io/odoc-demo/order/Order/index.html#module-type-Equal1>). I can fix that in a separate PR, if that's no intentional.

lpw25:
> The package has some examples of functors

Thanks. The functors look good. The includes need some work. The tricky thing about includes is that you want the contents to line up with the surrounding context when they're expanded, whilst still being able to easily see where the include ends (even when nested within other includes). That's why I put the bracket things in the current style -- but I've never been very happy with them and would love to see them replaced with something nicer.

> Both lack of "styling for OCaml keywords on those lines" and the issue with "module type declarations" are a bit out of scope of this PR. I can address them separately.

Fair enough. Although I think the "styling for OCaml keywords" is just a css thing because the existing style handles them.

> The goal of this PR was to validate my direction

For a first draft this does look pretty nice, keep up the good work.

rizo:
I did try aligning the includes with the main content. As you said it gets very tricky with nested includes. I'll try some other ideas that might work. The main motivation to indent/box includes was to avoid the issue mentioned here: #95.

Regarding the keyword styling in declarations: yes, it's a minor technical thing. [highlightjs](https://highlightjs.org) is used for syntax highlighting in code examples but declarations already had their own markup annotations too. Anyway, suggestion noted, I'll fix that too! :)

lpw25:
> I'll try some other ideas that might work. 

In case it's useful, the only other idea I had on this front was something like:
```
    val foo : int
      A foo
    - - - - - - - - - - - - - - - - - - - - --------------------------------
    val bar : int                                                  Bar.S    |
       A bar                                                                |
    val baz : int                                                           |
       A baz                                                                |
    - - - - - - - - - - - - - - - - - - - - -----------------------------   |
     val nested : int                                        Nested.S    |  |
        A nested thing                                                   |  |
     - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -  - - - -   |
     val another_thing_in_bar : int                                         |
     - - - - - - - - - - - - - - - - -- - - - - - - - - - - - - - - - - - - 
     val something_outside_the_includes : int
```
The fundamental idea is to use the right-hand edge to indicate the includes and nesting, leaving the left-hand edge nicely aligned.

rizo:
Thanks for the nice diagram, @lpw25!

I've created a few different styles that keep all items aligned to the left as you suggested:

1. http://odis.io/odoc-include/aligned-border-medium-1.html
2. http://odis.io/odoc-include/aligned-border-small-1.html
3. http://odis.io/odoc-include/aligned-indicator-left-1.html
4. http://odis.io/odoc-include/aligned-indicator-right-2.html

In all cases the border is rendered as an outline avoiding any padding/indentation. The level-indicator idea was inspired by the text-editor folding feature, but to be honest I don't think it works well in practice.

I personally like option 1 and 2. Let me know if this is what you had in mind.

I have also applied the style to more packages and am fixing small rendering issues. Here's Lwt for example: <http://odis.io/odoc-lwt/lwt/Lwt/index.html>.

samoht:
I've tried your branch on Irmin and it seems only a short paragraph is generated for the main page. Not sure what is happening.

lpw25:
Those all look pretty good. I think I prefer 3 and 4 because they make it very easy to see the extent of the include. Those indicators are so much better than what we are using now.

The style of the side bar seems to have changed in your examples. I think the old version was nicer, but of course such things are subjective.

By the way, it seems like links to sections within pages have been broken. That Lwt page has internal links to "http://odis.io/odoc-lwt/lwt/Lwt/index.html#3_Quickstart" which seems to now instead have address "http://odis.io/odoc-lwt/lwt/Lwt/index.html#quick-start".

rizo:
@lpw25 I'm glad you liked the level indicator version! It's certainly good at highlighting the include scope. I prototyped both left and right variants because of the anchor links on the items: they are already fighting for the space on the left, so I think the indicator fits better on the right. Do you agree?

I'm not entirely satisfied with some minor visual details of the level indicator version, but I could revisit them in the future.

Regarding the link issue: I think it's unrelated to my PR. In Lwt's [`mli` file](https://github.com/ocsigen/lwt/blob/master/src/core/lwt.mli#L57), that link points to `{{: #3_Quickstart} Quick start}`, the section is titled as `{3 Quick start}` and it seems like odoc generates an automatic id/anchor `#quick-start`.

@aantron can probably elaborate on this, I remember seeing a commit that introduced automatic id generation for headings.

---
@samoht Thanks for trying my PR and reporting this! It's a known issue in odoc already present in `master`. See <https://github.com/ocaml/dune/issues/791> for more details. Essentially the pages now are not allowed to have more than one `{1 ...}` heading.

samoht:
@rizo ha yes, I haven't seen that and that makes sense. So after fixing this I can get the nice doss, that's really great!! 

Something which doesn't seem to be working is that the table of contents is not displayed by submodules, for instance: http://gazagnaire.org/odoc/irmin/Irmin/Merge/index.html

Drup:
My vote also goes for version 4. A potential nice addition would be able to click on the level to go back to the anchor of the include.

rizo:
> The style of the side bar seems to have changed in your examples. I think the old version was nicer, but of course such things are subjective.

The Lwt example style uses a different font, that's the only difference. I forgot to add the CSS import for that font (Fira Sans) so you're probably seeing it with Times New Roman (the default sans-serif font on most browsers) – I'm sorry, that most look awful! 😄

rizo:
@samoht This PR still has some issues but I'm glad you're happy with the result! 😃 The TOC is currently only included in the main page. I will work on fixing that in a separate PR as the scope of this one is already quite broad.

rizo:
@Drup I did think about that but making it work with pure CSS would take me ages. I might do that in the future with Javascript.

trefis:
> Essentially the pages now are not allowed to have more than one {1 ...} heading.

You might have already explained this (in which case: I'm sorry for missing), but I'm wondering why that restriction has been added. Would you mind justifying it again?

---
Note: despite my silence or possibly annoying questions I find it great that you are doing this work and the result is appealing :)

rizo:
@trefis I honestly have no idea. I've only started contributing to `odoc` and was not involved in any decisions nor have knowledge of how/why they were made.

Just to clarify, the single `h1` heading restriction is not part of this PR and is already present in `master`. I actually think it makes sense, but error reporting could be definitely improved. Right now the documentation generation simply terminates producing half-finished pages.

One thing I would like to help with next is producing and publishing documentation for odoc itself.

Please, keep making annoying questions – they bring clarity! :)

pmetzger:
FYI, I prefer 1 and 2, the way @rizo does...

aantron:
> You might have already explained this (in which case: I'm sorry for missing), but I'm wondering why that restriction has been added. Would you mind justifying it again?

I added the `<h1>` restriction so we can easily select the title of both `.mld`s and `.mli`s with CSS. For `.mli`s, we control the generation of the page title, so we could easily use classes or other markup instead. But for `.mld`s, I think the easiest way for the author to communicate the page title to odoc is by putting it in `{1 ...}`.

We could change this if you have another suggestion. For example, we could take only the first `{1 ...}`, and give it a class. I currently think that having title-level headings throughout a doc page is an authoring mistake, though, and that we should catch it.

By the way, this isn't an error in odoc right now, IIRC. odoc should be converting these to level-2 headings, with a warning. The error output is not clear because it is hacked together.

aantron:
@rizo Indeed about the ids, I'll take a look at that separately. I don't think it should block this PR :)

smolkaj:
@rizo: I like version 3 most. Version 4 is also good, but I'm a bit bothered by the grey box around the include statement. Could v4 be made to look more like v3?

What do other people think about the grey boxes? Am I the only one bothered by this?

lpw25:
> so I think the indicator fits better on the right. Do you agree?

I do. I also agree with @smolkaj that it would be good if the box around the include in version 4 could be made to look like version 3 instead.

> I'm sorry, that must look awful!

That may be part of it, but there is something more going on. I realized I had been looking at the pages on different machines and in different browsers. Indeed, the side-bar looks great in chrome at work, but in firefox and chromium on my laptop it is much worse. In chrome at work there is no background colour (or a maybe a very subtle one), nice fonts and nothing obvious between the side-bar and the content. In chromium on my laptop there is a light grey background, a worse font and a scroll bar between the side-bar and the content. In firefox on my laptop there is a light grey background, a worse font, and a weird grey bar between the side-bar and the content (which becomes a scroll bar when I zoom in).

hcarty:
I like version 4 - it indicates what comes from the include without making it looks too much like the included content is special.  From a user/reader perspective the fact that a definition comes from an include may not matter in many cases.  So it's nice to have the indicator visible but out of the way.

rizo:
I've pushed changes that add the `include` level indicator (version 4) on the right side of the content without boxes (as in version 3). 

I've updated the [demo project](http://odis.io/odoc-demo/order/Order/index.html) and the [lwt docs](http://odis.io/odoc-lwt/lwt/Lwt/index.html). You can also go directly to the [`include` example](http://odis.io/odoc-demo/order/Order/Ordered0/module-type-Extension/index.html).

There is one issue with the indicator approach though: currently the `details summary::-webkit-details-marker` selector is styled to move the `◀︎` symbol on top of the indicator (without it the indicator looks like a scrollbar!), but this is not supported in Firefox. I will try to come up with an implementation that is more portable.

---
@smolkaj The reason I was trying out boxes for includes is because they are clickable and act as buttons. I agree it looks cleaner without them. But again, I still think it could be improved, I just don't know how but I'll keep thinking about this :)

---
@lpw25 I am testing the PR in multiple browsers (Chrome, Chromium, Firefox, Safari, Opera – all on MacOS) but don't see anything like that. I think my last commit might solve the issue with the scrollbar in the sidebar, though. Also the current examples only use system fonts (namely Helvetica, DejaVu Sans, Arial or the default sans-serif). I'll revisit the font selection when other more important things get done.

Can you check if the issues are still there? If they are, could you post a screenshot, that would help a lot?

pmetzger:
I'll just note that having the include "bars" on the right is kind of unnatural, as English is read from the left, so you only notice that something is part of an include group after you get to the end of the line. I think the bars should be at the left. Ideally there would be an indication on both sides, but apparently I got outvoted on options 1 and 2.

hcarty:
If the include bars were necessary to understand the library I would agree @pmetzger. They are not though, they're more of an implementation detail from an end user perspective.

rizo:
One concrete and general issue with includes is that for type modules (like the one in my example) the page will have two (or even more) `type t` declarations – it's slightly confusing and also breaks the anchors because they all are identified as `#type-t`. I guess having a box at least explicitly defines the scope of those repeating types.

lpw25:
>  the page will have two (or even more) type t declarations

That's a long standing issue that will be resolved at some point, so hopefully not worth worrying much about when it comes to designing the layout.

trefis:
Btw, there is something I don't really like about the current proposition which is that the markup of includes (background color, etc), at least when they are folded, is completely different from the one of the other signature items. Actually it looks like the text coming from docstrings.

With the current CSS (that was kindly contributed by @dbuenzli), includes where styled like other signature items and very clearly not text. I like that better.

trefis:
(Oh, I apparently wasn't looking at the most recent version. I must correct: it doesn't exactly look like text coming from docstrings; but it still looks reasonably different from other signature items. I don't know whether we should care about it or not)

rizo:
@trefis You raised a good point. I did think about this and decided to style includes differently because in my understanding they are not on the same level as other items such as type, value and module declarations – they don't really exist in the library from user's perspective. They also don't have an id/anchor (should they?).

I do agree that they look a bit inconsistent currently so, as an experiment, I followed your suggestion and styled them just like other items. And I actually like the result, you can see it [here](http://odis.io/odoc-demo/order/Order/Ordered0/module-type-Extension/index.html). What do you think?

---
I like the original simple style, it works very well even with very complex functor definitions and includes. My only concern is that I find it a bit monotonic and hard to navigate visually – that's why I started this PR and added the sidebar. Of course it's very subjective, but I would love to know your opinion on this @dbuenzli! :)

---

I'm continuing applying the style to more packages and have tested it now with `base`, here's how it looks:

- <http://odis.io/odoc-base/base/Base/index.html>

_Note_: the sidebar is not being rendered because there are no headings in the documentation.

dbuenzli:
> I like the original simple style, it works very well even with very complex functor definitions and includes. My only concern is that I find it a bit monotonic and hard to navigate visually – that's why I started this PR and added the sidebar. Of course it's very subjective, but I would love to know your opinion on this @dbuenzli! :)

I refrain interacting too much on this kind of issues except on usability aspects precisely because they quickly delve into taste issues. But if you want them anyways, I agree everyone's tired with the browsers' default fonts. However putting font issues aside I somehow miss the monotonic aspect of the original stylesheet. One of its aims was to avoid too much framing and boxes around signature elements which I find quite distracting, especially when docstrings are short. I know people like how it looks --- which is easy to get impressed by, by glancing --- but it's more important to actually see how it reads, i.e. you have to actually use the documentation to see how it works.

I think the sidebar is an essential and very welcome addition.

I'm not sure I understand what @trefis is mentionning but if that's what I understand it's a side effect of the fact that signature elements are not framed (see above).

Regarding testing doesn't `odig` work with `odoc`'s master ? It would be nice to see the results on a wider package set (especially on [those](http://erratique.ch/software) that make heavy use of the documentation language's abilities).

rizo:
Thanks for sharing your opinion, @dbuenzli!

> I agree everyone's tired with the browsers' default fonts.

I do believe that a custom font could improve the look of the documentation and give it some character. I'm not certain yet which fonts to use. I also think the size of the font plays an even more important role in this case. I imagine you prefer smaller font sizes (14px), I'm currently using the default 16px.

I understand your preference for plain and simple styling. Currently odoc reminds me of Unix man pages – which are an example of minimal, practical design. But they are also very different from what most people expect to see on modern web pages. While it's probably a matter of taste and habit, I find _large_ monotonic pages hard to navigate both visually and from the content perspective. That was the main reason I added a small outline to the signature elements – to make it easier to focus on them.

I agree that elements with short documentation don't look as great in the current style. Decreasing the font size of the whole document could improve that by making them stand out less. If not, I can remove the outlines, but I'll leave this decision to the project maintainers because I personally like them.

> I know people like how it looks --- which is easy to get impressed by, by glancing --- but it's more important to actually see how it reads, i.e. you have to actually use the documentation to see how it works.

I really want believe it's possible to have a style that both looks great and works well, as I'd rather sacrifice neither. :)

> Regarding testing doesn't odig work with odoc's master ? It would be nice to see the results on a wider package set (especially on those that make heavy use of the documentation language's abilities).

Unfortunately no. Currently many packages I've tested are broken mainly because of the [single title-heading restriction](https://github.com/ocaml/odoc/commit/177af882bfb03a4b72c92038275f79acb7c8db6b) AFAIK. As a workaround I've patched some of your packages locally (removing `{1 ...}` titles) and am testing them now. I'll will publish the results soon.

rizo:
Just to complement...

Before I started working on this PR I've spend a lot of time researching documentation systems from other programming languages. Allow me to share a few examples of what inspired me (at the risk of you already being familiar with some of them):

- [Rust](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html) is known to have detailed technical documentation with many examples.
- [Racket](https://docs.racket-lang.org/reference/strings.html) has a very well-structured and balanced documentation on content and presentation levels.
- [Haskell](https://hackage.haskell.org/package/pipes-4.3.9/docs/Pipes-Prelude.html) typically has a very concise documentation with a formal feel.
- [D](https://dlang.org/phobos/std_range.html) has challenging metaprogramming constructs that are hard to document (similar to OCaml's includes).
- [Erlang](https://erldocs.com/current/stdlib/timer.html) has a very barebones documentation but with nice navigation.
- [Elm](http://package.elm-lang.org/packages/elm-lang/core/5.1.1/String) is known to be very beginner friendly in its docs.

I specially like the Rust and Racket documentations as they both look great and make it very easy to find things. A lot of them also outline code elements. I'm not saying we should copy them – OCaml is unique and different in it's requirements – but there are some good visual conventions that we could consider adopting.

dbuenzli:
> Currently many packages I've tested are broken mainly because of the single title-heading restriction AFAIK. As a workaround I've patched some of your packages locally (removing {1 ...} titles) and am testing them now.

That's a bit insane rather patch `odoc` than packages. I don't know what was the rationale behind this move but it makes absolutely no sense.

rizo:
See @aantron's comment on this [above](https://github.com/ocaml/odoc/pull/139#issuecomment-391782891).

I was surprised by this restriction but assumed it was the new way of doing things. I'm fine with it being a warning if it facilitates other things.

On the other hand....

# This H1 Works

...and is marked as `.markdown-body h1` by GitHub.

dbuenzli:
> We could change this if you have another suggestion. For example, we could take only the first {1 ...}, and give it a class. I currently think that having title-level headings throughout a doc page is an authoring mistake, though, and that we should catch it.

I strongly disagree with this. There is no such thing as  "title-level" heading. The title of a page belongs to the `title` element. Headings and their level simply structure content into sections put on a equal foot. 

When you render `mli`s the highest level heading should be the module name whose level should be inaccessible to the contents of the `mli` itself (in some sense it's `(** {0 } *)` which you'll likely map to `h1`). Then you should use the ocamldoc language to section the contents of the `mli` using the headings you wish and the most natural and done by hundreds of `.mli` is to simply start from `{1` (which you'll likely map to `h2`). With the scheme you propose the `{1` level would be basically useless for `.mli`s. For the `title` element you can simply derive its content from the module name and it's package. This is the way `odoc` and `ocamldoc` have always worked.

Now regarding `.mld`'s title where the file name might not make sense to use as in `.mli`s you can simply use the convention, to populate the document's `title` element, to use the contents of the first heading you find whatever level that may be.

> I was surprised by this restriction but assumed it was the new way of doing things. I'm fine with it being a warning if it facilitates other things.

I think this restriction should simply be reverted.

aantron:
> I strongly disagree with this. There is no such thing as "title-level" heading. The title of a page belongs to the title element. Headings and their level simply structure content into sections put on a equal foot.

This statement is about HTML, which is our output language. I am talking exclusively about the input language, and about how to achieve semantic and style consistency between mlis and mlds.

> When you render mlis the highest level heading should be the module name whose level should be inaccessible to the contents of the mli itself

Yes.

> (in some sense it's (** {0 } *) which you'll likely map to h1).

Precisely. I am saying we should use `(** {1 } *)` for this, in a direct (not "some") sense, because the mapping to `h1` is more obvious.

> Then you should use the ocamldoc language to section the contents of the mli using the headings you wish and the most natural and done by hundreds of .mli is to simply start from {1 (which you'll likely map to h2).

I suggest we should tell people to start from {2, which we will map to h2, and reserve {1 for mlds only. We tell people that the reason is that odoc controls {1 inside mlis. That way, everything is obvious and consistent, and easy for people to understand when styling docs with custom CSS.

Current odoc is supposed to convert {1 in comments to {2, but there seems to be a bug of some sort.

I don't agree with arguments based on current practice. For example, current practice is also to use {9 as the "lowest-level" heading, but it seems ridiculous to have to invent ten levels of styles for different section levels. We have an opportunity to give more sane guidelines about sectioning with odoc.

> With the scheme you propose the {1 level would be basically useless for .mlis

That's (part of) the intent of the scheme.

nilsbecker:
i hope i'm not intruding by listing my impressions about the styling. 

i'm comparing the new html documentation for lwt.Lwt linked at the beginning of the PR, with the same module documentation generated by the current opam version of `odig odoc`. obviously the sidebar is great in the new version. however there are some aspects that i think are worth carrying over from the old version.

- the old version has a smaller font size. i think this is actually an advantage since
  - it allows one to see more of the content at once. this makes it easier to orient oneself within the page. (i have to scroll around more to see what is going on in the new layout.)
  - it affords generous line spacing and general white space 
- the page appears visually calmer and is easy grasp at one glance, since typographical variations are more minimal. i think this increases the usability when browsing the documentation.
- i think the slightly shorter character number per line in the old version is preferable for reading comfort.
- i believe the font size increase towards higher header levels should be kept very small to avoid 'screaming'/distraction and clutter. the old version does this well.
- the distinction between text and inline code snippets is quite weak in the old version. i think a more contrasting pair of fonts would be helpful to easily see what is code. however i like that there is no change in background color. inline code examples with different background as in the new version, give the text an uneven feel.

overall, my impression is that the new version, while generally already nicely readable, could be improved by making it visually somewhat calmer. this could be achieved by reducing font size and contrast in font size, increasing white space e.g. around headings, reducing variations in text background and graphical elements such as lines, bars etc.

Octachron:
> I suggest we should tell people to start from {2, which we will map to h2, and reserve {1 for mlds only.

Note that this was the semantic that was supposedly chosen by ocamldoc; and it didn't work: people still used `{1`, and that's why I patched ocamldoc to move the title heading to `{0` in 4.06 . If you really want to have headings start at `{2`, you will have to fight against the natural tendency of starting heading at `{1`  by publicizing this choice of `{2` broadly and loudly.

aantron:
@nilsbecker Agreed. I think we should merge this PR as a starting point (eventually :p), and then we can apply a series of tweaks to the styling afterwards.

aantron:
Thanks @rizo for the PR, and all for reviews and discussion!

PR #143, opened 2018-05-31
Create links in module title

bluddy:
For deep modules, it's really nice to be able to skip up the hierarchy to a specific location. This PR turns the relevant parts of the module title into links up the hierarchy. A simple example is pictured:

image

### DIFF
--- a/src/html/html_tree.ml
+++ b/src/html/html_tree.ml
@@ -19,8 +19,6 @@
 module Html = Tyxml.Html
 module Paths = Model.Paths
 
-
-
 type kind = [ `Arg | `Mod | `Mty | `Class | `Cty | `Page ]
 
 type t = {
@@ -219,7 +217,7 @@ let render_fragment = Relative_link.Of_fragment.render_raw
 
 let page_creator ?kind ~path header_docs content =
   let rec add_dotdot ~n acc =
-    if n = 0 then
+    if n <= 0 then
       acc
     else
       add_dotdot ~n:(n - 1) ("../" ^ acc)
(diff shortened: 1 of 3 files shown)


rizo:
I like this addition.

One thing you can consider is separating the path from the title. So that the last entry in the path would be a global standalone title and the path would be placed near the "Up" navigation. I know @dbuenzli mentioned that this can lead to usability issues with small module names like `S` but I think it could be solved with `a`-tag padding and some styling. After all we already have single-character links for anchors `#` and those work just fine.

bluddy:
I'm ok with whatever we want to massage this into -- if it's breadcrumbs, that's fine.

I do think that intuitively, when I see a multi-part title composed of a path like what we have, I expect modern web conventions to allow me to click on parts of it to get to those paths. It feels 'right' and modern (though my notions of modern could always be off). So another option is to keep things as they are here, but to also improve the css style so we're looking at something nicer than just the regular blue links.

rizo:
This is a bit unrelated to your proposal and complements my suggestion: I find full module paths in the `h1` headings quite intrusive specially for complex libraries like Base. Consider the following example:

screen shot 2018-05-31 at 16 22 27

I know that this style uses large fonts, but still, having that as a title is very distracting. My suggestion is to only use `Indexable` for the title and move the path closer to navigation with links to intermediate pages.

bluddy:
I think that's a great point @rizo -- your suggestion makes sense in this context. I'll try to play around with this.

lpw25:
I think your probably right, although the number of pages whose title will be `S` makes me slightly sad. Can't think of a more sensible alternative though.

rizo:
@lpw25 It can be:

## Module type `S`

...which is not that bad, IMO.

bluddy:
Another option is to define some maximum length to abide by, and to abbreviate it in the main title, while keeping the breadcrumbs at the top with the full title:


``` Base >> Binary_searchable_intf >> indexable ```
## Module type Base.Bi...Indexable

rizo:
I think completely hidding information should be the last resort in user interfaces. I see it acceptable on very small screens but not as a fixed restriction on the title.

Module type `S` is such a common convention that I think we could render it as `Parent.S` in the title.

lpw25:
I'd rather not special case output based on conventions. It'll just confuse someone who uses `module type T` or `module type S2`. How about something along the lines of:

    <h2> Foo.Bar.Baz. <h2>
    <h1> S <h1>

so still showing the full path but de-emphasising the earlier parts of it.

rizo:
I think that’s very close to what I suggested originally. Albeit with links for intermediate modules in `h2`.

I can propose a few different styles for this after I finish my other styling PR.

bluddy:
Pulling off @lpw25's suggestion should be easy. I have another branch where I created breadcrumbs at the top like @rizo suggested, and I can make it show modules Leo's way. I just need @rizo's fix to be merged so I can actually test on real pages that generate properly.

bluddy:
Closing this. I think the other PR (#150) is better.

Issue #141, opened 2018-05-28
Allow level-1 headings in mli files


bluddy:
I can't get any proper documentation output on 4.06.1. This happens even after pinning to dune master, which has an [issue](https://github.com/ocaml/dune/issues/791) which seemed to be similar to what I'm seeing.

Steps to reproduce:

1. Install 4.06.1 and dune/jbuilder.
2. Clone https://github.com/rizo/order.
3. Run `jbuilder build @doc`.
4. Open `_build/default/_doc/_html/index.html`
5. Click on the `order` package. 
6. Click on the `Order` module.
7. Click on the `Ordered` interface. You should get a 404 -- this resource wasn't created.

Running on `odoc` itself produces similar issues -- nothing is created inside the odoc package.

(To try out pinning to `dune` master, you currently have to use `dune` as the package name and prompt opam to create it, or both `dune` and `jbuilder` will not be available.)

bobbypriambodo:
I can reproduce this using the steps given above. Some pages are not generated.

| dune ver.       | odoc ver.        | docs working? |
| ------------- | ------------- | -------------- |
| 1.0+beta20    | 1.2.0               | Yes                   |
| dev                 | 1.2.0               | Yes                   |
| 1.0+beta20    | dev                 | No                    |
| dev                 | dev                 | No                    |

FWIW, there were several warnings emitted when using `odoc.dev`:

```
$ jbuilder build @doc
        odoc _doc/_odoc/lib/order/Order.odoc
File "src/Order.mli", line 322, character 53 to line 323, character 12:
line break is not allowed in '[...]' (code)
File "src/Order.mli", line 364, character 61 to line 365, character 31:
line break is not allowed in '[...]' (code)
File "src/Order.mli", line 389, character 75 to line 390, character 8:
line break is not allowed in '[...]' (code)
```

But even after fixing those errors (deleting the line breaks) the docs are still not generated properly, now without warnings or errors.

rizo:
@bluddy thanks for reporting this and sorry for the delay.

This is still related to the `{1 ...}` error. The mentioned issue in dune has nothing to do with this because it's odoc that has this restriction. The submitted [fix in dune](https://github.com/ocaml/dune/commit/b2dd90484776774e5f9f2204ab5d9ff0bb9e575b) simply stops using title-level headings avoiding the restriction.

I've also "fixed" this now in the `order` library by pushing a change that doesn't use `{1 ...}`.

@aantron I think we should relax this restriction somehow.

[...]

PR #149, opened 2018-05-31
Relax heading restriction

rizo:
This PR relaxes the heading level restriction reported in #141 which affects many existing packages. It does not change the current behaviour in which (a) the heading levels in docstrings are clamped to the interval [2-4] and (b) only one level-1 heading is allowed in manual pages.

The current implementation had a bug which resulted in the produced HTML page being cut off after parsing the first level-1 heading.

Based on my research I concluded that the bug was caused by the fact that the HTML generation code assumed that there was only one `` `Title `` heading. Specifically the [To_html_tree.Top_level_markup.section_comment](https://github.com/rizo/odoc/blob/fd68d5b7a479cce21cf2353304178c9162c152a6/src/html/to_html_tree.ml#L842) function checks for section nesting by comparing levels. In the case of duplicate `` `Title `` levels the function will terminate leaving unprocessed comment items.

This PR fixes that by mapping repeated `Title` levels to `Section` levels, thus avoiding the described issue. A warning is produced when this happens (previously it resulted in an error).

#### Note on Design

The sole intent of this PR is to fix the issue in the existing design in order to allow testing of packages with odoc. It was [suggested](https://github.com/ocaml/odoc/issues/141)  that the current design is not ideal and might be changed in the future; these concerns are out of scope of this PR.

### DIFF
--- a/src/parser/semantics.ml
+++ b/src/parser/semantics.ml
@@ -237,6 +237,12 @@ let section_heading
   in
   let label = Model.Paths.Identifier.Label (status.parent_of_sections, label) in
 
+  let is_page =
+    match status.parent_of_sections with
+    | Model.Paths.Identifier.Page _ -> true
+    | _ -> false
+  in
+
   match status.sections_allowed, level with
   | `None, _ ->
     warning status (Parse_error.sections_not_allowed location);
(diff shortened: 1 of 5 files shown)


aantron:
Thanks for looking into this and fixing it!

PR #150, opened 2018-06-01
Breadcrumbs + 2-level module path

bluddy:
Rather than put the links in the title, I followed @rizo's suggestion and created breadcrumbs, restricting the module path in the title to 2 levels to avoid overflow. I think this looks good, though it could use some styling.


rizo:
Since it's hard to reproduce this before #149 is merged, this is how it looks:
odoc-breadcrumbs

pmetzger:
What are "breadcrumbs"? I haven't been following...

rizo:
See: [Breadcrumb (navigation)](https://en.wikipedia.org/wiki/Breadcrumb_(navigation)).

aantron:
Ok, #149 and #139 from @rizo are both now in `master`, so you should be able to rebase this PR and see the effect :)

bluddy:
@aantron I think this looks good enough to merge now. We can improve the style later as needed.

aantron:
Ok, just going to update the expect tests, then will merge it.

aantron:
Thank you!

Issue #4, opened 2016-08-08
CSS handling story


dbuenzli:
Current CSS handling is sub-optimal, a link to a CSS file called `odoc.css` located above `--output-dir` is generated in the HTML. 

Here's a suggested approach:
1. By default the link to the CSS file should be `odoc.css` in the `--output-dir`. A default CSS file located in `etc` should be copied over in the `--output-dir`. [These instructions](http://erratique.ch/software/topkg/doc/Topkg.html#config_store) show how `topkg` can be used to work with an `etc` directory both in release and development mode.
2. A `--css-uri URI` option should be added to override the default CSS link (`odoc.css`) generated in the HTML.
3. A `--css-file FILE` option should be added to be able to specify the contents of the `odoc.css` file in `--output-dir`
4. A `--no-css-file` option should be added to specify that no CSS file should be copied over to `--output-dir`.

trefis:
Just to make I understand, point 1 says that the default is to link to an `odoc.css` but leave it to the user to ensure that this file is present (e.g. through his build system), correct?

dbuenzli:
No point 1. copies an `odoc.css` located in the `etc` dir. You'd need to use `--no-css-file` or provide one with `--css-file` if you don't want that file to be used.

trefis:
Ah, right. Got it.

dbuenzli:
N.B. 1. is similar to what `ocamldoc` does by default, it's useful when you want to quickly generate a doc set for a project.

lpw25:
I think that 2 makes sense as part of the `link` command, but the others aren't quite right. We don't want to copy over a css file every time we `link` something. I think the best way to handle this is to add a `css` command or similar, which copies the default file or a specified file to the appropriate location.

dbuenzli:
> We don't want to copy over a css file every time we link something.

But you only link once no ? I think that between all the file you'll generate once you do the indexing of values, copying a `css` file there will be negligible and the above has the advantage of making things simple by default.

lpw25:
> But you only link once no ?

You link once per module (the compilation analogy breaks down a little there).

> negligible and the above has the advantage of making things simple by default

I'm not keen on compilation chains which can write the same file at different stages. It can screw up parallel build, and is generally hard to express to a build system. I'd rather that `odoc` was precise than convenient, the convenience is meant to be provided by tools like `opkg` or by build rules within a build system.

dbuenzli:
[...]

Issue #89, opened 2017-10-21
Add a `--fragment` option to `odoc html`


dbuenzli:
That outputs only the markup in the `body`, see https://github.com/ocaml-doc/odoc/issues/61#issuecomment-308731899. This is needed for https://github.com/dbuenzli/odig/issues/19 and https://github.com/dbuenzli/odig/issues/18.

trefis:
Currently `odoc html foo.odoc` will output as many html files as there are (sub)modules in the unit `Foo`. What would you like the output to be when `--fragment` is passed?

dbuenzli:
Each file with only the internal markup.

dbuenzli:
Any news on that front ?

rizo:
@dbuenzli I think I understand _what_ needs to be done but I'm not 100% sure _why_ this needs to be done. I did read all linked discussions but some comments seem a bit incomplete or outdated.

So please bare with me while I'm trying to understand this feature.

Here's the *what*:

- Add the `--fragment` option to `odoc html`.
- It will represent exactly the same HTML tree as the one generated currently.
- The same `--output-dir` will be used to save the files.
- Only the content of the `body` tag will be present (so no `head`, no CSS links. _etc_).

Here's my understanding of *why*:

- We want to add HTML items (like headers, sidebars, footers?) around the content.
- This will allow allow us to effectively inject the HTML tree into existing web templates.

Am I missing something? If the above is correct, I don't really understand how the complementary HTML items will be added. Will people have to provide a "chrome" file (as you called it) to `odig` during doc generation? Will this somehow impact the existing template (_e.g._ we already have a sidebar)?

dbuenzli:
> Am I missing something? If the above is correct, I don't really understand how the complementary HTML items will be added.

Yes. However now that I'm thinking about we nowadays have `.mld` pages I may be able do without this, though I didn't look at how .mld generation works yet. 

At least for customized package landing pages, I think I can simply take the package's `$DOCDIR/$PKG/odoc-pages/index.mld` create a new mld document which surrounds that content with the `odig` chrome (that is package information as seen for example [here](https://docs.mirage.io/alcotest/index.html)) and render the result as the index page. That would do without `--fragment`.

[...]

Issue #59, opened 2017-01-11
Support for package documentation assets


dbuenzli:
In https://github.com/dbuenzli/odig/commit/90f8660f2b386f0707ea4148778bf31cf6d50050 I have implemented a first shot at support for documentation assets (used by Vg since this commit https://github.com/dbuenzli/vg/commit/e01ae8eb6214725c8aae0a3fd5a001787fc89782) . 

I do however think that `odig` is the wrong place to standardize this and that the scheme should be in `odoc` (since without support in the ocamldoc language it influences the links you make in your docs) and be simply driven by `odig`. As we discussed in https://github.com/dbuenzli/odig/issues/6, unless something happens quickly at the ocamldoc level I think we should go for something straightforward, usable now. 

One idea would be to simply allow to do what `odig` does at the moment by adding an `--assets-dir DIR`  option to the `html` command whose semantics is to simply copy over whatever is in `DIR` in an `_assets` directory at the root of the package HTML after having generated the HTML. However though that may be a bit build system unfriendly, an `odoc asset --pkg PKG file` that copies `file` over to `_assets` would be more.

Alternatively we could do nothing until there is support in the ocamldoc language/odoc and rely on `odig`'s scheme meanwhile.

dbuenzli:
This issues becomes more acute now that we have the manuals (#61).

trefis:
Ping.

I've just hit this, as I was thinking about how a merlin manual (currently being written) might render on the new v3.ocaml.org
It seems *pretty important* to have this.

trefis:
(And indeed https://v3.ocaml.org/p/vg/0.9.4/doc/Vg/index.html#paths vs https://erratique.ch/software/vg/doc/Vg/index.html#paths )

PR #154, opened 2018-06-21
Add theme URI option

rizo:
This attempts to improve https://github.com/ocaml/odoc/issues/4 and enable https://github.com/dbuenzli/odig/issues/21.

More concretely the new `--theme-uri=URI` option in `odoc html` will change the path used to locate theme assets (currently `odoc.css` and `highlight.pack.js`).  This overrides the default behaviour in which the assets are looked up in the output's parent directory. If the option is _not_ provided `odoc` behaves without changes.

Additionally the current assets are installed as the "default" theme under `$(opam var odoc:share)/odoc-theme/default` (as suggested by https://github.com/dbuenzli/odig/issues/21#issuecomment-397427905).

To better illustrate what it does, consider this example:

```
$ odoc compile --package=fmt _build/src/fmt.cmti
$ odoc html _build/src/fmt.odoc \
      --theme-uri=$(opam var odoc:share)/odoc-theme/default \
      --output-dir=/tmp/odoc-output/
```

The generated HTML files in `/tmp/odoc-output/fmt/Fmt/` have the following asset references:

```
<link rel="stylesheet" href="/Users/rizo/.opam/4.06.1/share/odoc/odoc-theme/default/odoc.css">
<script src="/Users/rizo/.opam/4.06.1/share/odoc/odoc-theme/default/highlight.pack.js"></script>
```

## Notes and Questions

- Should the option be called `--css-uri` or `--theme-uri`? I think `theme-uri` makes more sense.
- I'm not sure if `theme-uri` should be typed as an actual URI (with the `uri` package?) or as a path. Currently it has a basic validation only, let me know if that's unacaptable. @dbuenzli can you give examples of how you will use this?
- The existing asset management is left unchanged. Will this require any changes in the way dune handles assets?
- I updated the main HTML testing function to handle the theme option as that was the easiest solution. Should I invest more time in making the tests more focused and modular?
- The change does not apply to `mld` pages in `Html_page.from_mld`. Is this needed given that there's a deprecation note for the `--index-for` option?
- Currently the implementation assumes the existance of the `highlight.pack.js` file (and we might add more files later). Of course custom themes will not have it. Does it make sense to load `odoc.js` that would bundle all JavaScript dependencies? Images and fonts can be relatievely referenced from CSS so those are not an issue.

(CC @dbuenzli @aantron)


trefis:
I will do a review of this tomorrow.

(But on principle it looks awesome, and I'm excited! Thanks!)

dbuenzli:
Thanks @rizo a few comments.

> Should the option be called --css-uri or --theme-uri? I think theme-uri makes more sense.

It depends a bit. In my view of things everything for a theme should be defined through `odoc.css`, in which case it should simply be called `--css-uri` and only used to define the `href` of the stylesheet.

Now this `highlight.pack.js` file is news to me, I guess it's there to perform code highlighting in verbatim sections but in some sense that should be performed by `odoc` itself.  In any case I don't think it's a good idea for each theme to have to copy this file over. This means that an `odoc assets` command should be introduced that works like `odoc css` to copy these `.js` files over at the generation destination.

> I'm not sure if theme-uri should be typed as an actual URI (with the uri package?) or as a path.

Just a string is fine. Let's avoid pulling too many deps into `odoc`. A bonus point would be to try to detect a relative path and do the relativisation according to the location in the module hierarchy, but if that's too complex just document that this has to be an absolute uri for now.

> The existing asset management is left unchanged. Will this require any changes in the way dune handles assets?

I don't understand this question and what relates to `dune`. 

> Currently the implementation assumes the existance of the highlight.pack.js file (and we might add more files later).

I think it's better if 1) themes do not include any javascript files of any sort 2) odoc's pages does not rely too much on `js` files.

> The change does not apply to mld pages in Html_page.from_mld. Is this needed given that there's a deprecation note for the --index-for option?

Not familiar with this but just make sure the thing applies to `.mld` generated pages.

dbuenzli:
> but if that's too complex just document that this has to be an absolute uri for now.

Actually this would not allow to move the generated docs or easily publish it at an arbitrary endpoint on the www so the relativisation stuff should somehow be done, but I bet `odoc` already has functions for doing this internally.

rizo:
(Sorry for such a late reply, I've been failing to find time to work on this unfortunately.)

> In my view of things everything for a theme should be defined through `odoc.css`, in which case it should simply be called `--css-uri` and only used to define the `href` of the stylesheet.

Putting JavaScript files aside, I agree that everything for a theme should be defined _through_ `odoc.css`, but the theme itself _is_ more than just the CSS file. Themes can have fonts and potentially images. I think using `--css-uri` would make sense if the URI pointed directly to a CSS file and thus could be literally used as the `href` of the stylesheet. Otherwise I think `--css-uri` is a bit misleading. Does this make sense?

> I guess it's there to perform code highlighting in verbatim sections but in some sense that should be performed by odoc itself.

I agree that JavaScript files shouldn't belong to the theme. In this particular case `highlight.pack.js` adds CSS classes that can be customised in the CSS anyway.

> This means that an `odoc assets` command should be introduced that works like `odoc css` to copy these `.js` files over at the generation destination.

The `odoc css` command already does that: it copies both `odoc.css` and `highlight.pack.js`. In this case again I think the name is confusing, it should be renamed to `odoc assets` as you suggested.

Actually, is there a particular reason why the asset copying isn't done as part of the `odoc html` command? They are known and referenced from the generated HTML, why not copy them over to the `output-dir` at the same time?

> I don't understand this question and what relates to dune.

I wasn't sure whether this change affects dune in some way. For example, does it make sense for dune to use the `--theme-uri` and thus support configurable themes too? I don't consider this important, to be honest.

rizo:
@trefis Thank you  for your review!

I followed your suggestion and added support for relative/absolute URIs with a dedicated type. Relative URI resolution should work as expected. I decided to use `Relative "./"` and not `Relative ""` because the empty URI already has a [reserved meaning](https://tools.ietf.org/html/rfc3986#section-4.4). In this case `"./"` refers to the HTML output directory (see manpage description).

I wasn't sure what was the best place to defined the `uri` type, let me know if `Html.Html_tree` is not the best place for it.

aantron:
@rizo The main reason assets were under `odoc css` is probably that `odoc html` gets run a large number of times when building the docs for a project. `odoc css` is run once.

rizo:
Oh, of course. My bad, I was thinking about it in the context of a single project.

aantron:
AFAIK this is true for a single project as well, as `odoc html` is run once for every page that is output.

rizo:
Based on what was discussed about JavaScript files, the last commit uses `output-dir` instead of `theme_uri` to find the `highlight.pack.js` file.

***

The complete list of steps to have a working HTML tree with a custom theme is:

```
$ odoc compile --package=fmt _build/src/fmt.cmti
$ odoc html _build/src/fmt.odoc \
      --theme-uri=$(opam var odoc:share)/odoc-theme/default \
      --output-dir=/tmp/odoc-output
$ odoc css --output-dir=/tmp/odoc-output
```

There're a couple of problem with this:

- Technically the last command is only needed to copy the JavaScript file, but currently it also copies the `odoc.css` file.
- The `--theme-uri` option will, as expected, override the `odoc.css` reference, meaning that the copied `odoc.css` file will **not** be used.
- Dune [requires](https://github.com/ocaml/dune/blob/f7a4e51/src/odoc.ml#L190) the `odoc.css` file to be copied to `--output-dir`.

I think what @dbuenzli [described](https://github.com/ocaml/odoc/pull/154#issuecomment-399193455) as a new `odoc assets` command could partially solve this issue. The `odoc css` command could be changed to _only_ copy the `odoc.css` file. A new `odoc assets` command could be introduced to copy the JavaScript file but _not_ the CSS file.

But unfortunately even that only partially solves the problem. Copying only the CSS file is not enough because themes may contain files too (like fonts or images). Currently odoc's default theme is just the `odoc.css` file but I would like to suggest adding custom fonts to it soon. If that happens the `odoc css` command should copy those too. And should probably be renamed to `odoc theme` or `odoc copy-theme`...

Essentially the `odoc css` command should provides the default content equivalent to the custom content found in `--theme-uri`.

Sorry if this is confusing, but if we want to have custom themes I think it's worth making them work consistently everywhere and not just for `odig`.

aantron:
IMO, it's too early for custom theme support in odoc.

1. I'm not aware of any "organic" demand for custom themes. @dbuenzli, was this requested by someone for odig?
2. We still have a lot of things to work out in the presentation of odoc. While we should keep future support for custom themes in mind, it seems like we should focus on getting the default presentation decent, rather than supporting customization at this point.

dbuenzli:
@aantron I don't think it's too early, the mecanism will have to be done at a certain point and I don't think there's anything complex here. I nowadays can only work on `odig` on my free time which I happen to have some now and would like to solve this long standing issue during the current rewrite without hacks if possible. Since @rizo is willing to help on this and @trefis to review I don't think there's a problem in designing this now. More specifically:

Regarding 1. I had requests in the past to provide a dark theme for `odig` and support for custom styling the whole presentation for publishing on the web.

Regarding 2. we had done a great deal trying to stabilize a markup with @trefis, now apparently at some point, for reasons unknown to me, some people decided it was a good idea to redo the whole thing again rather than incrementally improve the existing. Things have to move forward at a certain point, it's at least the fourth iteration on the documentation markup I have had to handle and debug during the past five years (that's of course not your fault) and I have to admit that it became a little bit tiring with the time.

So @rizo I'd suggest the following:

1. Keep the `--theme-uri URI` option and the way it works, this is the good name and a forward looking mecanism for potential later expansion (e.g. allowing theme to have a custom `js` file). For now the only thing it guarantees is that `URI/odoc.css` will be the stylesheet link in the generated pages.
1. Remove the `odoc css` command. 
2. Add the `support-files` command (`assets` we might want to keep for something else) which by default copies the essential files (`highlight.js` for now) *and* the default theme (`odoc.css` and whatever that will be made of).
3. Add a `--without-theme` option to `support-files` that copies the essential files but not the default theme.

I think that this should solve all the problems.

rizo:
> IMO, it's too early for custom theme support in odoc. (...) We still have a lot of things to work out in the presentation of odoc.

I agree with the sentiment that custom themes are not essential at this stage and other things have higher priority. On the other hand, if we can come up with an acceptable model for themes, it's certainly worth completing this PR.

> now apparently at some point, for reasons unknown to me, some people decided it was a good idea to redo the whole thing again rather than incrementally improve the existing. (...)  it's at least the fourth iteration on the documentation markup (...)

This concerns me because themes (CSS really) directly depend on HTML structure. Maybe in the future themes could be limited to some sort of color palettes or something similar.

@dbuenzli Replying to your individual comments:

1. I'm glad we agree about `--theme-uri`.
2. Yes, I think it makes sense to deprecate `odoc css`.
3. I'm not sure about naming here. The only other kind of "assets" I can think of for odoc is user-supplied images referenced from docs. To me `support-files` is almost a synonym for `assets`, I think it will be confusing to have both commands in the future. In reality all these files are assets (see below).
4. Yes, optionally excluding the theme files will work.

### Assets

It is a common practice to call supporting files for HTML pages "assets". Those typically include CSS/JS files and images. What kind of assets do we have in `odoc`?

1. **Theme assets** (CSS file, fonts, _etc_). These can be either:
    - The default theme copied over to `output-dir` or;
    - A custom theme linked with `theme-uri`
2. **Private assets** (JS files like `highlight.pack.js` and potentially a base CSS file). These are private to `odoc` (_i.e._ users shouldn't be aware of them) and should _always_ be copied.
3. **User assets** (probably just images). Files that don't belong to `odoc` at all and are referenced from `mli`/`mld` documentation. These should to be tracked and _always_ copied to `output-dir`.

Based on this, my suggestion is to add the following command:

```
$ odoc assets [--without-theme] --output-dir=DIR
```

This will always copy private and user assets but might exclude the theme, in case `--theme-uri` was provided in `odoc html`. What do you think?

dbuenzli:
The issue tracking what you call "user" assets and what I'd call documentation assets or simply assets is here https://github.com/ocaml/odoc/issues/59 (follow the links for a bit further discussion). 

I'd prefer to keep `assets` for handling documentation assets;  that is unless you find a better name for the latter. There are quite a few things one might want to do there beyond simply copying them (e.g. extracting asset references from `odoc` files), let's reserve the name for that.

`odoc support-files` seems fine to me and quite different from documentation assets, it's the files the generated html actually *needs* in order to render them correctly.

dbuenzli:
Btw. @rizo feel free to propose better names, it's just that we will have to distinguish these things as a matter of fact (and I personally don't think user asset is a good name, `odoc` handles documentation, not users).

bluddy:
How about just odoc html-init?

rizo:
@dbuenzli "User assets" as opposed to "system assets", where "system" is `odoc` in this case. But never mind, I agree it's not a great name (specially if I have to explain it...). As long as "assets" are only used for documentation assets I think there will be no confusion with "support files".

Thus as discussed, the last commit:

1. Adds the `support-files` command with the `--without-theme` option.
2. Makes `css` an alias for `support-files` leaving a deprecation note in the description.
3. Renames the `assets` module to `support_files`.

@dbuenzli Do you think this implementation is now sufficient for your needs?

@trefis Let me know your thoughts on this, if you have time.

rizo:
> How about just odoc html-init?

@bluddy if you're referring to the `odoc support-files` command, I think it's more like `html-finish`. The HTML tree is already present in output dir, the last missing step is copying the support files.

dbuenzli:
@rizo Thanks ! I think this should be right.

src/odoc/bin/main.ml
```diff
@@ -45,19 +45,19 @@ let odoc_file_directories =
     "Where to look for required .odoc files. \
      (Can be present several times)."
   in
-  Arg.(value & opt_all convert_directory [] @@
+  Arg.(value & opt_all convert_directory [] &
```
pmetzger: Is this the `&` operator from pervasives? If so, it's now supposed to be `&&`...

rizo: No. That's the [function application defined in Cmdliner](http://erratique.ch/software/cmdliner/doc/Cmdliner.Arg.html#VAL(&)). It's equivalent to [`(@@)` in Pervasives](https://docs.mirage.io/ocaml/Pervasives/index.html#val-(@@)). AFAIK Cmdliner introduced the `&` operator before `@@` was added to Stdlib.

The reason I changed `@@` to `&` was because the current usage is inconsistent and most people use `&` with Cmdliner.

trefis:
@rizo I'll try to get back to you before the end of the week!
Thanks for the patience (and for all the work you're doing)!

src/odoc/bin/main.ml
pmetzger: So, you can see my confusion. Perhaps we should encourage everyone to go into the other direction given that it really means `@@`?

bluddy: I agree. We should move away from operators that duplicate the functionality of later standard operators. I think `&` should probably be deprecated in Cmdliner.

rizo: @pmetzger @bluddy while I tend to agree with you, in this case prefer following the current established convention. Both `&` and `@@` are mixed in current code which is more confusing than either of them exclusively IMO.

Feel free to open an issue for this on the Cmdliner repo (CC @dbuenzli).

rizo:
@trefis I'm not in a hurry and can work on other things meanwhile. Take your time :)

src/odoc/bin/main.ml
pmetzger: We can handle it by moving everything to `@@` surely, yes?

lpw25: You'll just confuse everyone who's used to cmdliner (which is most OCaml programmers). It's DSL is well established, and most people just know that you put `&` there and don't necessarily know that it is apply.

pmetzger: Gotcha.

aantron:
I suggest not installing themes in the opam switch. I don't think users will want to put their custom themes into switches because...

- the switch `share` directory is obscure,
- is not version-controlled,
- switches are ephemeral,
- the installation is per-switch, not per-project, and will have to be repeated for each switch.

That leaves only built-in themes. Rather than having a `share` directory containing only built-in themes, I think it's better to inline those into the `odoc` binary, as done now. See also #155 about improving how we inline assets.

dbuenzli:
@aantron I don't understand your objections. This scheme allows themes to be distributed by piggy backing on the opam distribution infrastructure whose packages can perfectly be versioned, either formally or informally via a pin.

`share` is not obscure it is where these kind of files go in a standard unix install. `odig` works both with `opam` and without `opam` and the proposed scheme is the most natural way of operating in a system install. 

If you want to built-in the default theme in the `odoc` binary, why not, but personally I think it's better if these things are handled in a uniform manner -- at least it makes `odig`'s job easier.

aantron:
I see the point about distributing themes as opam packages, I had missed that idea.

However, I think it's valuable to allow odoc to work independently of opam switches, hence the desire to have the default themes built in. I don't think supporting themes as opam packages should cause odoc to require opam, or gain any filesystem dependencies.

We could install the default theme in `share` for reference and not use it.

The complexity and conflicting considerations make me think, again, that we are ahead of time trying to design this.

dbuenzli:
> However, I think it's valuable to allow odoc to work independently of opam switches,

That will be hard in any case, anything that depends on `compiler-libs` makes that difficult in practice and as far as I know the `opam` relocation story remains spotty for the time being; I wouldn't worry too much about that for now.

That being said there's nothing hard about runtime configuring an `odoc` with an `ODOC_SHARE_DIR` environment variable that is discovered in a reasonable fashion when unset, see for example how [`omod`](https://github.com/dbuenzli/omod/blob/e6e9547c9f129e7bc3e01017f6940d2351cd2da0/src/omod_bin.ml#L453-L469) handles this (the `odig` rewrite on the way works in a similar fashion to find its `lib, `doc` and `share` directories).

> We could install the default theme in share for reference and not use it.

Why not but as far as I'm concerned I would eschew the binary embedding altogether.

aantron:
> anything that depends on `compiler-libs` makes that difficult in practice

I'm not sure which part this refers to. The build does need `compiler-libs`, but the built binary `odoc` is the only artifact of package `odoc`, and the binary `odoc` is currently fully self-contained AFAIK and has `compiler-libs` statically linked in. So, it does not depend on opam, or anything else in the filesystem.

dbuenzli:
> I'm not sure which part this refers to.

It means that it's only able to deal with the `cmt` files of a particular compiler version. Most of your switches have different compiler versions.

trefis:
I believe the discussion about the relation to opam, and whether to bundle support files in the executable or not, although it's interesting, is getting us slightly off track.

This PR seems like a clear improvement to me, and I believe the implementation is fine.
So I'm in favor of merging it ASAP.

@aantron: do you want to take care of merging, or should I just click the big green button?

aantron:
@trefis I'm ready to do the merge whenever, just waiting for the go-ahead from @rizo :)

trefis:
In the end I didn't wait (the only thing that might have changed is that some commits might have been squashed).

@dbuenzli: this is now on master, enjoy!

And thanks again @rizo!

rizo:
(Sorry for the delay – been a bit busy dealing with some minor health issues...)

@trefis Thank you for the review! :-)

@dbuenzli Let me know if this feature is enough to unlock your work in odig, or if there's anything else you need my help with.

@aantron I understand the portability concerns, if you think it's important we can open a separate issue to discuss this. Honestly if it's not important _now_, I'd follow @dbuenzli's suggestion and "eschew the binary embedding altogether" as it avoids the need for https://github.com/ocaml/odoc/issues/155 and makes theme handling easier.

rizo:
One last note about portability: supplying a custom `theme-uri` makes the content of `output-dir` _not_ portable (by design). I think it's an understandable limitation given the gain in flexibility.

An alternative approach could be to use `theme-path` instead and copy that directory to `output-dir` turning the generated output completely self-contained and portable.

dbuenzli:
Thanks all. @rizo if you want to help the ability to generate html snippets from `.mld` files would be nice (https://github.com/ocaml/odoc/issues/89).

Issue #129, opened 2018-04-17
Reason/OCaml syntax switching


aantron:
This is about outputting docs in Reason syntax.

Nothing needs to be done at the parsing end, because odoc reads docs from parsed ASTs (after the OCaml or Reason parser is done running), so the source syntax doesn't affect odoc.

https://discuss.ocaml.org/t/1841/13, https://discuss.ocaml.org/t/1841/47. The Ocsigen website has an example of switching (http://ocsigen.org/lwt/3.2.1/api/Lwt). cc @Drup

ryyppy:
Hey 👋 
I am currently investing some time into this and try to submit a PR soon!

Hope this is okay?

rizo:
The way I see this feature implemented and run is:

- Add the `--syntax=ml|re` option to `odoc html`.
- Implement an alternative HTML tree encoder for Reason.
- Generate both HTML trees.
- Have a button that changes the directory between `ml` and `re` trees.

@aantron What do you think about this approach?

Drup:
Yeah, or you could just compile refmt to js and do the conversion of the fly.

rizo:
@Drup I think it would be harder to achieve that for the markup. For examples, yes. But the markup is not purely textual, it has links and styling.

aantron:
> Hope this is okay?

It is, and quite welcome :)

@rizo, I think that works. My original thought was to generate both OCaml and Reason output side-by-side in each HTML file, and use CSS to hide/show the right one. But this is probably bad for accessibility, navigation, etc., so separate output trees are likely to be better.

aantron:
@ryyppy added Reason syntax output in #156. Right now, odoc generates one of either ML or Reason syntax each time it is run. What's left on this topic is to decide whether we want to always generate both.

ryyppy:
@aantron @ostera So my idea to complete all of this was to basically generate both versions, but always hide one unselected version via CSS. By providing the `--syntax` flag, you will be able to decide on a default language to display whenever you load the docs for the first time.

As soon as this is done, we can add a JS-driven toggle for both syntaxes in the doc header.

lpw25:
Not hard to too but can we make sure to have an option that only generates one or the other rather than both. Personally, I think that setting should be the default, but that's less important than having the option.

rizo:
[...]

PR #156, opened 2018-07-09
Add Reason support (WIP)

ryyppy:
Tracking Branch for #129 

Things that are being worked on:

- [x] Add `--lang` flag for html command ~~(currently not building b/c of type error)~~
- [x] Copy test cases to reason syntax & implement as much Reason syntax generation as possible
- [x] Apply `refmt` on code snippets inside block comments
- [x] ~~Use location info to determine `ml` or `re` formatting for `Code_block`s~~
- [x] ~~Provide mechanics to configure refmt's heuristic constructor list (prevent `explicit_arity` attributes) -> use `ODOC_REFMT_HEURISTICS` env variable~~

Test Cases:

- [X] Val
- [x] Markup
- [x] Section
- [x] Module
- [x] Interlude
- [x] Include
- [x] Mld
- [x] Type
- [x] External
- [x] Functor
- [x] Class
- [x] Stop


ryyppy:
The most intrusive changes are in the `src/html/documentation.ml` module.
I introduced a `~from_lang` / `~to_lang` parameter for block_elements which give context on which parser / printer logic to use. Rest of the work is dedicated to the signature markup rendering part.

src/html/markup.mli
```diff
@@ -19,19 +19,36 @@
 module Html = Tyxml.Html
 
 
-
 val keyword : string -> [> Html_types.span ] Html.elt
 
-module Type : sig
-  val path :
-    [< Html_types.span_content_fun ] Html.elt list ->
+module ML : sig
```
ryyppy: I introduced namespaces for Markup... common Markup is still in the toplevel of the module, e.g. `keyword`

src/html/markup.mli
```diff
@@ -19,19 +19,36 @@
...
+  (** "->" with a non breaking hyphen, styled as a keyword. *)
+
+  val label : Model.Lang.TypeExpr.label -> [> `PCDATA ] Html.elt list
+  (** For optional arguments adds a word joiner between the question mark and the
+      label. *)
 end
 
-val arrow : [> Html_types.span ] Html.elt
-(** "->" with a non breaking hyphen, styled as a keyword. *)
+module RE : sig
```
ryyppy: It's probably worth introducing a common module interface to make the signature more DRY

src/html/to_html_tree.ml
```diff
@@ -125,25 +125,25 @@ struct
     -> Model.Lang.TypeExpr.t -> ('inner, 'outer) text Html.elt list
   = fun ?(needs_parentheses=false) t ->
     match t with
-    | Var s -> [Markup.Type.var ("'" ^ s)]
-    | Any  -> [Markup.Type.var "_"]
+    | Var s -> [Markup.ML.Type.var ("'" ^ s)]
```
ryyppy: I guess this kind of pattern can be solved with first class modules as well? dropping the namespace and just reify a specific module implementation for each language? This solution was quicker and the least intrusive

src/html/to_re_html_tree.mli
```diff
@@ -0,0 +1,22 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *)
+
+val compilation_unit : ?theme_uri:Html_tree.uri -> Model.Lang.Compilation_unit.t -> Html_tree.t
```
ryyppy: This should probably just reuse the module definition of `to_html_tree.mli`... I was not too sure how to do this here, also I thought I would let this be for now and redesign this in a more unified way

src/html/utils.ml
```diff
@@ -10,3 +10,25 @@ let rec list_concat_map ?sep ~f = function
     match sep with
     | None -> hd @ tl
     | Some sep -> hd @ sep :: tl
+
+(**
```
ryyppy: I left this code in here, since there were some discussions on how to introduce `refmt` without coupling it to the odoc core

src/model/location_.ml
```diff
@@ -20,6 +20,9 @@ let at : span -> 'a -> 'a with_location = fun location value ->
 let location : 'a with_location -> span = fun {location; _} ->
   location
 
+let file_ext : span -> string = fun span ->
```
ryyppy: Required to be able to determine the `from_lang` attribute... if `.mli` -> OCaml, if `.rei` -> Reason

src/html/html_tree.ml
```diff
@@ -19,8 +19,14 @@
 module Html = Tyxml.Html
 module Paths = Model.Paths
 
+type lang = OCaml | Reason
```
ryyppy: I introduced the language type in here, not sure if this is the right module

src/odoc/html_page.ml
```diff
@@ -16,7 +16,17 @@
 
 open StdLabels
 
-let from_odoc ~env ?theme_uri ~output:root_dir input =
+let to_html_tree_page ?theme_uri ~lang v =
+  match lang with
+  | Html.Html_tree.Reason -> Html.To_html_tree.RE.page ?theme_uri v
```
ryyppy: I don't like this... I felt like we should use something like first-class-modules, but I was not entirely sure how to do this the right way... feedback welcome!

src/odoc/html_page.ml
```diff
@@ -16,7 +16,17 @@
...
+  | Html.Html_tree.OCaml -> Html.To_html_tree.ML.page ?theme_uri v
+
+let to_html_tree_compilation_unit ?theme_uri ~lang v =
+  match lang with
+  | Html.Html_tree.Reason -> Html.To_html_tree.RE.compilation_unit ?theme_uri v
+  | Html.Html_tree.OCaml -> Html.To_html_tree.ML.compilation_unit ?theme_uri v
+
+let from_odoc ~env ?(lang=Html.Html_tree.OCaml) ?theme_uri ~output:root_dir input =
```
ryyppy: The CLI has an optional `lang` parameter, which defaults to `OCaml`

src/html/jbuild
```diff
@@ -3,4 +3,4 @@
 (library
  ((name html)
   (preprocess (pps (bisect_ppx -conditional)))
-  (libraries (model tyxml))))
+  (libraries (model tyxml reason))))
```
aantron: Do we still have a dep on `reason` here? I think the code that uses `reason` is commented out, but not sure if I missed some.

src/html/to_html_tree.ml
ryyppy: I don't know how you would factor this out... I was thinking of using `include` statements, but most bindings of this module are recursively `and`ed... not sure if you can override specific functions then?

aantron: IIRC, in the diff between `to_html_tree.ml` and `to_re_html_tree.ml`, the skeleton of the functions is the same, and only some details change (what delimiter to use in records, variants, whether there is a semicolon, etc). So, we would factor out these details into `ML` and `Re` modules, and turn `To_html_tree` into a functor that contains the remaining common code.

test/html/expect/ml/markup.html
```diff
@@ -1,5 +1,5 @@
...
...shorthand link: <a href="#">#</a>. The text is also filled in by odoc in this case.</p><p>This is a reference to <a href="index.html#val-foo"><code>foo</code></a>. References can have replacement text: <a href="index.html#val-foo"><span>the value foo</span></a>. Except for the special lookup support, references are pretty much just like links. The replacement text can have nested styles: <a href="index.html#val-foo"><span><b>bold</b></span></a>, <a href="index.html#val-foo"><span><i>italic</i></span></a>, <a href="index.html#val-foo"><span><em>emphasis</em></span></a>, <a href="index.html#val-foo"><span>super<sup>script</sup></span></a>, <a href="index.html#val-foo"><span>sub<sub>script</sub></span></a>, and <a href="index.html#val-foo"><span><code>code</code></span></a>. It's also possible to surround a reference in a style: <b><a href="index.html#val-foo"><code>foo</code></a></b>. References can't be nested inside references, and links and references can't be nested inside each other.</p></header></section><section><header><h2 id="preformatted-text"><a href="#preformatted-text" class="anchor"></a>Preformatted text</h2><p>This is a code block:</p><pre><code class="ml">let foo = ()
```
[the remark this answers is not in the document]
ryyppy: I introduced a class attribute for each `pre` block (see https://github.com/ocaml/odoc/pull/156/files#diff-41083926a76829c938b1f88df53c48d7R274)

aantron:
Looks like the remaining problem is a [failure on 4.03](https://travis-ci.org/ocaml/odoc/jobs/413650273#L807), because [`Filename.extension`](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Filename.html#VALextension) was only added in 4.04. Since odoc depends on fpath, we can probably deal with this by replacing `Filename.extension` with some combination of [these functions from fpath](https://github.com/dbuenzli/fpath/blob/f679f231ee994e30ca2f5f2b327b3750e9bdf82d/src/fpath.mli#L380).

aantron:
You could also use [`Filename.check_suffix`](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Filename.html#VALcheck_suffix) instead of extracting the extension.

test/html/expect/ml/markup.html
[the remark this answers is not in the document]
aantron: Thanks, I had seen that but not made the connection. I gave myself a TODO to get these expected strings indented nicely, so that the diffs are actually helpful.

aantron:
@diml, @rgrinberg, We'd like to be able to choose, through Dune, whether odoc outputs OCaml syntax, or Reason syntax. This is controlled by passing `--lang=re` or `--lang=ml` to `odoc html` (`--lang=ml` is the default). What do you recommend for achieving this?

In the future, we might want to generate both outputs, and give users the option to switch between them. I guess Reason output wouldn't require support from Dune at that point.

rgrinberg:
@aantron you should open a dune ticket to discuss this. There's a couple of options here. The most natural one is to create separate targets for reason/ocaml documentation and then use 2 aliases to control them. For example, `@doc` and `@reason-doc`. Would you like to take a stab at implementing this in dune?

diml:
What about the following simple rule: if the original source file is a .mli, output in OCaml syntax, if it is a .rei, output in reason syntax?

PR #157, opened 2018-07-09
Fix markup.ml dev dependency

ryyppy:


aantron:
Thanks!

trefis:
I don't know the whole story, but it seems like the library is only used for a "test" executable.
If that's the case, shouldn't the dep be marked as a test-dep in the opam file?

ryyppy:
@trefis I complied to the format as stated here: https://github.com/ryyppy/odoc/blob/1590f17fe100133af885f53b73b0a4cd4a125e40/odoc.opam#L27

How does one define a dev dependency in opam files?

trefis:
Ah, I hadn't seen this.
In merlin's opam file I have [this](https://github.com/ocaml/merlin/blob/master/merlin.opam#L23), and I believe it works.

aantron:
@trefis, @rypppy, there is no practical way (AFAIK and IIRC) to define a dev dependency in an opam file today. The `{test}` deps are installed only if you do `opam install --build-test`, which not only installs the `{test}` deps, but runs the tests during opam install. The biggest problem with `--build-test` is that it runs the tests recursively for deps of the package you are installing, which often leads to false and irrelevant test failures due to poor maintenance of test suites. My experience with `{test}` is that I tried it for a long time in Markup.ml itself, and Lambda Soup, and it doesn't "work."

So, the way we define test deps is we put a comment in the opam file, and make sure to remove the deps manually when sending the opam file to the opam repo.

aantron:
Oh, the other problem with `--build-test`, due to it being recursive, is that it leads to a large expansion in the dependencies that need to be installed, because it installs all the `{test}` deps of the deps of your package.

Some packages configure themselves for your system automatically, but their test suites assume a configuration, or assume support for all features that need to be tested, etc.

rizo:
Would this be solved by having an option to disable recursive test execution in OPAM?

Seems like `--build-test` is deprecated. OPAM's manual is not very clear about this IMO. Here's what `opam install --help` says:

```
       -t, --with-test, --build-test
           Build and run the package unit-tests. This only affects packages
           listed on the command-line. The --build-test form is deprecated as
           this also affects installation.
```

So it should _only_ affect the explicitly listed packages, right?

aantron:
@rizo I think you have a newer version of opam. The problems of `--build-test` have been discussed and are at least partially addressed in 2.0. My `opam install --help` says:

```
       -t, --build-test
           Build and run the package unit-tests. This is equivalent to setting
           $OPAMBUILDTEST to "true".
```


This is opam 1.2.2, which is what most developers will have at this point.

> So it should *only* affect the explicitly listed packages, right?

I believe so, but, again, we have to wait for opam 2.0 to be released, and get into system package managers, etc.

rizo:
Yes, sorry, I understand that this is specific to OPAM 2.0. Just wanted to make sure this is a know issue and it seems like it's already a solved one too.

trefis:
Thanks for the explanation!

PR #163, opened 2018-08-10
Rename --lang to --syntax parameter

ryyppy:


ryyppy:
^ No idea why the tests are failing... seems like to be a timeout in the registry?

aantron:
Yeah, probably the PPA that contains opam 1.2 is not working right now for some reason.

aantron:
Thanks!

Issue #27, opened 2016-09-26
`odoc html` allow to specify the base URI for resolving links to external identifiers.


dbuenzli:
When you develop your package locally you'd like to be able to generate its documentation in its say `_build` directory but also to check the accuracy of external links against the current switch docs.

For this we need an option to specify the base URI to use for resolving links to other packages (typically this base uri will be a relative file path to `$(opkg cache path)/odoc`).

aantron:
No, I didn't explicitly consider this. It should be possible to add with some effort, though. I think my refactoring was mostly orthogonal to this issue.

dbuenzli:
Ok thanks for the answer. I don't think this is a complex problem as long as you carefully isolate the functions that generate links. It would be nice to have though so that you can properly test inter-package links.

trefis:
Things don't seem to have changed to much in that area during the reorganization, and it should indeed be relatively easy to change from what I can see.

rizo:
I'm considering implementing this feature. What would be the best place to start? Any pointers would be very appreciated as, given the current implementation, I don't see a clean and easy way to add override the base URL. The code base seems to pervasively assume that there's a root package.

This is needed to implement HTML fragments that don't belong to a particular package and can set the root to the docsets path for cross-reference resolution. See https://github.com/ocaml/odoc/pull/166#issuecomment-415421238 for more details.

rizo:
@dbuenzli Can we decide the interface for this feature? I might be able to implement partial support for fragments.

It seems like if I intercept the `Html.Html_tree.Relative_link.Id.href` function call during HTML generation I can supply a custom base URI.

What about?

```
$ odoc html-fragment --root-uri=URI -o FRAGMENT.html -I INCLUDE_DIR PAGE.mld
```

This is for the `html-fragment`, we could add it later to other commands too.

rizo:
As a side note (and to remind myself), it would be useful if this option could be integrated with dune when it's done.

dbuenzli:
Great ! I'm a little bit unsure about the `--root-uri` name though, when I read the cli it seems to me that it defines the root of the generated document and that's not the case, it defines the URI w.r.t. to which cross-references are resolved in the resulting document. Maybe `--xref-root-uri` or `--xref-base-uri` ?

rizo:
[...]

PR #166, opened 2018-08-22
Add html --fragment option and Odoc.Html_fragment module.

rizo:
A very straightforward solution for https://github.com/ocaml/odoc/issues/89.

This adds an option to generate HTML fragments for documentation pages contained in odoc files. Compilation units are not currently supported as they're both less important and harder to implement.

The full command sequence to convert an `mld` file into an HTML fragment file is:

```
$ odoc compile --package=PKG -o page-PAGE.odoc PAGE.mld
$ odoc html --fragment -o PATH page-PAGE.odoc
```

This will generate a file located at `./PKG/PAGE.html` containing only the HTML markup without the document wrapper.

Note that the `--package` option is required to resolve cross-references.

### DIFF
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -177,13 +177,19 @@ module Html : sig
 end = struct
 
   let html semantic_uris closed_details _hidden directories output_dir index_for
-        syntax theme_uri input_file =
+        syntax fragment theme_uri input_file =
     Html.Html_tree.Relative_link.semantic_uris := semantic_uris;
     Html.Html_tree.open_details := not closed_details;
     let env = Env.create ~important_digests:false ~directories in
     let file = Fs.File.of_string input_file in
     match index_for with
+    | None when fragment ->
+      Html_fragment.from_odoc ~env ~output:output_dir file
     | None -> Html_page.from_odoc ~env ~syntax ~theme_uri ~output:output_dir file
+    | Some _ when fragment ->
+      Printf.eprintf "ERROR: The --index-for option cannot be used with --fragment. \
+                      Please compile the mld file first.\n";
+      exit 1
     | Some pkg_name ->
       Html_page.from_mld ~env ~syntax ~output:output_dir ~package:pkg_name file
 
@@ -201,6 +207,12 @@ end = struct
                  be closed by default."
       in
       Arg.(value & flag (info ~doc ["closed-details"]))
+    in
+    let fragment =
+      let doc = "Produce an HTML fragment instead of a full document. \
+                 Currently only works with documentation pages."
+      in
+      Arg.(value & flag (info ~doc ["fragment"]))
     in
     let index_for =
       let doc = "DEPRECATED: you should use 'odoc compile' to process .mld \
@@ -224,7 +236,7 @@ end = struct
       Arg.(value & opt (pconv convert_syntax) (Html.Html_tree.OCaml) @@ info ~docv:"SYNTAX" ~doc ["syntax"])
     in
     Term.(const html $ semantic_uris $ closed_details $ hidden $
-      odoc_file_directories $ dst $ index_for $ syntax $ theme_uri $ input)
+          odoc_file_directories $ dst $ index_for $ syntax $ fragment $ theme_uri $ input)
 
   let info =
     Term.info ~doc:"Generates an html file from an odoc one" "html"
--- a/src/html/html.ml
+++ b/src/html/html.ml
@@ -1,4 +1,5 @@
 module Html_tree = Html_tree
+module Documentation = Documentation
 module To_html_tree = struct
   module ML = To_html_tree
   module RE = To_re_html_tree
--- a/src/odoc/html_fragment.ml
+++ b/src/odoc/html_fragment.ml
@@ -1,0 +1,26 @@
+
+let from_odoc ~env ~output:root_dir input =
+  let root = Root.read input in
+  match root.file with
+  | Page page_name ->
+    let page = Page.load input in
+    let odoctree =
+      let resolve_env = Env.build env (`Page page) in
+      Xref.resolve_page (Env.resolver resolve_env) page
+    in
+    let pkg_name = root.package in
+    let pkg_dir = Fs.Directory.reach_from ~dir:root_dir pkg_name in
+    Fs.Directory.mkdir_p pkg_dir;
+    let content = Html.Documentation.to_html odoctree.content in
+    let oc =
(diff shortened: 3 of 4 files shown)


rizo:
(I will add tests for this feature after making sure everyone's happy with the API.)

rizo:
Should we make the fact that the generated file is a fragment more explicit? The file path would be `PATH/PKG/fragment-PAGE.html`.

dbuenzli:
Thanks ! 

> Note that the --package option is required to resolve cross-references.

I guess something is needed to indicate where you are, however a package is not what you want here. What you'd like is to be able to specify how to access the root of the docset from the snippet. In my particular use case I'll use the generated fragment at the root of the docset *not* in a package. 

Somehow this issue may be related to https://github.com/ocaml/odoc/issues/27

> Should we make the fact that the generated file is a fragment more explicit? The file path would be PATH/PKG/fragment-PAGE.html.

I think `--fragment` should imply output on stdout and/or on a file you specify via `-o`.

rizo:
> I think --fragment should imply output on stdout and/or on a file you specify via -o.

@dbuenzli I considered that originally. I don't want to change how `odoc html` currently works. It assumes the input is an odoc file and the output is a directory.

In my [last commit](https://github.com/ocaml/odoc/pull/166/commits/1be66509d0486fa6c1e2cf2df6884bf552ea725c) I added a new `odoc html-fragment -o FRAGMENT.html PAGE.mld` command that translates the content of the page into a fragment. We can set the input/output files to stdin/stdout by default.

> I guess something is needed to indicate where you are, however a package is not what you want here. What you'd like is to be able to specify how to access the root of the docset from the snippet. In my particular use case I'll use the generated fragment at the root of the docset not in a package.

This makes sense in theory. In practice I have no idea how to do that. The code makes strong assumptions about the presence of a package root. In particular see [this comment](https://github.com/ocaml/odoc/pull/166/files#diff-97dcad485808b851bebe5a10cccac4d7R77) for an example of something non-obvious that took me an hour to figure out.

I feel like I'll need to invest a lot of time to understand how to properly implement https://github.com/ocaml/odoc/issues/27, which would certainly help in this case. I can start doing that, but meanwhile if you have any other ideas or suggestions for this feature, I'd be glad to hear them.

rizo:
To fully summarise the current status:

* Both `odoc html --fragment` and `odoc html-fragment` command invocations are supported with the behaviour described abode. I included both to test and discuss them, only one of them will be included in the final implementation.
* Cross-reference resolution requires a concrete root and, given the current implementation, a package. Fragments do not belong to a particular package which requires changes to odoc (like https://github.com/ocaml/odoc/issues/27).
* Currently in this PR the `odoc html-fragment` command uses [hardcoded package and page names](https://github.com/ocaml/odoc/pull/166/commits/3010be947e8356441b4d7a2b9d2a0717776e143c#diff-97dcad485808b851bebe5a10cccac4d7R32). This is not ideal because it assumes that the fragment belongs to a package (see previous option).

dbuenzli:
Thanks for you updates. Don't really know what to tell you but basically for my use case: 

> Cross-reference resolution requires a concrete root and, given the current implementation, a package.

This makes the result unusable.

rizo:
Ok, I managed to change the root URI for fragments.

Given a documentation page `test.mld` with the following content:

```
{1 Title}
This is just a {e test}.

Reference: {!type:Fmt.t}.
```

Running the following:

```
$ odoc html-fragment -o /dev/stdout -I $(odig cache path)/cache/fmt test.mld  
```

Produces on the standard output:

```
<h1 id="title"><a href="#title" class="anchor"></a>Title</h1>
<p>This is just a <em>test</em>.</p>
<p>Reference: <a href="./fmt/Fmt/index.html#type-t"><code>Fmt.t</code></a>.</p>
```
Additionally, if `--root-uri=$(odig cache path)/odoc/` is passed, the reference in the above example on my machine will resolve to:

```
/Users/rizo/.opam/4.06.1/var/cache/odig/odoc/fmt/Fmt/index.html#type-t
```

In other words it works. I'm not sure if setting `--root-uri` to `.` is the best choice for `html-fragment`.

rizo:
The only thing that annoys me is the need to define dummy [page and package names](https://github.com/ocaml/odoc/pull/166/files#diff-97dcad485808b851bebe5a10cccac4d7R40). They are not used in reference resolution but are needed to populate the Root/Page types. Since no `odoc` files are produced I don't think it's critical...

I'll try to polish the code and write some tests. Would be great if someone could review and validate this approach. I'm not sure how hard it will be implement this for `odoc html`. If this approach gets approval I can extend it to fully solve https://github.com/ocaml/odoc/issues/27.

rizo:
Ping. Could someone do a quick review of this PR, please?

rizo:
@dbuenzli since you know exactly how this feature needs to work, would you mind glancing over the code to make sure I'm not doing anything silly?

rizo:
I removed the `odoc html --fragment` version in favour of `odoc html-fragment` that supports custom root URI.

src/odoc/bin/main.ml
```diff
@@ -224,12 +224,48 @@ end = struct
...
+    let root_uri =
+      let last_char = String.get root_uri (String.length root_uri - 1) in
+      if last_char <> '/' then root_uri ^ "/" else root_uri
+    in
+    Html_fragment.from_mld ~env ~root_uri ~output:output_file input_file
+
+  let cmd =
+    let output =
+      let doc = "Output HTML fragment file" in
+      Arg.(required & opt (some string) None &
```
dbuenzli: Required optional arguments are kind of a contradiction. It would be better to simply output to stdout if no argument is provided. That would make it a better unix tool aswell.

src/odoc/bin/main.ml
```diff
@@ -224,12 +224,48 @@ end = struct
...
+          info ~docs ~docv:"file.html" ~doc ["o"; "output-file"])
+        in
+    let input =
+      let doc = "Input documentation page file" in
+      Arg.(required & pos 0 (some file) None & info ~doc ~docv:"file.mld" [])
+    in
+    let root_uri =
+      let doc = "Root URI used to resolve cross-references. Set this to the \
+                 root of the global docset during local development." in
+      Arg.(value & opt string "./" & info ~docv:"URI" ~doc ["root-uri"])
```
dbuenzli: If using `""` as a default does not poses a problem I think it would be preferable. Having relative links generated as a `./...` is a bit odd.

ryyppy:
@rizo #193 is now merged, how is the state of this?

PR #167, opened 2018-08-23
Always flush the HTML output.

rizo:
While working on https://github.com/ocaml/odoc/pull/166 I noticed that small mld pages did not produce any output because the formatter wasn't properly flushed.

This explicitly flushes the generated HTML output, even though in practice it should be automatically flushed given the relatively large output produced by HTML documents.


aantron:
Thanks :)

PR #175, opened 2018-08-31
print doc preamble for modules, module types, etc.

trefis:
This is a quick hack to reenable the printing of documentation module (and module types, and classes, and class types) bindings, that was disabled for some reason.

It's probably not the right way to do it, and I'm not suggesting that this PR should necessarily be merged as is. It's just here to point out that it's needed.

PS: I'm not sure why all these gets stuffed inside `<article>` compared to other signature items, but whatever.

### DIFF
--- a/src/html/to_html_tree.ml
+++ b/src/html/to_html_tree.ml
@@ -621,9 +621,9 @@ sig
     item_to_id:('item -> string option) ->
     item_to_spec:('item -> string option) ->
     render_leaf_item:('item -> rendered_item * Comment.docs) ->
-    render_nested_article:('item -> rendered_item * Html_tree.t list) ->
+    render_nested_article:('item -> rendered_item * Comment.docs * Html_tree.t list) ->
     ((_, 'item) tagged_item) list ->
-      rendered_item * toc * Html_tree.t list
+    rendered_item * toc * Html_tree.t list
 
   val render_toc :
     toc -> ([> Html_types.flow5_without_header_footer ] Html.elt) list
@@ -767,7 +767,7 @@ struct
     item_to_id : 'item -> string option;
     item_to_spec : 'item -> string option;
     render_leaf_item : 'item -> rendered_item * Comment.docs;
-    render_nested_article : 'item -> rendered_item * Html_tree.t list;
+    render_nested_article : 'item -> rendered_item * Comment.docs * Html_tree.t list;
   }
 
   let finish_section state =
@@ -811,12 +811,26 @@ struct
           }
 
       | `Nested_article item ->
-        let html, subpages = state.render_nested_article item in
+        let html, maybe_docs, subpages = state.render_nested_article item in
         let html, maybe_id = add_anchor state.item_to_id item html in
         let a = add_spec state.item_to_spec item maybe_id in
+        let html =
+          match maybe_docs with
+          | [] -> Html.article ~a html
+          | docs ->
+            let docs = Documentation.first_to_html docs in
+            let docs = (docs :> (Html_types.div_content Html.elt) list) in
+            let adoc =
+              [ Html.a_class ["doc"]
+              ; Html.a_style "padding: 0.75em; padding_bottom: 0;" ]
+            in
+            Html.article
+              [ Html.div ~a html
+              ; Html.div ~a:adoc docs ]
+        in
         section_items section_level {state with
             input_items;
-            acc_html = (Html.article ~a html)::state.acc_html;
+            acc_html = html :: state.acc_html;
             acc_subpages = state.acc_subpages @ subpages;
           }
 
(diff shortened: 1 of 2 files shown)


aantron:
@trefis Would you be willing to add or modify a test case, to exercise this code?

src/html/to_html_tree.ml
```diff
@@ -811,12 +811,26 @@ struct
...
+          match maybe_docs with
+          | [] -> Html.article ~a html
+          | docs ->
+            let docs = Documentation.first_to_html docs in
+            let docs = (docs :> (Html_types.div_content Html.elt) list) in
+            let adoc =
+              [ Html.a_class ["doc"]
+              ; Html.a_style "padding: 0.75em; padding_bottom: 0;" ]
+            in
+            Html.article
```
rizo: I'm not sure why this needs to be an `article`. Other signature items are generated as `dl/dt/dd` tags, I think modules shouldn't be an exception.

src/html/to_html_tree.ml
```diff
@@ -811,12 +811,26 @@ struct
...
         let a = add_spec state.item_to_spec item maybe_id in
+        let html =
+          match maybe_docs with
+          | [] -> Html.article ~a html
+          | docs ->
+            let docs = Documentation.first_to_html docs in
+            let docs = (docs :> (Html_types.div_content Html.elt) list) in
+            let adoc =
+              [ Html.a_class ["doc"]
+              ; Html.a_style "padding: 0.75em; padding_bottom: 0;" ]
```
rizo: The style should be added to the CSS file. Also ideally we should reuse existing rules (see my comment about `dl` tags).

test/print/print.ml
```diff
@@ -16,7 +16,7 @@ struct
       List ((Atom "compilation_unit")::(Atom name)::hidden)
 
   let root : Root.t -> sexp = fun {package; file; digest} ->
-    List [Atom package; odoc_file file; Atom digest]
+    List [Atom package; odoc_file file; Atom (Digest.to_hex digest)]
```
rizo: Sorry, I'm not sure why this is needed. It also seems like the tests are failing because of this with `[invalid] Digest.to_hex`.

rizo:
@trefis I understand this wasn't meant as a final implementation but I left some comments anyway. I can improve/polish this PR if you want, I just need access rights (or I can make a complementary PR...).

trefis:
I basically agree with all of your comments, but I have no time to work on this right now.
So feel free to send a fresh PR implementing this correctly :)

test/print/print.ml
aantron: @rizo I fixed that in `master`. Only the second commit in this PR is properly part of the PR, the first one was rewritten in `master`.

src/html/to_html_tree.ml
aantron: I think the original motivation for `<article>` (or something other than `<dt>`) was that for inlined modules and classes, we will have the content of almost an entire separate page inside the tag. For other kinds of signature items, the item goes in `<dt>` and the docs in `<dd>`. However, the docs for an inline module don't follow the whole module. They are nested inside the module, immediately after the module name.

rizo: > I think the original motivation for <article> (or something other than <dt>) was that for inlined modules and classes, we will have the content of almost an entire separate page inside the tag. 

I think the non-`dl` tag only makes sense for _actually_ inlined modules (do we have any plans for this, btw?). For the cases where the module is not inlined, and only has the brief, I'd still use a description list.

aantron: That was the plan, but I was originally going to inline all modules by default last year, so `<article>` was in preparation for that. I agree with undoing it for the time being, if anyone wants to.

rizo: I'd be happy to do it! :)

---

Can you elaborate on the trade-offs between inlining/referencing nested items?

In some cases referencing creates an additional indirection for the users. In larger libraries having all modules inlined is too verbose. Would be very useful to allow users to pass inline hints for odoc. Additionally the inlined modules could be collapsable similar to current includes.

aantron: These are the same tradeoffs I had in mind.

The basic argument for inlining by default was that if it was inline in the `.mli`, and looked fine, it will probably look fine if inlined in the HTML.

rizo: @dbuenzli I’m curious to know what’s your opinion on this, given that most of your libraries tend to expose one single interface.

Would you be happier with inlined modules, the same way they are defined in interface files?

dbuenzli: No I would not be happy at all. I think I had this discussion with almost every person that worked on the markup after all these years and some earlier version actually did it and it made the docs of my libraries unreadable. Then usually it would be proposed to have them in a details element, but then I would point out this would break anchor linking, usually it would then be proposed that this should be controlled via an annotation -- which is fine to me as long as the default is kept the way `ocamldoc` works. The latter is the argument to have here which is to follow the way `ocamldoc` has worked for all these years since most people who care about documentation actually designed their documentation interaction around its output structure.

rizo: @dbuenzli I'm glad I asked then. As you can imagine I wasn't aware of any previous discussions. Feel free to point me to any relevant discussions you think I should be aware of.

I'm fine with keeping ocamldoc's convention here as the default behaviour.

rizo:
I suggest closing this in favour of https://github.com/ocaml/odoc/pull/194.

Issue #181, opened 2018-09-12
Internal error due to truncated mld names


Octachron:
Trying to compile a mld file named `a.mld` generates an uncaught error

```
odoc compile a.mld --package Test
```
>odoc: internal error, uncaught exception:
      Invalid_argument("String.sub / Bytes.sub")

This stems from `compile.ml`  assuming that all output name for `mld` file are prefixed by `page-`, which is not the case if the option `-o` is not specified.

rizo:
Maybe we should just make the output required in this case?

I also I don't really understand why the `page-` prefix is a requirement and not a convention in odoc.

trefis:
> This stems from compile.ml assuming that all output name for mld file are prefixed by page-, which is not the case if the option `-o` is not specified.

Something may have changed, but at some point it should have been the case even if `-o` wasn't specified, at least according to https://ocaml.github.io/odoc/manual_usage.html

> I also I don't really understand why the page- prefix is a requirement and not a convention in odoc.

Namespacing. So you can have `foo.mld` and `foo.ml` and everything is going to be fine.

rizo:
FYI, I’m preparing a PR to fix this.

@trefis thanks for the clarification.

PR #183, opened 2018-09-13
Always use "page-" prefix when compiling mld files

rizo:
Fixes https://github.com/ocaml/odoc/issues/181.

In addition I corrected some outdated help comments in compile command and added a log line to show the input/output paths while compiling.

### DIFF
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -78,6 +78,12 @@ module Compile : sig
   val info: Term.info
 end = struct
 
+  let has_page_prefix file =
+    file
+    |> Fs.File.basename
+    |> Fs.File.to_string
+    |> Astring.String.is_prefix ~affix:"page-"
+
   let compile hidden directories resolve_fwd_refs output package_name input =
     let env =
       Env.create ~important_digests:(not resolve_fwd_refs) ~directories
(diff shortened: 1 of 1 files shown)


rizo:
The "page-" prefix will not be added if the input mld filename already contains it.

I must say I think the output option (`-o=PATH`) should be required with the "page-" prefix in the basename. Not a huge fan of implicit promotions like this one.

But this would be a breaking change, so maybe it's not worth it.

aantron:
@rizo I think the output option already requires the `page-` prefix, see https://github.com/ocaml/odoc/pull/183/commits/4f9c88c8498e54c17f7ea4080fec8568bfb05a38#diff-c5e9bcfdf9d516ccf8b1af0d19ccb68dL94.

src/odoc/bin/main.ml
```diff
@@ -87,17 +93,27 @@ end = struct
...
+            let directory = Fs.File.dirname input in
+            let name = Fs.File.basename input in
+            let name = "page-" ^ Fs.File.to_string name in
+            Fs.File.create ~directory ~name
+          else input
+        in
+        Fs.File.(set_ext ".odoc" output)
     in
+    Fmt.epr "Compiling... input=%s output=%s@."
+      (Fs.File.to_string input) (Fs.File.to_string output);
```
aantron: Is this a temporary print?

rizo: It was initially, but then I decided to leave it. Is there a reason why odoc doesn't have any info-level logging?

I can remove it if it feels inadequate.

rizo:
> I think the output option already requires the page- prefix

Oh, sorry that's not what I meant. I suggested the output option to be both required _and_ have the "page-" prefix. Only the latter is true now.

src/odoc/bin/main.ml
aantron: No particular reason, except that we haven't set up proper logging yet. Looking at this print, I wonder if the user always sees it. If so, I imagine it's noisy.

rizo: Having proper logging would be useful for reporting/debugging, let's add it. I'll remove this message until then.

aantron:
Thanks!

Issue #165, opened 2018-08-22
Run an HTML validator as part of the test suite


aantron:

pmetzger:
Do you have one in particular in mind?

aantron:
No, we need to pick one. Ideally something easy to install on multiple platforms. If not, then it should be easy to install in Ubuntu in Travis.

rizo:
I'd add that it should support the heading-level outline algorithm similar to one in https://validator.w3.org/nu.

aantron:
That makes me wonder if there is a validator we can access using curl/wget.

dbuenzli:
The one @rizo mentioned is available as a cli tool (but you'll need java). Other than that there's at least [`tidy`](http://www.html-tidy.org/). Which I is available in brew as `tidy-html5` and in ubuntu as [`tidy`](https://packages.ubuntu.com/bionic/tidy).

ostera:
+1 for using `tidy` as well for the expected outputs. Will make updating the expected HTML in PRs much clearer as well.

pmetzger:
If the consensus is tidy, what do we need to do in order to make that happen?

rizo:
I think before the integration, the first step would be to run tidy on the existing [HTML expected files](https://github.com/ocaml/odoc/tree/master/test/html/expect/ml) and find all issues with it. Those can be reported as issues and fixed separately. After that we can update the expected files and integrate tidy as part of the validation for actual output.

ostera:
After running:

```sh
tidy -ashtml test/html/expect/*/*.html 2>&1
  | grep line \
  | awk -F':' '{ print $2 }' \
  | sort -u
```

I see the following problems:

  1. `<h2>` anchor "empty-section" already defined
  2. discarding unexpected `</dt>`
  3. missing `<dd>`
  4. missing optional end tag `</dd>`
  5. nested emphasis `<em>`
  6. trimming empty `<code>`

Seems small enough to be addressed in a single PR.

ostera:
This can be closed now.

aantron:
Thanks, @ostera :)

Addressed by @ostera in #185 and #190.

PR #185, opened 2018-09-14
Generate Valid HTML

ostera:
Hello! I'll open this PR for visibility. The overall goal is to get valid HTML output from `odoc`.

What needs to be done, from what I understood from #165, is:

- [x] Run `tidy` on the expected html files
- [x] Install `tidy` on Travis
- [x] Make `tidy` part of running `make test`
- [x] Fix validation errors in output in library:
    - [x] `<h2>` anchor "empty-section" already defined
    - [x] ~~discarding unexpected `</dt>`~~
    - [x] ~~missing `<dd>`~~
    - [x] missing optional end tag `</dd>`
    - [x] ~~nested emphasis `<em>`~~
    - [x] trimming empty `<code>`

I'll be putting some time on this over the weekend.

I'd also like the output to be readable, instead of mangled into a single line, but we can have that discussion once it's valid HTML since it'll either tie the execution of `odoc` to `tidy` (which I do not like at all) or make the HTML generation code aware of indentation, et al.

Feel free to edit this description if i'm missing anything!


ostera:
cc @pmetzger @dbuenzli @rizo

aantron:
> I'd also like the output to be readable, instead of mangled into a single line, but we can have that discussion once it's valid HTML since it'll either tie the execution of odoc to tidy (which I do not like at all) or make the HTML generation code aware of indentation, et al.

It will also make the diffs much more legible when someone updates an expected output file. I've worked on this a bit in the past. If we pretty-print HTML, we have to make sure that the whitespace changes are not significant to the interpretation of HTML. For example, inserting newlines around tags in `<em>foo</em>bar` splits "*foo*" and "bar" into two words, whereas without newlines, this is one word "*foo*bar" that just has the first syllable emphasized.

TyXML, our HTML output library, has a simlpe pretty-printer (ocsigen/tyxml#187), but we have to make check that it doesn't insert any significant whitespace (cc @Drup).

We definitely shouldn't use tidy for this. We'll need some OCaml code, or a library, to get proper indentation.

I'll probably open a new issue for this later, and move this comment there.

aantron:
Some notes on the validation errors:

> missing `<dd>`

This might be because a signature item didn't have documentation.

> nested emphasis `<em>`

The ocamldoc source language allows `{e foo {e bar}}`, which is nested emphasis. HTML also seems to allow this: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em

ostera:
@aantron, is it me or we actually _are already_ pretty printing when we do the test diffing? https://github.com/ocaml/odoc/blob/master/test/html/test.ml#L40-L70

aantron:
Yes we do, but this pretty-printing is not sensitive to the meaning of whitespace (it will ruin `<em>foo</em>bar`). So, we can't use it for "production," or save the expected output in that form. It's just for helping the dev see what changed during testing. Also, it uses a tab width of zero to avoid diffs from indentation.

ostera:
```bash
ostera/odoc λ make test
dune build
dune build @test/parser/runtest --no-buffer -j 1
dune build @test/html/runtest --no-buffer -j 1
Done: 597/599 (jobs: 1)Testing html.
[OK]                support-files                 0   support-files.
[OK]                html (ml)                     0   val.
[OK]                html (ml)                     1   markup.
[OK]                html (ml)                     2   section.
[OK]                html (ml)                     3   module.
[OK]                html (ml)                     4   interlude.
[OK]                html (ml)                     5   include.
[OK]                html (ml)                     6   mld.
[OK]                html (ml)                     7   type.
[OK]                html (ml)                     8   external.
[OK]                html (ml)                     9   functor.
[OK]                html (ml)                    10   class.
[OK]                html (ml)                    11   stop.
[OK]                tidy validation (ml)          0   val.
[ERROR]             tidy validation (ml)          1   markup.
[ERROR]             tidy validation (ml)          2   section.
[ERROR]             tidy validation (ml)          3   module.
[OK]                tidy validation (ml)          4   interlude.
[ERROR]             tidy validation (ml)          5   include.
[OK]                tidy validation (ml)          6   mld.
[ERROR]             tidy validation (ml)          7   type.
[OK]                tidy validation (ml)          8   external.
[OK]                tidy validation (ml)          9   functor.
[OK]                tidy validation (ml)         10   class.
[OK]                tidy validation (ml)         11   stop.
[OK]                theme_uri                     0   module.
[OK]                theme_uri                     1   val.
[OK]                theme_uri                     2   include.
[OK]                theme_uri                     3   section.
[OK]                html (re)                     0   val.
[OK]                html (re)                     1   markup.
[OK]                html (re)                     2   section.
[OK]                html (re)                     3   module.
[OK]                html (re)                     4   interlude.
[OK]                html (re)                     5   include.
[OK]                html (re)                     6   mld.
[OK]                html (re)                     7   type.
[OK]                html (re)                     8   external.
[OK]                html (re)                     9   functor.
[OK]                html (re)                    10   class.
[OK]                html (re)                    11   stop.
[OK]                tidy validation (re)          0   val.
[ERROR]             tidy validation (re)          1   markup.
[ERROR]             tidy validation (re)          2   section.
[ERROR]             tidy validation (re)          3   module.
[OK]                tidy validation (re)          4   interlude.
[ERROR]             tidy validation (re)          5   include.
[OK]                tidy validation (re)          6   mld.
[ERROR]             tidy validation (re)          7   type.
[OK]                tidy validation (re)          8   external.
[OK]                tidy validation (re)          9   functor.
[OK]                tidy validation (re)         10   class.
[OK]                tidy validation (re)         11   stop.
```

Now proceeding to fix each validation 🙌

Drup:
@aantron That's an interesting. Is that the only part that is problematic ? If yes, that is relatively easy to patch in a way that doesn't increase the complexity of the code too much.

In any case, I still think we should add necessary pieces in tyxml to have some API easy to plug with markup's printers.

aantron:
@ostera I suggest running tidy as part of each expect test, not as separate tests.

aantron:
@Drup 

> Is that the only part that is problematic ?

I'm not sure, I'd have to go tag-by-tag and check, I think. If there is enough demand, I can do it in Markup.ml, so nobody has to worry about it again :)

ostera:
@antron @rizo I realized that most of the validations will be warnings rather than errors, according to [tidy's sources](https://github.com/htacg/tidy-html5/blob/86b52dc1081ca4b0582c7bad279bf254bad268e1/src/message.c#L262)

This comes with a tradeoff — if `tidy` finds something that raises warnings, it will exit with status code 1.

We can mute warnings, but `tidy` will still return 1 if it found some.

So ignoring all warnings (checking for return code > 1, where 2 is some errors) would make most of the useful checks in that table linked above not cause test failures.

Not ignoring all warnings means we may not be able to allow for nested `<em>`s; at least not without circumventing the `tidy` checks for it.

aantron:
@ostera It's not pretty, but we may be able to capture tidy's `STDERR`/`STDOUT`, and fail on non-muted warnings that way.

ostera:
>  capture tidy's `STDERR/STDOUT`, and fail on non-muted warnings that way.

Sounds pretty brittle. What's the use case for nested emphasis? (Other than backwards compatibility, which should be preserved by these changes anyways).

test/html/cases/section.mli
```diff
@@ -3,7 +3,7 @@
 
 (** This is the module comment. Eventually, sections won't be allowed in it. *)
 
-(** {2 Empty section} *)
+(** {2 First Empty section} *)
```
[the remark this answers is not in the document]
ostera: Understood — I missed that odoc can generate a table of contents where the anchors should be automatically disambiguated.

This change can be reverted easily but we would have to address the disambiguation within this PR, or we could keep it as is and track the disambiguation as a separate issue. What would you like to do?

test/html/test.ml
```diff
@@ -223,7 +238,7 @@ let () =
   in
 
   (* Custom theme URI tests. *)
-  let theme_uri_tests : unit Alcotest.test =
+  let _theme_uri_tests : unit Alcotest.test =
```
ostera: This worried me the most. I had to disable this tests to get a deterministic output on the test runs, otherwise the `section.mli` test would show up with a style path of `../../b/c/odoc.css` which was clearly part of the themed tests.

It's also possible that the behavior of `validate_html` messed up the updating of the `_scratch/actual` file and that's why I was seeing that altered path, but I haven't confirmed yet.

test/html/cases/markup.mli
```diff
@@ -27,9 +27,9 @@
     {2 Styling}
 
     This paragraph has some styled elements: {b bold} and {i italic},
-    {b {i bold italic}}, {e emphasis}, {e {e emphasis} within emphasis},
-    {b {i bold italic}}, super{^script}, sub{_script}. The line spacing should
-    be enough for superscripts and subscripts not to look odd.
+    {b {i bold italic}}, {e emphasis}, {b {i bold italic}}, super{^script},
```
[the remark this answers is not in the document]
ostera: Left a comment on this in the main thread.

test/html/cases/section.mli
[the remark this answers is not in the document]
aantron: A separate PR is probably better, as it is a qualitatively different kind of work from what we're doing here. But if you'd like, you can fix it in this PR as well :)

aantron:
I can be for banning nested emphasis. If we do that, we should detect it in the parser and refuse to accept it/warn on it. Let's see if anyone else comments on it. Perhaps a separate issue should be opened.

> Sounds pretty brittle, considering we'd have to parse it and check for NESTED_EMPHASIS and other tolerable warnings.

I assumed we can totally mute the warnings we don't want, so that we only need to check whether the output is empty or not. Perhaps I should install tidy locally and look at its actual output :)

test/html/cases/section.mli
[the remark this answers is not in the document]
aantron: Opened #186 to track. it

src/html/to_html_tree.ml
```diff
@@ -684,15 +684,14 @@ struct
...
-        begin match maybe_docs with
-        | [] ->
-          consume_leaf_items_until_one_is_documented items acc
-        | docs ->
-          let docs = Documentation.to_html docs in
-          let docs = (docs :> (Html_types.dd_content Html.elt) list) in
-          let docs = Html.dd docs in
-          List.rev (docs::acc), items
-        end
+        let docs = Documentation.to_html maybe_docs in
```
aantron: Does this change make it so each `<dt>` tag is followed by `<dd>`, whether there are docs or not?

If so, I don't think we should do it.

Sometimes, there are multiple signature items, that are summarized by one doc comment:

```ocaml
type yes
type no
(** Phantom types for some purpose. *)
```

In this case, I think we'd like

```html
<dt><!-- type yes --></dt>
<dt><!-- type no --></dt>
<dd><!-- docs for both --></dd>
```

This helps with styling as well, as we can use CSS to avoid vertical space between the two types, making them look visually grouped.

Here's a live example: http://aantron.github.io/lambda-soup/#TYPEelement

WhatWG suggests this is acceptable as well: https://html.spec.whatwg.org/multipage/grouping-content.html#the-dl-element, see "In the following example, one definition is linked to two terms."

ostera: Yes, that's exactly what it's doing to compensate for the `missing <dd>` warnings.

This is starting to point towards capturing the output and searching for the warnings we consider intolerable.

aantron: @ostera, I just installed tidy locally. It looks like with the right `--mute` options, the exit code is still 1, but STDERR is empty. So we don't have to parse the output, just check that it's empty. If non-empty, we show it to the user and fail the test.

aantron:
Do we want to force developers to install tidy? I think it's okay to run it only in CI, but interested in the opinions of others.

If not installing is okay, we should have the tester check if `tidy --version` works, or something, and just not try to run tidy if it is not installed. But if tidy is not installed, and [`TRAVIS=true`](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) is in the environment, the whole tester should probably just fail immediately with `exit 1`.

pmetzger:
> Do we want to force developers to install tidy?

Probably not, but making use of it in the CI and allowing them to use it if they want it and it is installed seems appropriate.

src/html/to_html_tree.ml
ostera: Updated.

test/html/test.ml
```diff
@@ -107,16 +107,49 @@ let () =
...
+      let (stdout, stdin, stderr) = Unix.open_process_full cmd [||]  in
+      let errors =
+          let rec r acc = match input_line stderr with
+          | s -> r ((s ^ "\n") :: acc)
+          | exception End_of_file -> List.rev acc
+          in
+          r []
+      in
+      let exit_status = Unix.close_process_full (stdout, stdin, stderr) in
+      match exit_status with
```
ostera: Code-style I'm not a fan of nested matches like this, but I haven't seen the codebase rely too much on operators (`>>=`, `<|>`) and I figured I wouldn't start now.

Also first time consuming outputs from a process with `Unix.*` so let me know if there's a more straightforward way of doing this!

test/html/test.ml
```diff
@@ -107,16 +107,49 @@ let () =
     let reference_file = expect_directory // lang // (file_title ^ ".html") in
 
     let validate_html file =
+      let label = "tidy" in
+      let muted_warnings = String.concat "," [
```
ostera: These are the 3 warnings that we said we didn't want to fail on — however, `MISSING_STARTTAG` is going to pop up for _a lot_ of tags. Some of which may or may not be optional by https://html.spec.whatwg.org/ but that `tidy` will consider not optional.

Exhaustively listing all of these can be done by parsing the HTML Spec, but I'd refrain from doing all that work unless we absolutely need to.

aantron: What is triggering `MISSING_STARTTAG` in the odoc output? Both `MISSING_STARTTAG` and `DISCARDING_UNEXPECTED` look uncomfortably broad to me (we may have to accept that, though :/)

test/html/test.ml
aantron: The `Unix.*` looks fine. For the nested matches, you could do

```ocaml
match exit_status, errors with
| WEXITED 0, _
| WEXITED 1, [] ->
  ()
| WEXITED _, _ ->
  ...
| _, _ ->
  ...
```


One of that advantages is that none of the case code will be duplicated.

test/html/test.ml
ostera: The missing `<dd>` tags on the `<dl>`'s are triggering `MISSING_STARTTAG` and some `<dt>`s are triggering `DISCARDING_UNEXPECTED`.

I haven't managed to find what exactly was, within the `<dt>`, that triggered the error, but my thesis is that `tidy` is behind what the HTML Spec considers [Flow content](https://html.spec.whatwg.org/multipage/dom.html#flow-content-2) (which is what's allowed within `<dt>`).

aantron: If there's no way to silence the warning only for `<dd>`, we can leave it as in this PR now. Presumably, we still get a lot of value from tidy's other warnings.

aantron:
Thanks for this!

I opened #189 about making tidy optional in the future.

PR #190, opened 2018-09-17
Run tidy if installed, fail if it's not and we're in travis

### LINKED ISSUE #189: Make the tidy command optional
tidy is used by the test suite, but we don't want to ask all developers to install it.

We do want to fail if it's not installed in CI, though (if `TRAVIS` environment variable is `true`).

See https://github.com/ocaml/odoc/pull/185#issuecomment-421617855.

cc @ostera

ostera:
Close #189.

### DIFF
--- a/test/html/test.ml
+++ b/test/html/test.ml
@@ -23,7 +23,7 @@ let running_in_travis = match Sys.getenv "TRAVIS" with
   | _ -> false
   | exception Not_found -> false
 
-let found_tidy = match Sys.command "which tidy" with
+let found_tidy = match Sys.command "which tidy > /dev/null" with
   | 0 -> true
   | _ -> false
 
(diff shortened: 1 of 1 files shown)


ostera:
I can use `Unix.open_process_full` to capture all the output?

aantron:
Thank you!

test/html/test.ml
```diff
@@ -198,7 +205,12 @@ let () =
     (* Running the actual commands for the test. *)
     let run_test_case () =
       generate_html ();
-      validate_html html_file;
+
+      match running_in_travis, found_tidy with
+        | true, false -> Alcotest.fail "Could not find `tidy` in PATH!";
+        | _, true -> validate_html html_file;
+        | _, _ -> ();
```
rizo: FYI, this currently breaks all the tests locally.

In particular when the matched pattern is `false, true` (second line), only the HTML validation will be run. This has to do with the OCaml's precedence rules. The content after this line is included in the last case after `(); ...`.

rizo: Maybe we should consider running ocamlformat on odoc :)

PR #191, opened 2018-09-17
Fix tidy check bug. Fail early if tidy is not installed in CI.

rizo:
Fixes https://github.com/ocaml/odoc/pull/190/files#r218243322 and checks if tidy is installed in the CI early.


aantron:
I missed that too. I actually bothered to run a build in Travis with tidy not installed to see what happens there, but neglected to try it locally :)

rizo:
Actually it wasn't properly working even in Travis. Or to be precise only HTML validation was running, not the actual tests.

PR #194, opened 2018-09-19
Print preamble for nested items

rizo:
This PR reimplements https://github.com/ocaml/odoc/pull/175 in a similar way.

The main difference is in the generated markup, where `dl` tags are used for the nested item and its preamble documentation. Regular `div` tags are used for nested items without documentation.

Thanks to @trefis for the original effort!

### DIFF
--- a/src/html/html_generator.ml
+++ b/src/html/html_generator.ml
@@ -168,7 +168,7 @@ module type Html_generator = sig
          item_to_id:('item -> string option)
       -> item_to_spec:('item -> string option)
       -> render_leaf_item:('item -> rendered_item * Comment.docs)
-      -> render_nested_article:('item -> rendered_item * Html_tree.t list)
+      -> render_nested_article:('item -> rendered_item * Comment.docs * Html_tree.t list)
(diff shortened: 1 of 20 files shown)


rizo:
I'll update the tests when https://github.com/ocaml/odoc/pull/193 is accepted as diffing/promotion is currently broken in master.

rizo:
I also want to note that using `dt_content` instead of `div_content` for [rendered items](https://github.com/ocaml/odoc/blob/394dba10f56eff6b75ff819fb87fe2a1ab758811/src/html/to_html_tree.ml#L29) should now be possible, in theory. In practice though we can't do that, because our variant representation uses tables and those are not allowed in `dt`.

aantron:
@rizo, I don't believe the `dt_content` fix (ocsigen/tyxml#193) has been released yet. It's only in TyXML `master`.

rizo:
Rebased from master and added a test case. This should be mergable now.

@aantron can you please glance through this to make sure everything looks good?

aantron:
LGTM, thanks!

PR #197, opened 2018-09-21
Automatically support ligature fonts if installed

ostera:
This would let docs fallback to ligature fonts if the user already has them, but not force them upon users who don't have them installed.

I put the 4 most common I've seen people use but more can be added (or some removed).

@rizo let me know if there's any important ligature fonts missing here!

And I can't remember @aantron's stance on this 🙈

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
   - OCAML=4.07.0
 
 before_script:
-  - sudo add-apt-repository -y ppa:robert7/nixnote21
+  - sudo add-apt-repository -y ppa:robert7/tidy-html5
   - sudo add-apt-repository -y ppa:avsm/ocaml42+opam12
   - sudo apt-get update
   - sudo apt-get install opam tidy


aantron:
My stance was that I suggested this as well :)

ostera:
@rizo ah, good point. I assumed that a Helvetica-like grotesque would remain as our body typeface. Hence I prioritized reader-familiarity over consistency in the transition to ligatures. After all, from a holistic perspective, the difference in monospaced fonts will unlikely grant a reevaluation of the body typeface, Helvetica is quite safe. If we were to change this then I'd be more weary of our monospaced choices!

**That being said**, I'd love to see a complete redesign of the current doc format to make better use of space as well :)

rizo:
Regardless of our body typeface, I'm worried that inline code elements will have unbalanced size because the font size is defined equally for all font families. For example, even with Helvetica, using PragmataPro feels very odd with the current font sizes. The size would have to be adjusted individually for each monospaced font accordingly to the body font.

With my suggestion the same font size would work perfectly because Fira Mono and Fira Code are essentially the same font.

>  I'd love to see a complete redesign of the current doc format to make better use of space as well :)

Can you elaborate or give some examples?

[Here](http://odis.io/odoc-demo/order/Order/index.html)'s a preview of my WIP changes by the way.

CC @pmetzger – I know you dislike ligatures but I'm sure you can explain this better than me.

ostera:
@rizo I'm sold on your changes. I'll remove the style and keep the renaming of the ppa 👌

PR #198, opened 2018-09-24
Testing: commit pretty-printed HTML output for nicer git diffs

### LINKED ISSUE #187: Pretty-print the HTML output
This is a bit tricky, because, e.g., `<em>foo</em>bar` cannot be indented, as a newline around `</em>` will be treated as a word break. So, the pretty-printer needs to be aware of the meaning of whitespace in various kinds of HTML tags.

The code for this probably belongs in Markup.ml or TyXML.

cc @ostera @Drup

aantron:
This requires pinning `markup` and `tyxml` for now:

```
opam pin add --dev-repo markup
opam pin add --dev-repo tyxml
```

The TyXML pin is needed because of a depopt from released of TyXML on Markup.ml. There is a breaking change in Markup.ml (I added the `` `Raw`` signal), but only the TyXML PPX depends on Markup.ml, and we don't use the PPX. cc @Drup.

cc @ostera, @rizo

Resolves #187.

### DIFF
--- a/odoc.opam
+++ b/odoc.opam
@@ -29,7 +29,7 @@ depends: [
   # opam repo.
   "alcotest" {>= "0.8.3"}
   "bisect_ppx"  {>= "1.3.0"}
-  "markup" {>= "0.7.6"}
+  "markup" {>= "0.8.0"}
   "lambdasoup"
   "ocamlfind"
   "sexplib"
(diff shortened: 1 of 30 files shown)


src/odoc/html_page.ml
```diff
@@ -16,6 +16,42 @@
 
 open StdLabels
 
+let write_html_to_channel : Pervasives.out_channel -> Tyxml.Html.doc -> unit =
+    fun channel page ->
+
+  let ns name = (Markup.Ns.html, name) in
```
ostera: Is this deconstructing the tuple on the right? If so, what's the difference in the bound `name` here?

ostera:
@aantron it seems that `No package matches markup>=0.9.0.` 😨

src/odoc/html_page.ml
rizo: "Deconstruction" or pattern-matching is always on the left side, _e.g._ see the [pattern position](https://caml.inria.fr/pub/docs/manual-ocaml/expr.html#let-binding) in let-bindings. In this case it's a tuple _construction_ with the `name` value as the second field.

dbuenzli:
Which specific problem does this PR solve ? #187 provides no motivation for this feature. 

Lacking this, in my eyes this only adds a new dependency to `odoc` whereas we should rather strive to remove dependencies.

src/odoc/html_page.ml
ostera: Yes, I understand that, but wouldn't this be the same as binding `Markup.Ns.html` to `ns` and letting `name` be? 🤔 

That's where my confusion arose.

ostera:
@dbuenzli it would make reviewing updated expect files considerably easier, since the diffs would actually pop up on the PR.

src/odoc/html_page.ml
aantron: It's a function that expands `name` by adding the HTML namespace. If you suggest a clearer way to write it, I can change it.

aantron:
> it seems that No package matches markup>=0.9.0. 

Indeed, it's not released yet. I had to modify Markup for this PR. That's why you have to do

> ```
> opam pin add --dev-repo markup
> ```

for now :)

src/odoc/html_page.ml
ostera: A-ha! I did not see that the left hand side is being evaluated as well.

This clears it up 👍🏼

rizo:
If adding a new library is really an issue, I think we might be able to only generate pretty HTML in the testing environment and keep markup as a dev dependency.

dbuenzli:
> If adding a new library is really an issue, I think we might be able to only generate pretty HTML for the testing environment and keep markup as a dev dependency.

Why not simply have a tool like `xmltrip` that pretty prints and run it on the outputs before diffing ? That way markup remains a dev dependency.

aantron:
We were doing that. It doesn't address git diffs, including in GitHub.

aantron:
We can commit the expected files permanently in pretty-printed form, and run the pretty-printer on the odoc output only during testing. However, I'd like the opinion of others on the importance of not adding this as a normal dependency.

aantron:
Actually, I'll change the PR over to pretty-print only in the tester, and commit the pretty-printed outputs (the reason we didn't commit pretty-printed output before is that we didn't have a pretty-printer that respected significant whitespace that we were willing to use).

If anyone wants the "production" output to be pretty-printed, please say so.

ostera:
After giving it some more thought, 👍 for checking in pretty-printed expected files and keeping `markup` as a dev dependency.

We can always reformat after generation, and we also avoid opening the can of worms of pretty-printing config flags.

aantron:
The pretty-printer is run only by the tester now, and the pretty-printed files are committed.

aantron:
It does require Markup.ml 0.8.0 though, which, I think, is only visible to opam 2, not opam 1.2. So, opam 2 is required for development of odoc at this point. odoc release should still work with opam 1.2.

I updated our Travis script to use opam 2.

PR #201, opened 2018-09-25
Refactor HTML Generation

ostera:
After talking to @aantron I figured I'd take this task for a dance and I'd attempt to make a functor to generate the syntax backends.

Seems to be going well so far 👌 but wanted to open the PR early so if there's anything outrageous going on is caught sooner than later.

Should drastically reduce the code needed for the two backends, but the diff is just awful to look at right now — it also caught a bug in the generated syntax for modules type substitutions 👍 (we were printing `params type` instead of reason's `type(params)`).

#### To do:

- [x] Finish tweaking functor to make `to_re_html_tree.ml` pass all tests
- [x] Use functor for `to_ml_html_tree.ml`

#### For another PR:
- [ ] Document functor modules (Syntax, Html_generator)


aantron:
A good way to make this reviewable is to leave the main "skeleton" code as untouched as possible, including with the same whitespace. Don't indent it, just surround it with new `module Foo (Bar) = ...` at the same indentation level. Don't apply ocamlformat to it or reformat it in any way, just change the code that calls into the argument modules or otherwise needs to be changed.

That way, reviewers can do a horizontal diff between the old files and the new functor. This is what I did when reviewing @ryyppy's original PR that added Reason support. Even though the patch diff looked huge, the horizontal PR between the new code and the original code emitting OCaml syntax was very, very manageable. https://github.com/ocaml/odoc/pull/156#discussion_r208438074 

We can apply formatting in separate commits, probably in separate PRs to avoid losing focus.

I'm slightly disappointed in the current ocamlformat formatting, it introduces way too much left whitespace. It looks like it needs several PRs of its own before it is ready for "production."

ostera:
@aantron from https://github.com/ocaml/odoc/pull/201/commits/157f21b845aca3528269b141ea487f095afa7a5d onwards you can see the changes one-by-one-ish

I'm _not happy_ with the way this PR ended up, but what's done is done. I'll leave the other 2 TODO points for another PR to not bloat the review time.

aantron:
I think you can delete `to_html_tree.ml` and `to_html_tree.mli`, since you renamed them to `to_ml_html_tree.ml(I)?`.

ryyppy:
So I looked through the code and especially through the code paths I worked on back then, and from my POV it looks quite solid? 

I wished there would be a way to get rid of those massive recursive declarations... not sure if this is doable in some way?

aantron:
Thanks! This should make odoc much easier to maintain.

PR #203, opened 2018-09-26
Documentation for Odoc

ostera:
Opening this to gather early feedback and comments 🗣 

I feel it's mandatory for `odoc`, since it's goal is to generate documentation, to bootstrap it's own. So `odoc` is documented using `odoc`. 

Hence I'm opening this PR. I started writing a new index manual, but i'd like a few more things to get done before we merge it.

For the time being I'm hosting it here: https://ostera.github.io/odoc/

#### TO DO
- [x] Scaffold Manuals
  - [X] Index + Getting Started
  - [X] High-level overview of `odoc`
  - [x] Using `odoc`
  - [X] Odoc Internals: In-depth flow of compilation 
  - [x] Build System Integration
  - [x] Contributing to `odoc`
- [x] ~~Automatically publish docs to Github Pages~~
- [x] `make docs`
- [x] Build docs on CI

#### Follow-up Work
- [ ] Point .md files in repo to docs
- [ ] Make sure all references to the modules are correct
- [ ] Make sure all flows are correct
- [ ] Proofread
- [ ] Table of Contents is generated
- [ ] Odoc Internals: In-depth flow of html generation


ostera:
cc @aantron @rizo

src/odoc/page-index.mld
```diff
@@ -0,0 +1,252 @@
+{1:top odoc}
+
+The documentation compiler for OCaml, Reason, and BuckleScript.
```
rizo: Does BuckleScript belong here? You could as well say dune in that case. Any tool can use odoc really, I'd keep just "OCaml and Reason".

src/odoc/page-index.mld
```diff
@@ -0,0 +1,252 @@
...
+
++ {{:#overview} What is [odoc] really? }
++ {{:#flow} How does it work? }
++ {{:#deep-dive} Digging Deeper }
+
+{2:overview What is [odoc] really?}
+
+[odoc] is not your regular documentation tool.
```
rizo: I'd mention that it's designed to be build-tool friendly. With focus on parallelism, caching, etc. I don't mind adding those changes separately though.

src/odoc/page-index.mld
```diff
@@ -0,0 +1,252 @@
...
+[
+  Game.cmti;
+  Player.cmti;
+  Monster.cmti;
+  page-intro.mld;
+]
+|> map ~f:compile_to_odoc  (* here we have Game.odoc, Player.odoc, etc *)
+|> iter ~f:compile_to_html (* here we have Game/index.html, and so on  *)
+
+]}
```
rizo: Wouldn't it be better to show this in shell syntax? Otherwise people will assume there's actually an OCaml API for this.

src/odoc/page-index.mld
```diff
@@ -0,0 +1,252 @@
...
+
+{2:overview What is [odoc] really?}
+
+[odoc] is not your regular documentation tool.
+
+Instead of source code, it works with {i compilation units}; that is, it works
+on compiler outputs and turns them into compiled documentation, which then
+becomes gorgeous HTML.
+
+A regular [odoc] execution transforms [cmt], [cmti], and [mld] files into [odoc]
```
rizo: Maybe you can explain what `odoc` files are before mentioning them here? Surely people will ask "what is an odoc file?".

src/odoc/page-index.mld
```diff
@@ -0,0 +1,252 @@
...
+Off we go!
+
+
+{4:step-1 1 — Compiling Player.cmti to Player.odoc}
+
+Superficially, what we need to do is straighforward.
+
+{[
+# We must tell odoc what the name of this Package is!
+λ odoc compile --package Game -o ./src/Player.odoc ./src/Player.cmti
```
rizo: Just a minor thing, but `$` would look more familiar to most people.

src/odoc/page-index.mld
```diff
@@ -0,0 +1,252 @@
...
++ And lastly {!Compilation_unit.save} takes care of saving this compilation
+  unit into the [Player.odoc] file in marshalled format.
+
+
+{4:step-2 2 — Compiling Player.odoc to Player/index.html}
+
+Now we can compile this to an HTML file:
+
+{[
+λ odoc html -I src -o . ./src/Player.odoc
```
rizo: The `-o` is very important. I would explain that it's typically the root path of the generated documentation for a package.

rizo:
Don't have time to do a thorough review but I like the idea in general.

I would try to clarify who is the target audience for the manual sections. For example, the low-level compilation information is IMO only useful for people who need to integrate odoc into a build tool – I would try to make it more formal and less example-driven.

Alternatively, for end users, it's more important to know how to write good documentation with the available features (pages, indexes, references, etc).

src/odoc/page-index.mld
pmetzger: Agreed.

src/odoc/page-index.mld
pmetzger: Rather than "regular" I'd say "ordinary". We hope it becomes everyone's _regular_ tool. :)

grabbou:
Thank you very much for taking care of this! If there's anyway I can help or get involved, please let me know! 

It took me a really long time to figure out the process of building a documentation website with `odoc`. Great that it will be all in one place from now. 

> Alternatively, for end users, it's more important to know how to write good documentation with the available features (pages, indexes, references, etc).

I can only speak for myself, but I wish I knew all that upfront.

aantron:
Agreeing with @rizo here, the text so far is for people doing build system integration (and we do need that text). Hopefully, fewer and fewer end users will need to see this over time. odoc developers will always need this.

src/odoc/page-index.mld
ostera: As long as we convey properly that this can be used with BuckleScript compilation units, I'm up for rephrasing.

It's not obvious to everyone, particularly since Reason can be used with the OCaml toolchain.

rizo: I think this can be explained later, i.e., the fact that odoc is designed to be integrated into existing build toolchains like BuckleScript and dune. A "documentation compiler" operates on syntax; BuckleScript is a compiler, that's why I find it confusing.

Also, ideally the end users shouldn't worry about how odoc works at all. That's why I think it's important to have a separate guide that describes the documentation features for documentation authors.

rizo:
One extra comment about the name change from `index.mld` to `page-index.mld`: dune treats the `index.mld` file in a special way – it's used to produce the module index page. For that reason and because the `page-*` prefix is only required for `.odoc` files I don't recommend renaming the file. Thus for consistency other documentation pages should also avoid the prefix. I also find it redundant anyway, the fact that the file is a page is already clear from the extension.

ostera:
I split the manual into smaller pieces that we can extend independently over time: https://ostera.github.io/odoc/Odoc/

However, I don't expect this PR to elaborate on all of them _immediately_, so I'll mark the Manuals in the TODO list as done.

Over the next few days I'll get the build+publish process in place, and then we can review more easily.

ostera:
@aantron @rizo thoughts on running `make docs` on every commit? pre-commit hook that can be setup? keep it manual for now?

If we keep it manual and use the master branch's `docs` folder then I can check those two TODO's off ✅

aantron:
I usually publish docs manually, because we want to leave the version for the last release up until the next release.

If/when odoc supports version switching, we can start publishing the dev docs automatically (and document that, so other projects can do it :) ).

ostera:
Alright, rewritten history without automatic publishing.

@aantron can we split this PR so I focus on getting the rest of the TODO's in another one?

aantron:
Yes, of course. Let me know when this one is ready for review.

ostera:
The last one point I'll make is about publishing the docs with Github Pages.

I'd prefer if we used the simpler mechanism they offer, where the docs site is hosted within the `master` branch, inside of the `docs` folder. This allows for:

1. Cleaner PRs that update both the code and the documentation in one go
2. Simpler workflow, where at the end of my changes I can run `make docs` and commit that with "Update docs".

Alternatively we could publish to `gh-pages` branch, which would mean an update to documentation will have to come in two steps: a PR to master, and a commit to `gh-pages`.

If we opt for the latter, then we should have a straightforward way of doing so (_a-la_, `make docs-publish`).

aantron:
@ostera Given we are going with republishing the docs only at release time (and for any serious errata), I think we want to *avoid* 1 and 2. We don't want PRs to `master` to affect the published docs, we want to discourage contributors from being misled into trying to republish them, and the publishing workflow is only going to be carried out by whoever is doing the release (in the coming release, myself).

I think we just need instructions for how the generated dosc (in a `.gitignore`d directory) can be reviewed locally.

I suggest to leave the generated docs out of `master` for now, and we will put them in `gh-pages` later. I can create the `Makefile` target for that now, or next week during release cleanup.

ostera:
Makes sense. There’s a `make docs` target already, I’ll drop the commit with the docs and add them to .gitignore when I get home 👍🏼

ostera:
@aantron should be good now 👍

aantron:
I see a `.swp` file committed, perhaps you should remove it? If it might be a problem in the future, maybe we should `.gitignore` it.

Makefile
```diff
@@ -17,6 +17,7 @@ npm-test :
 
 .PHONY : docs
 docs :
+	mkdir docs
```
aantron: IIRC this will fail if `docs/` already exists, you need `mkdir -p docs` to avoid that.

ostera: I need to stop working late at night. Thanks for catching this 🙌

aantron:
Thanks!

PR #204, opened 2018-09-27
Support references in Lists

### LINKED ISSUE #202: References in lists are broken
**Input**
```ocaml
+ {!Module.name}
```

**Expected Output**
A link to `Module.name` in a numbered list.

**Actual Output**
A numbered list with an unresolved reference to `Module.name`

**Workaround**
If I remove the list, so the input becomes simply `{!Module.name}` the reference works just fine.

Possibly related to #200

ostera:
This will recurse into List tags, but will only allow for reference links on:

-> Paragraphs
-> Modules
-> Lists

Code blocks and Verbatim tags will not be recursed since they only contain a `string`.

Should close #202.

### DIFF
--- a/test/html/expect/test_package+re/Markup/index.html
+++ b/test/html/expect/test_package+re/Markup/index.html
@@ -228,7 +228,13 @@ let bar =
          lists
         </li>
         <li>
-         can be nested.
+         can be nested
+        </li>
+        <li>
+         and can include references
+        </li>
+        <li>
+         <a href="index.html#val-foo"><code>foo</code></a>
         </li>
        </ul>
       </li>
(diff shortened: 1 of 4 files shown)


test/html/expect/test_package+re/Markup/index.html
```diff
@@ -228,7 +228,13 @@ <h2 id="lists">
...
         </li>
         <li>
-         can be nested.
+         can be nested
+        </li>
+        <li>
+         and can include references
+        </li>
+        <li>
+         <a href="index.html#val-foo"><code>foo</code></a>
```
ostera: Aaah, @aantron, this is infinitely better. Thanks for #198 ❤️

ostera:
cc @grabbou

Issue #199, opened 2018-09-24
Bring Reason Syntax up to par with Refmt v3


ostera:
I've been meaning to bring up the reason syntax printing on part with `refmt`, so I'm opening this issue to track the work that needs to be done, and we can reference it from smaller PRs until this is more or less completed.

- [ ] Compact arguments:
    `'a => 'b => 'c` should be `('a, 'b) => 'c`

- [ ] Optional arguments:
    `?l: 'a => 'b` should be `(~l: 'a=?) => 'b`

- [ ] Long signatures should break arguments into individual lines (cc @grabbou):
    `('a, 'b, 'c, 'd, ... /* ad nauseum */) => 'z` should be 
```
(
  'a,
  'b,
  'c,
  'd,
  ...
) => 'z
```

- [x] Space _around colons_ is normally a single space _after_ the colon:
  - [x] Record attribute name `{ a : 1 }` should be `{ a: 1 }`
  - [x] Object attribute name `{ a : 1 }` should be `{ a: 1 }`
  - [x] Function signature `let f : 'a => 'b` should be `let f: 'a => 'b`
  - [x] Tweened Type annotations `(a : unit)` should be `(a: unit)`

-----

If you have some more please comment below and I'll add them to the top-level list so we get some coverage 🙌 

cc/ @ryyppy @aantron

Drup:
@ostera If you want to move this forward, the new implementation should be more accessible, and thus easier to tweak for this.

PR #207, opened 2018-09-27
Fix all label separators in Reason syntax

ostera:
As it turns out, label separators seem to be pretty uniform within each
syntax, so I figured I'd pull that out of the `Syntax.Type` submodule
and solve it once for all of them.

Can be specialized if need be, but solves the issue for now.

Work on #199

cc @ryyppy @aantron

### DIFF
--- a/src/html/html_generator.ml
+++ b/src/html/html_generator.ml
@@ -109,8 +109,6 @@ module type Syntax = sig
 
     module Record : sig
       val field_separator : string
-
-      val label_value_separator : string
     end
 
     val var_prefix : string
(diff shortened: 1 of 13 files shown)


aantron:
LGTM. One request: can you rename the value? It is no longer just a label separator, but a general type annotation separator.

ostera:
@aantron `type_annot_separator` doesn't work very well because we are using it for objects and record labels as well.

I renamed to `separator`, which from context should be fairly obvious, but I'm open to better names if you can think of one 🙌

aantron:
Thanks!

Issue #155, opened 2018-07-03
Improve support-file management


rizo:
Currently this is how odoc manages support files (CSS+JS):

- https://github.com/ocaml/odoc/blob/3f1db7ad83c5be9af30b0ffebb49f7a11b45111c/src/odoc/jbuild#L10
- https://github.com/ocaml/odoc/blob/3f1db7ad83c5be9af30b0ffebb49f7a11b45111c/src/odoc/assets.ml#L12

Essentially it copies the content of the support files into OCaml modules that get linked with the odoc binary. This avoids having them as a file-system dependency making odoc effectively portable.

As we add more files (like fonts or other JS files) it will become a bit messier to do this because of all the boilerplate. I suggest to bundle the files into a directory for easier distribution but to preserve portability we might need to use something similar to https://github.com/mirage/ocaml-crunch.

rizo:
I think we need to make a decision regarding this. I'd like to add more support files to odoc and the current approach is not ideal.

@dbuenzli [noted](https://github.com/ocaml/odoc/pull/154#issuecomment-403926759) that odoc is already not portable across different compiler versions. Which to me implies that odoc is better distributed as part of a particular opam switch. Unless we want to invest time into making it portable in the future.

I believe we have two options:

1. Assume odoc is not portable and get rid of embedded support/theme files. Both would be stored as simple directories and copied over to the HTML output directory by default.
2. Integrate a tool like ocaml-crunch to distribute the support/theme files. This could either mean adding ocaml-crunch as a dependency or reimplementing it in odoc.

What do you think?

P.S. Just a reminder that in https://github.com/ocaml/odoc/pull/154 we decided to separate "assets" into "support files" and "theme files", leaving "assets" for files embedded in documentation.

aantron:
@rizo Even though odoc is not portable across compiler versions, there is still (IMO) a reason to embed the files. At some future time, we may want to suggest using odoc with BuckleScript and NPM. It will be easier to do this if odoc does not depend on directories in opam (or anywhere), at least for basic odoc usage.

BuckleScript is also usually fixed at one version of OCaml, so we would be compiling and distributing odoc binaries for the corresponding compiler version. Lack of portability is not a problem here, and indeed, I think it is a red herring for this issue.

So, I support some variant of (2).

PR #208, opened 2018-10-06
Add custom remote fonts. Small visual adjustments.

rizo:
This makes small adjustments to the CSS based on the feedback I received for the current theme. 

Custom fonts are included here as an experiment, we'll include them as static files when https://github.com/ocaml/odoc/issues/155 is resolved.

Here're some examples of packages with this change: [order](
http://odis.io/odoc-demo/order/Order/index.html), [lwt](
http://odis.io/odoc-demo/lwt/Lwt/index.html), [fmt](
http://odis.io/odoc-demo/fmt/Fmt/index.html), [cmdliner](
http://odis.io/odoc-demo/cmdliner/Cmdliner/index.html) and [base](
http://odis.io/odoc-demo/base/Base/index.html).

CC @aantron @ostera @ryyppy

### DIFF
--- a/src/odoc/etc/odoc.css
+++ b/src/odoc/etc/odoc.css
@@ -296,7 +296,7 @@ pre code {
 
 /* Indent the second line in multiline spec definitions. */
 .spec:not(.type) > code {
-  display: inline-block;
+  display: block;
   padding-left: 4ex;
   text-indent: -4ex;
 }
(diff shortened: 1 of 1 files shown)


pmetzger:
You need at least a monospace font for examples, a body font, and a display font. That's three families to start with. Each will need to be available in normal/roman, italic, bold, and likely bold italic. That's a minimum. Also, given that all of these together are tinier, combined, than a decent sized image, and get cached, I think economizing here is probably not really useful...

rizo:
Merging. We still have many details to polish but it can be done gradually. If someone isn't happy with the font choices/sizes, please open a separate issue/PR.

Issue #195, opened 2018-09-19
Publish as an `npm` installable package


ostera:
Now that `odoc` has 4.02 support and works with BuckleScript, I was thinking it'd be more reachable for people working from the js-side of the ecosystem if it was an `npm install odoc` away.

If this worked, then perhaps it could be included with the set of binaries in `bs-platform`.

Opening the issue to gather comments and feedback 🗒

**Update / To do**.
- [x] Equip repo to build `npm` package — #214 
- [ ] Figure out multi-platform packaging
- [ ] Publish `npm` package

rizo:
I think it sounds reasonable if that's the preferred installation for BuckleScript projects.

Here are some questions I have?

* Does this need to be done (and managed in the future) officially by the odoc project?
* Does this require any changes to the project structure, etc?
* How would the installation work (i.e. do we need to precompile binaries)?
* Are there any existing tools (esy?) that can help us with this?

aantron:
If we use esy, we don't need to precompile binaries. However, the installation could take a pretty long time, depending on whether the user has already installed a compatible compiler with esy.

We shouldn't have to change the project, except to add the NPM and esy boilerplate. And yes, we will have to maintain this.

aantron:
@rizo for a really simple example, see [Repromise_lwt](https://github.com/aantron/repromise_lwt). In the repo, you'll find a really simple, ordinary opam+Dune project, packaged for both regular opam and esy.

ostera:
@aantron without introducing another tool into the mix — what would be a _from first principles_ approach to publishing precompiled binaries as an `npm` package?

Given that the audience installing with `npm` _will be using 4.02.x+buckle-master_, perhaps we could only publish a precompiled bytecode with embedded runtime?

Looking for the least _magic_ 🔮approach.

aantron:
@ostera I think with an esy setup, the user won't see that esy is running during the build. Not 100% sure at this point. This will probably be the least magic approach for us. It does have the downsides of a long install time and potential build failures, though. We can minimize the risk of failures with `esy.lock`.

We should be able to publish odoc as a binary compiled to bytecode, and to make it portable, we can run it with `ocamlrun` from the BuckleScript install.

This is one of the reasons I wanted odoc to be one self-contained binary (we had a discussion about this here https://github.com/ocaml/odoc/pull/154#issuecomment-403892559).

[...]

PR #211, opened 2018-10-08
Add info do Readme about pid odoc

czystyl:
Added information about how to pin the `odoc` in ReasonML projects. I placed information below installation step because I don't want to change the maim installation section.

#195

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,12 +60,12 @@ This means we follow the same installation than above, but using the
 ```sh
 λ opam switch 4.02.3+buckle-master
 λ eval `opam config env`
-λ opam install odoc
+λ opam pin add odoc.dev git+https://github.com/ocaml/odoc
 λ odoc --version
 1.2.0
 ```
 
-> if you have problem with installing odoc by `opam install odoc` the try to use `opam pin add odoc.dev git+https://github.com/ocaml/odoc`
+> BuckleScript support currently requires the latest development version of odoc.
 
 Now with that working, we can point `odoc` to the path where BuckleScript saves
 the compiled code that we can use to generate our documentation. This path is
--- a/README.md
+++ b/README.md
@@ -51,6 +51,8 @@ The generated docs can then be found locally at
 
 ## Using `odoc` with BuckleScript/Reason
 
+> BuckleScript support currently requires the latest development version of odoc.
+
 While the BuckleScript/Reason toolchain relies on `npm`, `odoc` at the moment
 needs to be used from a working OCaml toolchain.
 
@@ -63,8 +65,6 @@ This means we follow the same installation than above, but using the
 λ opam pin add odoc.dev git+https://github.com/ocaml/odoc
 
 ```
-
-> BuckleScript support currently requires the latest development version of odoc.
 
 Now with that working, we can point `odoc` to the path where BuckleScript saves
 the compiled code that we can use to generate our documentation. This path is
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # odoc
 
-**odoc** is a documentation generator for OCaml. It reads *doc comments*,
+**odoc** is a documentation generator for OCaml. It reads _doc comments_,
 delimited with `(** ... *)`, and outputs HTML. See example output at
 [docs.mirage.io][mirage-docs].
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,6 +65,8 @@ This means we follow the same installation than above, but using the
 1.2.0
 ```
 
+> if you have problem with installing odoc by `opam install odoc` the try to use `opam pin add odoc.dev git+https://github.com/ocaml/odoc`
```
rizo: Even if people do install current stable odoc successfully with opam, Reason syntax will not be supported! I think this should say something like: "BuckleScript support currently requires the latest development version of odoc." and replace the `opam install odoc` line with the pin command instead.

README.md
```diff
@@ -1,6 +1,6 @@
 # odoc
 
-**odoc** is a documentation generator for OCaml. It reads *doc comments*,
+**odoc** is a documentation generator for OCaml. It reads _doc comments_,
```
rizo: We all like underscores better! 😛

README.md
czystyl: 👍

README.md
czystyl: This was done by my prettier, sorry for that :P

rizo: I'll leave this decision to you :)

README.md
```diff
@@ -60,11 +60,13 @@ This means we follow the same installation than above, but using the
 ```sh
 λ opam switch 4.02.3+buckle-master
 λ eval `opam config env`
-λ opam install odoc
+λ opam pin add odoc.dev git+https://github.com/ocaml/odoc
 λ odoc --version
 1.2.0
```
rizo: The output for `--version` will be different for the pinned installation. Currently it's `v1.1.1-743-gb6e4edc`. You can either remove this line entirely or put the current one as an example.

czystyl: It's better to remove this line because after the next push to develop branch it could be different.

dbuenzli: To be honest that kind of information is becoming so outdated so quickly I'm not sure it's the right place to put this.

czystyl: I was thinking about it. The first change was to stay with `opam install odoc` and put information below the instruction.

rizo: @dbuenzli are you referring to the installation by pinning or the version?

In any case, many people seem eager to try odoc with BuckleScript, this instructions will be helpful for them. I'll make sure we update the README once the next version of odoc is released.

README.md
```diff
@@ -60,11 +60,12 @@ This means we follow the same installation than above, but using the
...
 λ opam switch 4.02.3+buckle-master
 λ eval `opam config env`
-λ opam install odoc
-λ odoc --version
-1.2.0
+λ opam pin add odoc.dev git+https://github.com/ocaml/odoc
+
 ```
 
+> BuckleScript support currently requires the latest development version of odoc.
```
ostera: This feels like something we should include in the very first paragraph of this title. I'd want to know from the get go that I need the dev version of odoc.

Good catch otherwise!

README.md
dbuenzli: Pinning switch creation that includes a particular version, the toy shell script to generate the docs. etc. etc. I think this information should rather go to somewhere were bucklescript users refer to when they need documentation about their platform.

README.md
czystyl: 👍

README.md
rizo: Conceptually I agree, it's entirely out of odoc's scope to worry about BuckleScript's integration. In practice though, odoc is being only experimentally used by some packages (like [bs-react-native](https://github.com/reasonml-community/bs-react-native#readme)). It's currently not the officially recommended documentation generator for BuckleScript (it still uses ocamldoc AFAIK).

When things get more stable I would expect to have a proper external integration with BuckleScript's build system on the same level as we have with dune. For now I personally don't see any risks in _provisionally_ helping with the integration as long as it stays out of odoc's core.

Submitting a PR to BuckleScript's website with instructions to use odoc is another option. I don't know if @ostera and @aantron considered this when the BuckleScript section/script were added.

In any case, I will merge this PR as it's both harmless and useful. Further decisions can be discussed separately.

rizo:
Merging, thanks @czystyl!

czystyl:
@rizo No problem :)

PR #214, opened 2018-10-10
Setup NPM Packaging

ostera:
Opening this PR for visibility/referencing, but I'll be modifying the history heavily until I get something I'm comfortable with.

Setting up `esy` seemed straightforward, although adding dependencies one by one manually is always annoying.

This is missing a few things:

- [x] Pin tyxml/markup/etc to their appropriate versions (some tests are failing due to sexp indentation)
- [x] Run on Travis for switch 4.02.3
- [x] Trim down package (@jordwalke's suggestion)
- [x] ~~Publish from Travis for switch 4.02.3 on master branch tag~~

But so far I've managed to install the npm package locally and it runs alright 👍 

```sh
ostera/odoc λ make npm
esy install
info install 0.2.11
esy make build
dune build
esy release
info release 0.2.11
dune build
info Creating npm release
# ...so much output here exporting packages
info Exporting ocaml-4.2.3005-a85217cf: done
info Configuring release
info Done!

ostera/odoc λ cd _release && npm --global install .
/usr/local/bin/odoc -> /usr/local/lib/node_modules/odoc/bin/odoc

> odoc@1.2.0 postinstall /usr/local/lib/node_modules/odoc
> node ./esyInstallRelease.js

importing: opam__slash__uutf-1.0.1-1e184456
# another good deal of imports here
importing: esy_ocaml__slash__substs-0.0.1-441fe663
Done!
+ odoc@1.2.0
updated 1 package in 4.38s

odoc/_release λ /usr/local/lib/node_modules/odoc/bin/odoc
Available subcommands: compile, html, html-fragment, support-files, css, compile-deps,
html-deps, compile-targets, html-targets
See --help for more information.

odoc/_release λ /usr/local/lib/node_modules/odoc/bin/odoc --version
%%VERSION%%
```

If you can please give it a go!

cc/ @aantron @rizo @lpw25 @ryyppy @grabbou @avsm @ulrikstrid 

ref #195

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
-language: generic
+language: node_js
+node_js:
+  - "10"
 
 env:
   - OCAML=4.02.3
@@ -8,25 +10,19 @@ env:
   - OCAML=4.05.0
   - OCAML=4.06.1
   - OCAML=4.07.0
+  - ESY_BUILD=YES ESY__CACHE=/home/travis/.esy
 
-before_script:
-  - sudo add-apt-repository -y ppa:robert7/tidy-html5
-  - sudo apt-get update
-  - sudo apt-get install tidy
-  - wget https://github.com/ocaml/opam/releases/download/2.0.0/opam-2.0.0-x86_64-linux
-  - sudo mv opam-2.0.0-x86_64-linux /usr/local/bin/opam
-  - sudo chmod a+x /usr/local/bin/opam
-  - "opam init -y --compiler=$OCAML --disable-sandboxing"
-  - eval `opam config env`
-  - opam --version
-  - ocaml -version
+cache:
+  directories:
+  - /home/travis/.esy
+  - /home/travis/.nvm
 
-script:
-  - opam pin add -y --no-action odoc .
-  - opam install -y --deps-only odoc
-  - make test
-  - opam pin add -y --dev-repo dune
-  - make dune-test
+install:
+  - ./test/ci/install-deps.sh
+  - "if [[ $OCAML ]]; then eval `opam config env`; fi"
+  - ./test/ci/print-versions.sh
+
+script: ./test/ci/build.sh
 
 notifications:
   email:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,13 +14,18 @@ env:
 
 cache:
   directories:
+  - /home/travis/.opam
   - /home/travis/.esy
   - /home/travis/.nvm
+  - ./_build/default
 
 install:
   - ./test/ci/install-deps.sh
   - "if [[ $OCAML ]]; then eval `opam config env`; fi"
   - ./test/ci/print-versions.sh
+
+before_script:
+  - ./test/ci/clean-scratch.sh
 
 script: ./test/ci/build.sh
 
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,19 @@
 build :
 	dune build
 
+.PHONY : npm-package
+npm-package : npm-build
+	esy release
+
+.PHONY : npm-build
+npm-build :
+	esy install
+	esy build
+
+.PHONY : npm-test
+npm-test :
+	esy make test
+
 .PHONY : test
 test : build
 	dune build @test/parser/runtest --no-buffer -j 1
--- a/package.json
+++ b/package.json
@@ -1,0 +1,59 @@
+{
+  "name": "odoc",
+  "version": "1.2.0",
+  "description": "The OCaml/Reason Documentation Generator",
+  "repository": "https://github.com/ocaml/odoc",
+  "authors": [
+    "Thomas Refis <trefis@janestreet.com>",
+    "David Sheets <sheets@alum.mit.edu>",
+    "Leo White <leo@lpw25.net>"
+  ],
+  "esy": {
+    "build": "make build",
+    "release": {
+      "releasedBinaries": [
+        "odoc"
+      ],
+      "deleteFromBinaryRelease": [
+        "alcotest",
+        "bisect_ppx",
+        "js-build-tools",
+        "lambdasoup",
+        "markup",
+        "oasis",
+        "ocaml-*",
+        "ocamlfind",
+        "sexplib"
+      ]
+    }
+  },
+  "dependencies": {
+    "@opam/alcotest": "0.8.3",
+    "@opam/astring": "0.8.3",
+    "@opam/bisect_ppx": "1.3.4",
+    "@opam/bos": "0.2.0",
+    "@opam/cmdliner": "1.0.2",
+    "@opam/cppo": "1.6.5",
+    "@opam/dune": "1.3.0",
+    "@opam/fpath": "0.7.2",
+    "@opam/js-build-tools": "113.33.04",
+    "@opam/lambdasoup": "0.6.3",
+    "@opam/markup": "*",
+    "@opam/oasis": "0.4.11",
+    "@opam/ocamlfind": "1.8.0",
+    "@opam/result": "1.3",
+    "@opam/sexplib": "*",
+    "@opam/tyxml": "4.2.0",
+    "ocaml": "4.2.3005"
+  },
+  "resolutions": {
+    "@opam/markup": "aantron/markup.ml:markup.opam#9f8e77",
+    "@opam/sexplib": "janestreet/sexplib#9e6e44e"
+  },
+  "devDependencies": {
+    "@opam/merlin": "^3.0.3",
+    "ocaml": "4.2.3005"
+  },
+  "license": "ISC",
+  "private": false
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # Jbuilder output.
-_build/
+_build
 *.install
 *.merlin
 
@@ -11,3 +11,7 @@ scratch/
 
 # opam v2
 _opam/
+
+# Esy outputs
+_release
+_esy
--- a/esy.lock.json
+++ b/esy.lock.json
@@ -1,0 +1,880 @@
+{
+  "hash": "37f4154c38037928d007ef913df46d06",
+  "root": "root@path:./package.json",
+  "node": {
+    "root@path:./package.json": {
+      "record": {
+        "name": "root",
+        "version": "path:./package.json",
+        "source": "path:./package.json",
+        "files": [],
+        "opam": null
+      },
+      "dependencies": [
+        "@opam/alcotest@opam:0.8.3", "@opam/astring@opam:0.8.3",
+        "@opam/bisect_ppx@opam:1.3.4", "@opam/bos@opam:0.2.0",
+        "@opam/cmdliner@opam:1.0.2", "@opam/cppo@opam:1.6.5",
+        "@opam/dune@opam:1.3.0", "@opam/fpath@opam:0.7.2",
+        "@opam/js-build-tools@opam:113.33.04", "@opam/lambdasoup@opam:0.6.3",
+        "@opam/markup@github:aantron/markup.ml:markup.opam#9f8e77",
+        "@opam/merlin@opam:3.2.1", "@opam/oasis@opam:0.4.11",
+        "@opam/ocamlfind@opam:1.8.0", "@opam/result@opam:1.3",
+        "@opam/sexplib@github:janestreet/sexplib#9e6e44e",
+        "@opam/tyxml@opam:4.2.0", "ocaml@4.2.3005"
+      ]
+    },
+    "ocaml@4.2.3005": {
+      "record": {
+        "name": "ocaml",
+        "version": "4.2.3005",
+        "source":
+          "archive:https://registry.npmjs.org/ocaml/-/ocaml-4.2.3005.tgz#sha1:6f718145cc874d09e631538708af5eecc3b77c50",
+        "files": [],
+        "opam": null
+      },
+      "dependencies": []
+    },
+    "@opam/yojson@opam:1.4.1": {
+      "record": {
+        "name": "@opam/yojson",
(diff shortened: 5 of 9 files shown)


jordwalke:
HI, I'm curious if this is the kind of package that actually benefits from being a globally installable  command line utility from npm? (I am not yet familiar with odoc but eager to learn).

I did notice that you did not trim out any of the binary artifacts, so here's an example of how to do that. It makes the prebuilt release much leaner: https://github.com/esy-ocaml/hello-reason/blob/master/package.json#L13

dbuenzli:
> HI, I'm curious if this is the kind of package that actually benefits from being a globally installable command line utility from npm? (I am not yet familiar with odoc but eager to learn).

I'd say this is unlikely `odoc` which relies on `cmt/cmti/cmi` files is tied to the version of the compiler you used to generate these artefacts.

ostera:
Hej @jordwalke as it is right now using it at all requires `opam`, which is a hindrance if you're using an `bsb`-only setup.

I'll grant you that the current usage flow is far from ideal (as you can see from the suggested script in the readme), but having a release pipeline for `npm` makes it easier to iterate and gather feedback from the other side of the community.

And 👍 will trim down the package when I get home 📦

aantron:
@ostera I don't think this package requires a *switch* that users need to be aware of. IIRC we can just choose the proper compiler version in `esy.json`, so that the binary is built on 4.02.3, and then it can be used with BuckleScript.

I don't think we need an opam switch around at runtime.

ostera:
@aantron not at runtime, but they won't be able to generate docs from `.cmt[i]`s built with a different switch.

rizo:
Once again working on BuckleScript and dune projects on the same machine is going to be hard with global installations...

aantron:
@ostera It's not the switch but the compiler version that matters (actually, the format of `.cmt[i]` files). As long as the OCaml version our build internally uses to build the odoc binary is the same as what's being used by BuckleScript (4.02.3), the user doesn't need to know about it. They only need an easy way of finding the binary, probably through NPM.

.travis.yml
```diff
@@ -1,8 +1,11 @@
-language: generic
+language: node_js
+node_js:
+  - "10"
 
 env:
   - OCAML=4.02.3
   - OCAML=4.02.3+buckle-master
+  - OCAML=4.02.3+buckle-master ESY_BUILD=YES ESY__CACHE=/home/travis/.esy
```
jordwalke: Can you help me understand what this accomplishes?

.ci/install-esy.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [[ $ESY_BUILD == YES ]]; then
+  npm --global install esy
```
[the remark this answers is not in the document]
andreypopp: yeah, `esy@0.3.x` would be relatively safe (considering this project requires features new in 0.3.x, otherwise it's better to pin to 0.2.x).

.travis.yml
ostera: We need to build `odoc` using this version of the compiler to be able to read `.cmti` files compiled with it.

ostera:
Alright! This looks a little tidier now, let's do another round of reviews ⚔️ — I took the liberty to optimize the CI process, not to have to wait ~7 mins on every change, but instead ~2. Feedback welcome there too (I'm open to submitting a separate PR for those changes if need be).

@aantron: Just like with the automatic publishing of docs, let's keep this publishing manual. Any advice on caching the opam/ocaml installations any futher?

@jordwalke, @andreypopp: Could you provide insights on how to properly cache the build directory of esy? I'm noticing that `dune` runs again, rebuilding and retesting the whole project rather than doing it incrementally.

@rizo: is there anything you'd like changed about this PR that would make you more comfortable with the way we would publish the `odoc` binary? I've been thinking about this and I think the approach I'd like _best_ would be an `odoc` release that understands several compiler versions (so you only need one for every compiler), but I might be missing some underlying complexity of developing/building/distributing such a binary.

@ryyppy @grabbou @dbuenzli: would like to hear your thoughts on this too 🗣

ostera:
> @jordwalke: I just wanted to avoid people getting a really bad error when they try to run odoc on
> a local Dune project on 4.06.

Shouldn't they be running it with `esy dune build @doc`? which should use the opam installed version that's appropriate for their compiler version, if I understood that particular `esy` usage.

.ci/install-esy.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [[ $ESY_BUILD == YES ]]; then
+  npm --global install esy@0.3.x
+  esy --version
+else
+  echo "Nothing to be done."
```
rizo: Is this `else` branch needed? Seeing this message in the log is not informative IMO.

.ci/install.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
rizo: This might be some convention but I feel like this script should be called `setup`, `init` or `install-deps`.

.travis.yml
```diff
@@ -1,32 +1,32 @@
...
-  - wget https://github.com/ocaml/opam/releases/download/2.0.0/opam-2.0.0-x86_64-linux
-  - sudo mv opam-2.0.0-x86_64-linux /usr/local/bin/opam
-  - sudo chmod a+x /usr/local/bin/opam
-  - "opam init -y --compiler=$OCAML --disable-sandboxing"
+cache:
+  directories:
+  - /home/travis/.opam
+  - /home/travis/.esy
+  - /home/travis/.nvm
+  - ./_build/default
```
rizo: Is this desirable? Do we want to keep cache even for tests executions across builds?

Makefile
```diff
@@ -2,6 +2,19 @@
 build :
 	dune build
 
+.PHONY : npm
```
rizo: `make npm` seems a bit ambigious to me. Again maybe this is some sort of convention but I’d call it `npm-publish`.

Makefile
```diff
@@ -2,6 +2,19 @@
...
+npm : npm-build
+	esy release
+
+.PHONY : npm-build
+npm-build :
+	esy install
+	esy build
+
+.PHONY : npm-test
+npm-test :
```
rizo: This seems very meta (ie a make command that calls esy that alls make again). How is it different to `make test`?

package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "odoc",
+  "version": "1.2.0",
+  "description": "The OCaml/Reason Documentation Generator",
+  "main": "index.js",
```
rizo: Hm, are you trying to sneak in some uncommited JS? 😛

rizo:
>  I might be missing some underlying complexity of developing/building/distributing such a binary

The ability to work with `cmt`/`cmti`/`cmi` files produced by different compiler versions is currently hard because of inconsistencies across typed parsetrees. We would need a typed version of ocaml-migrate-parsetree which doesn’t exist AFAIK. The ad-hoc approach would be too painfull.

aantron:
> The ability to work with `cmt`/`cmti`/`cmi` files produced by different compiler versions is currently hard because of inconsistencies across typed parsetrees. We would need a typed version of ocaml-migrate-parsetree which doesn’t exist AFAIK. The ad-hoc approach would be too painfull.

@rizo This might not actually be a problem for odoc, in principle.

odoc already has enough code to load `Typedtree`s from compiler versions 4.02-4.07, and largely convert them into its own representation. (`src/model`).

This leaves the remaining direct uses of `Typedtree`. `Typedtree` is used only in the loaders themselves `cmt.ml`, `cmti.ml`, and `ident_env.ml`. That's where (almost?) all of our conditional compilation is, too, but it doesn't look too bad.

It may be possible to link loaders for all OCaml versions into a single odoc binary, and reconcile them somehow. The big issue would be how to avoid duplicating the vast majority of the shared code. Perhaps we can do it with a functor or some preprocessing. We would also likely need to vendor a copy of the relevant files in compiler-libs for each version of the compiler.

.travis.yml
jordwalke: I had the same question. You only want to cache builds of your dependencies typically (the `~/.esy/` directory.

rizo:
@aantron Thanks for clarifying that. I’m not sure I like the megabinary idea :) But having cross-version portability would be very nice indeed.

.ci/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
aantron: I suggest to rename directory `.ci/` to `test/ci/`, so it doesn't have to be hidden, and doesn't clutter the GitHub repo.

.travis.yml
aantron: @ostera AFAIK esy will use its own internal oap switch to build the `odoc` binary.

.travis.yml
```diff
@@ -1,32 +1,32 @@
...
+
+install:
+  - ./.ci/install.sh
```
aantron: I recommend not breaking this up into files. If you'd like to move it to a shell script (for conditionals?), I suggest to create one shell script, and do things there. Perhaps one for building and one for installing, if caching happens between those two steps.

With many scripts, and some commands in the `.travis.yml` file, people have to look through a lot of little files to understand what is going on.

Makefile
aantron: I don't think this should be called `npm-publish`, because, as I understand, it just packages things in a local `_release` directory. But it shouldn't be called `npm` either.

Makefile
aantron: @rizo This runs `make test` in the esy environment. It doesn't seem too bad.

package.json
aantron: I believe that will be a generated file.

jordwalke:
> > @jordwalke: I just wanted to avoid people getting a really bad error when they try to run odoc on
> > a local Dune project on 4.06.
> 
> Shouldn't they be running it with `esy dune build @doc`? which should use the opam installed version that's appropriate for their compiler version, if I understood that particular `esy` usage.

Perhaps but when BS upgrades to 4.06 the same thing will happen and we'll need to clearly make sure the user knows what version their globally installed odoc is compatible with. At least a good error message is in order. I don't have too much to add besides that. Happy to see odoc get into more hands either way.

.ci/install-esy.sh
ostera: It looked strange to see the script marked as ran but _no output whatsoever_. Perhaps I can change the text to `"Skipping esy installation."` ?

Makefile
ostera: `make npm-release` ?

package.json
ostera: This was generated. Should be removed from the `package.json`!

.travis.yml
ostera: @aantron a small problem with this approach is that the `eval $(opam env)` call has to be made from an individual line in the `.travis.yml` file to affect the rest of the build.

`eval` from within scripts only affects the current environment, not it's parent.

We can bundle the `install.sh` and `install-esy.sh`, but we'd need a separate script that is run after the `eval` to be able to output the `ocaml/opam -[-]version` commands.

```yaml
install:
  - ./.ci/install.sh
  - eval $(opam env)
  - ./.ci/print-versions.sh
```

Let me know what you think.

.travis.yml
ostera: Coming from `bazel` I'm used to letting the build tool cache very aggressively — assuming we trust `dune` to re-build and re-test incrementally, should this be a problem?

.travis.yml
ostera: @aantron @jordwalke I left that in there because I was installing opam/ocaml even for the esy build. I realize now that it's not needed at all, so I will remove it and make the installation optional.

.ci/install-esy.sh
ostera: Ended up joining the `install` scripts after removing the `OCAML` env variable from the esy build. This is no longer an issue.

Makefile
ostera: Alternatively, we can just call `esy make test`, but I fear that from looking at the files we currently have it won't be obvious that this should be run (at some point) to make sure the build works as an `npm` package.

We can also document it in the contribution guidelines.

.ci/install.sh
ostera: I like `install-deps`. Conveys a little more info about what's being installed.

.travis.yml
rizo: I would be fine with this if we cleaned the html `_scratch` directory between test executions. Otherwise the previously generated files that are not generated in next runs might interfere with the result (i.e., it should be an error in this case).

ostera:
@aantron @rizo I've addressed most of your suggestions. Let me know how you want to proceed!

I understand there's an open question of how would this `odoc` installation work with `opam` switching compilers, but from a practical point of view:

1. `dune`/`esy` users will run it through `dune build @doc`,
2. `npm` users can install it locally and run it via runscripts or `./node_modules/.bin/odoc` 

So I don't see it as a blocker at the moment. Happy to be shown wrong! 🙌

aantron:
@ostera I don't think there's any issue. The esy/NPM odoc does not interfere with opam switches. It's how you described in your list.

Makefile
aantron: `-release` has the danger of being interpreted as the verb, which makes it seem dangerous. How about `npm-package`?

.travis.yml
```diff
@@ -1,4 +1,6 @@
-language: generic
+language: node_js
+node_js:
+  - "10"
```
aantron: What is using this?

ostera: Installing `esy` requires a node environment, which we can either manage ourselves or let Travis manage for us.

aantron:
Thanks @ostera and the reviewers!

PR #216, opened 2018-10-12
Run CI on Linux/Windows/macOS

ostera:
Opening this for visibility and checking Travis logs. History might change heavily until it's ready for proper reviews.

cc/ @ryyppy @avsm @aantron


ryyppy:
So cool that Travis now supports Windows :)

aantron:
I recommend not building on every OCaml version on every platform, because whenever Travis is slow to start build rows, we will wait a long time. Also, there are spurious failures, such as temporary upstream problems. Having more build rows means more builds to manually restart when these occur.

The highest return is usually from building on every OCaml version on one platform (inevitably Linux), then building on one version only on each remaining platform.

So, I suggest to have this PR add only (1) opam+OCaml 4.07 and (2) esy+OCaml 4.02 to (A) macOS and (B) Windows.

Of course, if there is a reason to think a particular combination from the entire matrix tends to be especially buggy, we should add that, too. But generally, we get no extra information from building the remainder of the cross product of operating systems and OCaml versions.

avsm:
It's great to see Windows appear in Travis indeed! Agree with @aantron about reducing pressure on the build matrix -- just smoke tests on those two platforms to start with will get us going.

ostera:
Hey! Yes, the lowest hanging fruit was adding the OSes to the matrix, all the builds get generated off that automatically.

I’ll have a look tonight at how to reduce/limit but I’m not sure if conditional jobs are possible with matrix generated jobs — this might have changed since last I tried.

aantron:
I believe you can do it like this: https://github.com/ocsigen/lwt/blob/ac8f7b15733bd9e3992764736d8860ca6c9fa379/.travis.yml#L5-L24

ostera:
Slowly getting there! macOS build works 👌 — Windows build is causing some issues, and I think might be blocked by https://github.com/esy/esy/issues/526

cc/ @ryyppy @ulrikstrid

.travis.yml
```diff
@@ -2,18 +2,37 @@ language: node_js
...
+    - os: linux
+      env: OCAML=4.07.0
+    - os : linux
+      env: ESY_BUILD=YES ESY__CACHE=/home/travis/.esy
 
 cache:
   directories:
+  - /usr/local/Cellar/opam
+  - /usr/local/Cellar/tidy-html5
   - /home/travis/.opam
```
aantron: macOS users are under something like `/Users/travis` IIRC. I think you may be able to get this working on both Linux and macOS with `$HOME/.opam`, etc., instead. Not sure about Windows yet.

Alternatively, only for opam, you can try a local opam switch. But that won't help with `.esy` and `.nvm`, so it's probably not worth it.

ostera:
We seemed to be blocked by this: https://github.com/esy-ocaml/ocaml/issues/7

According to @jordwalke, @bryphe might be working on it. I'll reach out and see if it's something we can help with.

bryphe:
Thanks @ostera ! Yep, working on it now - I'm working on fixing the Windows issues with the current build script for the 4.02 compiler, and we also need some prebuilt flexdll binaries - working on that here: https://github.com/esy-ocaml/flexdll/pull/2

Will keep you posted - looking forward to being able to use `odoc` on Windows!

bryphe:
I've created the pre-built flexdll binaries in this NPM package: <https://www.npmjs.com/package/flexdll-prebuilt> - so that pre-req to getting the 4.02 compiler working is good to go.

Also have this PR out to get the `4.02` compiler working w/ those prebuilt binaries: <https://github.com/esy-ocaml/ocaml/pull/10> . Once that is in, we'll need to update our ocaml NPM package.

bryphe:
We brought in both PRs and @andreypopp released a new npm version w/ the fixes: https://www.npmjs.com/package/ocaml/v/4.2.3006

You'll need to update your `package.json` to pick up the latest version:
```
    "ocaml": "~4.2.0"
```
(in both the `dependencies` and `devDependencies` section).

Build gets pretty far! But I see one failure with `js-build-tools` at the moment, seems like a bug with our prebuilt `flexdll`. Investigating that issue.

aantron:
@bryphe We don't need that specific commit, I think. We need version at least 113.33..00, see this commit message for reasoning and a link to the sexplib changelog:

https://github.com/ocaml/odoc/commit/b6d24728dd3842ac87063d00e0c456288147b1b3#diff-e95a0c70d3bd8a7570f9603aeb137602

Taking a glance at the actual build issue.

bryphe:
Thanks @aantron for the info!

It appears the error message is coming from `oasis`:
https://github.com/ocaml/oasis/blob/6113a64d86502d79b2cfad37a2339db8636b909f/src/oasis/OASISHostPath.ml#L99 That patch seems to address some of the pathing issues in `OASISHostPath.ml`.

I wonder if we need to bring over this patch to esy's override, for Windows? https://github.com/fdopen/opam-repository-mingw/blob/master/packages/oasis/oasis.0.4.11/files/oasis-0.4.11.patch

Looks like the newer versions of `sexplib` use dune instead of oasis, so they are not be impacted by this issue. I verified this as I'm able to build the latest `sexplib` against `4.06` compiler w/o a problem on Windows.

I'll try out that patch on `oasis` and see what happens.

aantron:
...and that looks like a really old version of sexplib, pretty close to 113.33.00 (which is from April 2016). AFAIK that's just a lower bound constraint for us, so if you can make the build work with a newer version of sexplib, please go ahead (or we can try). Maybe @ostera can correct me.

I think to avoid `setup.ml`, you can constrain sexplib to at least v0.9.0, or use a commit after janestreet/sexplib@3f6ad53415eedfc3c2ab44e9da3429a1e8fda424. Note that he v.0.x.y versions are sorted after 113.x.y.

odoc itself (on Linux, with opam) works with the latest sexplib, and anything down to 113.33.00. In fact, I think it works with even older versions, and we need 113.33.00 just for test cases. So, this is only an issue for NPM and/or the new builds.

aantron:
Great, we cross-posted. Yeah, if you can use a later (try the latest) sexplib, then it should avoid those problems. I'm not sure if it will introduce something else (not sure if @ostera chose that commit deliberately). CI should reveal that, though, so there shouldn't be harm in trying.

bryphe:
Cool! I was able to get `odoc` building on Windows with `ocaml@4.06` and using the latest `sexplib` (v0.11.0):

```
E:\odoc [bryphe/windows-4.6 +1 ~2 -0 !]> esy b
info esy build 0.3.4
dune build
E:\odoc [bryphe/windows-4.6 +1 ~2 -0 !]> .\_build\install\default\bin\odoc.exe --version
%%VERSION%%
E:\odoc [bryphe/windows-4.6 +1 ~2 -0 !]> .\_build\install\default\bin\odoc.exe --help
NAME
       odoc

SYNOPSIS
       odoc COMMAND ...

COMMANDS
       compile
           Compile a cmti, cmt, cmi or mld file to an odoc file.
```

These are the changes I made to the `package.json` to support this: https://github.com/bryphe/odoc/commit/fe9b93c5910ad9e5d938ebf79e9fcc2076213e03

Next, I'll test including the oasis fixes in our https://github.com/esy-ocaml/esy-opam-override repo to see if that unblocks the 4.02 issue, or if there are any other issues beyond this.

aantron:
I think we have to use a 4.02 compiler in the esy builds, because, in those builds, we are trying to produce a binary that will work with BuckleScript. We need to link against a 4.02 version of `compiler-libs` in order to read 4.02.3 (BuckleScript) `.cmt` and `.cmti` files.

bryphe:
I was able to make progress here - have `odoc` building against the 4.02 compiler - I opened a PR for reference (#222), but feel free to simply incorporate the changes that make sense.

Needed a couple fixes:
- Ported over the windows patch for oasis here: https://github.com/esy-ocaml/esy-opam-override/pull/32 (and picked it up)
- The `sexplib` still had an additional issue where it depended on `cpp` being in the path. We don't have that in our cygwin environment - only the mingw toolchain is used. So I tweaked the `configure` script to check if the mingw compiler is available, and to prefer using that version of cpp. There might be better ways to handle this, but it unblocks the build across platforms:  https://github.com/bryphe/sexplib/commit/548f94de30d9bb484b4572d71e64522f1cc1e965548f94de30

I picked up that commit in the resolution, but perhaps you'd want to move it for another fork. @andreypopp also might have ideas for a better way to fix that using esy's new overrides functionality.

ostera:
@bryphe I've copied your changes over to this branch, to see how Travis does with them in the context of this CI setup.

If everything's green ✅, I'll squash this one so that changes, discussion, and CI results live all in one place.

However, the build is still failing flakily for 4.07 — and when it's not flaky it fails on running `opam` in the `git-bash` environment with a:

```
./test/ci/install-deps.sh: line 41: /c/program files/git/usr/bin/opam: cannot execute binary file: Exec format error
```

Need to look into that or disable that one build for now.

bryphe:
Thanks @ostera for porting over the changes! 👍 

> However, the build is still failing flakily for 4.07 — and when it's not flaky it fails on running opam in the git-bash environment with a:

I'm not sure about the OPAM issue, but we don't support the 4.07 compiler for `esy` yet - have this issue tracking:  https://github.com/esy-ocaml/ocaml/issues/8

The 4.06 compiler should work, though, if that's an option. Otherwise, once https://github.com/esy-ocaml/ocaml/issues/8 is addressed, we could turn on the 4.07 esy config here.

aantron:
@bryphe I think we are sticking with opam only for the 4.07 build, and using esy only with 4.02.3. So no pressure from us on esy+OCaml 4.07.

bryphe:
> @bryphe I managed to get something that does build quite a bit, but fails at building dune 😞

Progress! 😎 

> Do you have any thoughts on this? I'm looking into permission issues due to the _build folder being cached by Travis.

Ya, this does seem like a permission issue with the folder. You could try not caching it, to test if that is the root cause.

I spent some time with Travis's Windows build the other day and it still seems like there are some kinks to be worked out, so I can't really recommend Travis for Windows for now.

So I'd recommend, if you want something that works right away, to try either:
- __Appveyor__ - this has been our go-to Windows solution for a while. Works great with `esy`! Here's an example `appveyor.yml` I use for one of my projects: https://github.com/bryphe/reason-glfw/blob/master/appveyor.yml  The OCaml compiler itself uses Appveyor, so there's a precedent for it: https://ci.appveyor.com/project/avsm/ocaml

- __Azure CI__ - we're looking at migrating to Azure CI's infrastructure. Both esy and fastpack are leveraging this. Our config for esy is here: https://github.com/esy/esy/tree/master/.ci It has great cross-platform build support, and allows for even more advanced 'pipelines' - like you can run Windows/Mac/Linux builds, and then run _another_ job to take the build output from those and bundle it. Better concurrency than any other free offering. Only downside at the moment is no cache story that I'm aware of (@ulrikstrid had some ideas here though).

I have experience with both of those so I can help you get those set up if you'd like, @ostera

aantron:
@ostera Looks like this PR is stuck on the Windows build, but everything else works. Would you be ok with splitting out the non-Windows stuff and merging it, and we can do Windows in a separate PR later? I can do the work on this if you agree with me doing it :)

aantron:
I guess the remainder of the PR will just be about adding a macOS build, then.

ostera:
@aantron I've been rerunning the jobs for the windows+esy build and the errors come and go, mostly about some dependencies down the line that can't be build. It still looks like a small caching issue, but I haven't found time to work on it again due to vacations and moving apartments.

That said, feel free to  split it and figure out the windows build separately. Perhaps as @bryphe suggests, on Appveyor/Azure CI.

Issue #212, opened 2018-10-09
Release 1.3.0 near 15 October 2018


aantron:
We should start releasing again, so let's do it in one week, per https://github.com/ocaml/odoc/issues/129#issuecomment-427837523.

aantron:
Sent to opam repo in https://github.com/ocaml/opam-repository/pull/12811. We'll probably do the NPM release seperately, or start doing it from a later version.

Issue #210, opened 2018-10-08
[meta] Roadmap


ostera:
**Why.** There's quite a few things ongoing/planned at the moment but there isn't very good visibility of _what's next_ or dependencies across tasks. Grouping issues by milestones and resolving dependencies enables mainly 3 things:

1. current contributors can focus their efforts independently and efficiently,
2. new contributors get a clear picture of what they can jump on, and
2. odoc can communicate with stakeholders uniformly about what to expect in the near future

I talked to @aantron about this and he suggested using the Wiki when I proposed using Github Projects. In the past I've talked about this with @rizo as well, and recently with @ryyppy. Either way, I'll start this issue to discuss and we can move things to either tool.

**Proposal.** Here's some milestones with their respective issues/prs; dependencies are not by list order but indicated in each item.

----

### Milestone: Reason/BuckleScript-ready
**Brief.** Cut a release that is distributed in `npm` and lays the foundation for promoting `odoc` as the primary documentation tool for the Javascript side of the ecosystem.

- [ ] #195: Publish as an `npm` installable package.
   Might depend on #155 and using `ocaml-crunch`, and for easier adoption will require a high-level running script as suggested in #137. This script could be included only in the `npm` package and deal with other things like #209
- [ ] #199: Bring Reason Syntax up to par with Refmt v3 — this has been a problem for `bs-react-native`, who have started to use odoc but have stumbled upon inconsistencies that make the docs harder to go through. Related to #129 
- [ ] #129: Reason/OCaml syntax switching — I know @ryyppy has been further exploring this

### Milestone: Better Usability
**Brief.** Improve readability and accessibility of content in generated docs and in odoc's documentation itself by including common documentation features and facilitating guides for writing good documentation.

- [ ] #124: Docs-wide search
- [ ] #125: Make sure in-page search works, despite any markup
- [ ] #127: Improve navigation (TOC, index, etc.)
- [ ] #128: Link from docs to source files
- [x] #146: Document .mld files. — Having these documented makes an incredible difference for documentation writers, possibly resolvable through #203.
- [ ] #120: Document odoc, help people write good docs — linked to #146 and #203 
- [x] #161: Restore module summary for modules

### Milestone: Better Interop
**Brief.** Define a standard interoperability mechanism to allow other features to be built outside of the core of odoc.

- [ ] #137: odoc should drive itself
- [ ] #196: Stable structured output (a.k.a `odoc json ...`)
- [ ] _Missing Issue: @rizo's package proposal?_

[...]

PR #217, opened 2018-10-16
Implement new heading parsing logic

rizo:
Fixes https://github.com/ocaml/odoc/issues/141.

This took me a while to implement. I wasn't sure what's the best way to solve the heading-level problem in a consistent way. I think the proposed solution is relatively simple to understand, compatible with the current tradition and allows for good integration with other formats (like LaTeX or markdown).

The proposed approach can be formulated as:

1. In odoc's internal model, level-0 headings represent document titles for both API references (`mli` files) and documentation pages (`mld` files).
2. All odoc documents must have a level-0 heading (_i.e._ a title).
    - The level-0 heading for API references is automatically generated based on the module or module type name.
    - The level-0 heading for documentation pages must be manually defined by users.
3. Section headings start with level-1.
    - Subsection and subsubsection headings are level-2 and level-3, respectively.
4. New paragraph and subparagraph headings are added, inspired by the LaTeX's document model.
    - Unlike section headings, these are not meant to be included in the TOC.
5. The generated HTML is "shifted" by one: `{1 ...}` is `<h2>...</h2>`, `{2 ...}` is `<h3>...</h3>`, etc. The title heading `{0 ...}` is rendered as `<h1>...</h1>`.
    - This respects the [the HTML5 outline algorithm](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines#The_HTML5_outline_algorithm).

I also considered this in the scope of other markup formats. The following table represents the mapping where `#` and "Heading" represent odoc's model:

| # | Heading        | HTML                          | LaTeX                 |   Markdown   |
|---|----------------|-------------------------------|-----------------------|--------------|
| 0 | Title          | `<h1>...</h1>`  | `\chapter{...}`       | `# ...` |
| 1 | Section        | `<h2>...</h2>`  | `\section{...}`       | `## ...`      |
| 2 | Subsection     | `<h3>...</h3>`                | `\subsection{...}`    | `### ...`     |
| 3 | Subsubsection  | `<h4>...</h4>`                | `\subsubsection{...}` | `#### ...`    |
| 4 | Paragraph      | `<h5>...</h5>`                | `\paragraph{...}`     | `##### ...`   |
| 5 | Subparagraph   | `<h6>...</h6>`                | `\subparagraph{...}`  | `###### ...`  |

These mappings can be used in case we decide to add a LaTex backend or potentially a markdown frontend to odoc.

**Changes**:
- Allow `{1 ...}` headings for sections.
- Use `{0 ...}` for page titles.
- Shift heading levels by +1 in HTML output.
- Write documentation for documentation pages. (CC @ostera)
- Add "paragraph" and "subparagraph" headings (inspired by LaTeX).
- Update all tests to reflect the changes.

CC @aantron @dbuenzli


src/odoc/pages.mld
```diff
@@ -0,0 +1,68 @@
...
+
+For example, if you have a page called [my_page.mld], you can create a link to
+it with [{{:my_page.html} My page}] in your [index.mld] or anywhere else in
+your documentation.
+
+{2 Manual usage}
+
+Generic [odoc] build instructions apply to documentation pages. You can compile
+[mld] files manually with [odoc compile] (this is described in detaild in
+{{:manual_usage.html} Invoking [odoc] manually}).
```
rizo: I just realised that this link is now broken. @otera where should we point in this case?

src/odoc/pages.mld
```diff
@@ -0,0 +1,68 @@
+{0 Documentation pages}
```
ostera: Any chance we can elaborate on the current pages? We have a [`using-odoc` page](http://ocaml.github.io/odoc/odoc/using-odoc.html#custom-pages) that is intended to evolve and describe the features `odoc` includes.

src/odoc/pages.mld
```diff
@@ -0,0 +1,68 @@
...
+
+Additionally paragraphs can be annotated with level-4 or level-5 headings.
+Note that paragraph headings are {e not} be included in the generated table of
+contents and thus should be used to introduce examples, comments or other
+complementary notes.
+
+
+{1 Usage}
+
+The recommended way to setup documentation pages for your project is by using
```
ostera: In the same note, we have the [Build System Integrations # Dune](http://ocaml.github.io/odoc/odoc/system-integrations.html#dune) page where I'd expect to find this in some form.

src/odoc/pages.mld
```diff
@@ -0,0 +1,68 @@
...
+the Dune build system. It will automatically find and generate HTML for all
+[mld] files in your project. See
+{{:https://jbuilder.readthedocs.io/en/latest/documentation.html#documentation-stanza}
+Dune's configuration instructions} for more details.
+
+{2 Referencing pages}
```
ostera: As we move parts of the usage to other sections of the documentation, I'd expect this title to stay very close to the markup description of the page title and headings/sections.

src/odoc/pages.mld
ostera: Ah, a regression 😄 I'll make sure the [ Manual Usage page ](http://ocaml.github.io/odoc/odoc/system-integrations.html#manual-usage) includes this.

There's also a typo in `detaild`.

dbuenzli:
@rizo I didn't look at the implementation details but what I read in the summary looks really great, thanks. I really wish this had been fixed before the release that has just been made. 

One comment though:

> All odoc documents must have a level-0 heading (_i.e._ a title).
> * The level-0 heading for API references is automatically generated based on the module or module type name.

That sounds right.

> * The level-0 heading for documentation pages must be manually defined by users.

That kind of rule makes it harder to compose fragments you might want to render individually in a context and concatenate in another one to render a single page. Why not simply let the first header you hit (whichever level it is) define the page title ?

src/parser/parse_error.ml
```diff
@@ -41,9 +41,15 @@ let no_trailing_whitespace_in_verbatim : Location.span -> Error.t =
...
   Error.make "only one title-level heading is allowed"
 
+let page_title_required : string -> Error.t =
+  Error.filename_only "page title is required, add one with {0 ...}"
+
 let sections_not_allowed : Location.span -> Error.t =
   Error.make "sections not allowed in this comment"
 
+let titles_not_allowed : Location.span -> Error.t =
+  Error.make "titles not allowed"
```
aantron: Perhaps instead of "titles not allowed," this should say something like "titles ('{0 …}') are only allowed in .mld files"

test/parser/test.ml
```diff
@@ -467,43 +467,39 @@ let tests : test_suite list = [
...
+    t "bad-level-too-deep" "{6 Foo}";
     t "link-in-markup" "{2 {{:foo}}}";
     t "reference-in-markup" "{2 {!foo}}";
   ];
 
   "section-contexts", [
-    t "titles-allowed" "{1 Foo}"
+    t "titles-allowed" "{0 Foo}"
       ~sections_allowed:`All;
-    t "titles-zero-not-allowed" "{0 Foo}"
```
aantron: This test was deleted, so the expect file should be deleted, too. Perhaps, in a separate PR, we should have the tester scan for unused expect files.

test/parser/test.ml
```diff
@@ -467,43 +467,39 @@ let tests : test_suite list = [
...
-      ~sections_allowed:`All;
-    t "none" "{2 Foo}"
+    t "none" "{1 Foo}"
       ~sections_allowed:`None;
-    t "permissive" "{1 Foo}"
+    t "permissive" "{0 Foo}"
       ~permissive:true;
-    t "titles-allowed-permissive" "{1 Foo}"
-      ~permissive:true ~sections_allowed:`All;
-    t "titles-zero-not-allowed-permissive" "{0 Foo}"
```
aantron: The expect file for this test should be deleted.

test/parser/test.ml
```diff
@@ -857,7 +853,7 @@ let () =
       let actual =
         let {permissive; sections_allowed; location; parser_input; _} = case in
 
-        let dummy_filename = "f.ml" in
+        let dummy_filename = "f.ml" in (* FIXME: the .ml extension is not needed here. *)
```
aantron: Why change it?

test/parser/expect/section-contexts/titles-no-high-levels-permissive.txt
```diff
@@ -1,11 +1,13 @@
...
       (label
        ((root (f.ml (page f.ml) 00000000000000000000000000000000) f.ml)
         (label foo)))
       (((f.ml (1 3) (1 6)) (word Foo))))))))
  (warnings
   ( "File \"f.ml\", line 1, characters 0-7:\
-   \n'5': bad section level (2-4 allowed)")))
+   \n'6': bad heading level (0-5 allowed)"
+    "File \"f.ml.mld\":\
+   \npage title is required, add one with {0 ...}")))
```
aantron: This test is only accidentally triggering the check for a missing title (it's not the main purpose of the test). I suggest adding a test specifically for that, with `sections_allowed:`All`, and a valid non-title section inside. Perhaps we also should add a test that just has text, to make sure that it doesn't take parsing a section to notice that there is no title.

trefis:
> The only one applied is when the level greater than 5 is used, keeping it at 5.

Is a warning emitted in that case?

src/odoc/pages.mld
rizo: Ok, I'll point to that. That section is currently empty, should we recover the content of the `manual_usage.mld` page? This can be done separately.

rizo:
My last commit removes the requirement for `mld` pages to have at least one title. @dbuenzli I can revert this if you have a better proposal.

src/parser/parse_error.ml
rizo: Changed it to `level-0 headings (titles) are only allowed in pages`.

trefis:
> I meant any. By imposing such artificial restrictions you are imposing the fact that an .mld file should correspond to a "chapter", but there's no real need to impose such a restriction. Allowing .mld files to be catable is a good property which is prevented by this.

I agree with that.
For an example of .mld files compisition: at some points there was also a plan to be able to simply inline a .mld in another place (another mld, or a regular API doc page).

> In the html rendering case, for finding what you need to put in the title tag of an .mld file I suggest using the first header you find and "Untitled" if there is no header.

That sounds reasonable to me.

test/parser/test.ml
rizo: First of all, it's incorrect to have the extension. The root represents a page and not a compilation unit. In the error reporting I also assume the filename does not have an extension and add the `.mld` extension to it when printing an error.

I didn't change this in this PR because it would add a lot of noise in many (all?) parser expect files.

rizo:
> Is a warning emitted in that case?

[Yes](https://github.com/ocaml/odoc/pull/217/files#diff-22c1c7b0bd466e8a8de475b81a251449R284).

rizo:
> Frankly the way this project is currently managed is a bit concerning

I understand this was a rather forced release and I personally would prefer having a clear roadmap and only release on milestone completions. On the other hand, I also see value in releasing frequently as it generates more feedback. I also think that, even with this fix in, there are other relatively significant issues to be fixed before a "proper" stable release.

trefis:
> I also think that, even with this fix in, there are other relatively significant issues to be fixed before a "proper" stable release.

I agree with that.
I still think (despite all the nice work that has been going on) that the previous version (1.2) is in a better state than this one (or the 1.3 release).
There are still fairly major regressions that make it impossible to use the current version of odoc on codebases that were handled just fine with 1.2.

lpw25:
As a slight variation (or possibly a clarification) on Daniel's suggestion, how about using any heading as the contents of `<title>` but only if it is the first thing in the file. I think that if a file starts with some text and then a heading then the heading was probably not intended as the title.

Then a `{0` at the start of an mld will give you a heading styled appropriately for a title as well as setting the `<title>`.

rizo:
> I meant any. By imposing such artificial restrictions you are imposing the fact that an .mld file should correspond to a "chapter", but there's no real need to impose such a restriction. Allowing .mld files to be catable is a good property which is prevented by this.

I agree. I have now removed this restriction – `mld` files no longer require having a title. They gain meaning depending on the context, i.e., multiple `mld` files can be combined into a chapter, for example.

> In the html rendering case, for finding what you need to put in the title tag of an .mld file I suggest using the first header you find and "Untitled" if there is no header.

I don't entirely agree with this. I don't think for example that setting an arbitrary "subparagraph" heading (i.e., level-5, max allowed level) as a document title is reasonable (even if it's the first heading as @lpw25 suggested). If people want to have a document title, it should be explicitly set (with `{0 ...}`, that's how HTML works too.

Another clarification: currently the `<title>` tag for pages is set to the filename of the `mld` file. Given this, I think the automatic promotion of headings to `<title>`s is out of scope of this PR. I'd be glad to implement the promotion for `{0 ...}` but that can be done separately. Does this make sense?

test/parser/test.ml
rizo: Actually the error reporting situation no longer applies as I deleted the requirement on pages having titles. I still think it should be changed.

dbuenzli:
> it should be explicitly set (with `{0 ...}`, that's how HTML works too.

I prefer to reason in terms of how human work. It's good to have a proper distinguishing title in whatever you generate, if only to be able to distinguish them in your tabs.

rizo:
@dbuenzli The fallback (for not defined `{0 ...}`) could be using the filename, which is still better than "Untitled" (in case of no headings).

dbuenzli:
> @dbuenzli The fallback (for not defined `{0 ...}`) could be using the filename, which is still better than "Untitled" (in case of no headings).

The filename could be used in case of no initial heading but otherwise I'd rather avoid mentioning artefacts of the underlying producing technology; that's not very useful for humans. 

Requiring a `{0 ...}` for "setting a title" is again an uneeded artificial constraint. Headers are just a way of chunking (reasoning in terms of chapter/section/subsection/paragraph/subparagraph doesn't make much sense). Depending on your content/audience/context you might want to chunk at different levels or even not chunk at all; the proposed rule allows all of this in a convenient manner for authors.

lpw25:
I'm not necessarily suggesting it, but there is a similar restriction to your original one which is: an mld file must start with a heading, and all the other headings it contains must be at a lower level. That means that you can make a page corresponding to a chapter, a section, a subsection etc, but still enforces that they have a nested structure.

This means that you could not always `cat` them together. However, I think that being able to arbitrarily `cat` them together is not really a desirable property. The reason being that you cannot actually concatenate the files in a way that preserves their semantics. You always need to be careful that the headings of one file do not capture the lower headings of the next. Explicitly having each file represent a single unit of structure (e.g. a section) encourages you to think about how you compose them, which I think is a good thing.

rizo:
@lpw25 @dbuenzli I don't currently have any objections to the proposed solutions apart from what I already stated. I opened a separate issue (https://github.com/ocaml/odoc/issues/220) to discuss this matter.

This PR does not change the current behaviour (using the `mld`'s filename as a title) and fixes another significant limitation in the heading parsing. I suggest we focus on getting this fix reviewed and merged before considering https://github.com/ocaml/odoc/issues/220.

lpw25:
Sure, but my latest suggestion is not just about what you use for the `<title>`. It is a variation on the restriction that you originally had in this PR but then removed, in order to ensure that pages are properly nested. So I think it is worth considering implementing that aspect in this PR.

rizo:
Note that the current implementation is sufficiently flexible to allow `mld` concatenation, which was mentioned as a limitation of my original approach. The discussion I'm suggesting to delay is about making it stricter (_e.g._ enforcing a heading in the beginning of a file) and inferring a `<title>` for the document – which I consider less critical.

lpw25:
The reason to consider the restriction in this PR is that if there is a release with this change but without something for #220 then the restrictions will be different across three different versions of odoc. Not a disaster if it happens, but probably worth avoiding if we can.

rizo:
Some thoughts about heading enforcement (and _not_ title inference):

> an mld file must start with a heading, and all the other headings it contains must be at a lower level. 

What if users _do_ want to have multiple headings with the same level to be included in the parent file (that contains a chapter-level heading)?

> You always need to be careful that the headings of one file do not capture the lower headings of the next. 

Users always need to be careful about the way they manage headings and odoc won't help them avoid all possible mistakes – what are the benefits of being strict about this? Other documentation formats do not make such restrictions and delegate it to configurable linters and validator (not suggesting anything, just stating).

>  Explicitly having each file represent a single unit of structure (e.g. a section) encourages you to think about how you compose them

If we don't want `mld` files to represent "chapters", why would we want them to represent "sections"? What if people want to have separate `mld` files for menus, code examples, tables and other "units of structure" that might not need a section heading.

Overall, given the lack of concrete use-cases for `mld` pages, I prefer to keep them flexible and not impose any artificial constraints.

Of course, if people feel differently, I don't mind implementing the approach that gets the most consensus.

lpw25:
What I had in mind when thinking about the restriction was how you would compose all the pages of a project into a single volume (e.g. if you wanted latex output) and similarly how you would create a reasonable table of contents for a package. If you force the different pages to correspond to a particular unit of structure then it is easy to see how they can compose quite naturally without any difficulty. Whereas if the different files have no semantic meaning and are just arbitrary snippets then it is less clear to me that they compose naturally.

Also forcing pages to have a well-defined name makes the table of contents easier.

> If we don't want mld files to represent "chapters", why would we want them to represent "sections"? 

I can see two reasons:
- To be used with some kind of inclusion mechanism to be inlined into a chapter. Here having it be a proper structure ensures that the nesting within the inlined file remains unchanged.
- To be a section of the "introduction" of the package. Sections do not need to be part of a chapter, they can be stand-alone. For example, I may want to have a "changelog" mld. I don't necessarily want it to be given the same prominence in the table of contents as the actual modules, and nor do I want to put it on the front page where it clutters things up -- making it a section in its own file gives me that.

> Other documentation formats do not make such restrictions

Things like latex using a `begin`/`end` approach do essentially enforce something similar. Other formats like markdown are not really designed for constructing documentation across multiple pages.

lpw25:
Of course, the arguments based on latex output and table of contents aren't really that important. Both latex output and generated tables of contents always look pretty terrible for API documentation, so this certainly feels like re-arranging the deck chairs.

I do still have a preference for enforcing it so that every page has a sensible name, and so that pages have some meaning in terms of the structure of the documentation.

aantron:
My opinion on anything related to HTML fragments is:

1. We should keep the need for emitting them in mind, and try to redesign the core of odoc eventually so that it can naturally support them.
2. odoc cannot naturally support them today.

So, I think we shouldn't try to make this PR support fragments. I suggest to do something reasonable (like require title headings in `.mld`s), and we can merge that and release it.

We will work towards proper fragment support in some (far-?)future release of odoc.

In my view, odoc is basically in a beta state, despite the premature releasing of it as 1.x.x way back. In the future, people will need to put lower bounds constraints on which version of odoc they use. So, there is nothing wrong with not supporting fragments today.

dbuenzli:
> In my view, odoc is basically in a beta state, despite the premature releasing of it as 1.x.x way back.

Please, there's absolutely nothing beta in `odoc` and there was no premature releasing of 1. Thanks to @trefis' work we were almost one-to-one to ocamldoc + correctly handling many more funky cases before you started rewriting everything rather than gradually improving the existing and in the process making the same errors as everyone did before you, introducing regression and with little progress one year later on the long standing markup issues.

I have nothing against you rewriting things as long as you take the responsability for it, which means fixing what you are breaking. You direct us to https://github.com/ocaml/odoc/issues/212 where we should have said something about a release suddenly set at a random date; but it's not my job to tell you what you broke. I thought it would have been self-evident to you that a new `odoc` release would entail verifying your outputs on a large portion of the eco-system (which `odig` allows you to easily do, but instead you [simply broke](https://github.com/dbuenzli/odig/issues/39) without notice) and compare with what we had in the previous version to catch rendering regression. Regarding #212 I was actually expecting *you* to come up with a *release plan* in that issue before pushing whatever random state of the repo you had in your hands.

aantron:
@dbuenzli My main concern is that your communication style causes *other* contributors to be afraid to work on odoc. The example in this issue is when someone feels *at fault* because of your negatively-framed claims, the factual basis of which is also *unfounded*. You then take this as an invitation to go further.

My suggestions for you are:

- Constrain odig to require odoc < 1.3.0. We are not going to deliberately try to support it for a little while.
- *Find out* what happened before assuming, and definitely before complaining or accusing. See specifics below.
- Read carefully what people are actually saying. See specifics below.

What I am going to do next is:

- Post a clear description of where the project is and where it's going in CONTRIBUTING.md or the wiki.

Some perspective:

1. The "breakage" we are discussing *in this issue* is that certain headings cause a *warning* to be emitted.
2. There is some other breakage, which affects odig and JST. See below.

Specifics:

1. We are not trying to explicitly support odig *at the moment*. The reason is that it has unique requirements of odoc that take a noticeable amount of effort to satisfy. We need to focus on other gaps of odoc, which are huge.

    At the same time, it's possible that normal development of odoc will change the relationship between odoc and odig, we might find better ways to do things, etc. So, it seems pointless to slow down the rest of odoc development, for the sake of maintaining the current interface between odoc and odig.

    If you can easily refactor odig so that it doesn't require fragments, etc., then please do so. Otherwise, constrain it to odoc < 1.3.0. Later, you can constrain to odoc >= some_future_version.

2. > started rewriting everything

    Only the comment parser was rewritten, the rest was just refactored in ways that don't change the logic. The parser was rewritten precisely because there were outstanding issues with no clear fix, and writing tests for it revealed an even larger number of corner cases, that were difficult to fix with the previous design. There was a discussion about this somewhere back when it happened, I can't find it now. I'm not sure if you were in it.

3. > rather than gradually improving the existing and in the process making the same errors as everyone did before you, introducing regression and with little progress one year later on the long standing markup issues.

    What errors specifically? Open an issue or send a PR. We don't have time to fix every single issue in the parser to *your* level of satisfaction on *your* schedule, but we will fix them eventually. Some issues were resolved by the new parser, some regressions were introduced, but overall the design so far allows eventually handling everything we've considered in a pretty natural way. The previous design had some problems with that.

    The markup issues have little to do with the parser, and the markup emitter wasn't "rewritten." So most of the issues in it aren't "regressions."

4. > it's not my job to tell you what you broke

    See

    > Everyone posting in this issue, including me and @rizo, is aware that there are regressions in 1.3.0 and plenty of things to be worked on.

    In other words, **yes**. It's not your job. **We know** what is broken. I asked you to tell me if you think it's important enough to stop a release.

    Each thing you are complaining about is something we have discussed since months ago, including my resistance to fixing them in the near term.

    If a release benefits odoc's other users, can you solve your problem with a constraint?

    The way **you wrote** your sentence, it's as if somebody is asking you to fish through the odoc repo, looking for all kinds of breakage that someone else doesn't know about. Nobody is asking, or would ask, you to do that. You are being asked to explain if you think something is blocking, because we don't want to hold odoc release back forever on *every* flaw. And, we might disagree with you and release anyway because we don't see anyone fixing the issue in the near term.

5. **Your most telling assumption**:

    > I thought it would have been self-evident to you that a new odoc release would entail verifying your outputs on a large portion of the eco-system

    It **is** self-evident to me, which is why we have a [script](https://github.com/ocaml/odoc/blob/488e6677e11864793df7b85d72275ba686ce49f6/src/util/corpus.sh) for doing so. I ran this on ~1400 packages out of ~1800 that were in opam at the time, because that's how many could be consistently installed by opam, and I figured it was close enough. I went through every single error, warning, and failure, made several fixes to odoc, and decided whether the rest were important enough that they needed to fix immediately, or whether we can switch to other issues and fix them later on demand.

    This script is much easier to deal with than odig.

    You could have easily asked (or even checked) whether this thing exists, instead of assuming that something so "self-evident" is somehow *difficult* for others to see.

6. > I was actually expecting you to come up with a release plan in that issue before pushing whatever random state of the repo you had in your hands.

    Then please go to the issue and explain to us what kind of release plan you would like to see, and we can discuss whether we want to adopt it.

7. > thinking about the consequences of what is in or out of the release.

    We thought about the consequences. We have a different estimate of the consequences than you do, at least I do.

ostera:
@aantron @dbuenzli I took the liberty of filing an issue with a proposal on how we can make sure the next release's impact is transparent, verifiable, and end-user friendly. We can take that discussion on in #221 .

lpw25:
Firstly, can I ask people to stick to a more civil tone. Secondly, can we avoid arguments over what has happened and focus discussions on what people want to happen in the future -- since that's all anyone can actually do anything about.

@dbuenzli Whilst I agree that 1.3.0 has regressions from 1.20, if there is a 1.3.1 that fixes them then having one bad release lying around is hardly a major problem. Some constraints are needed on odig packages and maybe on some other packages whose documentation has been broken (e.g. ctypes), a pain but not the end of the world. I know that you like to keep a high (and sometimes opinionated) standard with the packages that you maintain, and your users are grateful, but please try not to get quite so aggressive when other project's maintainers don't meet your standards.

@aantron I know that you only changed the parser for 1.3.0, but the new parser is the source of the regressions that I'm aware of. None of them seem that hard to fix, but since you wrote the code I think people are somewhat waiting for you to fix it. You said that breakages only affect odoc and JST but that's the entire userbase of odoc 1.2.0. Beyond ordinary users they are also the people who wrote odoc 1.2.0 -- so you can understand where the frustration comes from when you dismiss their concerns and focus instead on new users.

> which is why we have a script for doing so. I ran this on ~1400 packages out of ~1800 that were in opam at the time

I'm surprised this hasn't found more of the problems, whenever we try to use new odoc on any sizable codebase it either crashes with uncaught exceptions or refuses to process some file because one of its comments is rejected by the validator. Fixing that issue is all we need at JST to start using the new version -- which we would very much like to do because it does have some of great stuff in it compared to 1.2.0.

avsm:
I'd like to reinforce @lpw25's request for a more civil tone on this thread, and to focus on next steps.  I really appreciate @aantron and @rizo's hard work on the recent release, and especially appreciate their focus on making it a project that's easier to contribute to.  (@dbuenzli: to answer your question about 'we', @aantron originally took on much of the odoc work at my request, in my role as the overall overseer of projects under the `ocaml/` organisation outside of the core compiler).

odoc clearly needs resources towards enforcing the more principled release discipline that @aantron is aiming for with his new test suite. It's very difficult to release it currently and get visibility into what will break, and this in turn reduces the motivation to work on the core tool itself.  To that end, I'd like to do two things:

- Make it easier to run the odoc over the entire opam codebase. To this end, we now have a `ocaml/opam2/opam-sources` container on the Docker Hub now that can be hooked in Travis CI and that contains convenient unpacked package sources in one place.  @aantron, is it feasible to have a odoc-lint binary that that runs over an ml/mli source file and displays any parsing inconsistencies? The corpus.sh seems to work on compiled files at present.

- Ensure that the Jane Street codebases passes the new parser as a regression test is also very valuable, since it covers many corners of the language. We also need this for the next Real World OCaml release, so I'll contact @lpw25 and @trefis out of band about arranging to get these run regularly in odoc.  Ideally, we'll get a set of contributed test cases from the results of this run that will give us complete language coverage (in addition to the very large set of new tests that @aantron wrote).

The opam constraint system in opam-repository is not perfect, but it exists exactly for the situation we are in with odoc. If a release ends up breaking some part of the universe, then we need to just establish upper bounds there, and it will not be picked up by the community.  What is *far* worse is a situation where we are unable to regularly release a particular subcomponent like odoc, as that leads to slow death.  So despite 1.3 having some regressions, let's focus on getting a point release out and just use the opam repository to avoid end-user breakage for now.

To be clear, odoc *is* in a beta state until it's a full replacement for ocamldoc, at which point we can also consider decoupling ocamldoc from the mainline OCaml distribution.  Until then, development continues and breaking changes may occur.

dbuenzli:
> What is _far_ worse is a situation where we are unable to regularly release a particular subcomponent like odoc, as that leads to slow death. 

Let's be honest here with a little bit of work across the community there wouldn't have been breakage, there are not that many direct users of `odoc`. Besides no one I saying one should not release, I actually wish there had been much more releases over the year with *gradual improvements*.

> in my role as the overall overseer of projects under the `ocaml/` organisation outside of the core compiler)
[...]
> To be clear, odoc _is_ in a beta state until it's a full replacement for ocamldoc, at which point we can also consider decoupling ocamldoc from the mainline OCaml distribution.

I think that a certain point you might want to reflect on your handling of the documentation problem over all these years at ocamllabs; keeping the documentation problem constantly in beta will not allow us to move forward. As I said [elsewhere](https://github.com/ocaml/odoc/issues/210#issuecomment-430927619) many people are relying on `odoc` and `odig` now for their daily programming, `odoc` is absolutely not beta software despite what you may think --- that is unless you do have no respect for the current users of these systems and all the work that many of us invested in building them.

lpw25:
Ok, if everyone has finished venting their frustration and antagonizing each other, we can get back to the contents of this actual PR. Poor @rizo has written a nice PR implementing functionality that everyone wants and we've graffitied it with an unrelated discussion.

That last remaining issue is what restriction to have on the headings in mld files. I suggest that we use the restriction I suggested because:

1. It avoids @dbuenzli's objections to the original restriction, and he's said that he is fine with this version.
2. I don't think anyone has any particular use case in mind that would require breaking this restriction
3. It will make implementing #220 in a way that is obviously correct easy

I think the other arguments I made earlier in support of this proposal are pretty weak, but I find 3 persuasive enough to go for it anyway. Anyone object to that plan?

src/parser/semantics.ml
```diff
@@ -290,6 +290,17 @@ let section_heading
     parsed_a_title, element
 
 
+let validate_first_page_heading status ast_element =
+  match status.parent_of_sections with
+  | Model.Paths.Identifier.Page ({file; _}, _) ->
+    begin match ast_element with
+      | Location.{value = `Heading (l, _, _); _} when l >= 0 && l < 4 -> ()
```
dbuenzli: I won't fight against this but I find that restriction to be quite artificial. I guess this is based on the `latex` interpretation since `4` would lead you to paragraphs and that you'd find absurd to have one `.mld` file per paragraph. 

But now if you get this interpretation out of your mind, that could simply be the valid description level of a certain subsystem in a layered architecture that you would like to document in its own file.

lpw25:
@rizo I don't think the latest commit doesn't quite implement the restriction I was suggesting -- which I'm assuming was its aim. It ensures that the page starts with a heading, but it doesn't check that all later headings in the file have a lower level. Note that the check that normal pages do not have a level-0 heading is actually subsumed by this other check -- since all normal pages start with an implicit level-0 heading -- so it should be pretty easy to implement it.

rizo:
@lpw25 I split my work in two commits: the one you saw required a heading in the beginning of pages and the second (just pushed) enforces the heading hierarchy. Sorry, I should have mentioned this.

rizo:
@lpw25 Your 3 point presented the most convincing argument to implement your proposal.

Here are some examples of failing cases handled by this PR (also encoded as tests of course).

### Cases that generate warnings

#### Title headings in references

```
{0 Foo}

Error: title-level headings {0 ...} are only allowed in pages
```

#### Duplicate title headings in pages
```
{0 Foo}
{0 Bar}

Error: only one title-level heading {0 ...} is allowed
```

#### Duplicate section headings in pages
```
{1 Foo}
{1 Bar}

Error: duplicate {1 ...} top level heading not allowed
```

#### Required title or section heading in pages
```
Some paragraph.

Error: pages must start with a title or a section heading (levels 0-3)
```

#### No heading hierarchy in pages
```
{2 Foo}
{1 Bar}

Error: heading level 1 is higher than top level 2
```

src/parser/semantics.ml
rizo: Yes, your interpretation is correct. One additional supporting argument is the fact that the TOC generated by odoc should not include levels 4 and 5 – this way the content in a page would always be equivalent to something that shows up in a TOC.

I can remove this restriction, I don't feel strongly about it.

rizo: Removed in [25ab291](https://github.com/ocaml/odoc/pull/217/commits/25ab29110979bfd637e2f5fb2bdc53e533061eac).

src/parser/parse_error.ml
```diff
@@ -39,10 +39,22 @@ let no_trailing_whitespace_in_verbatim : Location.span -> Error.t =
   Error.make "'v}' must be preceded by whitespace"
 
 let only_one_title_allowed : Location.span -> Error.t =
-  Error.make "only one title-level heading is allowed"
+  Error.make "only one title-level heading {0 ...} is allowed"
 
-let sections_not_allowed : Location.span -> Error.t =
-  Error.make "sections not allowed in this comment"
+let page_heading_required : string -> Error.t =
+  Error.filename_only "pages must start with a heading"
```
aantron: Could you add two tests for this: one testing the string `"foo"` (just a paragraph) in context "`All", and one testing `"foo\n{0 Bar}"` also in context "`All"?

I think this will cover

> #### Required heading in pages
>
> ```
> Some paragraph.
>
> Error: pages must start with a heading
> ```

rizo: I thought I’ve covered this case, thanks for spotting this. I’ll add the test cases ASAP.

aantron:
Thanks!

rizo:
Thanks everyone for review and discussion!

PR #222, opened 2018-10-18
Fixes for Windows build against 4.02 compiler

bryphe:


aantron:
This looks good to me.

In case we need this info in the future: I see that sexplib was marked as no longer supporting 4.02 in this commit, https://github.com/janestreet/sexplib/commit/2a5d87a6fd2c6bbeb44a06e9352e2774c97410e1. Assuming that constraint and when it was added are both accurate, the latest commit we could try basing the fix on is its parent, janestreet/sexplib@cba80ebe02962504fbe404b1baf484a5000f44b4, "114.01+04." This (indeed) still uses OASIS.

PR #223, opened 2018-10-19
Rename root-uri to xref-base-uri and fix empty string bug.

rizo:
This renames `root-uri` to `xref-base-uri` as suggested by https://github.com/ocaml/odoc/issues/27#issuecomment-427862082. And more importantly fixes a bug where the default value for this option resulted in out of bounds access.

@aantron this is a breaking change on the command-line interface level but I don't think anyone is using this in 1.3 yet.

CC @dbuenzli

### DIFF
--- a/src/html/documentation.ml
+++ b/src/html/documentation.ml
@@ -80,12 +80,12 @@ module Reference = struct
   let rec to_html
       : type a.
         ?text:(non_link_phrasing Html.elt) ->
-        ?root_uri:string ->
+        ?xref_base_uri:string ->
         stop_before:bool ->
         a Reference.t ->
           phrasing Html.elt =
 
-    fun ?text ?root_uri ~stop_before ref ->
+    fun ?text ?xref_base_uri ~stop_before ref ->
       let span' (txt : phrasing Html.elt list) : phrasing Html.elt =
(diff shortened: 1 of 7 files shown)


aantron:
No problem, fragments are probably the most "beta" part of odoc's interface (i.e., IMO, we shouldn't constrain the rest of odoc to the pace of their development).

LGTM provided CI passes. Will merge then provided this is ok with you, @dbuenzli.

dbuenzli:
@rizo next time would you be kind enough to directly add an entry to the changes files. It makes it easier on release and to track what is going with master.

dbuenzli:
Mmh my comment got lost. It was: 

Looks good to me. @aantron note that this is not only (or will not only be) related to fragments.

rizo:
@dbuenzli If I recall correctly, @aantron has asked me at some point to not update the changes file because he preferred to do it on his own before a release. Is that still the case @aantron?

dbuenzli:
I don't think it's needed to update the changes file on every tidbit but it's nice to document the breaking changes on the fly so that dependent tools can understand and plan with what is going on with master.

aantron:
What was and is the case is that IMO it's not necessary to update changes, but I will do changelog work in a batch at release time. If you'd like to add an entry, go ahead and do so.

Issue #147, opened 2018-05-31
Make a `ppx_odoc` that does the comment checking that `ppx_js_style` does


lpw25:
ppx_js_style is the Jane Street style checking ppx. One of the things it checks is that documentation comments have valid syntax using the documentation comment parser from the old version of odoc (octavius).

Now that the parser is no longer a separate library ppx_js_style can't use an up-to-date documentation parser. Now that odoc is getting more widely used there are probably people who want to check their documentation comments without using the other parts of ppx_js_style. So my suggestion is to implement a stand-alone ppx that just checks the syntax of documentation comments.

At the moment I think it would have to live inside the odoc repository, since that is where the parser is available, but it could also be implemented separately if odoc added a parser library. Either approach seems fine to me.

aantron:
Right now, it should be possible to make the `parser` sublibrary of odoc, in this repo, installable by just adding a `(public_name odoc.parser)` stanza to its `jbuild` file. ppx_odoc could then live in a separate repo, but depend on `odoc.parser`. I think `model` will also have to be installed, perhaps as `odoc.model`. For reference, right now, package `odoc` installs only the `odoc` binary.

pmetzger:
This seems like a very good idea.

rizo:
@aantron Dune currently does not permit having private dependencies for public libraries. This means that adding `(public_name odoc.parser)` would require making other odoc libraries public too.

rizo:
@dbuenzli mentioned in https://github.com/ocaml/odoc/pull/226#issuecomment-432581384 that parsing docstrings directly from `mli` files – which is what syntax rewriters does, of course – might not be reliable because other syntax rewriters might change the AST.

Is there a way to enforce the order of application of rewriters?

Is it possible to load a `cmti` file from a syntax rewriter for a given source file and only add warning attributes to the AST for merlin?

lpw25:
> Is there a way to enforce the order of application of rewriters?

You can give a `position` parameter to [Migrate_parsetree.Driver.register](https://github.com/ocaml-ppx/ocaml-migrate-parsetree/blob/master/src/migrate_parsetree_driver.mli#L61) to control the order the ppx passes are applied. However, using a non-zero position means that the AST must be processed an additional time which can slow down preprocessing. Personally, I think it would be fine to not worry about the order for this use case. It will still give warnings for any problems with user written comments, which is the main thing you want.

avsm:
[...]

Issue #108, opened 2018-03-23
New docstrings parser too strict


trefis:
Whenever it detects something it doesn't like it errors (actually, it raises `Failure` and that's that).
Up to now we'd arranged so that all such errors would be non fatal (i.e. you could print a warning and drop the docstring you don't like, or the part of the docstring).

It'd be nice to recover that behavior, as the current one makes it close to impossible to generate any documentation.

trefis:
Also, things like this:
```
Failure("File \"csvfields/xml-light/dtd.mli\", line 27, characters 26-29:\n'{ul ...}' (bulleted list) must begin on its own line")
```
probably don't need to be *that* strict.

aantron:
[The parser](https://github.com/ocaml/odoc/blob/d4b1b9479eb200a5b67b1e266c4d2423da889a8e/src/parser/parser_.mli#L3) isn't raising `Failure`, it is exactly designed to report local fatal errors in a `result`, and separate warnings, so that surrounding code can drop the docstring, exactly as you suggest:

https://github.com/ocaml/odoc/blob/d4b1b9479eb200a5b67b1e266c4d2423da889a8e/src/parser/parser_.mli#L9

The surrounding code is in too messy a state at the moment to integrate this properly, so it is weakly "forced" to just raise `Failure` for now as a cop-out. That will be fixed "naturally" as that surrounding code takes better shape.

For the other issue, we can add more error recovery to the parser over time. It already has some for the most common errors, so it's just a matter of priority. If this error occurs only once or a few times in the whole [corpus](https://github.com/ocaml/odoc/blob/d4b1b9479eb200a5b67b1e266c4d2423da889a8e/src/util/corpus.sh) of OCaml docs, then it's probably easier to fix the source of the project than to push back other tasks in odoc and also make the parser more complex in the long term.

aantron:
I suppose the new parser can temporarily be wrapped in a couple lines of code that convert a fatal error into a warning, and return empty docs in an `Ok` result, as a temporary measure to play nice with the surrounding code until it is ready.

avsm:
For development purposes, I find it very convenient to have the strict mode, so that new comments begin well-formed.  Hopefully we can retain that strict mode even if some of them become warnings in the future (similarly to `jbuilder build --dev`).

trefis:
[...]

PR #226, opened 2018-10-23
Add odoc lint command for mli and mld validation

rizo:
From https://github.com/ocaml/odoc/pull/217#issuecomment-430969591 by @avsm: 
> (...) is it feasible to have a odoc-lint binary that that runs over an ml/mli source file and displays any parsing inconsistencies?

This PR implements an `odoc lint` command that does exactly that. It parser `mli` and `mld` files and uses an AST mapper to go through all `[@@ocaml.doc ...]`/`[@@ocaml.text ...]` attributes running odoc's parser on them and producing errors (see the test expect files in this PR for examples).

I got excited about this and started implementing a ppx (https://github.com/ocaml/odoc/issues/147) to validate docstings and report errors with merlin. Unfortunately I was surprised to discover that merlin doesn't convert docstrings into documentation attributes (see https://github.com/ocaml/merlin/issues/876). After that is fixed implementing the ppx should be simple.

#### Changes
- Implement the `odoc lint FILE.{ml,mld}` command.
- Remove an unused `id` argument from `Loader.Attrs.read_attributes`.
- Expose `Attrs` module in `Loader` interface.
- Add `Compat.Filename.remove_extension`.
- Add tests for the new lint command.

#### Notes

- The current parser does not recover from errors. This means that for a documentation attributes or `mld` files that have multiple errors only the first one will be reported (related to https://github.com/ocaml/odoc/issues/108).
- Some old versions of the compiler don't seem to produce the `[@@ocaml.doc]` attribute on
  polymorphic variants (the CI should report all the culprits). I don't have a workaround for this.
- The column number reported in errors seems to be slightly shifted (I'll investigate this soon).
  - _Update_: seems like it is already present in `master`. The location of the attributes is off by -3, I assume it's because it's not counting `(**`.
- Support for more file formats (like `ml`) should be easy to implement and can be done separately.
- The current implementation could be improved but would require significant changes to the error/warning management in the whole project. @aantron has already plans for that AFAIK.
- I think the command could be extended to check for unresolved cross-references in the future.

Thoughts and suggestions are welcome!

### DIFF
--- a/src/compat/compat.ml
+++ b/src/compat/compat.ml
@@ -29,3 +29,17 @@ struct
 #endif
 
 end
+
+module Filename = struct
+  include Filename
+
+#if OCAML_MAJOR = 4 && OCAML_MINOR < 04
+  let extension filename =
+    let dot_index = String.rindex filename '.' in
+    String.sub filename dot_index (String.length filename - dot_index)
+
+  let remove_extension filename =
+    let dot_index = String.index filename '.' in
+    String.sub filename 0 dot_index
+#endif
+end
--- a/src/loader/attrs.ml
+++ b/src/loader/attrs.ml
@@ -41,7 +41,7 @@ let load_payload : Parsetree.payload -> (string * Location.t) option = function
   | _ ->
     None
 
-let read_attributes parent _id attrs =
+let read_attributes parent attrs =
   let ocaml_deprecated = ref None in
   let rec loop first nb_deprecated acc
       : _ -> (Model.Comment.docs, Model.Error.t) result =
--- a/src/loader/attrs.mli
+++ b/src/loader/attrs.mli
@@ -24,7 +24,6 @@ val empty : Model.Comment.docs
 
 val read_attributes :
   Paths.Identifier.label_parent ->
-  'kind Paths.Identifier.t ->
   Parsetree.attributes ->
     Model.Comment.docs
 
--- a/src/loader/cmi.ml
+++ b/src/loader/cmi.ml
@@ -512,7 +512,7 @@ let read_value_description env parent id vd =
   let container =
     Identifier.label_parent_of_parent (Identifier.parent_of_signature parent)
   in
-  let doc = read_attributes container id vd.val_attributes in
+  let doc = read_attributes container vd.val_attributes in
     mark_value_description vd;
     let type_ = read_type_expr env vd.val_type in
     match vd.val_kind with
@@ -534,8 +534,7 @@ let read_label_declaration env parent ld =
   let name = parenthesise (Ident.name ld.ld_id) in
   let id = Identifier.Field(parent, name) in
   let doc =
-    read_attributes (Identifier.label_parent_of_parent parent) id
-      ld.ld_attributes
+    read_attributes (Identifier.label_parent_of_parent parent) ld.ld_attributes
   in
   let mutable_ = (ld.ld_mutable = Mutable) in
   let type_ = read_type_expr env ld.ld_type in
@@ -562,7 +561,7 @@ let read_constructor_declaration env parent cd =
   let container =
     Identifier.label_parent_of_parent (Identifier.parent_of_datatype parent)
   in
-  let doc = read_attributes container id cd.cd_attributes in
+  let doc = read_attributes container cd.cd_attributes in
   let args =
     read_constructor_declaration_arguments env
       (Identifier.parent_of_datatype parent) cd.cd_args
@@ -623,7 +622,7 @@ let read_type_declaration env parent id decl =
   let container = Identifier.label_parent_of_parent
                     (Identifier.parent_of_signature parent)
   in
-  let doc = read_attributes container id decl.type_attributes in
+  let doc = read_attributes container decl.type_attributes in
   let params = mark_type_declaration decl in
   let manifest = opt_map (read_type_expr env) decl.type_manifest in
   let constraints = read_type_constraints env params in
@@ -654,7 +653,7 @@ let read_extension_constructor env parent id ext =
   let container =
     Identifier.label_parent_of_parent (Identifier.parent_of_signature parent)
   in
-  let doc = read_attributes container id ext.ext_attributes in
+  let doc = read_attributes container ext.ext_attributes in
   let args =
     read_constructor_declaration_arguments env
       (Identifier.parent_of_signature parent) ext.ext_args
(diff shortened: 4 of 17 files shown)


rizo:
As expected there are some failures with the 4.02 compiler in the CI. Specifically:

```
Done: 1168/1183 (jobs: 1)File "test/parser/lint/expect/ast.expected", line 1, characters 0-0:
Done: 1169/1183 (jobs: 1)--- test/parser/lint/expect/ast.expected	2018-10-23 23:27:26.397396290 +0000
+++ test/parser/lint/ast.output	2018-10-23 23:27:26.473397012 +0000
@@ -14,10 +14,6 @@
 unknown tag '@ConstructorDeclaration'
 File "cases/ast.mli", line 14, characters 1-17:
 unknown tag '@TypeDeclaration'
-File "cases/ast.mli", line 19, characters 6-10:
-unknown tag '@Tag'
-File "cases/ast.mli", line 20, characters 6-10:
-unknown tag '@Tag'
 File "cases/ast.mli", line 25, characters 12-29:
 unknown tag '@LabelDeclaration'
 File "cases/ast.mli", line 26, characters 13-30:
        diff (internal) (exit 1)
(cd _build/default && /usr/bin/diff -u test/parser/lint/expect/ast.expected test/parser/lint/ast.output)
```

What is **not** expected is the fact that only the 327.8 build job (the esy configuration) is reported as failed while 327.1 and 327.2 actually failed too (both use the 4.02 compiler). Other jobs don't seem to have any issues.

dbuenzli:
I'm not exactly sure what this is trying to achieve.

One of the improvement with `cmti` files is that it allowed to process documentation after any pre-processor has done its job (which was a problem with `ocamldoc`'s workflow for documenting software installs since even if you had installed the `.mli` you had lost the information about how you needed to pre-process it). 

Working again directly on `.mli` files you might report an inaccurate state of what will actually be in it's corresponding file after it has been built.

rizo:
> It seems that the main differences between this and `odoc compile` is that `odoc lint` reads .mli files, while odoc compile reads .cmti (etc.) files 

The `lint` subcommand is for convenience only, the same effect can be achieved with `compile` and ignoring the result. Maybe we could add a `--dry-run` option to `compile`, but I see `odoc lint` as essentially an alias for that.

@aantron:
> This is simpler for the user, but will it be problematic in case of cppo being used, for example?

@dbuenzli:
> One of the improvement with cmti files is that it allowed to process documentation after any pre-processor has done its job

These are both good points, but `mli` is just one of the options. I mentioned that `odoc lint` can be extended to support other formats like `cmti` (or even `odoc` files). `odoc lint` is a convenient way to _"run the odoc over the entire opam codebase"_ (see https://github.com/ocaml/odoc/pull/217#issuecomment-430969591). The intent is to follow @avsm's suggestion and run this on all opam source files.

rizo:
For reference I assume this is an example of what you were talking about, @dbuenzli: https://github.com/dbuenzli/odig/issues/12.

rizo:
> IIRC it's the responsibility of the calling code to pass an adjusted "base" location into the parser (~location argument). Looking at the code in attrs.ml right now, we might not be adjusting it correctly even during odoc compile.

I verified that, yes.

> As a little note, we can't unconditionally add 3 because .mld files don't have (**.

The `mld` files are parsed by a different function (`Loader.Attrs.read_string`), so it could be done specifically for `[@@ocaml.doc ...]` attributes in `Loader.Attrs.read_attributes` after getting the location of the payload.

dbuenzli:
> `odoc lint` on `mli` files is a convenient way to _"run the odoc over the entire opam codebase"_ (see [#217 (comment)](https://github.com/ocaml/odoc/pull/217#issuecomment-430969591)). The intent is to follow @avsm's suggestion and run this on all opam source files.

I don't know what the "entire opam codebase" is. But you really rather want to check `odoc compile` on entire "consistant opam install bases" and thus rather on `cmti` files (and yes what you pointed to is an example of the problem I'm mentioning).

I can certainly see a use for this, but again I'm don't think it is the right answer to the problem that was highlighted there.

rizo:
> you really rather want to check `odoc compile` on entire "consistant opam install bases"

Ok. Are you also opposed to the idea of supporting `odoc lint FILE.cmti`? Or adding `--dry-run` to `odoc compile`?

I can close this PR if this doesn't seem useful.

dbuenzli:
Note that I'm not *opposed*, I'm just asking how this fits in the picture to understand where this should go -- if it should -- and what problem it solves. It's always easy to pile up the features and produce code, but it has costs.

Personally I still don't understand the problem it solves.

I think `odoc compile --dry-run` can make sense if it buys you enough w.r.t. `odoc compile -o /dev/null` but that I don't know enough about the actual work `odoc compile` does to know for sure.

aantron:
To further agree with the above, running `odoc lint` on all *source* files in opam is probably not feasible because some of them require preprocessing. We almost certainly want to run `odoc lint` on all `.cmti` files instead, and `odoc compile`, driven by Dune, already does that.

It's possible that in some projects conditional compilation, unmet optional dependencies, OS constraints, etc., will cause `.cmti` files not to be generated for part of the codebase, or the projects are not compiled with Dune. Then, `odoc lint` could extend coverage to the files of those projects, as long as they don't also require preprocessing.

However, are we actually having a problem with these uncovered `.mli` files at this point? We could stick with covering the duniverse, and see if there is actually a problem to be solved beyond that first.

We likely don't need `--dry-run` yet. A build process that runs `odoc compile` already produces intermediate files, `.cmti`s. It's probably not a big deal to produce some `.odoc` files, too. Once we know that we don't want `.odoc` files for a specific reason, perhaps organization or performance, we can add `--dry-run`.

So, agreeing with the above, I don't see a need for `odoc lint`, so, in the absence of further arguments, I'd suggest to close it.

For comparison, `corpus.sh` is a really *bad* way to run `odoc compile` on all `.cmti` files, but it already lints potentially all of opam, to the same degree that `opam lint` would (if we downloaded sources instead of trying to install a consistent switch). All we need to achieve this, at a minimum, is to write some separate script that will run `find` to discover all `.mli` and `.mld` files, compile them to a single `.cmti`, and run `odoc compile` on them.

aantron:
A new test in this PR, `parser_errors.mli`, would be quite valuable if ported to `odoc compile` and committed to `master`. It would catch the 3-column offset bug, as well as any differences in how many errors `odoc compile` reports, their formatting, etc. We don't have an end-to-end test for parser errors today, only end-to-end tests for overall odoc success (in `test/html`), and unit tests for parser errors that are largely insensitive to context (in `test/parser`).

rizo:
That makes sense, thanks for putting things into perspective. I'll submit a separate PR with `parser_errors` tests later.

PR #229, opened 2018-10-25
Adjust new error messages from the headings PR (#217) and fix a bug

aantron:
- Some of the new error messages and checks were redundant with each other, or with existing checks, so I combined or removed them. See "Simplify heading level validation and errors." The easiest way to start looking at the effect is to look at the diff to the expect files.
- Reduced error message noise. For example, if we tell the user that a heading has a bad level, it's probably not worth telling the user that it's the same as that of a top-level heading which is also bad.
- Fixed a bug where we were applying the new `.mld` restriction that each heading must have level lower than the first to module-level comments in `.mli` files. See the last commit.

See #217 and #141.


aantron:
The first commit in this PR had the side effect of allowing heading levels with leading zeroes, so I added two test cases for that and some code to reject it.

The code raises an error for now, but we can easily recover from this and issue only a warning. That will be easier to implement once we have more disciplined warning and error-recovery support in place, which I am working on next, so I want to leave this error "fatal" (in the scope of the comment) in this PR.

I/we will do a pass later through all errors anyway, and make as many recoverable as is feasible.

aantron:
Thanks, @rizo.

Indeed, the columns are 0-based, like in the OCaml compiler. I personally prefer 1-based columns, like in the text editors I use, and most other compilers, but I wanted to stay consistent. This can be changed subject to discussion.

PR #230, opened 2018-10-25
Fix #101, assertion failure on some optional arguments in structure items

### LINKED ISSUE #101: Assertion failed in doc-ock
Do:

- `opam install odoc.1.2.0`
- `opam install tensorflow.0.0.10`
- `` odoc compile `opam config var prefix`/lib/tensorflow/fnn/tensorflow_fnn__Fnn.cmt --package foo ``

the result is:

```
      File "src/docOckCmi.ml", line 355, characters 21-27: Assertion failed
      Raised at file "src/docOckCmi.ml", line 355, characters 21-33
      Called from file "src/docOckCmt.ml", line 53, characters 22-63
      Called from file "src/docOckCmt.ml", line 87, characters 4-41
      Called from file "list.ml", line 88, characters 24-34
      Called from file "src/docOckCmt.ml", line 94, characters 4-336
      Called from file "src/docOckCmt.ml", line 503, characters 25-62
      Called from file "list.ml", line 88, characters 24-34
      Called from file "src/docOckCmt.ml", line 501, characters 4-136
      Called from file "src/docOckCmt.ml", line 357, characters 38-69
      Called from file "src/docOckCmt.ml", line 412, characters 22-60
      Called from file "src/docOckCmt.ml", line 460, characters 16-51
      Called from file "src/docOckCmt.ml", line 503, characters 25-62
      Called from file "list.ml", line 88, characters 24-34
      Called from file "src/docOckCmt.ml", line 501, characters 4-136
      Called from file "src/docOckCmt.ml", line 510, characters 14-46
      Called from file "src/docOck.ml", line 150, characters 31-75
      Called from file "lib/compile.ml", line 21, characters 8-17
      Called from file "src/cmdliner_term.ml", line 27, characters 19-24
      Called from file "src/cmdliner.ml", line 106, characters 32-39
```

The relevant code seems to be:

https://github.com/ocaml-doc/doc-ock/blob/ae76154f6d7651f03d888ff1363092d5751a2e0b/src/docOckCmi.ml#L348-L360

I also observed this being raised by `aws-s3/aws_s3__S3.cmt` from `aws-s3`.

aantron:
@lpw25, It looks like we can fix the assertion failure by simply not checking which constructor we have, since:

- We don't print it. Whatever the constructor is in `?(foo : bar some_option)` in the `.cmt` file, we print `?foo:bar` in the output, since we are using the `val` syntax of signatures, rather than the `let` syntax of structures.
- I think we can rely on the compiler to have unified the constructor with `option`.

The commit includes a test which reproduces the failure, if the fix is left out. The output looks right to me, as `let foo (type a) ?(bar : a opt) () = ()` should indeed generate `val foo : ?bar:'a -> unit -> unit`.

Fixes #101.

### DIFF
--- a/src/loader/cmi.ml
+++ b/src/loader/cmi.ml
@@ -368,7 +368,7 @@ let rec read_type_expr env typ =
           let arg =
             if Btype.is_optional lbl then
               match (Btype.repr arg).desc with
-              | Tconstr(_path, [arg], _) -> read_type_expr env arg
+              | Tconstr(_option, [arg], _) -> read_type_expr env arg
               | _ -> assert false
             else read_type_expr env arg
           in
--- a/src/loader/cmi.ml
+++ b/src/loader/cmi.ml
@@ -368,9 +368,7 @@ let rec read_type_expr env typ =
           let arg =
             if Btype.is_optional lbl then
               match (Btype.repr arg).desc with
-              | Tconstr(path, [arg], _)
-                  when OCamlPath.same path Predef.path_option ->
-                    read_type_expr env arg
+              | Tconstr(_path, [arg], _) -> read_type_expr env arg
               | _ -> assert false
             else read_type_expr env arg
           in
--- a/test/html/cases/bugs.ml
+++ b/test/html/cases/bugs.ml
@@ -1,0 +1,4 @@
+type 'a opt = 'a option
+let foo (type a) ?(bar : a opt) () = ()
+(** Triggers an assertion failure when
+    {:https://github.com/ocaml/odoc/issues/101} is not fixed. *)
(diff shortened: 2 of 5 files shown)


aantron:
Thanks for the speedy review, @trefis. I'll merge when the CI completes, mainly for checking the tester.

lpw25:
This is fine. Technically you can deliberately trick odoc into printing the wrong thing here, but its very unlikely to happen by accident and printing the wrong type in an extremely unlikely corner case is not a disaster. Hopefully upstream will fix the original issue at some point and then the problem will disappear.

aantron:
@lpw25 if you have time, could you describe what the corner case is?

lpw25:
I haven't actually tried it, but something like:

```ocaml
type 'a opt = int option

let foo (type a) ?(x : a opt) () = ()
```

would probably do it. It should cause this code to print something like:

```ocaml
val foo : ?x:'a -> unit -> unit
```

when it should really be:

```ocaml
val foo : ?x:int -> unit -> unit
```

aantron:
Indeed, I just tried it, and that's what happens. Thanks. Maybe I'll add a test case with a link to this comment.

Issue #227, opened 2018-10-24
Incorrect location reported for errors in doc attributes


rizo:
The column number reported in errors found in documentation attributes `[@@ocaml.doc ...]` is off by -3 columns. It seeems like it's not counting `(**`.

Some relevant discussion can be found here: https://github.com/ocaml/odoc/pull/226.

PR #231, opened 2018-10-31
Port new error message tests from #226 to master

aantron:
@rizo This PR is the test cases for the `odoc lint` command you proposed, ported to work with `odoc compile` instead (https://github.com/ocaml/odoc/pull/226#issuecomment-432717405). I set you as the author, because you wrote these tests. I want to commit them now, so I can make use of them in work I want to do next.

These tests will probably need some quality-of-life improvements for developers, but I'll make them as I go, as needed by (at least) me.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,7 @@ serve :
 .PHONY : test
 test : build
 	dune build @test/parser/runtest --no-buffer -j 1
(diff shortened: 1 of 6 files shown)


aantron:
Thanks @rizo. Locations are about to change anyway, as I'm merging this to help working on #227 ("Incorrect location reported for errors in doc attributes").

Issue #153, opened 2018-06-14
highlight.pack.js script removed after first build


rizo:
Currently when generating docs with dune (`dune build @doc`) the generated HTML uses the `highlight.pack.js` javascript library for syntax highlighting. This script is copied to `_build/default/_doc/_html/` along with `odoc.css` by the `odoc/assets.ml/write` function.

For some reason the second, and all consecutive builds, remove the `highlight.pack.js` file.

To reproduce run the following in any dune-managed package:

```
$ dune clean
$ dune build @doc
$ ls _build/default/_doc/_html/
highlight.pack.js  index.html  odoc.css  <package>
$ dune build @doc
$ ls _build/default/_doc/_html/
index.html  odoc.css  <package>        # <-- highlight.pack.js is missing
```

I find this strange because both the css and the javascript files seem to be handled equally in odoc, but only the javascript file disappears.

aantron:
This happens because `odoc.css` is a known target to Dune. By contrast, from Dune's point of view, it is an accident that odoc also emits `highlight.pack.js`. Dune then ends up deleting `highlight.pack.js` [here](https://github.com/ocaml/dune/blob/79076ae2f2493a9f1e95ea61e2e1017729faedd1/src/build_system.ml#L1152).

I see several ways to solve this:

1. List `highlight.pack.js` as a target (and dependency) in Dune ([here](https://github.com/ocaml/dune/blob/79076ae2f2493a9f1e95ea61e2e1017729faedd1/src/odoc.ml#L304) and [here](https://github.com/ocaml/dune/blob/79076ae2f2493a9f1e95ea61e2e1017729faedd1/src/odoc.ml#L220)).

    However, this way is a bad idea, because it will have to be repeated every time we change support files, resulting in an unreasonable coupling between odoc and Dune (and odig, and everything else). That will produce lots of version incompatibilities, constraints, etc. See also #174.

2. Add an `odoc support-files-targets` command (#174) that emits the names of the files odoc wants to generate.

3. We could instead have the convention that odoc writes all of its support files into a subdirectory of the HTML root, like `assets` or `static`, whose contents are opaque to the build system, and whose presence is all that is tested.

I suggest (3) because, in the future, odoc might need to output different sets of files depending on, for example, the theme. So, the implementation of `odoc support-files-targets` could become complicated. It seems to me that (3) is the solution that is easiest on both build tools and on odoc.

[...]

PR #232, opened 2018-10-31
Add odoc support-files-targets

### LINKED ISSUE #174: Add `odoc support-files-targets` command
To avoid build tools to making assumption about what the new `odoc support-files` compile command writes to the destination directory.

aantron:
With this PR:

```
$ odoc support-files-targets
odoc.css
highlight.pack.js
```

~~The command has no options~~.

I assume this output is suitable for consumption. cc @dbuenzli, @rgrinberg (though I will/can make a PR to Dune to use this).

Resolves #174.

See also https://github.com/ocaml/odoc/issues/153#issuecomment-434551361.

### DIFF
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -157,7 +157,7 @@ end = struct
       ~doc:"Compile a cmti, cmt, cmi or mld file to an odoc file."
 end
 
-module Support_files = struct
+module Support_files_command = struct
   let support_files without_theme output_dir =
     Support_files.write ~without_theme output_dir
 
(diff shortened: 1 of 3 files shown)


dbuenzli:
> The command has no options.

I think this needs an `-o DIR` option. This is not about listing the `support-files` it's about which file paths the `odoc support-files` will write.

aantron:
Ok, it now has the same options as `support-files`:

```
$ odoc support-files-targets --output-dir foo
foo/odoc.css
foo/highlight.pack.js
```

```
$ odoc support-files-targets --output-dir foo --without-theme
foo/highlight.pack.js
```

The only problem is that the code fails if the output directory doesn't exist. I'll undo that next, but I want to check whether the output and command line are otherwise reasonable.

dbuenzli:
Thanks ! Looks to me exactly what is needed.

Issue #178, opened 2018-09-04
uncaught exception: Parser__Helpers.InvalidReference


trefis:
Because of the reference on that line: https://github.com/ocamllabs/ocaml-ctypes/blob/master/src/ctypes/ctypes_types.mli#L367

```
odoc: internal error, uncaught exception:
      Parser___Helpers.InvalidReference("unknown qualifier `(@'")
      Raised at file "src/parser/helpers.ml", line 28, characters 14-72
      Called from file "src/parser/helpers.ml", line 258, characters 28-47
      Called from file "src/parser/helpers.ml", line 338, characters 10-38
```

aantron:
This is a duplicate of https://github.com/ocaml/odoc/issues/90#issuecomment-355248618.

aantron:
Actually, reopening this, because we want to keep the other issue for actually fixing the reference parser *syntax*.

The attached commit fixes the immediate issue of the current (and original) reference parser not being able to handle operators with dashes for the short term.

Issue #90, opened 2017-10-23
Disambiguating references


trefis:
Follow up to [#61](https://github.com/ocaml-doc/odoc/issues/61#issuecomment-319974223) : 

> This is a somewhat independent issue but: ocamldoc's syntax for references is vaguely broken as it doesn't (as far as we can see) allow to disambiguate the reference on the last line in the following:
> ```ocaml
> module M : sig
>   module type S : sig
>     type t
>   end
> 
>   module S : sig
>     type t
>   end
> end
> 
> (** Some reference to {!M.S.t} *)
> ```
> 
> Indeed the syntax ocamldoc provides only allows to disambiguate the last part of the reference, but not intermediate ones. So here one cannot say: "*I want `S` to be a module type*".
> 
> We're proposing to extend the syntax for references to be able to disambiguate intermediate parts of the path.
> In the previous example the fully disambiguated reference could look like this: `{!module-M.module-type-S.type-t}`.
> 
> N.B. An alternative syntax could be `{!M[module].S[module-type].t[type]}` but we have a preference for the first one (as it's closer to the urls generated by doc-ock-html). Or something better. Any ideas?

Actually, playing with pages and this new reference syntax, I realized that will using dashes to separate between the "kind" and the "identifier bit" is fine for any OCaml path, it might not be when referring to section labels, which might contain dashes.

There are a few options available:
1. use the other syntax mentioned previously (which I retrospectively find verbose and ugly)
2. use another separator (would `#` be OK? would we run into the same issues?)
3. keep using dashes for references and forbid it in section labels (erroring when we encounter one)

@aantron @dbuenzli @lpw25 : any preference?

dbuenzli:
> it might not be when referring to section labels, which might contain dashes.

I don't think this is the case, IIRC section labels with dashes make `ocamldoc` choke, you have to use underscore (I always suspected it was in fact expecting a valid ocaml identifier for these labels) so it's unlikely you'll find many of these in the wild.

> I realized that will using dashes to separate between the "kind" and the "identifier bit" is fine for any OCaml path,

This looks consistent with the fragment identifiers generated in the markup so maybe we should go for this.

trefis:
So, option 3 then?

dbuenzli:
That's what I would say.

aantron:
[...]

PR #237, opened 2018-11-06
Refactor the reference parser ({!type-foo} syntax)

aantron:
odoc introduces a new syntax for disambiguating references (see #90). Each component of a long identifier (such as `Foo.bar`) can be tagged with a kind:

```ocaml
(** {!module-Foo.type-bar} *)
```

I'm currently working on error recovery in the parser, and I needed to refactor the part that parses these tagged identifiers, hence this PR. It...

- Separates the reference parser into a lexical scan (https://github.com/ocaml/odoc/issues/90#issuecomment-355300304), and a separate pass for assembling the GADT path representation. The benefits are:
    - The code is a lot clearer.
    - It is now very easy to introduce proper error handling.
    - We can readily change the concrete syntax, should we continue discussing #90.
- Adds about 275 tests for the reference parser, exercising various paths through the GADT, and checking other corner cases. The reference parser was untested before this PR.
- Fixes a few bugs as a side effect, most notably:
    - The `.` in `{!(*.)}` was being treated as an identifier path separator (similar to #178 and https://github.com/ocaml/odoc/issues/90#issuecomment-355248618, in which `-` was being treated as kind-identifier separator, even in `{!(@->)}`).
    - Whitespace is trimmed correctly around path components.

This PR is mainly FYI rather than up for review. I will merge it relatively quickly because I need it for the error recovery branch, but any needed changes can be made in `master` after the fact.

cc @trefis FYI


trefis:
Out of curiosity: have you tried writing a simple ocamllex lexer instead of the manual one?

aantron:
No, because

1. This one is really simple and short.
2. We need to keep track of parentheses, and potentially parse right-to-left (though left-to-right should be possible as well), so I suspect the ocamllex lexer wouldn't be that much shorter, but would require a separate module, etc.

For anyone reading briefly, I want to clarify that the "main" parser does use an ocamllex lexer, this comment is just about the helper parser that handles references.

trefis:
Ack. Thanks :)

PR #238, opened 2018-11-06
Make the parser recover from all errors and continue parsing

aantron:
This PR turns all errors in the parser into warnings, after emitting each of which the parser recovers in some way, and continues parsing. The warnings are mostly the same as the errors that the parser used to emit (and stop), except for the reference parser, which has new detailed warnings replacing the generic exceptions it used to raise.

This is most of what is still needed for solving #108, "New docstrings parser too strict." I've been running the recovering parser on most of the repos in the `janestreet` organization with good results, but there is some other work to be done for #108 that is not as directly related to the parser, so I'll post in #108 when that is finished.

See the test cases for examples of recoveries. Most of them are straightforward (the parser can simply continue parsing). A few require new code that is not completely trivial. I think we will improve some of these latter over time, according to how often people actually need them.

The type of the parser has changed. The parser no longer returns a `result`, because it does not generate `Error _`.

cc @trefis, @lpw25

I don't think it's realistic to review the whole PR, though the main code changes (in `src/`) are not too bad. So, the cc is just FYI :) We can fix things before merge in this PR, and after merge in `master` as needed.


aantron:
Merged to avoid growing conflicts, since the changes, while not very complex, touch almost all parts of the parser a little bit, and touch every test case.

Issue #19, opened 2016-09-20
Markup for definitions


dbuenzli:
Note the following is about definitions not code fences.

The markup for definitions could be more semantic and the layout should be improved as it can be pretty bad on long definitions since it may wrap anywhere, something a sensitive OCaml programmer would never do. 

First these things would be good to have:
1. The whole definition should be wrapped by a `<code>` tag.
2. The identifier being defined should be `class`ified via a `<span>`.
3. The constructors of variants should be `class`ified via a `<span>` the same way they are classified in textual `code` spans.

Regarding the layout there are at least two solutions that can be tried. 

a) `<pre>` can be used instead of `<div class="def">` and all the formatting is done manually (this is the way `ocamldoc` proceeds), the disadvantage is that it will not be responsive to the page width. 

b) `<div class = "def">` is kept but we try to control the line breaking algorithm by only using spaces at the places where we would like to break (e.g. before `|`, after `->`) and using non breaking spaces U+00A0  where we don't want to break. I should then be able to make an indent via css' `text-indent`.

I have a slight preference to try what b) can give us.

dbuenzli:
> The constructors of variants should be classified via a <span> the same way they are classified in textual code spans.

In fact they are not at all in `odoc` I was looking at an `ocamldoc` generated page.

dbuenzli:
Also word-joiners (U+2060) should be introduced between the `?` and the identifier of optional labels.

trefis:
Do you have examples of `class`ification?

dbuenzli:
```
val get : key -> 'a t -> 'a

<span class="keyword">val</span> <span class="value-id">span</span> : <span "type-id">key</span>&nbsp;-> <span "type-var">'a</span>&nbsp;<span "type-id">t</span>&nbsp;-> <span class="type-var">'a</span>
```

`

trefis:
I did some work on that, it's most likely not perfect yet but I think you can already have a look.

dbuenzli:
It seems identifier (value and type) are not tagged, also you should avoid putting spaces inside the spans (e.g. in `val` keyword or in `->`). Linked identifiers should also be spanned and finally it would be nice not to tag `->` as `keyword` but `type-op`.  

For example:

```
<span class="keyword">val </span>list : 
<span class="type-var">'a</span> <a href="index.html#/testable.typ">testable</a>
<span class="keyword"> -&gt; </span>
[...]

PR #247, opened 2018-11-16
Fix exception spacing

### LINKED ISSUE #245: Missing space after exception declaration with 1.3.0 CSS
On firefox on linux (Xorg, intel GPU), no space is displayed between the `exception` keywords and the excpetion name : 
screenshot-2018-11-14 url ahrefskit url

Reproduced withFirefox 60 and nightly.

aantron:
Fixes #245 on Chrome (https://github.com/ocaml/odoc/issues/245#issuecomment-438677135). Not sure if that's the same problem that @jorisgio originally reported on Firefox, as well. @jorisgio, could you pin this branch, `fix-exception-spacing`, and try it?

@rizo, I removed the CSS for indenting lines in declarations:

```diff
-/* Indent the second line in multiline spec definitions. */
-.spec:not(.type) > code {
-  display: block;
-  padding-left: 4ex;
-  text-indent: -4ex;
-}
-
-.spec.exception > code {
-  display: inline-block;
-}
```

They were causing the issue in Chrome.

We have to indent by laying out the specifications correctly during HTML generation. Doing it with CSS gives undesirable results. For example, sometimes I saw line breaks in the middle of `->`, making the declaration difficult to read, and suppressing the ligature.

### DIFF
--- a/src/odoc/etc/odoc.css
+++ b/src/odoc/etc/odoc.css
@@ -300,17 +300,6 @@ dl, div.spec, .doc, aside {
   margin-bottom: 20px;
 }
 
-/* Indent the second line in multiline spec definitions. */
-.spec:not(.type) > code {
-  display: block;
-  padding-left: 4ex;
-  text-indent: -4ex;
-}
-
-.spec.exception > code {
-  display: inline-block;
-}
-
 dl > dd {
   padding: 0.5em;
 }
--- a/src/odoc/etc/odoc.css
+++ b/src/odoc/etc/odoc.css
@@ -279,7 +279,7 @@ pre code {
 /* Code lexemes */
 
 .keyword {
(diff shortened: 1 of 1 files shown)


dbuenzli:
> We have to indent by laying out the specifications correctly during HTML generation. Doing it with CSS gives undesirable results. For example, sometimes I saw line breaks in the middle of `->`, making the declaration difficult to read, and suppressing the ligature.

I think the right way of doing this would be to properly `span` the definitions as suggested in issue https://github.com/ocaml/odoc/issues/19. You can then solve the problem at the CSS level.

aantron:
I put #19 in one of the project to make it more visible.

aantron:
(and I confirmed this fixes the problem in Firefox)

Issue #236, opened 2018-11-05
Expose model, html, etc as public libraries


jfeser:
I have a small tool (https://github.com/jfeser/odoc2docset) that depends on odoc. I use code from `odoc.model` to extract names and `odoc.html` to handle html generated by odoc. Could these libraries be exposed publicly? Right now I have a fork, but I'd rather just use the released version.

aantron:
We can expose them with the understanding that their interface is not meant to be stable.

jfeser:
That works for me. Thanks!

aantron:
This seems to require giving the libraries globally unique names, like `Odoc__model`. The reason for the two underscores is that there are already modules with one. They come with the compiler and are part of ocamldoc.

This would make odoc's own code much uglier.

Still thinking about what to do here.

aantron:
Ok, the sublibraries are now exposed. All the sublibraries are installed upon `opam install odoc`.

The library in directory `model` is installed as ocamlfind package `odoc.model`, and the module name is `Odoc__model`. The rest of the sublibraries follow the same pattern.

To do this, I had to, of course, rename module `Model` to `Odoc__model`, etc. To avoid ugly names, and changing all our code for this issue, I added an alias module in `src/alias/odoc__alias.ml`. This allows us to use the renamed modules under their previous, neat names:

https://github.com/ocaml/odoc/blob/703b3344c0e01f68f30f4aa60cab9e0987205585/src/alias/odoc__alias.ml#L1-L7

This alias module is `-open`ed in all of odoc's code. It's a little bit dubious, but it's something we can undo quickly, if needed.

As mentioned, the interfaces of these libraries are unstable (including their names and existence), and we won't document them beyond what is needed by odoc. So, be careful :)

PR #252, opened 2018-11-20
Expose Html.Url module.

jfeser:
Minor addition to #236.


aantron:
I'm going to add a comment to it in `master`, as otherwise I or someone else is likely to remove this alias when we see that odoc isn't using it, during refactoring. Beware the human GC.

aantron:
Cherry-picked to `master`, as the base was set to the `expose-sublibraries` branch for this PR.

Issue #213, opened 2018-10-09
Test nested HTML pages


rizo:
Currently the HTML tests only cover the root module `M/index.html` page. The nested sub-pages are simply ignored. We can adapt the diffing to iterate on the full tree in the `expect` directory and validate it with the contents generated in `_scratch`.

rizo:
FYI there are currently 48 HTML "nested" files not covered by tests:

```
_build/default/test/html/_scratch/test_package+ml/Class/class-mutually'/index.html
_build/default/test/html/_scratch/test_package+ml/Class/class-recursive'/index.html
_build/default/test/html/_scratch/test_package+ml/Class/class-type-empty/index.html
_build/default/test/html/_scratch/test_package+ml/Class/class-type-mutually/index.html
_build/default/test/html/_scratch/test_package+ml/Class/class-type-recursive/index.html
_build/default/test/html/_scratch/test_package+ml/Functor/F1/argument-1-Arg/index.html
_build/default/test/html/_scratch/test_package+ml/Functor/F1/index.html
_build/default/test/html/_scratch/test_package+ml/Functor/F2/argument-1-Arg/index.html
_build/default/test/html/_scratch/test_package+ml/Functor/F2/index.html
_build/default/test/html/_scratch/test_package+ml/Functor/F3/argument-1-Arg/index.html
_build/default/test/html/_scratch/test_package+ml/Functor/F3/index.html
_build/default/test/html/_scratch/test_package+ml/Functor/F4/argument-1-Arg/index.html
_build/default/test/html/_scratch/test_package+ml/Functor/F4/index.html
_build/default/test/html/_scratch/test_package+ml/Functor/module-type-S/index.html
_build/default/test/html/_scratch/test_package+ml/Functor/module-type-S1/argument-1-_/index.html
_build/default/test/html/_scratch/test_package+ml/Functor/module-type-S1/index.html
_build/default/test/html/_scratch/test_package+ml/Include/module-type-Inlined/index.html
_build/default/test/html/_scratch/test_package+ml/Include/module-type-Not_inlined/index.html
_build/default/test/html/_scratch/test_package+ml/Include/module-type-Not_inlined_and_closed/index.html
_build/default/test/html/_scratch/test_package+ml/Include/module-type-Not_inlined_and_opened/index.html
_build/default/test/html/_scratch/test_package+ml/Module/M'/index.html
_build/default/test/html/_scratch/test_package+ml/Module/module-type-S/index.html
[...]

PR #261, opened 2018-12-03
Render TOC in nested pages

rizo:
Currently the table of contents is only rendered for top-level modules. This PR implements the rendering for all nested items (modules, module types, classes, functors and functor parameters) and makes some other minor adjustments.

### Full list of changes

* Render TOC (if not empty) inside header in all nested pages.
* Fix a bug where functor arguments were listed in reverse order ([diff](https://github.com/ocaml/odoc/compare/master...rizo:render-nested-toc?expand=1#diff-46e526e8d708dd25c7f790e20662ad2eR1371)).
* Remove the `doc` CSS class from class items, to be consistent with other nested items ([diff](https://github.com/ocaml/odoc/compare/master...rizo:render-nested-toc?expand=1#diff-46e526e8d708dd25c7f790e20662ad2eL1171)).
* Render functor parameters as `ul` (and not as a `dl`).
* Extend the "nested" HTML test to cover the proposed changes.
  * The TOC rendering for nested pages is not currently tested because of https://github.com/ocaml/odoc/issues/213.

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -1355,13 +1355,14 @@ struct
           | None -> acc
           | Some arg ->
             let arg, arg_subpages = functor_argument ?theme_uri arg in
-            (args @ arg, subpages @ arg_subpages)
+            let arg = Html.li arg in
+            (args @ [arg], subpages @ arg_subpages)
         )
         ([], []) args
       in
       let html =
         Html.h3 ~a:[ Html.a_class ["heading"] ] [ Html.txt "Parameters" ] ::
-        Html.dl (List.map Html.Unsafe.coerce_elt params) ::
+        Html.ul (List.map Html.Unsafe.coerce_elt params) ::
         Html.h3 ~a:[ Html.a_class ["heading"] ] [ Html.txt "Signature" ] ::
         sig_html
       in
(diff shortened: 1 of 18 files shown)


aantron:
I suggest...

- Addressing #213 first.
- Splitting up the one commit into the individual changes that were made, so each can be reviewed trivially. Note that links to code diffs don't also link to diffs in the test cases, and the net effect of combining everything is to partially defeat the advantage to review offered by having expected output committed to the repo.

I can take care of #213, but please let me know if you have intent to do it.

rizo:
Thanks for the suggestions and apologies for bundling everything (the changes seemed rather small). I changed the history to reflect the individual changes accordingly.

> Addressing #213 first.

There's a large number of other nested files (40+) the are not currently tested. Although I agree that #213 must to be fixed, I don't see why this PR should be blocked until then. I can add the generated nested expect files if that helps for now.

I'm currently working on other PRs, so feel free to work on #213.

rizo:
It is possible to test this now! I'll add the generated files later today.

rizo:
@aantron I think this can be merged now, unless you have any objections/suggestions.

rizo:
Thanks, @dbuenzli.

I think @aantron had some potential suggestions with regards to testing in this PR. @aantron, if that's the case, please let me know and I'll work on that separately.

aantron:
Yes. If the changes are trivial, I'll just make them.

PR #262, opened 2018-12-03
Create output directory when generating HTML.

rizo:
Fixes https://github.com/ocaml/odoc/issues/171.

PR #264, opened 2018-12-03
Fix output dir creation

rizo:
Fixes https://github.com/ocaml/odoc/pull/262#issuecomment-443858040.

### DIFF
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -68,10 +68,10 @@ let hidden =
   in
   Arg.(value & flag & info ~docs ~doc ["hidden"])
 
-let dst =
+let dst ?create () =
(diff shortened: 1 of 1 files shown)


src/odoc/bin/main.ml
```diff
@@ -69,8 +69,7 @@ let hidden =
   Arg.(value & flag & info ~docs ~doc ["hidden"])
 
 let dst ?create () =
-  let doc = "Output directory where the HTML tree is expected to be saved. \
-             Will be created if it doesn't exist." in
+  let doc = "Output directory where the HTML tree is expected to be saved." in
```
aantron: Just a note: I think this description is inaccurate for the usage of `dst` in `compile-targets`. This inaccuracy is not introduced by these PRs, however.

aantron:
Thanks again :)

PR #266, opened 2018-12-06
Test nested HTML pages

### LINKED ISSUE #213: Test nested HTML pages

rizo:
Attempts to ~~fix~~ #213.

I plan to split the work in two parts:

- [x] Adapt the runner to diff multiple outputs per test case.
- [ ] ~~Add, review and promote actual HTML files.~~ (see comments below)

Maybe it's better to split the second point in multiple commits to make review easier.

### DIFF
--- a/test/html/test.ml
+++ b/test/html/test.ml
@@ -64,19 +64,20 @@ module Case = struct
     kind: [ `mli | `mld | `ml ];
     theme_uri: string option;
     syntax: [ `ml | `re ];
+    outputs: string list;
   }
 
-  let make ?theme_uri ?(syntax = `ml) basename =
-    let name = Filename.chop_extension basename in
+  let make ?theme_uri ?(syntax = `ml) (input, outputs) =
+    let name = Filename.chop_extension input in
     let kind =
-      match Filename.extension basename with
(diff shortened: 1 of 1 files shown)


aantron:
Thanks, I'll take a close look tomorrow!

I recommend adding a minimum of extra files, i.e. omitting as much of the list in #213 as possible.

aantron:
I suggest doing the second part as part of #261. We don't need to test every nested page. Many are generated as unneeded side effects of having dummy modules, for example. We just need to test nested pages in those tests where a useful feature is exercised, such as one page for #261.

aantron:
(...which means I think this is ready for merge so that #261 can be rebased over it, let me know if you agree)

rizo:
I agree @aantron, thanks for the review. Let's merge this and rebase #261.

I would suggest only closing #213 when we decide what other nested pages should be tested.

aantron:
Thanks! #213 is still open after this merge.

Issue #148, opened 2018-05-31
odoc: Internal error, uncaught exception in `src/odoc/env.ml`


ejgallego:
Using the tree at https://github.com/coq/coq/pull/6857 I get from `jbuilder build @doc`:

```
(cd _build/default/_doc/_html && /home/egallego/.opam/4.06.1/bin/odoc html -I ../_odoc/lib/coq.checker -I ../_odoc/lib/coq.clib -I ../_odoc/lib/coq.config -I ../_odoc/lib/coq.engine -I ../_odoc/lib/coq.grammar -I ../_odoc/lib/coq.interp -I ../_odoc/lib/coq.kernel -I ../_odoc/lib/coq.lib -I ../_odoc/lib/coq.library -I ../_odoc/lib/coq.parsing -I ../_odoc/lib/coq.pretyping -I ../_odoc/lib/coq.printing -I ../_odoc/lib/coq.proofs -I ../_odoc/lib/coq.stm -I ../_odoc/lib/coq.tactics -I ../_odoc/lib/coq.toplevel -I ../_odoc/lib/coq.vernac -I ../_odoc/lib/coq.vm -o . ../_odoc/lib/coq.engine/termops.odoc)
WARNING: digest of ../_odoc/lib/coq.checker/names.odoc doesn't match the one excepted for file Names
odoc: internal error, uncaught exception:
      File "src/odoc/env.ml", line 82, characters 6-12: Assertion failed
      Raised at file "src/odoc/env.ml", line 82, characters 6-38
      Called from file "src/odoc/env.ml", line 128, characters 8-45
      Called from file "src/xref/component_table.ml", line 276, characters 16-35
      Called from file "src/xref/component_table.ml", line 349, characters 19-47
      Called from file "src/xref/component_table.ml", line 349, characters 19-47
      Called from file "src/xref/component_table.ml" (inlined), line 596, characters 4-32
      Called from file "src/xref/resolve.ml", line 155, characters 32-63
      Called from file "src/xref/resolve.ml", line 368, characters 12-50
      Called from file "src/model/maps.ml", line 1895, characters 19-33
      Called from file "src/model/maps.ml", line 1887, characters 21-35
      Called from file "src/model/maps.ml", line 1887, characters 21-35
      Called from file "src/model/maps.ml", line 1502, characters 16-30
      Called from file "src/model/maps.ml", line 1192, characters 19-29
      Called from file "list.ml", line 82, characters 20-23
[...]

PR #268, opened 2018-12-07
Scan include directories for odoc files eagerly

jonludlam:
Goes some way to addressing #148.

This is an RFC, particularly the error handling. As is, this still results in a nasty backtrace spewed to the terminal so I don't think it should be merged yet. Before I just wrapped the invocations of Env.create in an exception handler I thought I'd better ask whether there was a better plan in mind for this sort of non-recoverable one?

### DIFF
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -53,6 +53,13 @@ let convert_uri : Html.Tree.uri Arg.converter =
 
 let docs = "ARGUMENTS"
 
+let handle_fatal_err f =
+  match Model.Error.catch f with
+  | Ok x -> x
+  | Error y ->
+    Printf.eprintf "FATAL: %s\n%!" (Model.Error.to_string y);
+    exit 2
+
 let odoc_file_directories =
   let doc =
     "Where to look for required .odoc files. \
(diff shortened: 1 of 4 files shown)


aantron:
We do eventually plan to have more disciplined error handling at the top level, but so far we've been working "bottom-up" by introducing proper error handling in the parser first, and gradually propagating new error handling up to the command-line interface level. This process has reached only the parser and parts of the loader so far, however.

I'd say to do whatever you think is easiest here :) It could be the right solution for the long term, but even if not, that's okay, because then we will still address it later :)

jonludlam:
OK. I'll stick with raising the exception from Model.Errors and catch it right at the top.

I'll need to stick in some more error handling around the eager loading too, since having just tested it on Coq I find it's failing with `End_of_file`, because dune is running a whole bunch of odoc instances in parallel all writing out `odoc` files and hence we can't guarantee to be able to read them.

lpw25:
> because dune is running a whole bunch of odoc instances in parallel all writing out odoc files and hence we can't guarantee to be able to read them.

You should be able to use the imports list to restrict the eager scanning to just those files we will actually use. This should remove issues with parallelism and also prevent non-reproducible documentation. Essentially you should only include files whose name matches those in the [import list](https://github.com/ocaml/odoc/blob/master/src/model/lang.ml#L444).

There could still be problems if there are two files with the same name, and dune thinks only one of them is a dependency. So some protection for `End_of_file` should still be there, but the problem should then be extremely rare and also a symptom of an actual problem with how the build system is handling dependencies.

jonludlam:
Ah, this sounds much better. I'll have a look.

jonludlam:
This is the wrong solution.

aantron:
@jonludlam Could you link to something, or state in one brief sentence what the core of the reasoning is?

jonludlam:
I was planning on doing so in the follow-up PR, which I have cooking. I just didn't want to waste anyones time reading this when I'm intending to do something different

aantron:
Fair enough, thanks.

Issue #209, opened 2018-10-08
[BuckleScript] Automatic namespacing breaks linking of modules


ostera:
BuckleScript will by default do automatic namespacing of every module when compiling down `.cmt[i]` files, using the `npm/bsconfig` package name as a _suffix_.

I.e, if the package name is `BsReactNative`, and the module name is `ActivityIndicator`, the compiled `.cmti` file will be `ActivityIndicator-BsReactNative.cmti`.

Personally, I do my namespacing by hand, but a lot of people have it set up by default, which means that they can't really reference modules like `ActivityIndicator` above since they don't seem exist. What's worse, the suffixed name is _not a valid OCaml module name_, so it can't be referenced either.

In fact, [bs-react-native](https://github.com/reasonml-community/bs-react-native/) has been experiencing this, as shown in https://github.com/reasonml-community/bs-react-native/pull/259.

What I suggest is fixing this at the boundaries of `odoc/bsb`: the running script for `odoc`, as published by the npm package, could fix this by moving a few things around (i.e, copying and renaming .cmti files before running).

Alternatively, we could consider a `--namespace-suffix` flag that, if present, would strip the suffix from the file name when it builds up the compilation unit.

cc/ @aantron @rizo @grabbou

lpw25:
Before rushing in with a feature like `--namespace-suffix` could we have some description of what Bucklescript is actually doing to implement namespacing? It sounds to me like the currently proposed solutions are basically undoing Bucklescript's namespacing -- so that you'll get conflicts if you try to generate documentation for multiple Bucklescript libraries that have modules with the same name.

Note that dune does namespacing of OCaml modules and odoc is able to correctly handle that without adding ad-hoc features that diverge from OCaml's semantics.

ostera:
@lpw25 good insight! I’ll try to get a word from @chenglou / @bobzhang and see if there’s a way for BuckleScript to do their namespacing like dune does, and how and why they diverge.

ostera:
I recently learned in a discussion with @bobzhang that the reason to do suffix-namespacing is so that the generated .cmt[i] files are __purposefully__ not valid OCaml names.

This means you can't import an "internal" module, such as `ActivityIndicator-BsReactNative` because well, it's not a valid OCaml module name. Instead, you're forced to use it through the name of the package: `BsReactNative.ActivityIndicator`.

I'll explain more in a PR to BuckleScript that I intend to open tomorrow. Will backlink here.

PR #269, opened 2018-12-07
Switch bash driver with bsdoc

ostera:
After long playing around with the bash script used to drive odoc for
BuckleScript projects, I decided it could be turned into a cli tool,
npm-installable, that would do the same thing without requiring copying
files around or leaking how odoc needs to be orchestrated.

This could pave the way to a deeper integration with the BuckleScript
platform, perhaps in the future `bsdoc` can be absorbed into it.

So here I just replace the references to such script with references to
the tool instead.

Let me know what you think!

cc/ @aantron @trefis @ryyppy @bobzhang

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,74 +54,50 @@ The generated docs can then be found locally at
 
 ## Using `odoc` with BuckleScript/Reason
 
-> BuckleScript support currently requires the latest development version of odoc.
-
-While the BuckleScript/Reason toolchain relies on `npm`, `odoc` at the moment
-needs to be used from a working OCaml toolchain.
-
-This means we follow the same installation than above, but using the
-`4.02.3+buckle-master` version of the OCaml compiler.
+You can use the [`bsdoc`](https://ostera.github.io/bsdoc) npm package to use
+`odoc` in your BuckleScript projects.
 
 ```sh
-λ opam switch 4.02.3+buckle-master
-λ eval `opam config env`
-λ opam pin add odoc.dev git+https://github.com/ocaml/odoc
+$ cd MyProject
+MyProject $ yarn add bsdoc --dev
+MyProject $ yarn build
 
+MyProject $ yarn run bsdoc support-files 
+yarn run v1.12.3
+$ ./node_modules/.bin/bsdoc support-files
+info: Copying support files (CSS, JS) into ./docs
+info: Done ✅
+✨  Done in 0.60s.
+
+MyProject $ yarn run bsdoc build MyProject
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -54,74 +54,50 @@ The generated docs can then be found locally at
...
-`4.02.3+buckle-master` version of the OCaml compiler.
+You can use the [`bsdoc`](https://ostera.github.io/bsdoc) npm package to use
+`odoc` in your BuckleScript projects.
 
 ```sh
-λ opam switch 4.02.3+buckle-master
-λ eval `opam config env`
-λ opam pin add odoc.dev git+https://github.com/ocaml/odoc
-
+$ cd MyProject
```
rizo: Given that the instructions are already provided in [[bsdoc]'s README](https://github.com/ostera/bsdoc#getting-started), I don't think it's necessary to repeat them here. Each tool should be responsible for its own documentation, otherwise things will get out of date eventually.

rizo:
It's good to see progress in this direction. Even though this is much better than the bash script `bsdoc` replaces, I still think we should make efforts to integrate `odoc` with BuckleScript officially (if that is what the community wants).

I understand there might be some obstacles on both sides but it would be good to enumerate them. Specifically:
* What is currently missing in odoc in terms of features to support BuckleScript projects?
* What can we do to help with the integration? `odoc` is a build system friendly tool so it shouldn't be to hard to achieve this, but there might be some issues like https://github.com/ocaml/odoc/issues/209.

ostera:
@rizo, opened https://github.com/BuckleScript/bucklescript/issues/3198 to keep track of the BuckleScript side of the integration.

Hopefully it pans out like the `dune` one and there's nothing to be changed in `odoc`.

I'll update the commits to just point to the `bsdoc` docs!

ostera:
> @rizo: What is currently missing in odoc in terms of features to support BuckleScript projects?

From looking at the [Reason and BuckleScript](https://github.com/ocaml/odoc/projects/2) project I see two big outstanding things we should work on:

1. https://github.com/ocaml/odoc/issues/199 — bring reason syntax up to par is important, since most BuckleScript users would come from the Reason side of the community.

2. https://github.com/ocaml/odoc/issues/129 — switching syntaxes would then again make all docs readable for both parts of the ecosystem.

But I don't see any _blocking_ issues on the `odoc` side. The namespacing one stands out but it's on bucklescript's court.

rizo:
Thansk @ostera! We can continue the discussion on the issue you opened.

PR #270, opened 2018-12-14
Env: handle module name clashes

jonludlam:

### DIFF
--- a/src/odoc/env.ml
+++ b/src/odoc/env.ml
@@ -87,7 +87,18 @@ module Accessible_paths = struct
     | exception Not_found ->
       let paths = find_file_by_name t filename in
       (* This could be the empty list *)
-      let roots = List.map Root.read paths in
+      let filter_map f l =
+        List.fold_right (fun x acc -> match f x with | Some y -> y::acc | None -> acc) l []
+      in
+      let safe_read file =
+        try Some (Root.read file)
+        with
+        | End_of_file ->
+          let warning = Model.Error.filename_only "End_of_file while reading" (Fs.File.to_string file) in
+          prerr_endline (Model.Error.to_string warning);
+          None
+      in
+      let roots = filter_map safe_read paths in
       Hashtbl.add t.file_map filename roots;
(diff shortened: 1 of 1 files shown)


src/odoc/env.ml
```diff
@@ -38,61 +54,83 @@ module Accessible_paths = struct
...
-    | exception Not_found -> Not_found
-    | root ->
-      match root.file with
-      | Compilation_unit {hidden; _} -> Xref.Found {root; hidden}
-      | Page _ -> assert false
+    match Accessible_paths.find_root ap ~filename:target_name, digest with
+    | [], _ -> Xref.Not_found
+    | [r], _ -> handle_root r (* Already checked the digest, if one's been specified *)
+    | r :: rs, None ->
+      Printf.fprintf stderr "Warning, ambiguous lookup. Please wrap your libraries. Possible files:\n%!";
```
aantron: In the future, we will probably want to provide more context about which source file depended on the ambiguous compilation unit. Just a note.

src/odoc/env.ml
```diff
@@ -38,61 +54,83 @@ module Accessible_paths = struct
...
           let filename = "page-" ^ page_name in
-          find_root ~digest:root.digest t ~filename
+          check_optional_digest ~digest:root.digest filename @@ find_root t ~filename
         | Compilation_unit { name; _ } ->
-          find_root ~digest:root.digest t ~filename:name
+          check_optional_digest ~digest:root.digest name @@ find_root t ~filename:name
       in
-      assert (Model.Root.equal root r);
-      Model.Root.Hash_table.find t.root_map r
+      file_of_root t root
```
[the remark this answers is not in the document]
jonludlam: Ah! yes, remnants of a previous version. Will fix.

src/odoc/env.ml
jonludlam: Yes, that'd be good. We either need a richer return type or to pass in the location into here. I don't know which is better.

PR #271, opened 2018-12-18
Fix odoc compile-deps.

### LINKED ISSUE #162: `odoc compile-deps` does not work on `.cmt` files
```
> odoc --version
v1.1.1-710-gdd47a98
> odoc compile-deps int64.cmt
      Cmi_format.Error(_)
      Raised at file "typing/cmi_format.ml", line 74, characters 6-21
      Called from file "src/odoc/depends.ml", line 31, characters 18-43
      Called from file "src/odoc/bin/main.ml", line 218, characters 17-72
      Called from file "src/cmdliner_term.ml", line 27, characters 19-24
      Called from file "src/cmdliner.ml", line 106, characters 32-39
```

dbuenzli:
Make it work on cmt files ~~and do not report the unit itself as as a dependency~~. Closes #162.

### DIFF
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -306,7 +306,8 @@ module Depends = struct
   let info =
     Term.info "compile-deps"
       ~doc:"List units (with their digest) which needs to be compiled in order \
-            to compile this one."
+            to compile this one. The unit itself and its digest is also \
+            reported in the output."
   end
 
   module Html = struct
--- a/src/odoc/depends.ml
+++ b/src/odoc/depends.ml
@@ -26,14 +26,21 @@ module Compile = struct
   let digest t = t.digest
 end
 
-let for_compile_step file =
-  let input = Fs.File.to_string file in
-  let cmi_infos = Cmi_format.read_cmi input in
-  List.fold_left ~f:(fun acc -> function
-    | _, None -> acc
-    | unit_name, Some digest -> { Compile. unit_name; digest } :: acc
-  ) ~init:[] cmi_infos.Cmi_format.cmi_crcs
+let add_dep acc = function
+| _, None -> acc (* drop module aliases *)
+| unit_name, Some digest ->  { Compile.unit_name; digest } :: acc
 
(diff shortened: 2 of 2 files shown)


dbuenzli:
I wonder if the digest of the unit itself should not be kept in the end. The cmt format read functions used by `compile-deps` are quite slow (on my machine a non parallel `odoc compile-deps` on ~3670 files takes about 43secs). 

However a good dependency resolver will need to know about the actual digest of the unit and the way of getting it is to use these same functions `compile-deps` is using. This means we end up using the slow functions twice. If we keep the unit's digest in the output we can invoke the tool to give us both the unit digest and deps.

Thoughts ?

dbuenzli:
I discussed this privately with @trefis, he has no opinion on the subject but sympathises with the issue. We decided that maybe best was to first print the unit digest's on the first line followed by the actual dependencies in the current format and document the format in the manpage.

dbuenzli:
In the end keeping the same format for all lines felt more natural. I do mention however in the command doc string that a self-dependency is reported.

jonludlam:
I'm all for efficiency so this makes perfect sense to me. I'm don't think dune is using compile-deps yet - is odig/bsdoc?

dbuenzli:
> I'm all for efficiency so this makes perfect sense to me. I'm don't think dune is using compile-deps yet - is odig/bsdoc?

The next version of `odig` should. I'm trying to get rid of the compilation object index I was performing myself and rely solely on `odoc` commands to drive the generation.

In any case if there are existing users of `compile-deps` this PR change nothing for them except enabling the command on `cmt` files.

(Also aside improving the performance of this command is not trivial since they rely on `Cmt_format.read` which relies on `Marshal`, we'd need to change the way `cmt` files themselves are read and written upstream).

aantron:
I restarted the macOS build. If it fails again, I think we can merge, and hope upstream fixes the macOS setup in some days. If not, we will debug the macOS build then.

aantron:
Ok, macOS build failed with the same problem, and I think this is ready for merge.

PR #272, opened 2018-12-18
Cli: be more lenient with directory existence.

### LINKED ISSUE #25: References in doc strings
are not hyperlinked and references to identifiers not wrapped in a `code` tag. 

Contrast the links in the preamble of this module: 

http://docs.mirage.io/odoc/astring/Astring/index.html
http://docs.mirage.io/astring/Astring.html

Contrast the doc string of the two first values in the `String` section here:

http://docs.mirage.io/odoc/astring/Astring/index.html
http://docs.mirage.io/astring/Astring.html

### LINKED ISSUE #35: `-I $DIR` option needs $DIR to exist
Currently the `-I $DIR` option requires $DIR to exist. I'm not sure this is a good idea. Any reason why this is the case ?

### LINKED ISSUE #170: `odoc html-target --dir DIR` needs `DIR` to exist.
Somehow I don't think this is a good idea. We had a similar discussion [here](https://github.com/ocaml/odoc/issues/35).

dbuenzli:
Do not require -I directories to exist in `compile` command
(closes #35). Do not require the -o directory to exist in
`*-targets` commands (closes #170).

I have checked the transitive usage of `convert_directory` via `dst` a few lines below and I believe this to be correct.

### DIFF
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -19,7 +19,7 @@ let convert_syntax : Html.Tree.syntax Arg.converter =
   (syntax_parser, syntax_printer)
 
 let convert_directory ?(create=false) () : Fs.Directory.t Arg.converter =
-  let (dir_parser, dir_printer) = Arg.dir in
+  let (dir_parser, dir_printer) = Arg.string in
   let odoc_dir_parser str =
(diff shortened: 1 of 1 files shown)


aantron:
Thanks

PR #273, opened 2018-12-18
Fix compile-targets.

### LINKED ISSUE #172: Fix or remove `odoc compile-targets`
Somehow it seems it didn't evolve with the corresponding `odoc compile` command (`-o` is still a required argument and a directory in `compile-targets`). 

Unless it is envisioned for `odoc compile` to produce multiple outputs in the future I'd move on to simply remove the command.

dbuenzli:
Rather than duplicate output logic in the Target module, we
delete it and expose both cli arguments and logic from the compile
command for usage by the compile-targets command. This decreases
the probability of those getting out of sync. Closes #172.

### DIFF
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -74,44 +74,46 @@ let dst ?create () =
        info ~docs ~docv:"DIR" ~doc ["o"; "output-dir"])
 
 module Compile : sig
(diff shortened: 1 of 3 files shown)


aantron:
Thanks, @dbuenzli!

aantron:
...and @jonludlam  :)

PR #278, opened 2019-01-15
render externals as function signature

### LINKED ISSUE #275: Improving `external` output
@ryyppy and I are currently looking into using odoc (bsdoc) for multiple BuckleScript projects.

As an example I generated docs for ReasonReact. See a preview below.

I'm wondering what would be the best course of action:
- Should we simply ignore the right hand side for external?
- Is there a way to present it nice? If so would this include BuckleScript FFI or only the native one?

I'm happy to contribute to a solution. First thing would be to identify the desired output.

screenshot 2019-01-08 at 17 38 38

nikgraf:
Closes #275 

This is how the ReasonReact docs would look like after this change:

screenshot 2019-01-15 at 10 15 59

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -631,12 +631,10 @@ struct
   let external_ (t : Model.Lang.External.t) =
     let name = Paths.Identifier.name t.id in
     let external_ =
-      keyword "external " ::
+      keyword (Syntax.Value.variable_keyword ^ " ") ::
       Html.txt name ::
       Html.txt Syntax.Type.annotation_separator ::
-      type_expr t.type_ @
-      Html.txt " = " ::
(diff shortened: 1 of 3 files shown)


rizo:
I think not leaking the external identifier is a good idea. But I would argue that knowing that a value is  external is still useful. Maybe keeping the `external` keyword and hiding the identifier is a more informative option.

I don't feel strongly about this though.

aantron:
@rizo I would argue against making `external` visible. In particular, a non-external value, documented as a `let`/`val` could become `external` in a future version of the library. I don't think the docs should change in that case, so that the value is now documented as `external`.

aantron:
Thanks for the PR!

PR #279, opened 2019-01-15
ignore node modules

nikgraf:
When following the CONTRIBUTION.md guide during the install process a `node_modules` directory is created. I believe it should be ignored for git

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ _esy
 
 # Generated docs
 docs/
+
+# Node Modules
+node_modules/


ostera:
I can't seem to reproduce this — both following the contribution steps and running `esy` do not end up in us creating a `node_modules` folder 🤔 for me.

How'd you end up with it?

nikgraf:
good question, I followed the contribution guide. Just tried again from the beginning, but didn't run into the issue.

PR #280, opened 2019-01-17
Read ODOC_SYNTAX env var in odoc html command.

rizo:
A simple workaround until https://github.com/ocaml/odoc/issues/129 is properly implemented. Setting `ODOC_SYNTAX=re` before calling `dune build @doc` will generate docs with reason syntax.

CC @ryyppy


aantron:
Thanks!

PR #287, opened 2019-01-29
Update package.json to be less restrictive

ulrikstrid:
This is has less restrictive versions and I changed to build command to just run `dune build` directly instead of via `make`

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "Leo White <leo@lpw25.net>"
   ],
   "esy": {
-    "build": "make build",
+    "build": "dune build",
     "release": {
       "releasedBinaries": [
         "odoc"
@@ -25,35 +25,28 @@
         "ocamlfind",
         "sexplib"
       ]
-    },
-    "buildsInSource": "_build"
+    }
   },
   "dependencies": {
-    "@opam/alcotest": "0.8.3",
-    "@opam/astring": "0.8.3",
-    "@opam/bisect_ppx": "1.3.4",
-    "@opam/bos": "0.2.0",
-    "@opam/cmdliner": "1.0.2",
-    "@opam/cppo": "1.6.5",
-    "@opam/dune": "1.3.0",
-    "@opam/fpath": "0.7.2",
-    "@opam/js-build-tools": "113.33.04",
-    "@opam/lambdasoup": "0.6.3",
-    "@opam/markup": "*",
-    "@opam/oasis": "0.4.11",
-    "@opam/ocamlfind": "1.8.0",
-    "@opam/result": "1.3",
-    "@opam/sexplib": "*",
-    "@opam/tyxml": "4.3.0",
-    "ocaml": "~4.2.0"
-  },
-  "resolutions": {
-    "@opam/markup": "aantron/markup.ml:markup.opam#9f8e77",
-    "@opam/sexplib": "bryphe/sexplib#548f94d"
+    "@opam/bisect_ppx": "^1.3.4",
+    "@opam/bos": "*",
+    "@opam/dune": "^1.3.0",
+    "@opam/tyxml": "^4.3.0",
+    "ocaml": "*"
   },
   "devDependencies": {
+    "@opam/alcotest": "^0.8.3",
+    "@opam/astring": "^0.8.3",
+    "@opam/cmdliner": "^1.0.2",
+    "@opam/cppo": "^1.6.5",
+    "@opam/fpath": "^0.7.2",
+    "@opam/markup": ">= 0.8.0",
     "@opam/merlin": "^3.0.3",
-    "ocaml": "~4.2.0"
+    "@opam/ocamlfind": "^1.8.0",
+    "@opam/result": "*",
+    "@opam/rresult": "*",
+    "@opam/sexplib": "*",
+    "ocaml": "*"
   },
   "license": "ISC",
   "private": false
--- a/package.json
+++ b/package.json
@@ -28,25 +28,25 @@
     }
   },
   "dependencies": {
-    "@opam/bisect_ppx": "^1.3.4",
+    "@opam/astring": "*",
     "@opam/bos": "*",
-    "@opam/dune": "^1.3.0",
+    "@opam/cmdliner": "^1.0.0",
+    "@opam/cppo": "*",
+    "@opam/dune": "*",
+    "@opam/fpath": "*",
+    "@opam/result": "*",
+    "@opam/rresult": ">= 0.3.0",
     "@opam/tyxml": "^4.3.0",
-    "ocaml": "*"
+    "ocaml": "^4.4.0"
   },
   "devDependencies": {
-    "@opam/alcotest": "^0.8.3",
-    "@opam/astring": "^0.8.3",
-    "@opam/cmdliner": "^1.0.2",
-    "@opam/cppo": "^1.6.5",
-    "@opam/fpath": "^0.7.2",
+    "@opam/alcotest": ">= 0.8.3",
+    "@opam/bisect_ppx": "^1.3.0",
     "@opam/markup": ">= 0.8.0",
-    "@opam/merlin": "^3.0.3",
-    "@opam/ocamlfind": "^1.8.0",
-    "@opam/result": "*",
-    "@opam/rresult": "*",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -25,35 +25,28 @@
...
+    "@opam/cmdliner": "^1.0.2",
+    "@opam/cppo": "^1.6.5",
+    "@opam/fpath": "^0.7.2",
+    "@opam/markup": ">= 0.8.0",
     "@opam/merlin": "^3.0.3",
-    "ocaml": "~4.2.0"
+    "@opam/ocamlfind": "^1.8.0",
+    "@opam/result": "*",
+    "@opam/rresult": "*",
+    "@opam/sexplib": "*",
```
aantron: We had a problem with sexplib on 4.02 previously, see https://github.com/ocaml/odoc/pull/216#issuecomment-430823206. Solving it resulted in the resolution entry, and also dependencies on OASIS and js-build-tools. This `*` constraint looks like it might cause a regression.

package.json
```diff
@@ -25,35 +25,28 @@
...
+    "@opam/bisect_ppx": "^1.3.4",
+    "@opam/bos": "*",
+    "@opam/dune": "^1.3.0",
+    "@opam/tyxml": "^4.3.0",
+    "ocaml": "*"
   },
   "devDependencies": {
+    "@opam/alcotest": "^0.8.3",
+    "@opam/astring": "^0.8.3",
+    "@opam/cmdliner": "^1.0.2",
```
aantron: I believe this can be relaxed down to 1.0.0.

package.json
```diff
@@ -25,35 +25,28 @@
...
+    "@opam/cppo": "^1.6.5",
+    "@opam/fpath": "^0.7.2",
+    "@opam/markup": ">= 0.8.0",
     "@opam/merlin": "^3.0.3",
-    "ocaml": "~4.2.0"
+    "@opam/ocamlfind": "^1.8.0",
+    "@opam/result": "*",
+    "@opam/rresult": "*",
```
aantron: I believe we require rresult 0.3.0 or higher.

package.json
```diff
@@ -25,35 +25,28 @@
         "ocamlfind",
         "sexplib"
       ]
-    },
-    "buildsInSource": "_build"
```
aantron: Could you briefly say why this is not needed?

package.json
```diff
@@ -25,35 +25,28 @@
...
-    "@opam/sexplib": "bryphe/sexplib#548f94d"
+    "@opam/bisect_ppx": "^1.3.4",
+    "@opam/bos": "*",
+    "@opam/dune": "^1.3.0",
+    "@opam/tyxml": "^4.3.0",
+    "ocaml": "*"
   },
   "devDependencies": {
+    "@opam/alcotest": "^0.8.3",
+    "@opam/astring": "^0.8.3",
```
aantron: I think we can use `*` here.

package.json
```diff
@@ -25,35 +25,28 @@
...
+    "@opam/cmdliner": "^1.0.2",
+    "@opam/cppo": "^1.6.5",
```
aantron: `*`

package.json
```diff
@@ -25,35 +25,28 @@
...
+    "@opam/fpath": "^0.7.2",
```
aantron: `*`

package.json
```diff
@@ -25,35 +25,28 @@
...
+    "@opam/markup": ">= 0.8.0",
     "@opam/merlin": "^3.0.3",
-    "ocaml": "~4.2.0"
+    "@opam/ocamlfind": "^1.8.0",
```
aantron: `*`

package.json
```diff
@@ -25,35 +25,28 @@
...
-    "@opam/tyxml": "4.3.0",
-    "ocaml": "~4.2.0"
-  },
-  "resolutions": {
-    "@opam/markup": "aantron/markup.ml:markup.opam#9f8e77",
-    "@opam/sexplib": "bryphe/sexplib#548f94d"
+    "@opam/bisect_ppx": "^1.3.4",
+    "@opam/bos": "*",
+    "@opam/dune": "^1.3.0",
+    "@opam/tyxml": "^4.3.0",
```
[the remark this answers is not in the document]
ulrikstrid: My thoughts was that anything that had `{build}` in `odoc.opam` was okay to be in devDependencies which I realize now is wrong, but I moved everything I needed to build this as a dependency in another project to dependencies.  But I guess only those with `dev` in the opam file should be devDependencies.

I'll move them to mirror what's in the opam file.

package.json
```diff
@@ -25,35 +25,28 @@
...
+    "ocaml": "*"
```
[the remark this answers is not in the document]
ulrikstrid: Okay, I think the version should be `^4.2.0` then so it's any "patch" over `4.2.0`.

package.json
ulrikstrid: I tested this on windows and it built, but I built with a later compiler, will test again to see if it has been fixed upstream.

package.json
jordwalke: If you use Dune, then Dune has support for esy's out of source package builds, which means that esy can speed things up during local development. By default `buildsInSource` is assumed to be `true`, and you can set it to `false` or `_build`. If you use `false` it copies inefficiently but works with everything all the time. If you use `_build` it builds into `_build` with a small perf hit. If your build system supports out of source builds, you leave it as the default `true` and get the best perf.

package.json
ulrikstrid: I tested by setting ocaml to `~4.2.0` and sexplib to the same as in the opam file `>= 113.33.00` but then it couldn't find versions that were okay.
The resolution we had before is broken as there is not manifest. When I update to the latest commit we get a dependency in `sexplib0` that needs `>4.4` of the compiler.

Would a good trade-off be that we don't need to support down to `4.2` when using esy?

ostera: @ulrikstrid if we are to build binaries to be used with BuckleScript `.cmt` files, we need to support building on `4.02`

ulrikstrid: @ostera Will these be built with `esy`? And as BuckleScript is actively working in switching to `4.06`, is this a issue?

package.json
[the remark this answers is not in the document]
ulrikstrid: It's because of what I wrote in the other discussion, sexplib (even in the resolution) is depending on sexplib0 that in turn needs >= 4.4.0. If someone can find a set of resolutions to make this work I would love it but I can't get it to work without forking a bunch of libraries.

package.json
ostera: If BuckleScript actually moves to `4.06` then sure, but until that happens (and for everyone that's isn't running the latest release when that happens), we should support `4.02`.

Ideally the odoc binaries would be built with `esy` too, yes 👍

package.json
```diff
@@ -25,35 +25,28 @@
...
-    "@opam/oasis": "0.4.11",
-    "@opam/ocamlfind": "1.8.0",
-    "@opam/result": "1.3",
-    "@opam/sexplib": "*",
-    "@opam/tyxml": "4.3.0",
-    "ocaml": "~4.2.0"
-  },
-  "resolutions": {
-    "@opam/markup": "aantron/markup.ml:markup.opam#9f8e77",
-    "@opam/sexplib": "bryphe/sexplib#548f94d"
```
ostera: You'll need this resolution to get a `sexplib` version that builds with `4.02` afaik

ulrikstrid: There are 2 issues here, first is that it doesn't point to the opam file so it can't be resolved with esy. The second issue is that when you point to the opam file it doesn't contain metadata so it fails anyway.

package.json
ulrikstrid: I might have come at this from the wrong angle. I used this package.json when linking from another project to test out a PR I was working on. But you probably want this package.json to just build binaries and don't care about dependants as they should depend on the opam file right?

ostera:
@ulrikstrid if the goal was to build with `dune` directly, do we need to touch the dependency pinnings?

ulrikstrid:
I think we can close this PR, I was changing the package.json to be able to link to it locally when trying out another PR but this is supposed to be used to build `odoc` itself.
I still think this change is good: `"build": "dune build",` but can easily be added outside of this.

PR #291, opened 2019-02-05
Update CONTRIBUTING file with tidy installation instructions

ryyppy:
This fixes an issue I had when I ran `make test` with my stock `tidy` binary shipped with MacOS.
Installing `tidy-html5` via `brew` made the tests green :-)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,8 +49,8 @@ To make edits to the HTML generation, run the following commands:
     opam install --deps-only odoc
     ```
 
-    Also make sure to install a recent version of `tidy-html5` (required for
-    pretty-printing html):
+    Also make sure to install a recent version of
+    [tidy](http://www.html-tidy.org/) (used for HTML validity testing):
 
     ```
     # On MacOS (should be version 5.6.0 by the date of this writing)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -49,6 +49,17 @@ To make edits to the HTML generation, run the following commands:
     opam install --deps-only odoc
     ```
 
+    Also make sure to install a recent version of `tidy-html5` (required for
```
ostera: Could we add a link to the tidy website? (http://www.html-tidy.org/)

aantron: Also note that tidy is not required for the test suite. It's optional, the test suite detects it and uses it if installed.

It is used for HTML validity testing, not for pretty-printing.

ryyppy: @aantron So it seems that for MacOS, `tidy` is installed by default, which caused my tests to fail because of a missing CLI flag. Don't you think it should be installed, no matter what?

aantron:
Go ahead, @ostera. If someone doesn't read the instructions and doesn't have tidy, or has a recent version, they won't have a problem. If they do have tidy, have a problem, and go to read the instructions, the instructions will help them to fix it.

CONTRIBUTING.md
aantron: We can certainly put that in the instructions. Thanks for doing it :)

Issue #294, opened 2019-02-11
odoc pages have no TOC


dbuenzli:
It would be nice to have some following the same rules as the `.mli` file. 

Btw. I'm not very fond of calling the preamble to the table of contents `Topics` in the rendering, that's not a topic index, it's really a table of contents. I think `Contents` would be a better name (that's what wikipedia uses and what documentation preparation systems like `LaTeX` generate for you).

rizo:
@dbuenzli I'm currently working on this. I was waiting for https://github.com/ocaml/odoc/pull/261 to be merged (looks like you already did merge it, thanks) before submitting a PR.

Regarding "Topics": I thought about it and the reasoning is that I wanted to omit paragraph and subparagraph headings (like "Examples" or "Notes") in the listing, in which case it wouldn't be a _complete_ "table of contents". Normally the different section headings represent library topics, in my opinion. On the other hand, "Contents" is a more general term that would imply that the module symbols would also be included, and that's not the case.

I don't feel strongly about this and would be happy to rename "Topics" to "Contents".

dbuenzli:
> in which case it wouldn't be a _complete_ "table of contents".

I don't see the problem with this, when you make table of contents in documents and books you don't necessarily include all sub-headings levels . 

> Normally the different section headings represent library topics, in my opinion.

Frankly it feels a bit a weird I dont' really understand what a "library topic" is. It's even weirder in pages where you could have an `Introduction` and a `Conclusion`.

> On the other hand, "Contents" is a more general term that would imply that the module symbols would also be included, and that's not the case.

Not really, I would expect this to be found under *indexes*. I think `Contents` is a better term here that follows established document production.

rizo:
"Contents" it is then :)

> Not really, I would expect this to be found under _indexes_.

Just to clarify: do you mean the global index as the one generated by ocamldoc ([example](https://caml.inria.fr/pub/docs/manual-ocaml/libref/index.html))?

dbuenzli:
> Just to clarify: do you mean the global index as the one generated by ocamldoc ([example](https://caml.inria.fr/pub/docs/manual-ocaml/libref/index.html))?

Yes these indexes. Issue https://github.com/ocaml/odoc/issues/52 tracks this.

trefis:
@rizo any news? Do you still have your code lying around somewhere? How close is it from being finished? :)

rizo:
@trefis See linked PR :)

[...]

PR #295, opened 2019-02-11
Rename Topics to Contents.

rizo:
See https://github.com/ocaml/odoc/issues/294.


rizo:
> I didn't realize that was done in CSS.

It's a common technique but I think we should generate an actual heading at some point.

Issue #297, opened 2019-02-11
The `{!modules}` directive should splice in the module synopsis


dbuenzli:
These [kind](https://b0-system.github.io/odig/doc/containers/index.html) of lists are a bit bland and not very user friendly. Having the module synopsis next to the name would be nice.

Also it would be nice to classify the generated `ul` a bit along the way so that it can be easily styled.

rizo:
> it would be nice to classify the generated `ul`

What would be an appropriate class for the `ul`? Would `class="modules"` be ok?

dbuenzli:
That's what they used to be yes.

rizo:
> Having the module synopsis next to the name would be nice.

Do you have a preference on the markup for this? I looked at the output from ocamldoc and it seems to generate a table.

dbuenzli:
Tables are nice to align things if you know how long these things will be. Since in general we don't know how long module identifiers are going to be I'd rather not do this other wise you may get
```
M                      M's synopsis
M_with_a_very_long_id  That synopsis
```
which is difficult to scan. 

A good workaround which I once found in a typographic book is to align as follows:
```
                    M  M's synopsis
M_with_a_very_long_id That synopsis
```
but it doesn't look good or natural in all context. Personally I'm fine with:
```
M M's synopsis
M_with_a_very_long_id  That synopsis
```
so there's no need for tables. However it's always good to properly classify with `span`s. `span`ing the synopsis with a `synopsis` class won't hurt.

dbuenzli:
Also adding an anchor would be nice.

greedy:
Wouldn't definition lists like
```
<dl>
  <dt>M</dt>
  <dd>M's synopsis</dd>
  ...
</dl>
```
be appropriate for this?

gasche:
ocamldoc generates a list of each module with its synopsis (formatted as a table). As mentioned in the Discuss thread [How to get pleasant documentation for a library using Dune](https://discuss.ocaml.org/t/how-to-get-pleasant-documentation-for-a-library-using-dune/7121), this feature `{modules: ...}` is very useful for front pages... with the ocamldoc rendering. Could we get the module synopsis(es?) with odoc as well?

Julow:
I've implemented this using description lists: https://github.com/ocaml/odoc/pull/597

PR #299, opened 2019-02-11
Add `modules` class to the ul generated by {!modules: ...} directive.

rizo:
See https://github.com/ocaml/odoc/issues/297.

### DIFF
--- a/test/html/expect/test_package+ml/Markup/index.html
+++ b/test/html/expect/test_package+ml/Markup/index.html
@@ -70,6 +70,9 @@
        <a href="#raw-html">Raw HTML</a>
       </li>
       <li>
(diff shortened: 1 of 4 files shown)


test/html/expect/test_package+ml/Markup/index.html
```diff
@@ -298,6 +301,30 @@ <h2 id="raw-html">
      </blockquote>
     </header>
    </section>
+   <section>
```
dbuenzli: These `section` elements look dubious to me but they are not related to the PR.

rizo: Yeah, both `section` and `header` don't seem necessary. Each (section) heading starts a section, but I don't think that's needed.

PR #301, opened 2019-02-11
Fix travis builds using esy

jonludlam:
- Use newer esy version
- Update syntax for resolution

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
   },
   "resolutions": {
     "@opam/markup": "aantron/markup.ml:markup.opam#9f8e77",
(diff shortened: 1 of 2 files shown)


jonludlam:
I've updated the syntax for the resolution of sexplib, but I don't know why we're using @bryphe's 'windows-4.2.0' branch...? @ostera, any idea?

aantron:
Thanks! I'll move those build rows out of `allowed_failures`.

Regarding sexplib, https://github.com/ocaml/odoc/pull/216#issuecomment-431172087 and https://github.com/ocaml/odoc/pull/222#issuecomment-431179931 might be helpful.

PR #302, opened 2019-02-11
Fix reporting of locations in standalone comments

### LINKED ISSUE #253: Fix location offsets in signature-level (unattached) comments
Same as #227, but the fix for that issue only fixed locations in comments that are attached to signature items. We need to also fix locations for signature-level comments.

jonludlam:
Closes #253


aantron:
Thanks!

Issue #130, opened 2018-04-17
Extract all examples and run them (e.g., in CI)


aantron:
This would ensure that they remain up-to-date, and therefore useful to readers.

https://discuss.ocaml.org/t/1841/42, https://discuss.ocaml.org/t/1841/43. cc @grayswandyr @trefis.

dbuenzli:
I thought there was already an issue and a design for this but I can't find it anyway I'll propose here what I had in mind. It's quite an important issue since for now this is absolutely not DRY, I usually have a copy of the examples in my packages that I compile but as you guess this has all the chances of becoming outdated and is painful to maintain by hand. It will also be very useful for tutorial style `.mld` documents.

The goal of the proposal is to allow:

1. Multiple examples in a single `.mld`/compilation unit. See e.g. the [Cmdliner examples](http://erratique.ch/software/cmdliner/doc/Cmdliner.html#examples)'s examples.
2. Literate programming style that interleaves code and prose. See e.g. [Vg's basics](http://erratique.ch/software/vg/doc/Vg.html#basics).

The idea is that verbatim/code fences should allow the specification of an identifier (don't know exactly how though, people into the parsing of `ocamldoc` comments certainly know better). The extracted file corresponding to an identifier is simply the concatenation of all the fences that bear the identifier .

```ocaml
{id1[let x = ...]}
blabla
{id1[let y = ...]}
{id2[ other example ]}
{id1[ still belongs to id1]}
```
Now odoc is extended as follows: 
```
odoc extract --id-list CMTI # print the ids of CMTI one by line
odoc extract --id ID -o FILE CMTI   # extract the contents for id ID in CMTI to FILE
```

dbuenzli:
Btw. I don't know if this is possible due to the semantics of stop comments but allowing: 
```ocaml
(**/**)
(** {1id[This will be in id1]} *)
(**/**)
```
would be useful aswell.

rizo:
@dbuenzli Can you elaborate on how the code for particular identifiers would be executed after the extraction? I assume you would feed them into the compiler after that somehow, but which tool would do that? Also what is the need to have the identifiers explicitly in the code blocks?

FYI, there's an ongoing attempt to [add support for "literate programming" in dune](https://github.com/ocaml/dune/issues/895) with help from [mdx](https://github.com/samoht/mdx). I am interested in trying to use mdx with `mli` and `mld` files for a potential integration with odoc (See https://github.com/samoht/mdx/issues/17).

[...]

PR #303, opened 2019-02-12
[WIP] Extract code blocks by ID

rizo:
This is a work in progress attempt to implement testable code examples for odoc. Or to be more precise, the extraction of annotated code blocks from `mli` and `mld` files.  See https://github.com/ocaml/odoc/issues/130 for a proposed design and discussion. I might work on tools complementary to odoc to facilitate the execution of the examples and the integration with dune.

### DIFF
--- a/notes/testable_examples.md
+++ b/notes/testable_examples.md
@@ -1,0 +1,208 @@
+# [DRAFT] Testable examples
+
+Library authors are encouraged to include examples and short snippets of code
+in documentation to demonstrate how to effectively use their library. Such code
+snippets are included in docstrings as code blocks and therefor cannot be
+executed and tested in the same way regular source files are. This leads to
+code duplication for library authors who want to make sure their examples can
+be correctly executed, and to out of date examples when they forget to update
+them, as the library’s API changes.
+
+To address this problem odoc implements the ability to extract code blocks from
+documented interfaces and documentation pages (`mli` and `mld` files
+respectively) into source code files. With this build systems can implement
+user-friendly workflows for execution, testing and even promotion of corrected
+examples. In addition, the extracted examples can be installed as documentation
+assets and thus avoid the need to duplicate them as separate files for
+distribution.
+
+## Named code blocks
+
+In the new version of odoc code blocks can be annotated with a file name. This
+file name is used by odoc to group related code blocks for extraction, and also
+to correctly annotate the markup for syntax highlighting.
+
+The following table demonstrates the two variants of code blocks: the
+traditionally supported *anonymous* code blocks and the new *named* code
+blocks.
+
+| **Anonymous code block** | **Named code block**                |
+| ------------------------ | ----------------------------------- |
+| `"{[" <content> "]}"`    | `"{" <filename> "[" <content> "]}"` |
+
+
+### Code extraction
+
+Both named and anonymous code blocks can be extracted by odoc via the
+command-line interface. Code blocks with the same file name in a given
+documentation file will be concatenated and written into a file with that name.
+Optionally a different output file name for a given group can be provided.
+Users are always required to provide an output file name for extraction of
+anonymous code blocks.
+
+To facilitate debugging and allow the tooling to implement expect-style
+promotions, popularized by cram and dune, the extracted examples can be
+optionally annotated with line numbers and the source file name (see [Line
+number directives](https://caml.inria.fr/pub/docs/manual-ocaml/lex.html#sec86)
+in the OCaml manual).
+
+**Note**: unrelated code blocks do not need to have a unique file name, it is
+recommended to group them by using a file name like `examples.ml` or similar.
+
+The described functionality will also be exposed as a library to facilitate
+integration with build systems and test promotion tooling.
+
+### Syntax highlighting
+
+The file names used to annotated code blocks are also used by odoc to decide
+what language should be used for syntax highlighting in the generated HTML. The
+language is decided based on the file name’s extension.
+
+**Warning:** code blocks without a file name will not have syntax highlighting.
+Once this feature is implemented, the currently used automatic language
+inference should be disabled.
+
+
+## Command-line interface
+
+The following simplified manual page defines the command-line interface:
+
+
+    odoc-extract-code(1)              Odoc Manual             odoc-extract-code(1)
+
+
+    NAME
+           odoc-extract-code - Extract code blocks included in documentation files.
+
+    SYNOPSIS
+           odoc extract-code [OPTION]... FILE
+
+    OPTIONS
+           --name=NAME
+               The name of the code block to extract.
+
+           -o PATH, --output=PATH
+               Output file path. If omitted, the provided NAME will be used.
+               Required for extraction of anonymous code blocks.
+
+           --anonymous
+               Extract code blocks without name. Cannot be used with the `--name'
+               option.
+
+           --with-line-numbers
+               Include line number and file name of the extracted code blocks.
+
+           FILE (required)
+               Input cmti, cmt, cmi, mli or mld file.
+
+
+    Odoc 11VERSION11                                          odoc-extract-code(1)
+
+
+## Dune integration
+
+Here is an excerpt from a documented interface file that demonstrates named code blocks.
+
+**Io_utils.mli**
+
+```ocaml
+val read_file : string -> string
+(** [read_file path] is the content of the file located at [path] read into a string.
+
+    {4 Examples}
+    Given a text file with the content:
+
+    {letters.txt[abcdef]}
+
+    The following example will print the number of letters in the file:
+
+    {count_letters.ml[
+    # let letters = read_file "assets/letters.txt";;
+    val letters : string = "abcdef"
+    # String.length letters;;
+    - : int = 6
+    ]} *)
+```
+
+The user wants to test the two code blocks in this example and all the
+anonymous code blocks. To achieve this, the library stanza can be instructed to
+extract and execute the code blocks from the documentation:
+
+
+```dune
+(library
+  (public_name io-utils)
+  (name Io_utils)
+  (libraries base bos)
+  (documentation
+    (extract_code
+      (letters.txt as assets/letters.txt)
+      count_letters.ml
+      (:anonymous as examples.ml))
+    (execute_code examples.ml count_letters.ml)))
+```
+
+Here is a detailed description of these options:
+
+- `(extract_code <filenames>)` where `<filenames>` field follows the [Ordered
+  set language](http://#). This is a set of code block names found in `mli`
+  files of the library that should be extracted into files. Where `:standard`
+  refers to all annotated code blocks found in the library. Optionally the name
+  of the extracted file can be changed by using the following form:
+  `(<code_block> as <filename>)`, for example, `(letters.txt as
+  assets/letters.txt)`. Untitled code blocks can be extracted by providing a
+  file name to a special `:anonymous` name: `(:anonymous as <filename>)`.
+- `(execute_code <filenames>)` where `<filenames>` field follows the [Ordered
+  set language](http://#). This is a set of extracted code files that will be
+  compiled and executed during documentation generation. Currently only the
+  files with the `ml` and `re` extensions are supported.
+
+With these two options it is possible to precisely control what gets extracted
+and what gets executed. Furthermore the extracted files can also be installed
+by dune.
+
+The top-level `documentation` stanza for `mld` files can also be extended to
+support these options.
+
+----------
+
+## Requirements
+
+- Allow the errors to be highlighted in examples in the original file. Might require
+  https://github.com/ocaml/odoc/issues/147
+- Produce `.corrected` files to allow dune (or other build systems) to support
+  promotion of corrected files.
+- The code block name should contain the language information for syntax
+  highlighting.
+
+
+## Questions
+
+- Should odoc require code block annotations to be filenames with extension?
+  The extension could be used to identify the language and correctly do code
+  highlighting. On the other hand the code blocks could be annotated only with
+  the language name (*i.e.* `{ocaml[...]}`), but this would limit the scope of
+  the feature. In particular this would make it impossible to:
+  1. Write examples in code blocks that read input from files extracted from
+     other code blocks;
+  2. Explicitly select the examples that should be compiled (ignoring others);
+  3. Install multiple extracted examples without compiling them.
+- Should “execution” of `mli` files be supported too? Might be useful for basic
+  type-checking of the signature items.
+- Should code blocks with the same name from different `mli` and `mld` files
+  (in the same library) be extracted into the same file? This might be
+  problematic with anonymous code blocks. On the other hand the
+  `--with-line-numbers` can be used to keep track of the name of the original
+  file.
+
+
+## Alternatives
+
+- Examples could be loaded from existing files into odoc's output. This is more
+  limited than the current proposal because it does not allow to interleve
+  comments and code. But, on the other hand, would not need any additional
+  build tooling as the examples can be directly compiled/tested.
+- Introduce something like `mlt` files where code is mixed with comments. These
+  files could be converted into `mld` files for HTML rendering. See
+  https://github.com/janestreet/toplevel_expect_test
+
--- a/notes/testable_examples.md
+++ b/notes/testable_examples.md
@@ -58,9 +58,9 @@ The file names used to annotated code blocks are also used by odoc to decide
 what language should be used for syntax highlighting in the generated HTML. The
 language is decided based on the file name’s extension.
 
-**Warning:** code blocks without a file name will not have syntax highlighting.
-Once this feature is implemented, the currently used automatic language
-inference should be disabled.
+**Note:** code blocks without a file name will be assumed to be in OCaml for
+syntax highlighting purposes. To completely disable syntax highlighting,
+verbatim blocks should be used (`{v ... v}`).
 
 
 ## Command-line interface
--- a/notes/testable_examples.md
+++ b/notes/testable_examples.md
@@ -2,15 +2,15 @@
 
 Library authors are encouraged to include examples and short snippets of code
 in documentation to demonstrate how to effectively use their library. Such code
-snippets are included in docstrings as code blocks and therefor cannot be
+snippets are included in docstrings as code blocks and therefore cannot be
 executed and tested in the same way regular source files are. This leads to
 code duplication for library authors who want to make sure their examples can
-be correctly executed, and to out of date examples when they forget to update
+be correctly executed, and to out-of-date examples when they forget to update
 them, as the library’s API changes.
 
 To address this problem odoc implements the ability to extract code blocks from
 documented interfaces and documentation pages (`mli` and `mld` files
-respectively) into source code files. With this build systems can implement
+respectively) into source code files. With this, build systems can implement
 user-friendly workflows for execution, testing and even promotion of corrected
 examples. In addition, the extracted examples can be installed as documentation
 assets and thus avoid the need to duplicate them as separate files for
@@ -18,7 +18,7 @@ distribution.
 
 ## Named code blocks
 
-In the new version of odoc code blocks can be annotated with a file name. This
+In the new version of odoc, code blocks can be annotated with a file name. This
 file name is used by odoc to group related code blocks for extraction, and also
 to correctly annotate the markup for syntax highlighting.
 
@@ -36,7 +36,7 @@ blocks.
 Both named and anonymous code blocks can be extracted by odoc via the
 command-line interface. Code blocks with the same file name in a given
 documentation file will be concatenated and written into a file with that name.
-Optionally a different output file name for a given group can be provided.
+Optionally, a different output file name for a given group can be provided.
 Users are always required to provide an output file name for extraction of
 anonymous code blocks.
 
@@ -168,6 +168,11 @@ support these options.
 
 ## Requirements
 
+- In the common case users should be able to execute all the code blocks in a
+  single environment. This behaviour should be the default.
+- It should be possible to assign different environments to code blocks. Code
+  blocks with the same environment will be executed in the same toplevel
+  session.
 - Allow the errors to be highlighted in examples in the original file. Might require
   https://github.com/ocaml/odoc/issues/147
 - Produce `.corrected` files to allow dune (or other build systems) to support
--- a/notes/testable_examples.md
+++ b/notes/testable_examples.md
@@ -40,6 +40,17 @@ Optionally, a different output file name for a given group can be provided.
 Users are always required to provide an output file name for extraction of
 anonymous code blocks.
 
+#### Hidden code blocks
+
+In some cases it might be useful to execute code blocks that contain
+configuration or setup logic. If those code blocks are not essential for
+documentation, they can be hidden by placing them between the special stop
+comment (see [The Stop special
+comment](https://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#sec355) in
+the OCaml manual). Note that hidden code blocks will still be executed, to
+prevent the execution of a code block `execute=false` can be added to the
+annotation. ([TODO] Not currently covered by the proposal.)
+
 To facilitate debugging and allow the tooling to implement expect-style
 promotions, popularized by cram and dune, the extracted examples can be
 optionally annotated with line numbers and the source file name (see [Line
@@ -65,7 +76,8 @@ verbatim blocks should be used (`{v ... v}`).
 
 ## Command-line interface
 
-The following simplified manual page defines the command-line interface:
+The following simplified manual page defines the command-line interface for
+code extraction:
 
 
     odoc-extract-code(1)              Odoc Manual             odoc-extract-code(1)
@@ -81,13 +93,20 @@ The following simplified manual page defines the command-line interface:
            --name=NAME
                The name of the code block to extract.
 
-           -o PATH, --output=PATH
-               Output file path. If omitted, the provided NAME will be used.
-               Required for extraction of anonymous code blocks.
-
            --anonymous
                Extract code blocks without name. Cannot be used with the `--name'
                option.
+
+           --all
+               Extract all code blocks from the input file (including anonymous
+               code blocks). The output PATH will be treated as a dirctory when
+               invoking this option.
+
+           -o PATH, --output=PATH
+               Output path. If omitted and the `--name=NAME' option is
+               provided, NAME will be used as an outupt file path. Required for
+               extraction of anonymous code blocks. Must be a directory path
+               when `--all' code blocks are extracted.
 
            --with-line-numbers
                Include line number and file name of the extracted code blocks.
@@ -97,6 +116,10 @@ The following simplified manual page defines the command-line interface:
 
 
     Odoc 11VERSION11                                          odoc-extract-code(1)
+
+Similar to other odoc commands that produce output, a complementary `odoc
+extract-code-targets` command will be added to list all code block names
+present in the input file.
 
 
 ## Dune integration
--- a/test/html/cases/markup.mli
+++ b/test/html/cases/markup.mli
@@ -88,6 +88,14 @@
           ignore foo
     ]}
 
+    Code blocks can have an identifier attached to them:
+
+    {ocaml[
+        # print_int (2 + 2);;
+        4
+        - : unit = ()
+    ]}
+
     There are also verbatim blocks:
 
 {v
--- a/src/html/comment.ml
+++ b/src/html/comment.ml
@@ -251,7 +251,8 @@ let rec nestable_block_element
   fun ?xref_base_uri ~to_syntax ~from_syntax -> function
   | `Paragraph [{value = `Raw_markup (`Html, s); _}] -> Html.Unsafe.data s
   | `Paragraph content -> Html.p (inline_element_list ?xref_base_uri content)
-  | `Code_block s ->
+  | `Code_block (_, s) ->
+    (* TODO(rizo): use code block id as a CSS class. *)
     let open Tree in
     (*
     TODO: This will probably be replaced by a proper plugin / PPX system.
--- a/src/model/comment.ml
+++ b/src/model/comment.ml
@@ -44,7 +44,7 @@ type inline_element = [
 
 type nestable_block_element = [
   | `Paragraph of (inline_element with_location) list
-  | `Code_block of string
+  | `Code_block of string option * string
   | `Verbatim of string
   | `Modules of Reference.module_ list
   | `List of
--- a/src/parser/ast.ml
+++ b/src/parser/ast.ml
@@ -22,7 +22,7 @@ type inline_element = [
 
 type nestable_block_element = [
   | `Paragraph of (inline_element with_location) list
-  | `Code_block of string
+  | `Code_block of string option * string
   | `Verbatim of string
   | `Modules of Reference.module_ list
   | `List of
--- a/src/parser/lexer.mll
+++ b/src/parser/lexer.mll
@@ -201,11 +201,11 @@ let emit_verbatim input start_offset buffer =
   let t = trim_trailing_blank_lines t in
   emit input (`Verbatim t) ~start_offset
 
-let code_block c =
+let code_block ?id c =
   let c = trim_leading_blank_lines c in
   let c = trim_trailing_blank_lines c in
   let c = trim_leading_whitespace c in
-  `Code_block c
+  `Code_block (id, c)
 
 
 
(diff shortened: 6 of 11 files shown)


pmetzger:
There's some facility like this that got built for the Real World OCaml book, isn't there?

lpw25:
So is the idea that `odoc` would extract the code samples and then some other mdx-like tool would call this feature of odoc as part of its processing? That could work, although mostly odoc works on `.cmt` files whilst I would expect that tool to work directly on mli/mld files. Also not all comments have an associated id, so its not clear how you'd want to deal with those.

Maybe you could outline how you see this working to give a clearer picture of how this part will be used?

dbuenzli:
> That could work, although mostly odoc works on `.cmt` files whilst I would expect that tool to work directly on mli/mld files.

`.mli` files are subject to pre-processing. That's the reason why we have `cmti` files no ? 

> Maybe you could outline how you see this working to give a clearer picture of how this part will be used?

I don't know what @rizo had in mind but following the link he provided leads to [this](https://github.com/ocaml/odoc/issues/130#issuecomment-384968939)

lpw25:
> `.mli` files are subject to pre-processing. That's the reason why we have `cmti` files no ?

Sure, and I would expect that a tool to run the examples in them would need to perform such pre-processing. Certainly similar tools like toplevel_expect_test do. Since we already need to know the various compilation options in order to run the examples that does not seem unreasonable.

Also I would expect such a tool to work like other "expect test" style tools, so to produce an updated version of the ml file, so that we can use dune's `promote` workflow -- that clearly requires taking the mli file as input.

> I don't know what @rizo had in mind but following the link he provided leads to [this](https://github.com/ocaml/odoc/issues/130#issuecomment-384968939)

There are a variety of suggestions on the thread @rizo linked so I wasn't sure which suggestion was being referred to. It does look like he has the comment you point to in mind. That comment though does not constitute a full proposal. It is not clear to me how you would use those commands to build a useful workflow.

lpw25:
I should say though, that I think any proposal is going to require the syntax to allow code blocks to be labelled in some way -- so that part seems useful regardless of how we build the surrounding tooling.

dbuenzli:
> That comment though does not constitute a full proposal. It is not clear to me how you would use those commands to build a useful workflow.

I'm not sure what you are refering to. Basically with the proposal you can see `.cmti` or `.mld` file as a tar archive with text files that can be extracted. The rest is yours. 

To give an example say you have this `example.mld` file:
```

{0 Examples}

The hello world program goes like this:

{hello.ml[
let () = print_endline "Hello"
]}

You may not need this data:

{data.json[
  [1; 2]
]}
```

Then `odoc extract --list-id` outputs:
```
hello.ml
data.json
```
Your build system can then pickup these file names to invoke:
```
odoc extract --id hello.ml -o hello.ml example.mld
odoc extract --id data.json -o data.json example.mld
```
At which point you may instruct your build system to do something with the generated `hello.ml` file.

lpw25:
Sure, but how do I map the results of using `hello.ml` back to the example.mld file afterwards? We lose the mapping between the comments and the code so you can't use this to implement an "expect test" style workflow -- which is I think the most desirable workflow for this kind of thing.

dbuenzli:
> Sure, but how do I map the results of using `hello.ml` back to the `example.mld` file afterwards?

One can add an option to `extract` to emit linenum directives in the output.

> We lose the mapping between the comments and the code so you can't use this to implement an "expect test" style workflow -- which is I think the most desirable workflow for this kind of thing.

I don't know exactly what this is but wouldn't the above suffice ?

lpw25:
Perhaps I should clarify what I mean by an "expect test" workflow. I think that one of the most useful aspects of being able to run examples from documentation is to be able to write things like:

```ocaml
{[
    # let foo x y = y;;
    val foo : 'a -> 'b -> 'b = <fun>
    # foo 1 3;;
    val - : int = 3
]}
```
and have a tool which checks that the results of running that code in the toplevel match what I've written there. In an "expect test" style workflow this is achdieved by having the tool print a new version of the `.mli`/`.mld` file with the actual generated output. Dune has special support for such tools -- it will show the diff between the actual and expected output as an error and you can run `dune promote` to copy the actual output over the original file.

lpw25:
As examples this is how the [mdx](https://github.com/realworldocaml/mdx) and [toplevel_expect_tests](https://github.com/janestreet/toplevel_expect_test) tools work.

dbuenzli:
Well I don't expect `odoc` to make the whole promotion business. Aren't `#linenum` directives surrounding each snippet sufficient for the promotion tool to do its business ?

lpw25:
> Well I don't expect `odoc` to make the whole promotion business.

Me neither. I don't think `odoc` needs to provide support for any of this. As long as its comment parser is available as a library all of these things can easily be implemented as separate tools -- including the feature being suggested here.

dbuenzli:
> As long as its comment parser is available as a library all of these things can easily be implemented as separate tools -- including the feature being suggested here.

Maybe but OTOH `odoc` is the tool that processes `.mld` files,  I'm not sure there's any gain to introduce yet another tool, especially if the scheme mentioned above is reasonably usable for most workflows.

lpw25:
I can see how this proposal would be better than what I'm suggesting for literate programming, is that what you have in mind? In which case, would only having the support for doing it for `.mld` files be sufficient? That version makes more sense to me.

My other main concern is the semantics of the "id"s. Most systems that have quoting of code fragments like this (markdown, org-mode, etc.) have support for specifying the language of the code fragment along with other data about how the code should be extracted. Specifying the language seems particularly useful for odoc to allow us to correctly do code highlighting in the resulting documentation. I'm worried that we won't easily be able to do that if we merge this patch as is because we'll already have specified a different semantics for the text that appears between `{` and `[`.

dbuenzli:
> I can see how this proposal would be better than what I'm suggesting for literate programming, is that what you have in mind?

That's one of the worfklow along with `.mld` based tutorials. I'm also interested in extracting my code samples from `.cmti` files and make sure they compile though. If I take these [examples](https://erratique.ch/software/uutf/doc/Uutf#examples) I'm not really interested in `topexpecting` them. Just extracting them with `linenum` directives and make sure they actually compile in my dev builds.

> My other main concern is the semantics of the "id"s. Most systems that have quoting of code fragments like this (markdown, org-mode, etc.) have support for specifying the language of the code fragment along with other data about how the code should be extracted.

Agreed I wouldn't be against more structure here. FWIW in CommonMark, the `id` is an "info string". and is it only *suggested* that the first word should specify the language.  See [here](https://spec.commonmark.org/0.17/#code-fence).

lpw25:
> If I take these [examples](https://erratique.ch/software/uutf/doc/Uutf#examples) I'm not really interested in `topexpecting` them

The expect test style approach wouldn't give you anything extra, but it would give you what you want in these cases. So if we assume that such support is on the way then I think you could support the `extract` command only for `.mld` files and you would have all the things you are after.

dbuenzli:
Not sure I understand the deep objection about making that on `cmti` files aswell :-) Presumably it's the same code path.

lpw25:
I just want to discourage people from trying to use it to implement the execution of code in comments in mli files since I don't think it is the best way to do it. I'd prefer it if odoc only had commands for things that have a clear use case and where the command is the best approach for addressing that use case.

Whereas the mld version has a clear use case in literate programming and, as you said, odoc is the main tool that understands mld files so it is the logical place to put the feature.

rizo:
I have pushed an early attempt to document this feature (see draft version [here](https://github.com/ocaml/odoc/blob/5a001e025b59c0b62ae6c53481dd0da157c3f0ff/notes/testable_examples.md)). It mostly focuses on user experience and integration with dune. I understand this is might be significantly out of odoc's scope but, as I mentioned previously, it is not my intention to exclusively focus on low-level odoc-specific interfaces. It is not clear (to me) how all the discussed details (extraction, execution, line numbers, test promotion, etc) connect together into something users can use. And that is my focus.

Let me know if it is preferred to move this discussion to dune.

CC @rgrinberg: you might have an opinion about this.

---

> So is the idea that `odoc` would extract the code samples and then some other mdx-like tool would call this feature of odoc as part of its processing? That could work, although mostly odoc works on `.cmt` files whilst I would expect that tool to work directly on mli/mld files.

I think both options could be supported. odoc's parser library is already exposed and it can be used to do the extraction.

> Also not all comments have an associated id, so its not clear how you'd want to deal with those.

I covered this in the proposal. Essentially all code blocks without an associated id can be extracted as a common "anonymous" group.

> Maybe you could outline how you see this working to give a clearer picture of how this part will be used?

I do not at this stage have a final plan for the fully integrated solution. One of the reasons I submitted this PR as WIP was to start this discussion. And as you mentioned, annotating code blocks is needed in any case.

I started experimenting with mdx and it seems like it could be used to build a tool (I'm calling it `odoc-mdx`) that would use odoc's parser as a library to extract the code blocks from mli/mld files, execute them and generated `.corrected` files. I'm hoping dune can be extended to natively support this workflow.

---

> My other main concern is the semantics of the "id"s. Most systems that have quoting of code fragments like this (markdown, org-mode, etc.) have support for specifying the language (...)

This is something I mention in my proposal. Essentially using filenames gives us both: the ability to group code blocks and the language information. It does require odoc to interpret code block annotations as file names, but I think it is a reasonable limitation.

>  I'm worried that we won't easily be able to do **that** if we merge this patch as is because we'll already have specified a different semantics for the text that appears between `{` and `[`.

By _that_ do you mean "other data about how the code should be extracted"? For OCaml, at least the execution options could be passed as `#` directives (similar to how toplevel_expect_test works).

> Agreed I wouldn't be against more structure here. FWIW in CommonMark, the `id` is an "info string". and is it only _suggested_ that the first word should specify the language.

Mdx actually relies on this. I think we could adopt this model. How would multiple files be grouped for extraction though?

Maybe something like the following could work? The extraction command would match on the `file` value.

```ocaml
{ocaml file=hello.ml[ ... ]}
```

---

> Not sure I understand the deep reason for objecting doing this on `cmti` files aswell :-)

> I just want to discourage people from trying to use it to implement the execution of code in comments in mli files

I'm not sure I understand what this means. Does this only apply to the CLI? Wouldn't the extraction from `cmti` (or `mli`) files be required to implement the expect-based workflow?

notes/testable_examples.md
```diff
@@ -0,0 +1,208 @@
...
+           --name=NAME
+               The name of the code block to extract.
+
+           -o PATH, --output=PATH
+               Output file path. If omitted, the provided NAME will be used.
+               Required for extraction of anonymous code blocks.
+
+           --anonymous
+               Extract code blocks without name. Cannot be used with the `--name'
+               option.
```
dbuenzli: Why not. Note however that my guess is that unless the author pays particular attention to her anonymous code block the result will likely be garbage most of the time (you may have code showing errors, or even get different languages...)

notes/testable_examples.md
```diff
@@ -0,0 +1,208 @@
...
+
+           --with-line-numbers
+               Include line number and file name of the extracted code blocks.
+
+           FILE (required)
+               Input cmti, cmt, cmi, mli or mld file.
+
+
+    Odoc 11VERSION11                                          odoc-extract-code(1)
```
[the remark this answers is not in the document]
rizo: This was in my plans but I forgot to include it. Thanks.

jsomers:
Thanks @rizo for putting together this design. It jibes with what I've been thinking about, though I think there is one key difference.

> Should odoc require code block annotations to be filenames with extension?

I don't think so. For an mli file, the workflow I imagined was:

1. In a doc comment that includes a code example you'd like to execute, you add a small annotation, either `{ocaml|` or `{x|` (for "execute" -- we can assume the block is OCaml; perhaps you could have another token to specify the language).

2. In your dune file in the mli's directory, you specify that you want to execute marked code blocks.

3. If your code blocks contain toplevel syntax, like

```ocaml
# 1 + 1
- : int = 2
```
then you go through the diff-and-promote workflow. (That is, if the output is incorrect, you generate a .corrected file and allow the user to easily diff against and accept this correction.) There shouldn't be any need to add line number directives -- the tool should figure out where in the mli file this code lives, and should be able to patch in the correction appropriately. (Toplevel_expect_test and mdx both do this.)

Otherwise, if you're not dealing with toplevel-style code, you simply throw a compiler error if the code block doesn't compile.

---

Note that this workflow is almost exactly the same as processing a markdown file with mdx, except that instead of executing the code in triple-backtick blocks, you're executing the code in `{ocaml|` or `{x|` blocks.

For mld files, the flow would be basically identical to mdx, except that instead of the base markup language being Markdown, it would be the markup language used by odoc.

By doing things this way, without filenames, you do give up the ability to write code in one block that references input from files extracted from other code blocks -- but I don't think that's nearly as important as being able to simply run small self-contained examples that use the library you're documenting.

As the user of the tool, I don't want to have to think about the code in my doc comments being extracted out into a separate file where it's then executed; I just want the code executed. If the tool needs to extract my code to separate files under the hood, that's fine, but I don't think that detail should be exposed to the user. The user (at least this user) just wants to say "execute this block" on some blocks.

dbuenzli:
@jsomers I don't think what is described here prevents from implementing a system that does exactly what you want. What is described here is rather the plumbing and a more general mecanism that allows you to implement the system you would like.

> There shouldn't be any need to add line number directives -- the tool should figure out where in the mli file this code lives, and should be able to patch in the correction appropriately.

Extracting with line directives is precisely what will allow the underlying tool to figure out where in the mli/mld file the code you are processing lived and, for example if you compile the extraction to actually report any compilation error in the *original* file and location rather than in the extracted file.

notes/testable_examples.md
dbuenzli: Echoing @jsomers desire for certain fileless senarios (though you could always specificy a filename to use by convention). This could be `--anonymous=TOKEN`, and would extract anonymous code blocks that have the token `TOKEN` (e.g. `ocamlx`). 

It would also prevent the likely garbage problem I mentioned, if completely unconstrained.

jsomers:
> > There shouldn't be any need to add line number directives -- the tool should figure out where in the mli file this code lives, and should be able to patch in the correction appropriately.
> 
> Extracting with line directives is precisely what will allow the underlying tool to figure out where in the mli/mld file the code you are processing lived and, for example if you compile the extraction, to actually report any compilation error in the _original_ file and location rather than in the extracted file.

Ah, I see -- my worry is that we would require the end user to add line-number directives by adding a special notation. But it sounds like you're saying we wouldn't.

dbuenzli:
> Ah, I see -- my worry is that we would require the end user to add line-number directives by adding a special notation. But it sounds like you're saying we wouldn't.

No, that would be absolutely terrible :-) Basically the lineno stuff would work as as follows. Suppose you have your `file.mld`:
```ocaml
Bla bla
{ocamlx[
let f x = x
]}
Ho Ho
{ocamlx[ 
let () = print_endliiine "bla.ml"
]}
```
Invoking `odoc extract-doc --with-line-numbers --anonymous=ocamlx -o ocamlx.ml file.mld` would produce the `ocamlx.ml` file 
```
#3 "file.mld" 
let f x = x
#7 "file.mld"
let () = print_endliiine "bla.ml"
```
Now you can try for yourself to compile that with `ocamlc` that's what gets reported:
```
File "file.mld", line 7, characters 9-24:
Error: Unbound value print_endliiine
Hint: Did you mean print_endline?
```

notes/testable_examples.md
```diff
@@ -0,0 +1,208 @@
...
+to correctly annotate the markup for syntax highlighting.
+
+The following table demonstrates the two variants of code blocks: the
+traditionally supported *anonymous* code blocks and the new *named* code
+blocks.
+
+| **Anonymous code block** | **Named code block**                |
+| ------------------------ | ----------------------------------- |
+| `"{[" <content> "]}"`    | `"{" <filename> "[" <content> "]}"` |
+
```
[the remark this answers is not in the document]
dbuenzli: But… do we get these in `.mld` files ?

lpw25:
Thanks for the design doc @rizo that makes your plan much clearer. I think the aim is a good one but in general I think that the proposed `extract-code` command is not the right approach to implement what you are aiming for. From your comments I think you already know what the disadvantages and alternatives are, but I'll make them explicit here anyway:

The need to talk about external files is an unnecessary burden for users. In the common case they simply want to execute all the comments in a single environment. This behaviour should be the default. If they wish to keep some comments separate there should be an optional argument like the `env` argument in `mdx` to specify named environments with no requirement that they be unique filenames, something like:
```ocaml
{ocaml env=foo[
...
]}
```
If they wish not to run a particular comment there should be an optional argument for that to:
```ocaml
{ocaml executed=false[
...
]}
```

The dune support should simply be something like `(execute-examples)` in the documentation stanza. Having to manage the extraction of multiple files is an unnecessary burden on the user.

The `extract-code` command is not sufficient to implement the "promote" aspect robustly. It provides no mechanism to pass arguments about how the code should be run to the actually testing tool -- for example mdx's `non-deterministic` and `version` options on code snippets would be useful. Line directives are not a robust means of communicating when blocks start and end -- if I put a line directive in my comment fragments the tool will simply break. To do this kind of thing robustly you would need some quoting mechanism and to use a format other than just an `.ml` file.

The `extract-code` command is also not necessary to implement the "promote" aspect. The comment parser is already available and the tool can simply use that on the parsetree of the mli file directly.

> I started experimenting with mdx and it seems like it could be used to build a tool (I'm calling it `odoc-mdx`) that would use odoc's parser as a library to extract the code blocks from mli/mld files, execute them and generate `.corrected` files. I'm hoping dune can be extended to natively support this workflow.

This is I think a better approach than `extract-code` and I think it would be better to focus on this.

> Maybe something like the following could work? The extraction command would match on the `file` value.
>
```ocaml
{ocaml file=hello.ml[ ... ]}
```

I think that would be better for implementing `extract-code`. As I said above, `extract-code` on `mld` files *is* as useful command for literate programming and doing it this way will prevent it from getting in the way of the `mdx`-based alternative.

dbuenzli:
> I think that would be better for implementing `extract-code`. As I said above, `extract-code` on `mld` files _is_ as useful command for literate programming and doing it this way will prevent it from getting in the way of the `mdx`-based alternative.

I'm not exactly sure how something generic like [this](https://github.com/ocaml/odoc/pull/303#discussion_r256563229) gets in the way of `mdx`-based alternatives. Let's seperate concerns here: 

1. Basically the `{[]}` has to be extended in a useful way. So let's agree on the syntax. What's your proposal ? Simply a list of uninterpreted tokens ? I'm fine with this aswell though we should be careful that such a token allows to express a file path.

2. Regarding `extract-code` if you think it's not useful for you and you can do better using an external tool then by all means go build that tool. Personally I have quite a few uses for it and I would like to see it implemented in some way as proposed here.

notes/testable_examples.md
```diff
@@ -0,0 +1,208 @@
...
+recommended to group them by using a file name like `examples.ml` or similar.
+
+The described functionality will also be exposed as a library to facilitate
+integration with build systems and test promotion tooling.
+
+### Syntax highlighting
+
+The file names used to annotated code blocks are also used by odoc to decide
+what language should be used for syntax highlighting in the generated HTML. The
+language is decided based on the file name’s extension.
```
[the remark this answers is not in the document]
rizo: Clashes aside, I think a more structured annotation is needed indeed.

notes/testable_examples.md
```diff
@@ -0,0 +1,208 @@
...
+
+**Warning:** code blocks without a file name will not have syntax highlighting.
+Once this feature is implemented, the currently used automatic language
+inference should be disabled.
```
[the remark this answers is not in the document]
rizo: I generally dislike implicit promotions, but let's respect the history in this case. I updated the proposal to match your suggestion.

lpw25:
> I'm not exactly sure how something generic like [this](https://github.com/ocaml/odoc/pull/303#discussion_r256563229) gets in the way of `mdx`-based alternatives

Something like that seems fine. I was commenting on the proposal as written in the design doc.

> 1. Basically the `{[]}` has to be extended in a useful way. So let's agree on the syntax. What's your proposal ? Simply a list of uninterpreted tokens ? I'm fine with this aswell though we should be careful that such a token allows to express a file path.

I am suggesting an uninterpreted list of tokens, but I'm open to other suggestions. Anything that gives enough flexibility to put fairly arbitrary flags and options is fine.

> 2\. Regarding `extract-code` if you think it's not useful for you and you can do better using an external tool then by all means go build that tool. Personally I have quite a few uses for it and I would like to see it implemented in some way as proposed here.

My comments are in regard to the design document @rizo wrote. I think that `extract-doc` is the wrong way to implement what that proposal attempts. As I've said, I think `extract-doc` on `mld` files makes sense for literate programming, but I've not heard any use cases for `extract-doc` on `cmti` files. If you have a use case or motivation for it then I would like to know what it is. Without a specific use case in mind how are we supposed to judge whether it is the right feature or not?

dbuenzli:
> As I've said, I think `extract-doc` on `mld` files makes sense for literate programming, but I've not heard any use cases for `extract-doc` on `cmti` files.

[These samples](https://erratique.ch/software/cmdliner/doc/Cmdliner.html#examples) are full executables. They are cut and pasted from the mli to [here](https://github.com/dbuenzli/cmdliner/tree/master/test). I want to extract them once with lineno directives in my dev builds to make sure they compile, and once without lineno directives to install them in the `doc` directory of the package as sample code.

notes/testable_examples.md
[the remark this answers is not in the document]
rizo: Corrected the proposal, let me know if this is what you had in mind.

> But… do we get these in `.mld` files ?

Maybe the extended code block annotation could have an option for that...

notes/testable_examples.md
```diff
@@ -0,0 +1,208 @@
...
+  promotion of corrected files.
+- The code block name should contain the language information for syntax
+  highlighting.
+
+
+## Questions
+
+- Should odoc require code block annotations to be filenames with extension?
+  The extension could be used to identify the language and correctly do code
+  highlighting. On the other hand the code blocks could be annotated only with
```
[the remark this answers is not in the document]
rizo: I think this is a reasonable proposal. I also think this should be the focus for the actual work in the current PR. I'll elaborate on this a bit later.

notes/testable_examples.md
rizo: How is it _anonymous_ if it has a name (`ocamlx` in your example)?

But I do understand your concerns. I will try to address this in the new proposal for code block annotations.

rizo:
> As the user of the tool, I don't want to have to think about the code in my doc comments being extracted out into a separate file where it's then executed; I just want the code executed.

@jsomers Users should not think about any kind of extraction, unless they really need this feature (see the examples @dbuenzli referenced). I will clarify this in the proposal. I agree overall that the default behavior should be execute all `ocaml` code blocks without any extra configuration.

lpw25:
> [These samples](https://erratique.ch/software/cmdliner/doc/Cmdliner.html#examples) are full executables.

Thanks that makes things clearer. So you have some literate programming within the `mli` files of a library. That's a reasonable use case. My preference would be to have `extract-code` take the `.mli` file directly as input, since this should really be an operation on source files rather than installed files. It shouldn't be much harder to implement than the `cmti` version -- note that `dune` already takes care of preprocessing the `mli` file to a (marshalled) `.mli` file so you don't need to deal with `-pp` -- but if that is not the case then the `.cmti` file based version is a reasonable alternative.

notes/testable_examples.md
```diff
@@ -0,0 +1,208 @@
...
+Users are always required to provide an output file name for extraction of
+anonymous code blocks.
+
+To facilitate debugging and allow the tooling to implement expect-style
+promotions, popularized by cram and dune, the extracted examples can be
+optionally annotated with line numbers and the source file name (see [Line
+number directives](https://caml.inria.fr/pub/docs/manual-ocaml/lex.html#sec86)
+in the OCaml manual).
+
+**Note**: unrelated code blocks do not need to have a unique file name, it is
```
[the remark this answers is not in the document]
rizo: > However, why would it be good to put unrelated examples into one file?

If an interface has hundreds of functions each one of them with examples, will users want to provide a unique name to each code block to get code execution and syntax highlighting?

It is clear that filenames are not a satisfactory way to annotate code blocks, so this needs to be changed in the proposal anyway.

dbuenzli:
> I agree overall that the default behavior should be execute all `ocaml` code blocks without any extra configuration.

I don't think this concerns this actual proposal but I would rather not do this and rely on a special token for that (e.g. `ocamlx`). 

You still need a bit of ordering care when you start executing code blocks. When I don't want to care about this I prefer not having to care... by which I mean that having to specify an `exec=false` would mean I would have to care. For example I want to be able to simply add an `ocaml` code block anywhere in the document without having to think about it or have to define all its free variables so that it can get executed.

For all it's awesomeness and shininess I still think that executing code blocks in API docs remains quite a limited tool as soon as you escape the realm of documenting basic data structures.

rizo:
> The need to talk about external files is an unnecessary burden for users. In
> the common case they simply want to execute all the comments in a single
> environment. This behaviour should be the default. If they wish to keep some
> comments separate there should be an optional argument like the `env` (...)

I added this to the requirements in the proposal. To be clear the default
behavior in my current proposal is to execute all code blocks that have a file
name with the `ml` extension. Users don't have to think about extraction at all
(the `extract_code` option in dune is optional, and so is `execute_code`).

I do agree that annotating code blocks with the language and allowing arguments
like `env` is a better approach. I don't see why different environments
couldn't be extracted into files though, if users really want that. The two
uses cases being: (1) using other code blocks as data files for execution and
(2) installing examples for package distribution. Use case 1 would be done by
the mdx-style tool (_i.e._ it would extract the data files before execution),
and use case 2 would be done by dune (or other build systems). I think this
applies to both `mld` and `mli` files.

Non of this has to be implemented for the initial version, of course. We can start with a
simple mdx-style tool that executes `ocaml` code blocks (potentially with the `x` annotation) and
ignores all others. At this stage extraction for `mld` files can be implemented
to support self-contained literate programming documents described by @dbuenzli. Finally, the
extraction could be implemented for `mli` files, along with dune support.

I'll try to clarify this in my updated proposal.

rizo:
@lpw25:
> The `extract-code` command is also not necessary to implement the "promote" aspect.

I agree. As long as the code blocks are properly annotated with language and
execution options, the complementary tooling can implement the promotion
workflow without the `extract-code` command.

notes/testable_examples.md
rizo: Still regarding anonymous blocks:

> The only thing I'm a little bit sceptical about is the extraction of
> anonymous blocks, it will likely often be made of garbage so I wouldn't
> bother.

It was one of the @lpw25's concerns, I believe. I imagine for large code bases
it would be useful to operate on existing unmarked code blocks in some way. So
if people do want to execute them they might get an opportunity to fix the
"garbage".

dbuenzli: > How is it _anonymous_ if it has a name (`ocamlx` in your example)?

In the sense it has no file*name*.

rizo:
> I don't think this concerns this actual proposal but I would rather not do this and rely on a special token for that (e.g. `ocamlx`).

I personally don't like the `ocamlx` as an annotation. If we want to annotated code blocks with language identifiers they should be known language identifiers. We would need to do the same for reason syntax, by the way and having `reasonx` or `rex` doesn't look good to me.

But yeah, execution everything unconditionally is also quite extreme. I'll have this in mind and will make a proposal for code block annotations that allows this kind of flexibility.

lpw25:
> You still need a bit of ordering care when you start executing code blocks. When I don't want to care about this I prefer not having to care... by which I mean that having to specify an `exec=false` would mean I would have to care.

I suspect that the best thing to do here is to let the default be specified. I would expect both `extract-code` and the `mdx`-like tooling to take an option that said whether to assume `exec=false` or `exec=true` as the default.

notes/testable_examples.md
dbuenzli: Basically in the [two forms](https://github.com/ocaml/odoc/pull/303#discussion_r256563229) here. The first one is a named code block and the second one is an anonymous one. 

I quite like the idea of being able to extract by token, this leaves a lot of flexibity to authors to attribute meaning to what they want to extract.

rizo:
> I suspect that the best thing to do here is to let the default be specified.

Then it's not really a _default_, in that case ;-).

But I agree that this could be configured on a level higher than a code block. The question is more about what would be the default in case users don't provide this option.

dbuenzli:
> I suspect that the best thing to do here is to let the default be specified. 

I suspect this is really not a good idea. It's always better to keep documents self-describing to the readers. If you do this then I must hunt in the build system what default you choose and then remember it in my head when I read a code block.

I would really go for not executable by default and maybe simply `{$LANG exec[ ... ]}` for those blocks that you want to execute. That works with any language, is clear to the reader and not hard to type.

notes/testable_examples.md
rizo: I also like the idea of token annotations. Without wanting to commit to a concrete syntax, something like `{ocaml id=BLOCK_ID[]}` could work, where `BLOCK_ID` could be anything including a file name. This of course would be used during extraction. I'd make the `--output` option required though, to clarify that the `id` is not necessarily a file name.

lpw25:
There are going to be plenty of people who want to test all their code in comments with "odocx" and they're going to be pretty annoyed with having to write `ocaml exec=true` hundreds of times.

dbuenzli:
> `ocaml exec=true` hundreds of times.

That's the reason why I proposed `ocaml exec` ;-)

dbuenzli:
(even `{exec[]}` since code blocks default to OCaml)

notes/testable_examples.md
dbuenzli: I still think it's preferable to have to direct support for filenames.

XVilka:
There are conflicts now, so should be rebased.

test/html/cases/markup.mli
```diff
@@ -88,6 +88,14 @@
           ignore foo
     ]}
 
+    Code blocks can have an identifier attached to them:
+
+    {ocaml[
```
thomassa: Misleading example? Using `ocaml` as the identifier makes it looks like an annotation specifying the language, rather than an identifier to enable references to this specific block from elsewhere.
```suggestion
    {print-two-plus-two[
```

rizo: The current definition of annotations needs to be changed in my opinion as it has some limitations. Unfortunately I haven't been available to work on this in the last few months.

To clarify the misunderstanding here: the code block annotation is _not_ used to reference a specific block from elsewhere (it is not a unique identifier). It annotates a class of blocks to be extracted. It could be a language or a filename, or any other user-defined "class".

Not sure if this is helpful, but let me know if you have any other questions or suggestions.

dbuenzli:
Could we move on this ?

NoahTheDuke:
As discussed in https://github.com/ocaml/ocaml/pull/13792, I am interested in continuing this or similar work.

NoahTheDuke:
Reading the proposal in this PR and reading the lexer code that exists now, I believe a lot of this work already exists. As seen in https://github.com/ocaml-doc/odoc-parser/pull/2 and https://github.com/ocaml-doc/odoc-parser/pull/17, odoc now supports metadata in code blocks and delimited code blocks, along with associated output:

```ocaml
{@ocaml env=f1 version>=4.06 [code goes here]}

{delim@ocaml[foo]delim[output {b foo}]}
```

These features eliminate the need for named code blocks, as the name can be passed through metadata.

panglesd:
Hello @NoahTheDuke !

Thanks a lot for offering to work on this.

The "syntax" part of this PR is done, as you saw. Also, support for the odoc syntax by mdx has improved a lot! I think two independent things remain to be done:

-  Add some "layout" information to the internal representation of code blocks: odoc does some processing with the whitespace. For instance:
   ```
     {[
           aaa
             bb
     ]}
   ```
   is turned into the string `aaa\n  bb`: some whitespace was stripped before, after, but also "on the left"! There is currently no way to know how much with the location remembered in code blocks, currently.

   This is needed to have locations working when extracting with line directives. But it will also be very useful for mdx, since this missing information makes it impossible for mdx to not be sensitive to layout, and it starts an endless fight with ocamlformat.

- Implement the `extract` command. This requires deciding a few things (do we use `file=...` or `id=...`) which has a chance of getting stuck on discussion during 6 years! :upside_down_face: 

To be honest, I'm not sure it's the best first issue on odoc! Unless you want to work on that, I'll do it so that it can be included in the odoc 3 release.

NoahTheDuke:
hah I love a challenge, even if I fail. What better way to learn?

I think adding location data to code blocks is doable. I know where it happens and why (means you don't have to dedent code in code blocks), and improving the other tooling will be a nice benefit. Does ocamlformat change the layout of code in code blocks in comments? I didn't know that, but I guess it makes sense.

Implementing the `extract` command will be tougher, and I agree will open many bikeshedding discussions lol. Given that this sits somewhat outside of ocaml proper (compared to Rust's [rustdoc](https://rustc-dev-guide.rust-lang.org/rustdoc.html)), I wonder if it's even worthwhile to build such a tool. Maybe a driver would be a better fit? Given that this part is a significant effort and I'm just beginning, I'll not worry too much about it for now.

Let's see how far I get with the location data first.

rizo:
Thanks for taking over and completing this, @panglesd.

PR #305, opened 2019-02-15
Remove dependency on bos.

dbuenzli:
Tools like odoc at the roots of the eco-system should be lean on deps. This removes one of them.

In this case it seems only `Bos.File.read` was being used. I added something that is functionally equivalent.

### DIFF
--- a/src/odoc/fs.ml
+++ b/src/odoc/fs.ml
@@ -30,16 +30,52 @@ module File = struct
 
   let create ~directory ~name =
     match Fpath.of_string name with
-    | Error (`Msg e) -> invalid_arg ("Odoc.Fs.File.create: " ^ e)
-    | Ok psuf -> Fpath.(normalize @@ directory // psuf)
+    | Result.Error (`Msg e) -> invalid_arg ("Odoc.Fs.File.create: " ^ e)
+    | Result.Ok psuf -> Fpath.(normalize @@ directory // psuf)
 
   let to_string = Fpath.to_string
   let of_string s =
     match Fpath.of_string s with
-    | Error (`Msg e) -> invalid_arg ("Odoc.Fs.File.of_string: " ^ e)
-    | Ok p -> p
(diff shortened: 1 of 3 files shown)


aantron:
It looks like the replacement code should use `Result.Ok` instead of `Ok`, for 4.02 compatibility.

src/odoc/fs.ml
```diff
@@ -39,7 +39,43 @@ module File = struct
...
+      let buf = Bytes.create len in
+      really_input ic buf 0 len;
+      close_in ic;
+      Ok (Bytes.unsafe_to_string buf)
+    in
+    let input_stream ic =
+      let bsize = 65536 (* IO_BUFFER_SIZE *) in
+      let buf = Buffer.create bsize in
+      let b = Bytes.create bsize in
+      let rec loop () =
```
aantron: Would [`Buffer.add_channel`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Buffer.html#VALadd_channel) work here? Not that we necessarily need this to be more copy-efficient, but nonetheless. This may also do away with the need for the distinction between `input_one_shot` and `input_stream`, assuming some exception is raised if the `Buffer.t` gets too large.

dbuenzli: > Would [`Buffer.add_channel`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Buffer.html#VALadd_channel) work here?

Yes. 

> This may also do away with the need for the distinction between `input_one_shot` and `input_stream`

Kept one shot. This is really the most efficient you can get to allocation wise: the exact result.

dbuenzli:
> It looks like the replacement code should use `Result.Ok` instead of `Ok`, for 4.02 compatibility.

A bit odd there were quite a few other `{Error,Ok}` lying around that weren't prefixed. I don't think `bos` did anything special for that. The patch now prefix them aswell.

aantron:
Agreed, haven't looked into what's happening there.

Thanks!

PR #306, opened 2019-02-15
Remove dependency on rresult.

dbuenzli:
Same rationale as https://github.com/ocaml/odoc/pull/305. 

Here only the bind operator (`>>=`) was used and only at three specific points. I just inlined the bind definition.


aantron:
Thank you.

Issue #300, opened 2019-02-11
odoc compile misleading and incorrect warning


dbuenzli:
In presence of library variants `odoc compile` entices to wrap libraries which is obviously not the solution. Also when it reports the ambiguity it *sometimes* reports the same path twice. Here's an example output:
```
[spawn:21151] ['/Users/dbuenzli/.opam/4.03.0/bin/odoc' 'compile' '--pkg' 'ocaml' '-o' '/tmp/odig-cache/odoc/ocaml/page-index.odoc' '/tmp/odig-cache/odoc/ocaml/index.mld' '-I' '/tmp/odig-cache/odoc/ocaml/' '-I' '/tmp/odig-cache/odoc/ocaml/camlp4/' '-I' '/tmp/odig-cache/odoc/ocaml/camlp4/Camlp4Filters/' '-I' '/tmp/odig-cache/odoc/ocaml/camlp4/Camlp4Parsers/' '-I' '/tmp/odig-cache/odoc/ocaml/camlp4/Camlp4Printers/' '-I' '/tmp/odig-cache/odoc/ocaml/camlp4/Camlp4Top/' '-I' '/tmp/odig-cache/odoc/ocaml/compiler-libs/' '-I' '/tmp/odig-cache/odoc/ocaml/ocamldoc/' '-I' '/tmp/odig-cache/odoc/ocaml/ocplib-compat/' '-I' '/tmp/odig-cache/odoc/ocaml/ocplib-config/' '-I' '/tmp/odig-cache/odoc/ocaml/ocplib-debug/' '-I' '/tmp/odig-cache/odoc/ocaml/ocplib-file/' '-I' '/tmp/odig-cache/odoc/ocaml/ocplib-lang/' '-I' '/tmp/odig-cache/odoc/ocaml/ocplib-stdlib/' '-I' '/tmp/odig-cache/odoc/ocaml/ocplib-subcmd/' '-I' '/tmp/odig-cache/odoc/ocaml/ocplib-system/' '-I' '/tmp/odig-cache/odoc/ocaml/ocplib-unix/' '-I' '/tmp/odig-cache/odoc/ocaml/threads/' '-I' '/tmp/odig-cache/odoc/ocaml/vmthreads/']:
 Warning, ambiguous lookup. Please wrap your libraries. Possible files:
  /tmp/odig-cache/odoc/ocaml/topdirs.odoc
  /tmp/odig-cache/odoc/ocaml/topdirs.odoc
[...]

Issue #274, opened 2018-12-19
`odoc html-deps`'s interface


dbuenzli:
As it stands `odoc html-deps` takes a "package directory" as a parameter. Since as far as I know `odoc` files are not namespaced by directories this seems a bit odd. 

Is there any reason why `odoc html-deps` does not act on `odoc` files ?

dbuenzli:
`odoc html-deps` is nowadays called `odoc link-deps` but basically with the same interface does anyone know if this issue still makes sense :–) or is that part of the "old" odoc driving way ?

PR #307, opened 2019-02-15
html-deps: Fix package flatness assumption.

dbuenzli:
Make `html-deps` look recursively in the package directory for odoc files. Before this patch it only looked at the first level of the given package directory which is not what you want in practice. 

A lot of packages have a hierarchy in their install libdir (try e.g. `ls $(opam var odoc:lib)`). When you generate the `odoc` files you copycat that hierarchy. As mentioned `html-deps` would not lookup the hierarchy but only the first level of the hierarchy. For a package like `odoc` itself this would return an empty set of deps, which would then result in failures further down the pipeline.

### DIFF
--- a/src/odoc/fs.ml
+++ b/src/odoc/fs.ml
@@ -89,12 +89,25 @@ module Directory = struct
     | Error (`Msg e) -> invalid_arg ("Odoc.Fs.Directory.of_string: " ^ e)
     | Ok p -> Fpath.to_dir_path p
 
-  let ls t =
-    let elts = Sys.readdir (to_string t) |> Array.to_list in
-    List.fold_left elts ~init:[] ~f:(fun acc elt ->
-      let file = File.create ~directory:t ~name:elt in
-      if Fpath.is_file_path file then file :: acc else acc
-    )
+  let fold_files_rec ?(ext = "") f acc d =
+    let fold_non_dirs ext f acc files =
+      let is_dir d = try Sys.is_directory d with Sys_error _ -> false in
+      let has_ext ext file = Filename.check_suffix file ext in
+      let dirs, files = List.partition ~f:is_dir files in
+      let files = List.find_all ~f:(has_ext ext) files in
+      let f acc fn = f acc (Fpath.v fn) in
+      List.fold_left ~f ~init:acc files, dirs
+    in
+    let rec loop ext f acc = function
+    | (d :: ds) :: up ->
+        let rdir d = try Array.to_list (Sys.readdir d) with Sys_error _ -> [] in
+        let files = List.rev (List.rev_map ~f:(Filename.concat d) (rdir d)) in
+        let acc, dirs = fold_non_dirs ext f acc files in
+        loop ext f acc (dirs :: ds :: up)
+    | [] :: up -> loop ext f acc up
+    | [] -> acc
+    in
+    loop ext f acc ([Fpath.to_string d] :: []);;
 
   module Table = Hashtbl.Make(struct
       type nonrec t = t
--- a/src/odoc/depends.ml
+++ b/src/odoc/depends.ml
@@ -77,8 +77,6 @@ let deps_of_odoc_file ~deps input = match (Root.read input).file with
 
 let for_html_step pkg_dir =
   let deps = Hash_set.create () in
-  List.iter (Fs.Directory.ls pkg_dir) ~f:(fun file ->
-    if Fs.File.has_ext "odoc" file then
-      deps_of_odoc_file ~deps file
-  );
+  let add_deps () file = deps_of_odoc_file ~deps file in
+  Fs.Directory.fold_files_rec ~ext:".odoc" add_deps () pkg_dir;
   Hash_set.elements deps
--- a/src/odoc/fs.mli
+++ b/src/odoc/fs.mli
@@ -39,7 +39,10 @@ module Directory : sig
   val of_string : string -> t
   val to_string : t -> string
 
-  val ls : t -> file list
+  val fold_files_rec : ?ext:string -> ('a -> file -> 'a) -> 'a -> t -> 'a
+  (** [fold_files_rec ~ext f acc d] recursively folds [f] over the files
+      with extension matching [ext] (defaults to [""]) contained in [d]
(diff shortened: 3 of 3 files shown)


src/odoc/fs.ml
```diff
@@ -89,12 +89,25 @@ module Directory = struct
     | Error (`Msg e) -> invalid_arg ("Odoc.Fs.Directory.of_string: " ^ e)
     | Ok p -> Fpath.to_dir_path p
 
-  let ls t =
-    let elts = Sys.readdir (to_string t) |> Array.to_list in
-    List.fold_left elts ~init:[] ~f:(fun acc elt ->
-      let file = File.create ~directory:t ~name:elt in
-      if Fpath.is_file_path file then file :: acc else acc
```
dbuenzli: If anyone wonders w.r.t. what I'm replacing with that line was absurd. `Fpath` works syntactically it nevers queries the OS.

lpw25:
> When you generate the `odoc` files you copycat that hierarchy

Is this right? I would have expected to create a flat structure for the odoc files, the structure has no meaning for them.

dbuenzli:
> Is this right? I would have expected to create a flat structure for the odoc files, the structure has no meaning for them.

The problem is that nothing prevents a package from installing a cmti file with the same name in different parts of their hierarchy.

lpw25:
But IIRC then we can't generate the html for that package anyway, right?

dbuenzli:
> But IIRC then we can't generate the html for that package anyway, right?

You can. See https://github.com/ocaml/odoc/blob/master/src/odoc/env.ml#L20-L35

lpw25:
But don't the names of the html files for the modules collide and override each other?

dbuenzli:
> But don't the names of the html files for the modules collide and override each other?

Maybe I don't know about the details, I'll let people who work on odoc figure it out. 

I'm working this way for now in `odig` because real world packages do this (`ocaml` behind the first one) and I don't want to see it rewriting its own outputs.

In any case I don't think looking things up recursively in the directory is harmful, if you generate things flatly no cost is incurred by this patch, so I'm not sure what your objection is here. Besides you could also imagine a tool that works directly in the `libdir` in place, putting `.odoc` files alongside the other compilation objects in which case you want `odoc html-deps` on the package's libdir to report the appropriate outputs.

(Not that I think that `odoc html-deps` is a bit odd, see #274, but then maybe someone should have a closer look at the `*-{deps,targets}` design and bugs, which I guess might happen once the model rewrite happens).


(To be perfectly honest I think the whole thing

jonludlam:
The point of that bit in Env was to not die horribly, rather than to actively support the idea that having module name clashes within a package is a good idea.

The use in OCaml itself is an interesting one - threads and vmthreads are mostly interchangable, but not quite; so we can't quite use the virtual libraries idea without a bit of rearragement. @dbuenzli - do you happen to know if this is the only clash?

lpw25:
> In any case I don't think looking things up recursively in the directory is harmful, if you generate things flatly no cost is incurred by this patch, so I'm not sure what your objection is here.

No objection, I was just surprised that it was needed.

dbuenzli:
> @dbuenzli - do you happen to know if this is the only clash?

There are others like `ptime.clock` and `mtime.clock` but these are due to the way I implemented virtual libraries in the past (I would still allow cross unit inlining to occur --- the disadvantage being that if you wanted to make a *library* that depended on them, you also had to provide a variant for each of the one provided by the virtual lib you dependend on, which of course may combinatorially explode). 

Regarding `vmthreads` note that they are on their [way out](https://github.com/ocaml/ocaml/pull/2038). 

In other words I'm not sure I arealdy saw such a [second case](https://github.com/ocaml/odoc/issues/300#issuecomment-462492096) but regardless, I'll proceed that way for the time being. It will be  a single line change in `odig` to flatten the result in the result once these things have been thought over.

aantron:
The code looks good, but I defer to @lpw25, @dbuenzli, @jonludlam on whether recursive scan is justified.

dbuenzli:
Merging this. No strong objection it seems. Also useful in build scenarios where your might want to generate your `.odoc` files next to the `cmti` files which may be in sub-dirs.

trefis:
Meh.

Independently of whether people think there should be a flat structure to store odoc files or not, one thing that I don't understand is why you can't just pass the right include dirs to odoc.

Btw:
> The problem is that nothing prevents a package from installing a cmti file with the same name in different parts of their hierarchy.

In a case like this I think you're introducing issue by making the search recursive. And the example of "well ocaml has such a structure" is not convincing, or rather, I see it as a good example that you shouldn't do this! For example, ocaml expects you to pass a `-I +threads` or `-I +vmthreads`.

Does this make sense or did I misunderstand what's going on here?

dbuenzli:
> Independently of whether people think there should be a flat structure to store odoc files or not, one thing that I don't understand is why you can't just pass the right include dirs to odoc.

There is no include option in the `odoc html-deps` command. Moreover includes are usually used to denote where to look for dependencies not to denote where you should lookup the objects for which you want to report the dependencies... 

Again I would have found it much more natural for `html-deps` to apply directly on `odoc` files since it would have been more precise for incremental rebuilds -- when one uses the resolved dependencies of the current `html-deps` with `odoc html` on an `odoc` file you have as dependencies the union of all those `odoc` files of the package rather than just those of the `odoc` file itself which is much less precise. I don't know what was the idea here, see #274 and I'm happy to change things when these things have been properly clarified.

> In a case like this I think you're introducing issue by making the search recursive. 

I don't think it will introduce any issue. It will just mix the dependencies of all the odoc files as `html-deps` does at the moment, see #274 again.

> And the example of "well ocaml has such a structure" is not convincing, or rather, I see it as a good example that you shouldn't do this! For example, ocaml expects you to pass a `-I +threads` or `-I +vmthreads`.

Of course you can only link with only one of these units name. But a package may install more than one unit with the same name; if those have different documentation how do you document them ? One answer I'd be perfectly happy with would be this is illegal; but so far that's has not been the case.

trefis:
OK so I had just forgotten how `html-deps` was working, that's my bad.
Thanks for reminding me!

PR #308, opened 2019-02-18
Remove make from esy build step

ostera:
As @ulrikstrid mentioned before, depending on `make` hinders
the build on Windows.

Also bump version number from 1.2.0 to match current tag 1.3.0


aantron:
Thanks!

PR #309, opened 2019-02-19
OCaml 4.08 compatibility

### LINKED ISSUE #304: Odoc fails to compile with OCaml 4.08
The constructors of `Types.signature_item` have been extended with a `visibility` flag introduced in ocaml/ocaml#2147. On first glance it looks like we should just filter out items that are `Hidden`, but I'd like to double check that.

jonludlam:
Further work needs to be done to support the new typesubst/modsubst constructors.

Fixes #304

### DIFF
--- a/src/loader/cmt.ml
+++ b/src/loader/cmt.ml
@@ -79,8 +79,8 @@ let rec read_pattern env parent doc pat =
     | Tpat_lazy pat ->
         read_pattern env parent doc pat
 #if OCAML_MAJOR = 4 && OCAML_MINOR >= 08
-    | Tpat_exception _ ->
-      []
+    | Tpat_exception pat ->
+        read_pattern env parent doc pat
 #endif
 
 let read_value_binding env parent vb =
--- a/src/loader/cmt.ml
+++ b/src/loader/cmt.ml
@@ -78,6 +78,10 @@ let rec read_pattern env parent doc pat =
         read_pattern env parent doc pat
     | Tpat_lazy pat ->
         read_pattern env parent doc pat
+#if OCAML_MAJOR = 4 && OCAML_MINOR >= 08
+    | Tpat_exception _ ->
+      []
+#endif
 
 let read_value_binding env parent vb =
   let container = (parent : Identifier.Signature.t :> Identifier.LabelParent.t) in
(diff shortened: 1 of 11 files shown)


src/loader/doc_attr.ml
```diff
@@ -46,9 +46,15 @@ let attached parent attrs =
...
     function
+#if OCAML_MAJOR = 4 && OCAML_MINOR >= 08
+    | {Parsetree.attr_name = { Location.txt =
+          ("doc" | "ocaml.doc"); loc = _loc}; attr_payload; _ } :: rest -> begin
+        match load_payload attr_payload with
+#else
     | ({Location.txt =
           ("doc" | "ocaml.doc"); loc = _loc}, payload) :: rest -> begin
         match load_payload payload with
+#endif
```
trefis: You could limit the `#if .. #else .. #endif` to the pattern part if you make sure to use the same variable names on both sides.

src/loader/doc_attr.ml
```diff
@@ -93,9 +99,15 @@ let standalone parent
...
   function
+#if OCAML_MAJOR = 4 && OCAML_MINOR >= 08
+  | { attr_name = { Location.txt =
+        ("text" | "ocaml.text"); loc = _loc}; attr_payload; _ } -> begin
+      match load_payload attr_payload with
+#else
   | ({Location.txt =
         ("text" | "ocaml.text"); loc = _loc}, payload) -> begin
       match load_payload payload with
+#endif
```
trefis: Same as above.

src/loader/cmt.ml
```diff
@@ -78,6 +78,10 @@ let rec read_pattern env parent doc pat =
         read_pattern env parent doc pat
     | Tpat_lazy pat ->
         read_pattern env parent doc pat
+#if OCAML_MAJOR = 4 && OCAML_MINOR >= 08
+    | Tpat_exception _ ->
+      []
```
trefis: That doesn't look right. I think you should recurse in that case.

jonludlam: Yes, I think you're right.

src/loader/doc_attr.ml
jonludlam: True, that'll probably look nicer. Thanks!

aantron:
Thanks @jonludlam and @trefis. I'll merge once the CI is done. Looks stuck for now :/ I'll add a 4.08 build to Travis right after merging (or push into this PR if the current CI build still hasn't started by then).

jonludlam:
The 4.08 travis build might be tricky until `ocaml-migrate-parsetree` is available. We'd have to remove the bisect_ppx dependency.

aantron:
Looks like we also transitively depend on Ocamlbuild, which needs a release (https://github.com/ocaml/ocamlbuild/issues/294).

jonludlam:
True - though there at least pinning master works.

aantron:
@jonludlam or @trefis, would you mind opening a follow-up issue about "proper" support for the new 4.08 features?

avsm:
I've opened up #350 to track the new features in 4.08

PR #312, opened 2019-02-19
Enable bisect on model/xref

jonludlam:


dbuenzli:
Btw. is there a way to disable this on `pin` builds ? It depends on brittle part of the eco-system (ocaml-migrate-parse-tree) which is problematic when new ocaml releases occur.

aantron:
Yes, ideally we would have Bisect enabled in development trees, and disabled and not a dependency for pins and releases. Dune and opam don't currently support this as far as I know, though.

jonludlam:
Incidentally, is there a better way to do cppo+ppx?

dbuenzli:
> Yes, ideally we would have Bisect enabled in development trees, and disabled and not a dependency for pins and releases.

Isn't it possible to move it to a `test` dependency (`opam`) and discover it (`dune`) ? That would solve the problem as far as `opam` is concerned.

aantron:
We would have to generate `dune` files for that. I've preferred to avoid that so far. But this is also the first time OMP has become a problem for odoc.

PR #313, opened 2019-02-20
Fix breadcrumbs on odoc pages (closes #293).

### LINKED ISSUE #293: Breadcrumbs are broken on odoc pages
There are three problems on odoc pages:

1. The direct link to the package page jumps to the toplevel package list page rather than package page. See for example the `odoc` link [here](https://b0-system.github.io/odig/doc/odoc/contributing.html).
2. On non `index.mld` pages, the up link jumps to the toplevel package list rather than package page. See for example the `up` link [here](https://b0-system.github.io/odig/doc/odoc/contributing.html). It should jump to package page.
2. The breadcrumbs for `index.mld` should be specialized. I think best would be to have 
simply `Up - PKG` rather `Up - LINKEDPKG >> index` as it is now (see for example [here](https://b0-system.github.io/odig/doc/odoc/index.html)). That is we should simply have an `Up` link going to the toplevel package list page, followed by the *unlinked* package name (because that's the page we are on).

dbuenzli:


trefis:
Doesn’t that break the « up » link when you’re on the index page of a package?

dbuenzli:
No see [this line](https://github.com/ocaml/odoc/pull/313/files#diff-7fa4444761e99f03844c6f94d4f41316R329).

src/html/tree.ml
```diff
@@ -337,11 +340,16 @@ let page_creator ?kind ?(theme_uri = Relative "./") ~path header_docs content =
           ] @
             (* Create breadcrumbs *)
             let space = Html.txt " " in
-            let init =
-              if !Relative_link.semantic_uris then "" else "index.html"
+            let breadcrumb_spec = match kind with
+            | Some `Page -> (fun n x -> n, dot, x)
```
aantron: I would have expected the second element (here `dot`) to depend on `n`. Is this relying on an assumption about the nesting level of pages?

dbuenzli: Yes. Pages are flat at the moment and can only occur at the root of the package.

dbuenzli:
Note that if you want to see the effect w.r.t. what was written in #293 you can have a look at https://b0-system.github.io/odig/doc/odig/index.html which should have all the cases as corrected by this patch.

aantron:
Ok, thank you!

dbuenzli:
Thanks !

PR #314, opened 2019-02-20
Make bisect_ppx a build-time dependency

ostera:
I was trying to update the version of `odoc` being used in [an effort to build BuckleScript documentation with it](https://github.com/BuckleScript/bucklescript/pull/3379) and I ran into the following issue:

```
λ esy build
info esy build 0.5.6 (using package.json)
info building @opam/odoc@github:ocaml/odoc:odoc.opam#eea456edb503ca259718c745233fa4054ad185c5@d41d8cd9
error: build failed with exit code: 1
  build log:
    # esy-build-package: building: @opam/odoc@github:ocaml/odoc:odoc.opam#eea456edb503ca259718c745233fa4054ad185c5
    # esy-build-package: pwd: /Users/ostera/.esy/3__________________________________________________________________/b/opam__s__odoc-11be51b8
    # esy-build-package: running: 'dune' 'build' '-p' 'odoc' '-j' '4'
    File "test/inactive/core/jbuild", line 1, characters 0-0:
    Warning: jbuild files are deprecated, please convert this file to a dune file instead.
    Note: You can use "dune upgrade" to convert your project to dune.
    File "src/html/dune", line 4, characters 18-28:
    4 |  (preprocess (pps bisect_ppx -conditional))
                          ^^^^^^^^^^
    Error: Library "bisect_ppx" not found.
    Hint: try: dune external-lib-deps --missing -j 4 -p odoc @@default
    File "src/model/dune", line 14, characters 18-28:
    14 |  (preprocess (pps bisect_ppx -conditional))
                           ^^^^^^^^^^
    Error: Library "bisect_ppx" not found.
    Hint: try: dune external-lib-deps --missing -j 4 -p odoc @@default
    File "src/parser/dune", line 6, characters 18-28:
    6 |  (preprocess (pps bisect_ppx -conditional))
                          ^^^^^^^^^^
    Error: Library "bisect_ppx" not found.
    Hint: try: dune external-lib-deps --missing -j 4 -p odoc @@default
    File "src/xref/dune", line 4, characters 18-28:
    4 |  (preprocess (pps bisect_ppx -conditional))
                          ^^^^^^^^^^
    Error: Library "bisect_ppx" not found.
    Hint: try: dune external-lib-deps --missing -j 4 -p odoc @@default
    error: command failed: 'dune' 'build' '-p' 'odoc' '-j' '4' (exited with 1)
    esy-build-package: exiting with errors above...

  building @opam/odoc@github:ocaml/odoc:odoc.opam#eea456edb503ca259718c745233fa4054ad185c5
esy: exiting due to errors above
```

I noticed that this change https://github.com/ocaml/odoc/commit/eea456edb503ca259718c745233fa4054ad185c5#diff-be686d4352ebd00cf2bdeae00bcd6070L4 changed `bisect_ppx` to be used during build time instead of being just a development dependency.

So here's a PR to change it into a build dependency that makes `odoc` build again as a direct esy dependency.

### DIFF
--- a/odoc.opam
+++ b/odoc.opam
@@ -18,6 +18,7 @@ synopsis: "OCaml documentation generator"
 
 depends: [
   "astring" {build}
+  "bisect_ppx"  {build & >= "1.3.0"}
   "cmdliner" {build & >= "1.0.0"}
   "cppo" {build}
   "dune" {build}
@@ -27,7 +28,6 @@ depends: [
   "tyxml" {build & >= "4.3.0"}
 
   "alcotest" {dev & >= "0.8.3"}
-  "bisect_ppx"  {dev & >= "1.3.0"}
   "markup" {dev & >= "0.8.0"}
   "ocamlfind" {dev}
   "sexplib" {dev & >= "113.33.00"}


jonludlam:
Huh, that's interesting - that recent change only added `bisect_ppx` to the model and xref dune files - it was already in parser and html. I wonder how that was working before then?

ostera:
@jonludlam you're right. I misinterpreted your diff. I'm wondering how was it working before then...I'll investigate more.

aantron:
We manually remove the dependency on `bisect_ppx` from `dune` files as part of release. Since you pinned odoc, you are getting the development version from GitHub, which still depends on `bisect_ppx`. See also https://github.com/ocaml/odoc/pull/312#issuecomment-465333593.

ostera:
Perfect, that makes sense. Thanks @aantron 🙏🏼

If the behavior of odoc doesn’t change by keeping them there, I’d like to have this change merged to allow consuming this package with a pinning.

Perhaps @dbuenzli has something to say about this.

dbuenzli:
> Perhaps @dbuenzli has something to say about this.

Nothing, well almost. 

As I already said I would like to see this as a test dependency, I don't think test coverage tools should be required for building in any context. To be precise I wouldn't mind if their dependency was not problematic but `ppx` dependencies are and will always problematic when you want to build/work on the next version of the compiler. 

For tools like `odoc` which will invariably need some work and testing on the next compiler releases, if I manage to fix what the next compiler wants I shouldn't have to wait for fixes in other part of the eco-system or have to manually perform surgery on the build system to be able to pin or release it. Doing so puts needless barriers and delays to bring the support up to date for an essential tool (ideally the support should be ready before the compiler beta).

Also unless done automatically manually surgerging the `dune` file at release seems a bad idea. Releasing should be an [inane procedure](http://ocamllabs.io/projects/2017/02/23/topkg.html) and a matter of following the stupid steps as provided by `dune-release`.

Really please, make that dependency a test dependency and use `dune-release`, it seems all releases done after @trefis (which were done with `topkg` as far as I can see) were done manually which is a bad idea.

aantron:
This change doesn't, on its own, fix the error that is in the description.

The difficult (rather, annoying) part of fixing this is changing `dune` files to detect Bisect, instead of requiring it. I can do that, and then merge this PR after.

Integrating Bisect with Dune is a known problem and has been an outstanding issue for a while. See https://github.com/ocaml/dune/issues/57, where we discussed exactly this.

aantron:
Sorry, not merge this PR, but merge a PR to make Bisect_ppx a `test` dep.

I'll close this, and add open an issue about that.

PR #315, opened 2019-02-20
Doc improvements

### LINKED ISSUE #219: Add odig reference and instructions
Since most users will be using opam, it'd be nice to inform them about odig and how to use it to document their packages in the readme.

dbuenzli:
Among other things closes #219

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,12 +36,23 @@ a work in progress :)
 
 ## Using `odoc` with OCaml
 
+To install odoc with opam:
+
 ```
 opam install odoc
 ```
 
-The easiest way to use odoc right now is by having Dune drive it. This command
-should work in most Dune projects out of the box:
+If you want to use odoc on the packages you have installed in your
+opam switch type:
+
+```
+opam install ocaml-manual odig
+odig doc
+```
+
+When you are developing the easiest way to use odoc right now is by
+having Dune drive it. This command should work in most Dune projects
+out of the box:
 
 ```
 dune build @doc
--- a/README.md
+++ b/README.md
@@ -4,10 +4,10 @@
 [travis-img]: https://api.travis-ci.org/ocaml/odoc.svg?branch=master
 
 **odoc** is a documentation generator for OCaml. It reads *doc comments* ,
-delimited with `(** ... *)`, and outputs HTML. See example output at
-[docs.mirage.io][mirage-docs].
+delimited with `(** ... *)`, and outputs HTML.
+See [example output][odig-sample].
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -4,10 +4,10 @@
 [travis-img]: https://api.travis-ci.org/ocaml/odoc.svg?branch=master
 
 **odoc** is a documentation generator for OCaml. It reads *doc comments* ,
-delimited with `(** ... *)`, and outputs HTML. See example output at
-[docs.mirage.io][mirage-docs].
```
avsm: I'm refreshing docs.mirage.io to build with odoc-master, so I'd prefer to keep that here rather than elide the link completely.  It's still the most comprehensive source of odoc generated docs in terms of the package volume.

dbuenzli: > I'm refreshing docs.mirage.io to build with odoc-master,

Last time I looked it [seemed](https://github.com/mirage/docs) you had switched to using dune for generating it which AFAIK will provide a much less rich package information (e.g. no access to distribution documentation like change logs and readmes). I'm not sure it's worth multiplying the *sample* outputs.

> It's still the most comprehensive source of odoc generated docs in terms of the package volume.

There are quite a few packages that are not represented there (e.g. other alternatives standard libraries). I plan to have a larger set than there is now here by generating doc for a maximal set of packages as soon as the next `odoc`/`odig` combo is out which I hope will happen anytime soon.

dbuenzli: > It's still the most comprehensive source of odoc generated docs in terms of the package volume.

But in case you worry I just pushed an update there that has the docs for ~1000 packages of the opam repository :-)

aantron: I will merge this for now. We can add or change links again in a future PR.

avsm: > Last time I looked it [seemed](https://github.com/mirage/docs) you had switched to using dune for generating it which AFAIK will provide a much less rich package information (e.g. no access to distribution documentation like change logs and readmes). I'm not sure it's worth multiplying the _sample_ outputs.

That's not true -- the scripts in that repo plainly still invoke odig.  I would have preferred a ping that the site was out of date (which appears to be due to the CI service we were using stopping) than to silently remove references to them under a PR titled 'doc improvements'.

dbuenzli: > That's not true -- the scripts in that repo plainly still invoke odig.

Sorry, maybe you should have mentioned that in the README, it wasn't clear. In any case we have a larger doc set at the new link so it shouldn't be a problem. 

Besides odig's home is the logical place to provide that, if only because odig's devs can directly update it themselves as improvements go by rather than have to ping yet another third party.

PR #317, opened 2019-02-22
Remove docs when including modules

### LINKED ISSUE #87: Module descriptions should not appear in includes
When a module has a description comment:

```
(** This is the module that does foos *)
module Foo = struct
  val x : int
  ...
end
```

it gets included in the documentation of includes of that module:

```
include Foo

This is the module that does foos

val x : int
```

It probably shouldn't be since the comment describes the module itself rather than its contents, and if someone wants to know about `Foo` itself they can just follow the link and read about it.

jonludlam:
Fixes #87

Also adds a test for this issue as well as one for Aliases as my first
attempt at this broke them.

### DIFF
--- a/src/xref/expand.ml
+++ b/src/xref/expand.ml
@@ -914,13 +914,21 @@ let expand_module_type t mty =
         let expansion = force_expansion t root (expand_module_type t root mty) in
           { mty with expansion }
 
+let remove_docs_from_signature =
+  (* Remove bare doc comments from a signature *)
+  let open Signature in
+  function
+  | Comment (`Docs _) :: xs -> xs
+  | xs -> xs
+
 let expand_include t incl =
   let open Include in
     if incl.expansion.resolved then incl
     else begin
       let root = Identifier.Signature.root incl.parent in
         match expand_include t root incl with
-        | Expanded content ->
+        | Expanded content' ->
+            let content = remove_docs_from_signature content' in
             let expansion = {content;resolved=true} in
               { incl with expansion }
         | _ -> incl
--- a/test/html/cases/alias.ml
+++ b/test/html/cases/alias.ml
@@ -1,0 +1,7 @@
+module Foo__X = struct
+  (** Module Foo__X documentation. This should appear in the documentation
+      for the alias to this module 'X' *)
+  type t = int
+end
+
+module X = Foo__X
--- a/test/html/expect/test_package+ml/Alias/index.html
+++ b/test/html/expect/test_package+ml/Alias/index.html
@@ -1,0 +1,32 @@
+<!DOCTYPE html><html xmlns="http://www.w3.org/1999/xhtml">
+ <head>
+  <title>
+   Alias (test_package+ml.Alias)
+  </title>
+  <link rel="stylesheet" href="../../odoc.css">
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width,initial-scale=1.0">
+  <script src="../../highlight.pack.js"></script>
+  <script>
+   hljs.initHighlightingOnLoad();
+  </script>
+ </head>
+ <body>
+  <div class="content">
+   <header>
+    <nav>
+     <a href="../index.html">Up</a> – <a href="../index.html">test_package+ml</a> » Alias
+    </nav>
+    <h1>
+     Module <code>Alias</code>
+    </h1>
+   </header>
+   <div class="spec module" id="module-Foo__X">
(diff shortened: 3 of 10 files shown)


src/xref/expand.ml
```diff
@@ -914,13 +914,19 @@ let expand_module_type t mty =
...
+
 let expand_include t incl =
   let open Include in
     if incl.expansion.resolved then incl
     else begin
       let root = Identifier.Signature.root incl.parent in
         match expand_include t root incl with
-        | Expanded content ->
+        | Expanded content' ->
+            let content = remove_docs_from_signature content' in
```
trefis: That doesn't look right. Aren't you removing all the the floating doc comments from that signature? I think we just want to remove the first one.

jonludlam: It's just the first that wants to go is it? I wasn't completely sure. I'll push an update.

trefis:
At the risk of killing the current surge of enthusiasm around odoc, I feel like this PR might have been merged at bit too eagerly for my taste.

I'd have liked to look at it in more details. For example, why is the change presented here more correct / better, than changing [this line](https://github.com/ocaml/odoc/blob/master/src/xref/expand.ml#L253)? Or than reworking a bit the various types to better track what comes from where?

jonludlam:
As it happens, that line you point out was my first attempt, but it broke module aliases which I didn't notice until running it with odig. Hence the new test 😀. This is one thing I'd really like to see improved; that the current set of tests isn't really sufficient to give me confidence that a change is ok or not.

In general I'm very happy to discuss implementation alternatives. Did you have a specific change to the types in mind?

aantron:
@trefis No problem, my fault, we can keep discussing it, revert it, add more changes on top, etc.

Optional, but I suggest to review with "changes requested" (or whatever it is called) if you want to help communicate that you want to slow down a PR.

If this code is to be changed, I'd like to ask that `remove_docs_from_signature content'` be renamed, since the name is a stretch at this point. Otherwise, I'll do it in some refactoring/tweaking pass. Also, @jonludlam, I'd like to request to avoid force-pushes, as it makes conversations difficult to follow after they happen. For PRs where the merge is likely to be by squash anyway, there is rarely a reason to do a force-push.

trefis:
> Optional, but I suggest to review with "changes requested" (or whatever it is called) if you want to help communicate that you want to slow down a PR.

Ack, will do :)

> As it happens, that line you point out was my first attempt, but it broke module aliases which I didn't notice until running it with odig.

By that, do you mean that [here](https://b0-system.github.io/odig/doc/base/Base/index.html#module-Backtrace) the doc wasn't appearing?

---

Btw, the situation #87 describes is:
```ocaml
(** This is the module that does foos *)
module Foo = struct
  val x : int
  ...
end
```

And the code of this PR correctly drops the comment.
But, IIUC, it also drops the comment in this situation:

```ocaml
module Foo = struct

  (** Some text about Foo *)

  val x : int
  ...
end
```

which I'm not sure is something we want.

jonludlam:
@trefis yes exactly, those docs disappeared, as well as the top comments in the nested page - ie, [here](https://b0-system.github.io/odig/doc/base/Base/Backtrace/index.html)

trefis:
Ack.

Note for everyone else: we've discussed my example offline, it seems that the current behavior (as implemented in this PR) is the most reasonable one.
So let's just stop here, thanks everyone!

test/html/cases/alias.ml
```diff
@@ -0,0 +1,7 @@
+module Foo__X = struct
+  (** Module Foo__X documentation. This should appear in the documentation
+      for the alias to this module 'X' *)
```
trefis: That's incorrect, that docstring is actually attached to the type `t`, not to the module. You want to add a blank line.

test/html/expect/test_package+ml/Alias/index.html
```diff
@@ -0,0 +1,32 @@
...
+   <header>
+    <nav>
+     <a href="../index.html">Up</a> – <a href="../index.html">test_package+ml</a> » Alias
+    </nav>
+    <h1>
+     Module <code>Alias</code>
+    </h1>
+   </header>
+   <div class="spec module" id="module-Foo__X">
+    <a href="#module-Foo__X" class="anchor"></a><code><span class="keyword">module </span><a href="Foo__X/index.html">Foo__X</a> : <span class="keyword">sig</span> ... <span class="keyword">end</span></code>
```
trefis: That link is dead, for some reason no page was generated for `Foo__X`.

PR #319, opened 2019-02-23
HTML: clean up keyword markup

aantron:
- Move spaces inside keyword `<span>`s to the outside. This makes the markup friendly to stylesheets that use a background color or an underline to highlight keywords.
- Punctuation marks and parentheses are not keywords.

All changes are reflected in the expect files.

This PR is mostly FYI. I'll merge once CI passes.

I spotted several more bugs while doing this. I'll fix them after this PR.

### DIFF
--- a/src/html/ML.ml
+++ b/src/html/ML.ml
@@ -68,8 +68,8 @@ module ML = Generator.Make (struct
 
   module Mod =
   struct
-    let open_tag = "sig"
-    let close_tag = "end"
+    let open_tag = Generator.keyword "sig"
+    let close_tag = Generator.keyword "end"
     let close_tag_semicolon = false
     let include_semicolon = false
     let functor_keyword = true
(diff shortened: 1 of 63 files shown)


aantron:
There seems to be a difference in how we render comments in classes between 4.08 and other versions: https://travis-ci.org/ocaml/odoc/jobs/497594436#L1773. I'm going to drop this test from this PR, and look at the problem separately in follow-on work.

aantron:
It seems that before OCaml 4.08, comments like this were dropped:

```ocaml
object
  (** No blank line above this comment. *)

  method x : int
end
```

while comments like this were kept:

```ocaml
object

  (** Blank line above this comment. *)

  method x : int
end
```

4.08 keeps both. I suspect this was changed by https://github.com/ocaml/ocaml/pull/1693. @lpw25, does that seem right?

lpw25:
>  @lpw25, does that seem right?

I think ocaml/ocaml#1693 only affects cases where the class is empty other than comments. I think this was instead fixed by ocaml/ocaml#2151.

PR #330, opened 2019-03-02
odoc.css: Change copyright attribution.

dbuenzli:
The stylesheet was significantly changed since I created it and no longer represents my original design or one I would author.


aantron:
Thank you.

PR #331, opened 2019-03-05
theme-uri: propagate everywhere

### LINKED ISSUE #318: `--theme-uri` is not fully propagated
See for example here https://b0-system.github.io/odig/doc/astring/Astring/Char/Ascii/index.html

trefis:
Fixes #318.

This is a quick fix. A better version would make the argument mandatory instead of optional, to avoid such issues in the future.


dbuenzli:
Thanks @trefis !

Issue #327, opened 2019-02-26
Remove or reconsider <section>s, <header>s, <article>s


aantron:
See https://github.com/ocaml/odoc/issues/326#issuecomment-467455333.

dbuenzli:
I think a structure here would be as mentioned before to have as few nesting as possible and allow main elements to be easily retargeted on the page geometry e.g. using CSS grid.

The main logical elements that can be distinguished in ocamldoc generated documentation is a global navigation bar (the "breadcrumbs"), a local navigation bar (the "toc"), a preamble and actual contents (everything after the preamble). The preamble is defined by:

* For an `mli`, anything before the first structure item or header. 
* For an `odoc` page, anything before the first header preceeded by content.

This can then be rendered to a toplevel div or article with basically the following structure:

```xml
<article> 
 <nav>...</nav> <--! breadcrumbs -->
 <header>...</header>   <--! module preamble !-->
 <nav>...</nav> <--! toc -->
 <div>...</div> <--! contents -->
</article>
```

In the preamble and contents part the mapping to the ocamldoc language should be straightfoward. `ocamldoc` paragraphs map to paragraphs, lists to lists, headers to headers etc, structure item to block elements; all at the same level nesting level

Structure items map to a block element which distinguish the item and its doc string as block elements as well. In the doc string block, the straightforward mapping to the ocamldoc language is maintained.

This structure closely maps the mental model exposed by ocamldoc to the writer and thus should minimizes the discrepancy between what the user writes and what she gets or expect.

lpw25:
A couple of questions/suggestions: Shouldn't there be a `<main>` tag somewhere or do people not actually use those? I feel like there should be some semantic element for the grouping of an item and its attached documentation. Isn't it really it's own little `<article>` or `<section>`? Certainly merlin treats them as standalone articles when it displays them to users. Similarly our internal search engine has individual listings for the individual definitions.

dbuenzli:
> 1. Shouldn't there be a `<main>` tag somewhere or do people not actually use those?

One thing is that `<main>` should not be used as a descendent of an `article` and there should be only one in the document so I think it's better not to use it in case you want to compose the `odoc` output with other markup (you'd also need to find something else for the toplevel element).

> 2\. I feel like there should be some semantic element for the grouping of an item and its attached documentation. Isn't it really it's own little `<article>` or `<section>`?

[...]

PR #335, opened 2019-03-09
odoc.css: Update margin to improve centering

kevinji:
This seems to make the centering better, except between ~700px and 760px, where the page should probably switch to the smaller layout but sticks with the larger layout. Any improvements are welcome.

### DIFF
--- a/src/odoc/etc/odoc.css
+++ b/src/odoc/etc/odoc.css
@@ -46,10 +46,7 @@ body {
 
 .content {
   max-width: 90ex;
-  margin-left: calc(10vw + 20ex);
-  margin-right: 4ex;
-  margin-top: 20px;
-  margin-bottom: 50px;
+  margin: 20px auto 50px;
   font-family: "Noticia Text", Georgia, serif;
   line-height: 1.5;
 }


aantron:
Thanks!

It's possible we don't want centering in odoc output (at least at the moment). For example, applying this CSS rule to [this page](http://odis.io/odoc-demo/order/Order/index.html) moves the content far away from the sidebar, and the distance from the sidebar begins to vary with window width (worse, for this specific rule, the sidebar can occlude the content for some widths).

I think, for docs with a sidebar, a typical convention is to left-align everything. In terms of web layout, the content's left margin width is fixed (though to different values for different window widths), and is some function of the sidebar width. We could probably find lots of examples of this kind of layout in common documentation systems for other languages. The main example that comes immediately to mind is Sphinx:

- https://docs.python.org/2/library/string.html
- https://jbuilder.readthedocs.io/en/latest/ (I believe readthedocs is styled Sphinx output)

There are examples with centering and a sidebar, e.g. [Travis docs](https://docs.travis-ci.com/user/environment-variables/). However, their sidebar is not fixed in the window. I think not setting `left` inside a  `fixed` child of a `margin: X auto` block would center a fixed sidebar, but I'm not sure this is desirable.

kevinji:
Ah, good catch; I think I was working off an older version of odoc that didn't have the sidebar, so this makes a lot more sense.

dbuenzli:
This is tangentially related, but the current markup makes it hard to control these parameters in general. It would be nice to move the model mentioned [here](https://github.com/ocaml/odoc/issues/327#issuecomment-467593801) so that we can use the modern tools browsers give us to control the arrangement on the page.

aantron:
> Maybe it's worth it to include the sidebar on all pages?

Agreed, see #294.

For pages that have no internal sections, we will eventually be generating a sidebar with at least the global table of contents.

dbuenzli:
> For pages that have no internal sections, we will eventually be generating a sidebar with at least the global table of contents.

What do you mean by global table of contents ? I don't think anything global should go in the toc, the toc is for the current page and should remain so consistently. Global navigation should be left to the breadcrumbs `nav`.

If there are no internal section simply put nothing there (but do generate an empty `nav` element).

lpw25:
I haven't been keeping up with the styling but I was under the impression that at a certain width the side-bar got inlined into the main page. Do we centre the content in that case? I feel like we probably should although maybe the widths are too small for it to matter at that point.

dbuenzli:
> I haven't been keeping up with the styling but I was under the impression that at a certain width the side-bar got inlined into the main page. Do we centre the content in that case?

@lpw25 the page takes the whole width at that point so it is somehow naturally centered... 

@kevinji The absence of centering and unbalanced feeling is indeed due to the lack of `nav` representing the toc on certain pages. 

In my [own stylesheet](https://b0-system.github.io/odig/doc/) I actually took the decision to put the toc on the right because it entails a more balanced feeling when this "bug" happens (people are used to have blank on the right of the page).

However the exact reason for this was more a UX choice than a simple look and feel argument. The toc on the left disturbs the process of going from the end of the line to the start of the next line when you read the main content. In other words when you read the main content the toc block is easier to ignore at the end of the line rather than at the beginning. 

Also in a lot of modules the toc block will remain mostly empty, toc on the right makes it easier to ignore and makes a less "empty" feeling on the page while still making it easy to access when you need it.

dbuenzli:
> (people are used to have blank on the right of the page).

That is in a left-to-right language.

kevinji:
In any case, this isn't the change that I thought it would be, so I'm going to close this PR.

PR #344, opened 2019-03-31
Always use " = " separator for module type definitions

greedy:
All cases of module type definitions in both OCaml and Reason use an
equal sign to separate the identifier from the definitions.


aantron:
Thanks!

PR #345, opened 2019-03-31
Check that tidy supports the mute option

greedy:
macOS comes with tidy but it is an old version that does not support the `--mute` option. This leads to spurious test failures on macOS.

This adds a check to is_present_in_path that requires the available version of tidy support the mute configuration option before it will be used.

### DIFF
--- a/test/html/tidy.ml
+++ b/test/html/tidy.ml
@@ -13,7 +13,7 @@ let muted_warnings = [
 
 
 let is_present_in_path =
-  Sys.command "which tidy > /dev/null" = 0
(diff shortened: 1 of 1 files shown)


aantron:
Thanks. I think we'd eventually prefer a version check, but this works.

greedy:
@aantron a version check was the first thing I had in mind but the version string for the version on macOS El Captian (yes, I know, it's quite old) reads

HTML Tidy for Mac OS X released on 31 October 2006 - Apple Inc. build 15.17

where 5.6.0 built here says

HTML Tidy for Apple macOS version 5.6.0

so it's not too straightforward to compare the version strings.

aantron:
Argh, yes, fair enough. Thanks!

Issue #96, opened 2017-11-13
destructive substitution not being done


trefis:

greedy:
This was bothering me in some work I'm doing so I thought I'd take a look at fixing this. I tried the simplest thing that could possibly work:

```patch
diff --git a/src/xref/expand.ml b/src/xref/expand.ml
index ba36e36..9062ed4 100644
--- a/src/xref/expand.ml
+++ b/src/xref/expand.ml
@@ -305,8 +305,8 @@ and expand_module_type_expr ({equal; _ } as t) root dest offset expr =
                match subst with
                | TypeEq(frag, eq) -> refine_type ex frag eq
                | ModuleEq(frag, eq) -> refine_module ex frag eq
-               | TypeSubst _ -> ex (* TODO perform substitution *)
-               | ModuleSubst _ -> ex (* TODO perform substitution *))
+               | TypeSubst(frag, eq) -> refine_type ex frag eq
+               | ModuleSubst(frag, eq) -> refine_module ex frag (Module.Alias eq))
             ex substs
     | TypeOf decl ->
         expand_module_decl t root dest offset decl (* TODO perform weakening *)
```

Somewhat surprisingly, this seems to work right for my usecase. It even removes the substituted types and modules from the expansion. I imagine that it isn't actually this easy and there are cases which won't work right. If anyone could show some examples where this doesn't produce the desired results maybe I could make a more correct fix.

avsm:
That diff seems like a step forward even if it only works for structures! Would you like to convert it into a PR to this repo?

greedy:
I found out that the reason it seems to do something useful for includes at the top-level of a module interface is that the problem in #343 causes `refine_type` to fall into the \`Dot case and return None which seems probably incorrect. Note that this also happens with a non-destructive refinement in released odoc.

jonludlam:
Fixed with new model.

PR #349, opened 2019-04-08
[wip] Expand destructive substitution

greedy:
Work in progress to address #96. I think it's producing the right sort of results but I'm not at all sure there aren't some corner cases that are incorrectly handled. I've made a test case but not any expectation files until I'm sure what the right result should be. Of course any comments about the implementation qualities are welcome as well.

I plan to squash this and remove the debugging before this would be merged.

### DIFF
--- a/src/model/dump.ml
+++ b/src/model/dump.ml
@@ -1,0 +1,271 @@
+module Identifier =struct
+    module Signature = struct
+        let rec dump : Paths.Identifier.Signature.t -> string = function
+            | `Root(_, n) -> (Names.UnitName.to_string n)
+            | `Module(s, n) -> Printf.sprintf "%s.%s" (dump s) (Names.ModuleName.to_string n)
+            | `Argument(s, i, n) -> Printf.sprintf "(argument %s %d %s)" (dump s) i (Names.ArgumentName.to_string n)
+            | `ModuleType(s, n) -> Printf.sprintf "%s.%s" (dump s) (Names.ModuleTypeName.to_string n)
+    end
+    module ClassSignature = struct
+        let dump : Paths.Identifier.ClassSignature.t -> string = function
+            | `Class(s, n) -> Printf.sprintf "%s.%s" (Signature.dump s) (Names.ClassName.to_string n)
+            | `ClassType(s, n) -> Printf.sprintf "%s.%s" (Signature.dump s) (Names.ClassTypeName.to_string n)
+    end
+    module DataType = struct
+        let dump : Paths.Identifier.DataType.t -> string = function
+            | `Type(s, n) -> Printf.sprintf "%s.%s" (Signature.dump s) (Names.TypeName.to_string n)
+            | `CoreType n -> Names.TypeName.to_string n
+    end
+    module Parent = struct
+        let dump : Paths.Identifier.Parent.t -> string = function
+            | #Paths_types.Identifier.signature as s -> Signature.dump s
+            | #Paths_types.Identifier.datatype as t -> DataType.dump t
+            | #Paths_types.Identifier.class_signature as s -> ClassSignature.dump s
+    end
+    module LabelParent = struct
+        let dump : Paths.Identifier.LabelParent.t -> string = function
+            | #Paths_types.Identifier.parent as p -> Parent.dump p
+            | `Page(_, n) -> Printf.sprintf "(page %s)" (Names.PageName.to_string n)
+    end
+    module Module = struct
+        let dump : Paths.Identifier.Module.t -> string = function
+            | `Root(_, n) -> (Names.UnitName.to_string n)
+            | `Module(s, n) -> Printf.sprintf "%s.%s" (Signature.dump s) (Names.ModuleName.to_string n)
+            | `Argument(s, i, n) -> Printf.sprintf "(argument %s %d %s)" (Signature.dump s) i (Names.ArgumentName.to_string n)
+    end
+    module ModuleType = struct
+        let dump : Paths.Identifier.ModuleType.t -> string = function
+            | `ModuleType(s, n) -> Printf.sprintf "%s.%s" (Signature.dump s) (Names.ModuleTypeName.to_string n)
+    end
+    module Type = struct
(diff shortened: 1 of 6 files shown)


jonludlam:
This is looking really good thanks! I'll make some comments inline.

src/model/dump.ml
```diff
@@ -0,0 +1,271 @@
+module Identifier =struct
```
jonludlam: This module is very similar to src/test/print/print.ml - if you've got anything extra can you merge it into there please.

src/xref/subst.ml
```diff
@@ -371,3 +371,161 @@ end
...
+  method root x = x
+
+  method offset_identifier_signature x = x
+
+end
+
+let subst_type ~equal x y =
+  new subst_type ~equal x y
+
+class subst_module ~equal:_ (x:Path.Resolved.Module.t) (_y:Path.Module.t) : t = object (self)
```
jonludlam: _y -> y

src/xref/subst.ml
```diff
@@ -371,3 +371,161 @@ end
...
+        | _ ->
+          (match item with
+           | Lang.Signature.Value Lang.Value.{ id; type_=Lang.TypeExpr.Constr(tp, _); _ } ->
+             Printf.printf "Visiting value %s : _ %s\n" (Dump.Identifier.Value.dump id) (Dump.Path.Type.dump tp)
+           | _ -> ()
+          );
+          let item' = self#signature_item item in
+          item' :: items) s []
+
+  inherit Maps.paths as _super
```
jonludlam: _super -> super

greedy:
I'm not sure what to do to fix the test case I just added. My first thought was to rearrange the order of processing so that the substitution was done after expansion somehow. The problem here is that (for module substitution, at least) the substitution could target a module that is included and then maybe they'd need to be iterated to a fixed point to get completely correct results? Of course, it'd probably be okay to just approximate with subst-include-subst or something?

jonludlam:
Hmm yes - very interesting! I'll have a poke around and see if I can come up with an explanation of what's going on.

jonludlam:
I think I have an explanation of what's going on.

Right up at the top of `expand.ml` there's the `module_type` method:

```ocaml
method! module_type mty =
    let mty' = expand_module_type t mty in
      super#module_type mty'
```

The first bit of this essentially resolves anything that's not a signature already. In your case we've got two module types that are already signatures:

```ocaml
module type S = sig
  type 'a t
  val init : string t
end

module type X = sig
    include S
    val doit : 'a t -> unit
end
```

and one that isn't:

```ocaml
module type X2 = X with type 'a t := 'a list
```

The idea is that we turn `X` into `sig include S val doit : 'a t -> unit end` first, effectively by splicing in the definition of `X` into the `expansion` field of `X2` at this point and fixing up the identifiers by mapping `X -> X2`. We then recurse down this expansion in the call to `super#module_type`.

Unfortunately the way it's working now is that we do the call to `replace_type` in this first call to `expand_module_type`, which returns an expansion that has correctly had the type `t` erased. So at this point the expansion is now:

```ocaml
   Some
     (Module.Signature
       [Signature.Include
         {Include.parent =
           `ModuleType
             (`Root root,
              "X2");
          doc = [];
          decl =
           Module.ModuleType
            (ModuleType.Path
              (`Resolved
                 (`Identifier
                    (`ModuleType
                       (`Root root,
                        "S")))));
          expansion =
           {Include.resolved = false;
            content =
             [Signature.Value
               {Value.id =
                 `Value
                   (`ModuleType
                      (`Root root,
                       "X2"),
                    "init");
                doc = [];
                type_ =
                 TypeExpr.Constr
                  (`Resolved (`Identifier (`CoreType "list")),
                  [TypeExpr.Constr
                    (`Resolved (`Identifier (`CoreType "string")), [])])}]}};
        Signature.Value
         {Value.id =
           `Value
             (`ModuleType
                (`Root root,
                 "X2"),
              "doit");
          doc = [];
          type_ =
           TypeExpr.Arrow (None,
            TypeExpr.Constr
             (`Resolved (`Identifier (`CoreType "list")),
             [TypeExpr.Var "a"]),
            TypeExpr.Constr
             (`Resolved (`Identifier (`CoreType "unit")), []))}])}                   
```

Unfortunately in the `Include` the `expansion` field has `resolved=false`, so when we recurse through this expansion we hit:

```ocaml
let expand_include t incl =
  let open Include in
    if incl.expansion.resolved then incl
    else begin
      let root = Identifier.Signature.root incl.parent in
        match expand_include t root incl with
        | Expanded content' ->
            let content = remove_docs_from_signature content' in
            let expansion = {content;resolved=true} in
              { incl with expansion }
        | _ -> incl
    end
```

and because `resolved=false` we _reexpand_ the include, unfortunately skipping out the bit where we replace the type. So we end up with an expansion that has `X` respliced back in (though we do at least remember to rename it `X2`).

I'm not yet sure of the correct fix.

lpw25:
I haven't had a chance to look at any of this yet, but as a hint: IIRC when an include expansion has `resolved = false` it has come from the `cmti` file directly not from the work of the actual expander. It is supposed to be replaced by the proper expansion done by the expander and should only actually be used for anything if resolving the path of the include fails for some reason.

jonludlam:
I see, yes - the expansion of the `include` is loaded directly from the cmti file. We've already processed the expansion of `X` at this point but when we look up the path to `X` we get if from the original cmti rather than what we've already processed. So this could be a candidate for fixing with the improved expansion-ready components data structure?

lpw25:
> So this could be a candidate for fixing with the improved expansion-ready components data structure?

I think so yes. IIRC the current expansion code does nothing for destructive substitution, whilst the current resolving code does something that only works in the common case. The hope is that the new components data structure would do the correct thing and so solve both of these issues.

jonludlam:
This has been superseded by the work going on [here](https://github.com/jonludlam/odoc/tree/next)

jonludlam:
Thanks for the work you put in on this @greedy - the new model should be much more understandable and easier to hack on, so I look forward to your future contributions :-) Also this particular issue is fixed!

PR #353, opened 2019-04-26
Fix generated html for conjunctive types

Octachron:
This PR fixes the html generated for conjunctive types, for instance
```OCaml
type 'a t = [< `X of & int & float ] as 'a
```
which are currently printed as
```OCaml
type 'a t = [< `X of int * float ] as 'a
```
because the list of arguments of the polymorphic variant constructor is interpreted as a product instead of a conjunction.

### DIFF
--- a/test/html/cases/recent.mli
+++ b/test/html/cases/recent.mli
@@ -26,3 +26,11 @@ type polymorphic_variant = [
 ]
 
 type nonrec nonrec_ = int
+
+
+(* Conjunctive types: dune compilation scheme exposes a bug in old
+   versions of the compiler *)
+type empty_conj= X: [< `X of & 'a & int * float  ] -> empty_conj
+type conj = X: [< `X of int & [< `B of int & float ] ] -> conj
+val empty_conj: [< `X of & 'a & int * float  ]
+val conj : [< `X of int & [< `B of int & float ] ]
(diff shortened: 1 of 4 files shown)


jonludlam:
Well, you learn something new every day :-) I'd not come across these types before.

test/html/cases/recent.mli
```diff
@@ -26,3 +26,11 @@ type polymorphic_variant = [
 ]
 
 type nonrec nonrec_ = int
+
+
+(* Conjunctive types: dune compilation scheme exposes a bug in old
+   versions of the compiler *)
```
jonludlam: What does this refer to?

Octachron: I was thinking of https://github.com/ocaml/ocaml/issues/7629 , but there were other issues with conjunctive types in older version of ocaml : https://github.com/ocaml/ocaml/issues/7496 for instance. The short explanation is that conjunctive types are intended as type-checking artefacts and using them directly can make compilation break in unexpected way.

Octachron:
Indeed, conjunctive types are not really intended to be used directly. Honestly, the value of this PR is more in the documentation of the unusal interpration of constructor arguments for polymorphic variants than in the fix itself.

jonludlam:
Right. Thanks very much for this!

jonludlam:
Rebased in #358

PR #354, opened 2019-05-02
Reference.deconstruct

Julow:
Add the `Reference.deconstruct` and `Reference.Resolved.deconstruct` functions to easily print references.

While working with Odoc, I had to write these functions in order to print references.
They are very hard to write and most probably contain bugs. Implementing them in Odoc should solve these problems.

The interface is:

```ocaml
val deconstruct : t -> string * Paths_types.Reference.tag_any
```

### DIFF
--- a/src/model/paths.ml
+++ b/src/model/paths.ml
@@ -2523,6 +2523,41 @@ module Reference = struct
       | #Paths_types.Resolved_reference.s_label as x -> x
       | _ -> assert false
 
+    let rec deconstruct_ acc : t -> _ =
+      let dot t s =
+        let ds, _ = deconstruct_ (s :: acc) t in
+        ds
+      in
+      function
+      | `Identifier id -> Identifier.name id :: acc, `TUnknown
+      | `SubstAlias (_, r) -> deconstruct_ acc (r :> t)
+      | `Module (r, s) -> dot (r :> t) (ModuleName.to_string s), `TModule
+      | `Canonical (_, `Resolved r) -> deconstruct_ acc (r :> t)
+      | `Canonical (p, _) -> deconstruct_ acc (p :> t)
+      | `ModuleType (r, s) -> dot (r :> t) (ModuleTypeName.to_string s), `TModuleType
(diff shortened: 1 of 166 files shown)


lpw25:
What is this function for exactly? Without the use case it is not clear that this function produces the right string.

Either way I think you should split the function into two functions -- one to get the string and one to get the kind.

jonludlam:
The use case is essentially `unparse` - this is for formatting ocamldoc comments. The AST returned by the parser contains `Reference.t`s and we need to convert it back to a string to insert in the reformatted text.

Another suggestion was to keep the original string in the AST, but on reflection that didn't look right. I suppose another possibility is to use the location info in the AST to get it from the original string too.

lpw25:
> The AST returned by the parser contains Reference.ts

I suspect this is the problem. Really the parser should produce something that is pretty close to one-to-one with the syntax and we should translate them into `Reference.t`s when we store the information into the things in `Lang`. `Reference.t`s don't keep enough information for re-printing them as they were, and maintaining that information when processing them would be annoying.

Julow:
The function could be made a little more useful by returning the list of component, the kind and the label: `: t -> string list * tag_any * string option` and allow the user to concatenate it.

Otherwise in https://github.com/ocaml/odoc/pull/351, I suggest to expose `Parser_.Ast` and later to simplify it. This is one of the simplifications I had in mind. What do you think ?

lpw25:
> The function could be made a little more useful by returning the list of component, the kind and the label: : t -> string list * tag_any * string option and allow the user to concatenate it.

I don't think this does what you want for a reference like:

    Foo.module-type-Bar.t

The last part of the reference is not the only part that has a meaningful kind.

Also, you'll conflate:

    Foo.type-t

with

    type:Foo.t

which I imagine you'd prefer not to do (assuming you are trying to print the comments as they were written, rather than trying to normalize them).

Julow:
I see. Returning the type of each part is also not good because it would be too hard to known what should be printed and how.

PR #357, opened 2019-05-21
Do not use double underscores in library (or unit) names

trefis:
Basically, this undoes https://github.com/ocaml/odoc/issues/236#issuecomment-440248694 .

This has the benefit of simplifying the build rules, but the main reason for this change is that using double underscore in library and module names is a bad idea: various tools (including the compiler and odoc itself) understand "__" as "this name is generated, please do your best to not show it to me".

The various libraries are now named "odoc_whatever", and I've updated all the code in `src/` to refer to "Odoc_whatever" instead of "Whatever".


jonludlam:
This looks fine to me.

trefis:
Travis seems to be unhappy, but I don't understand why / it seems unrelated to my patch.
For instance, [the 4.07.1 test](https://travis-ci.org/ocaml/odoc/jobs/535686538) failed with:
```
<><> jbuilder.transition installed successfully <><><><><><><><><><><><><><><><>
=> Jbuilder has been renamed and the jbuilder package is now a transition package. Use the dune package instead.
+make dune-test
dune build
make: *** [build] Segmentation fault (core dumped)
/home/travis/.travis/functions: line 524: 13671 Terminated              travis_jigger "${!}" "${timeout}" "${cmd[@]}"
```
But that seems to have happened after having run the tests (which we can see the output of above in the log).

Also, it seems like the 4.08 tests are using beta1. Which seems like a bad idea, it should be switched to beta3.

jonludlam:
That's when it clones the dev version of dune and tries to run `make dune-test` - so possibly an issue with dune master? I'll check.

jonludlam:
https://github.com/ocaml/dune/issues/2187

trefis:
Ack. So the failure is unrelated to this issue and we can merge this?

Also, ping @Julow: the libraries name will change when we merge this. This shouldn't be too much of an issue for ocamlformat (i.e. a sed should do the job).

aantron:
I would have hesitated to use `Odoc_` prefixes (with one underscore), because of all these compiler (ocamldoc) modules: https://github.com/ocaml/ocaml/tree/trunk/ocamldoc.

I don't recall where they are packaged, `compiler-libs`, or just visible to ocamldoc plugins.

PR #359, opened 2019-05-24
<span> parentheses

trefis:
As suggested by @dbuenzli (:heart_eyes:) on https://github.com/ocaml/odoc/issues/19#issuecomment-470128041

Before this PR:
before

With this PR:
after

It's not the cleanest patch, but I encountered some type-checker bug when trying a cleaner version, so this will have to do.

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -69,40 +69,61 @@ struct
     : 'inner 'outer. Odoc_model.Lang.TypeExpr.Polymorphic_variant.t ->
         ('inner, 'outer) text Html.elt
   = fun (t : Odoc_model.Lang.TypeExpr.Polymorphic_variant.t) ->
-    let elements =
-      list_concat_map t.elements ~sep:(Html.txt " | ") ~f:(function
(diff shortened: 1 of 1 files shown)


trefis:
I added another commit to span polymorphic variants.
Moving from:
before
to:
after

The code is a bit convoluted, that's because when there is a pipe, I force it to be on the same line as the constructor (but I'm happy to push the argument to the next line).
If you don't do things that way, you end up with:
```ocaml
[ `Foo |
`Bar |
`Baz ]
```
which is less aesthetically pleasing.

trefis:
Note: I didn't update the testsuite because:
1. I didn't notice any obvious "promote" target in the Makefile
2. the generated html is not so nice right now, it generates extra (unneeded) spans in places.

aantron:
@trefis the test suite gives you instructions for promotion. For example, this is in the [Travis log](https://travis-ci.org/ocaml/odoc/jobs/537726810#L1792-L1793):

```
To promote the actual output to expected, run:
cp `cat _build/default/test/html/_scratch/actual` `cat _build/default/test/html/_scratch/expected` && make test
```

trefis:
Thanks @aantron, I had missed that.

PR #360, opened 2019-05-28
Span things

trefis:
#359 reloaded:
- it's closer to @dbuenzli's initial markup
- I also `<span>` type constructors and their parameters together
- the implementation is much cleaner.
- I've updated the testsuite!

The final result for the extreme example I was using doesn't look quite as nice:
markup

But it's reasonable, and Daniel seems to thing the markup proposed here will be better in general, which I tend to agree with.
I'm considering including some special handling for extreme cases like the one in the screenshot, but that will wait until a later date.

I think that unlike #359 this is in a reasonable state and is ready for review.

### DIFF
--- a/src/html/ML.ml
+++ b/src/html/ML.ml
@@ -18,7 +18,7 @@ module ML = Generator.Make (struct
 
     let handle_params name args =
       if args <> [ Html.txt "" ]
-      then args @ [ Html.txt " " ] @ name
+      then [Html.span (args @ [ Html.txt " " ] @ name)]
       else name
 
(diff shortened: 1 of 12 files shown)


trefis:
The failure is only on 4.08 and because one of our dependencies doesn't build.
But since we're using beta1 in travis (which is known to be broken), I don't think it's worth investigating. (4.08 is going to be release in the coming days, potentially even today, let's just update travis at that point).

aantron:
Thanks!

And thanks also for including the test diff with each commit.

PR #361, opened 2019-05-29
Reduce memory usage

trefis:
We have some files at janestreet where odoc was using over 40G of memory, this series of patch brings that back down to roughly 10G.
On these files odoc is about 5x slower though.

I expect the rewrite that @jonludlam is currently working on to improve the (resulting as well as preexisting) slowness issues.
In the meantime, these patches are necessary because odoc gets reliably OOM killed when ran from a build system.

### DIFF
--- a/src/model/maps.ml
+++ b/src/model/maps.ml
@@ -961,88 +961,133 @@ class virtual documentation = object (self)
       reference
 
   method private documentation_inline_element element =
-    match element with
+    let open Location_ in
+    match element.value with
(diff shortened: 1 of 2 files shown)


lpw25:
I think the first two commits are worth having. The third one probably just makes things worse for people other than us -- so we're happy to just keep that on our internal branch until the rewrite is done.

trefis:
I removed the last commit.

aantron:
Thanks!

PR #362, opened 2019-06-06
HTML TOC for mld pages

rizo:
Fixes https://github.com/ocaml/odoc/issues/294.

### Notes for Reviewers

- Added `Generator.Top_level_markup.lay_out_page` – an entry point for HTML generation of pages. This function will produce HTML content for pages, the header element and the TOC.
- Introduced the `comment_state` type that is used to track the state for HTML generation of both signatures and mld files.
- `page_section_comment` implements the same logic as `section_comment` with some changes:
  - The title heading is not included in TOC.
  - The title heading and the following content (until the next heading or end) are placed into a header element. This is similar to signatures.
  - The other headings and the following content are _not_ placed into section elements. I find this unnecessary for plain mld files. I also think we should simplify this in signatures (https://github.com/ocaml/odoc/issues/327).
- A potential improvement would be to avoid duplicating the comment traversal logic between signatures and pages, but that would require quite deep changes to the HTML generator.

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -725,6 +725,11 @@ sig
 
   val render_toc :
     toc -> ([> Html_types.flow5_without_header_footer ] Html.elt) list
+
+  val lay_out_page : Odoc_model.Comment.docs ->
+      ((Html_types.div_content Html.elt) list *
+       (Html_types.flow5_without_header_footer Html.elt) list *
+       toc)
 end =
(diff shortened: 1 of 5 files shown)


rizo:
The CI failure does not seem to be related to odoc tests. It's failing on 4.08 compiling base.

aantron:
Thanks!

Issue #355, opened 2019-05-10
[RFC] Simpler AST


Julow:
Hi,

First of all, thanks for this great library !

I'm trying to use Odoc in OCamlformat to format documentation comments.
It's been great so far but I ran into a few problems with the representation of comments (`Model.Comment`):

- The parser requires passing a `containing_definition` argument that is hard to understand.
- A lot of comments that were valid with Octavius now trigger warnings
- Some part of the AST are very hard to understand and pretty print
	References (see https://github.com/ocaml/odoc/pull/354)

I proposed in https://github.com/ocaml/odoc/pull/351 to expose `Parser_.Ast`.
This AST has some advantages:

- Parser is easier to call (no need to build dummy arguments)
- Less strict (no `non_link_inline_element`, etc..)
- Heading label is not auto generated yet

It could then be simplified by moving some code from the parser to the `semantics` pass that build `Model.Comment`:

- Represent references as strings
- List syntax (`-` or `+` vs `ul` or `ol`)
	This information could also be in `Model.Comment`.
- Keep whitespaces
- Allow any language or none in blocks (eg. `Raw_markup of string option * string`)

Alternatively, most of these could be fixed without exposing `Parser_.Ast` and only modifying `Model.Comment`.
But I think this AST will be much easier to change in the future to allow more use cases.
We will probably receive issues about documentation comments on OCamlformat and Odoc will be used by a lot of projects.

I'll be happy to hear your opinions on this.

jonludlam:
There are several things here to comment on.

Firstly, there's no question that the parser should be easier to call, so nobody has to make up a dummy Root.t and so on.

In general, I think the AST from Model.Comment should contain enough information to recover the original text - that's fundamentally what ocamlformat needs. It's not missing much 
- the whitespace info
- the list syntax used originally
- a way to recover the original reference text
- a way to know whether the heading was autogenerated or not
- unrecognised tags need to be labeled as such along with the original tag

I'm actually quite happy that it currently generates more warnings than octavius - providing that we can produce an AST in all cases octavius could. It'd be handy to feed those warnings back to ocamlformat in a useful way to be reported - it'd be neat if merlin could also do this! This is also why I'm not keen on leaving the references as strings, since warnings in parsing them couldn't be passed back to ocamlformat - essentially we keep all of the input validation happening at the same point.

[...]

PR #364, opened 2019-06-13
Simpler parser ast

Julow:
Hi!

As I explained in ocaml/odoc#355, `Model.Comment` is a bit hard to obtain and understand and too strict and lossy on the input.

Some of this issues could be solved by changing `Model.Comment` with a bit of work. But I think that in the future, this will make similar issues harder to solve as Odoc codebase and language change.

This PR simplifies `Parser.Ast` to make it closer to the concrete syntax and simpler by delaying some parsing to `semantics.ml`:

- References are represented as strings
- `Raw_markup` targets can be any string or none.
- Keep concrete syntax used for lists
- Keep whitespaces (Only when a `Space` element is generated)

This PR also removes some uses of `Model` types, as @lpw25 suggested, but not all yet.

This PR is supposed to follow ocaml/odoc#351 but if this approach is definitively rejected, it should be a good first step at improving `Model.Comment`.

What do you think?

Thanks!

### DIFF
--- a/src/parser/syntax.ml
+++ b/src/parser/syntax.ml
@@ -90,13 +90,13 @@ let rec inline_element
     fun input location next_token ->
 
   match next_token with
-  | `Space ws ->
+  | `Space _ as token ->
     junk input;
-    Location.at location (`Space ws)
+    Location.at location token
 
-  | `Word w ->
+  | `Word _ as token ->
     junk input;
-    Location.at location (`Word w)
+    Location.at location token
     (* This is actually the same memory representation as the token, complete
        with location, and is probably the most common case. Perhaps the token
        can be reused somehow. The same is true of [`Space], [`Code_span]. *)
--- a/src/parser/syntax.ml
+++ b/src/parser/syntax.ml
@@ -90,9 +90,9 @@ let rec inline_element
     fun input location next_token ->
 
   match next_token with
-  | `Space ->
+  | `Space ws ->
     junk input;
-    Location.at location `Space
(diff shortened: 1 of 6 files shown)


src/parser/syntax.ml
```diff
@@ -90,9 +90,9 @@ let rec inline_element
     fun input location next_token ->
 
   match next_token with
-  | `Space ->
+  | `Space ws ->
     junk input;
-    Location.at location `Space
+    Location.at location (`Space ws)
```
aantron: This looks like a good place for an `as`-binding to avoid an allocation, which is introduced by this diff.

Julow:
Thanks for your review !

I cleaned the code but did not manage to fix the CI with 4.08 yet.
Yes, I'd like to reopen #351 if possible, I just rebased the branch. The stability should not be an issue for us.

aantron:
Thanks! The 4.08 failure is not due to this PR, and is tracked in #367. It seems GitHub won't let me reopen #351, because the underlying branch has been rebased. Would you mind opening a new PR, or pinging me if you move the branch back to the last commit in #351 when it was open?

aantron:
Thanks! The 4.08 failure is not due to this PR, and is tracked in #367. It seems GitHub won't let me reopen #351, because the underlying branch has been rebased. Would you mind opening a new PR, or pinging me if you move the branch back to the last commit in #351 when it was open?

Julow:
Thanks for merging !
I changed the branch of #351.

Issue #365, opened 2019-06-13
Sections paragraphs are reversed in odoc manuals


dbuenzli:
On master it seems section paragraphs are rendered in the reverse order on mld pages with:
```
> odoc --version 
v1.1.1-969-gd40430bc
```

Steps to reproduce

```
opam pin odoc --dev
opam install odig 
odig doc -u odig # see the manual page and compare with:
                 # https://b0-system.github.io/odig/doc/odig/manual.html
```

rizo:
Good catch! https://github.com/ocaml/odoc/pull/366 should fix this, sorry about that!

PR #366, opened 2019-06-14
Fixes paragraph ordering in HTML for mld pages.

rizo:
Fixes https://github.com/ocaml/odoc/issues/365.

## Notes

* Added paragraphs to the mld test.
* There's no point in using `finish_comment_state` for `acc_html` because without section elements we're appending elements, not consing.


aantron:
Thanks @rizo!

PR #370, opened 2019-07-18
Fix trimming of leading whitespaces in codeblocks

Julow:
The trimming removes the maximum amount of leading spaces that every lines contain.

```ocaml
(** {[ foo
       bar
    ]} *)
```

In this example, it is 1 and the output will be

```
foo\n      bar
```

This PR fix that by using the location to compute the indentation of the first line, so the output is `foo\nbar`.

### DIFF
--- a/src/parser/lexer.mll
+++ b/src/parser/lexer.mll
@@ -70,7 +70,8 @@ let trim_trailing_blank_lines : string -> string = fun s ->
     in
     String.sub s 0 trim_from
 
-let trim_leading_whitespace : string -> string = fun s ->
+let trim_leading_whitespace : first_line_offset:int -> string -> string =
+ fun ~first_line_offset s ->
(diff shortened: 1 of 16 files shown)


jonludlam:
What happens with:

```ocaml
(** {[ foo
 bar
    ]} *)
```

?

Julow:
In that case, the first line will not be trimmed. That's a regression.

Julow:
Fixed.

jonludlam:
I think it'll make more sense to group the new tests with the other leading-whitespace tests above (and perhaps to name them similarly?)

Julow:
That makes sense. I cleaned the tests.

jonludlam:
Looks good to me now. Thanks!

aantron:
LGTM too.

PR #376, opened 2019-08-30
odoc/env: Fix bug in List.iter2 throwing Invalid_arg

bluddy:
If we encounter End_of_file, we get uneven lists for roots and paths.


aantron:
Thanks. The CI error looks unrelated to the PR.

PR #378, opened 2019-09-01
Add meta generator

### LINKED ISSUE #324: Readd versioned meta generator tag
The old markup had: 
```
<meta name="generator" content="doc-ock-html %%VERSION%%"/>
```
which is useful to know with which version of the markup generator `odoc` you are staring at. It would be nice to readd this.

dbuenzli:
Bring back the versioned meta generator tag (closes #324).

I made the test promotion as a separate commit. Btw. is there a way to promote things via a single diff. I had to run the tests, see one failing, promote, start over, etc. I'm sure I'm missing something.

Also I promoted a few things that seem unrelated to the change but is related to @trefis' spanning of sigs I think.

### DIFF
--- a/src/html/tree.ml
+++ b/src/html/tree.ml
@@ -283,6 +283,8 @@ let page_creator ?kind ?(theme_uri = Relative "./") ~path header_docs content =
     Html.head (Html.title (Html.txt title_string)) [
       Html.link ~rel:[`Stylesheet] ~href:odoc_css_uri () ;
       Html.meta ~a:[ Html.a_charset "utf-8" ] () ;
+      Html.meta ~a:[ Html.a_name "generator";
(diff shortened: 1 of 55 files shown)


dbuenzli:
The 4.02.3 failures are unrelated, apparently `dune` fails to compile.

aantron:
Thanks.

IIRC, there is some command you could manually write to promote in one go, but there is no direct "support" for it at the moment.

dbuenzli:
Thanks.

> IIRC, there is some command you could manually write to promote in one go,

Eventually got around with:
```
for i in {1..XXX}; do cp `cat _build/default/test/html/_scratch/actual` `cat _build/default/test/html/_scratch/expected` && make test; done
```
But it's still very slow...

PR #381, opened 2019-09-11
Handle 4.08 module and type substitutions

### LINKED ISSUE #350: support ocaml 4.08 features
Opening this as a followup to #304 and #310 which added basic OCaml 4.08 compat, but didn't support the new features.

From @trefis, `Tsig_typesubst` and `Tsig_modsubst` occurrences will result in unresolved references, so we need testcases that exercise the new features and support for resolving them.

(Note that as of #304, OCaml 4.08 can be used with odoc for resolving existing code that doesn't use the new 4.08 features).

jonludlam:
This PR will render the new syntactic elements as written in the mli files. This should fix #350 as it qualifies as 'minimal support'. We might want to consider removing the substituted modules and types entirely, depending on how these features end up being used.

### DIFF
--- a/src/loader/cmti.ml
+++ b/src/loader/cmti.ml
@@ -622,8 +622,6 @@ and read_signature_item env parent item =
         [ModuleSubstitution (read_module_substitution env parent mst)]
 #endif
 
-(*and read_type_subst _env _parent _tst = []*)
-
 and read_module_substitution env parent ms =
   let open ModuleSubstitution in
   let name = parenthesise (Ident.name ms.ms_id) in
--- a/src/loader/cmti.ml
+++ b/src/loader/cmti.ml
@@ -273,6 +273,9 @@ let read_type_declarations env parent rec_flag decls =
   in
     List.rev items
 
+let read_type_substitutions env parent decls =
+  List.map (fun decl -> Odoc_model.Lang.Signature.TypeSubstitution (read_type_declaration env parent decl)) decls
+
 let read_extension_constructor env parent ext =
   let open Extension.Constructor in
   let open Odoc_model.Names in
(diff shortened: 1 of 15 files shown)


src/loader/cmti.ml
```diff
@@ -613,10 +616,23 @@ and read_signature_item env parent item =
...
 #if OCAML_MAJOR = 4 && OCAML_MINOR >= 08
-    | Tsig_typesubst _
-    | Tsig_modsubst _ -> []
+    | Tsig_typesubst tst ->
+        read_type_substitutions env parent tst
+    | Tsig_modsubst mst ->
+        [ModuleSubstitution (read_module_substitution env parent mst)]
 #endif
 
+(*and read_type_subst _env _parent _tst = []*)
```
trefis: This can be deleted.

src/model/maps.ml
```diff
@@ -1175,6 +1175,28 @@ class virtual module_ = object (self)
...
+
+  method module_substitution subst =
+    let open ModuleSubstitution in
+    let {id; doc; manifest} = subst in
+    let id' = self#identifier_module id in
+    let doc' = self#documentation doc in
+    let manifest' = self#path_module manifest in
+    if id != id' || doc != doc' || manifest' != manifest then
+        {id = id'; doc = doc'; manifest = manifest' }
+      else subst
```
trefis: The indentation doesn't look quite right.

trefis: I'm sure Leo would *love* that.

aantron:
Thanks @jonludlam and @trefis. Please let me know if we need a speedy release. Otherwise, the next release should be around 1 October.

PR #383, opened 2019-09-15
Fix compilation on OCaml 4.09.0 beta2

### LINKED ISSUE #382: Build failure on 4.09
See https://travis-ci.org/ocaml/odoc/jobs/584430492#L403:

```
File "src/model/ident_env.cppo.ml", line 152, characters 41-57:
Error: Unbound value Compat.signature
```

It's probably because of this condition:

https://github.com/ocaml/odoc/blob/61ac0c9532ee6bb372dfd6643d73b13155f02953/src/model/compat.cppo.ml#L70

from #309. cc @jonludlam

avsm:
The constraint in the cppo conditional compilation also expands to 4.10.0~dev as well, since it should work there too.

Closes #382
cc @jonludlam

### DIFF
--- a/src/model/compat.cppo.ml
+++ b/src/model/compat.cppo.ml
@@ -67,7 +67,7 @@ and modtype_declaration =
 
 let opt conv = function | None -> None | Some x -> Some (conv x)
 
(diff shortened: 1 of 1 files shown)


aantron:
Thanks!

PR #385, opened 2019-09-17
Parser: handle colons in references to operators

### LINKED ISSUE #384: Cannot reference to infix operator, e.g OUnit2.(>::), in .mld files
I get an error, if I use the following piece of code in a .mld file:
```
{1 Foobar}
{!Test.(>::)}
```

The error is:
File "index.mld", line 1, characters 2-10:
Unknown reference qualifier 'Test.(>:'.

Not only does it strip half of the identifier, it also doesn't link '>::'. Using the same code in a .ml file, I got no errors. Doing the same thing with ocamldoc works.

Here is an archive where you can see the problem:
[test-odoc.tar.gz](https://github.com/ocaml/odoc/files/3616521/test-odoc.tar.gz)

Just extract the archive and run "make odoc" to see the problem.

aantron:
Fixes #384.

cc @gildor478

### DIFF
--- a/src/parser/reference.ml
+++ b/src/parser/reference.ml
@@ -321,7 +321,21 @@ let parse warnings whole_reference_location s : (Paths.Reference.t, Error.t) Res
   in
 
   let old_kind, s, location =
-    match String.rindex s ':' with
+    let rec find_old_reference_kind_separator index =
+      match s.[index] with
+      | ':' ->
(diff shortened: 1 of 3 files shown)


aantron:
Thanks @gildor478 for the report, and @trefis for the look and the diagnosis.

Issue #284, opened 2019-01-28
Add option to fail on warnings


Gbury:
Is there a way to make odoc fail on warnings ? Ideally, that would be useful in CI scripts to make sure doc builds without errors or warnings.

aantron:
Not at the moment. We should add this.

aantron:
Even odoc's own development would benefit. We currently have CI emitting warnings on odoc's own comments, which have gone unnoticed for a while: https://travis-ci.org/ocaml/odoc/jobs/503418122#L1798.

(Aside: those docs are being built unintentionally, but we should probably build them in a separate test).

aantron:
We should also expose this through Dune somehow (cc @emillon).

emillon:
Yes, that'll need to be opt in for now, either through a `(documentation)` parameter or a flag in `dune-project`, but if it works well that could become default around dune 3 ;)

PR #398, opened 2019-11-25
Option to turn warnings into errors

Julow:
Closes https://github.com/ocaml/odoc/issues/284

This PR adds the option `--warn-error` to turn warnings into errors.
It is implemented using a global flag and failing after warnings are printed.

This is the easier implementation I found and it has some problems:
- Very few warnings are accumulated before `shed_warnings` is called and the error is raised.
- It's a global variable that is very deep in the dependency chain

I also started looking into other ways that require refactoring:
- Passing the warning accumulator everywhere
  It's how the parser works but makes the code very verbose and introduce a lot of diffs. (~300 adds)
- Change `Error.catch` to also "catch" warnings
  This is also a lot of changes and requires to change the API of `Error` to avoid bad uses, which requires even more refactoring.
- Monad style would drastically change the whole codebase

What do you think?

### DIFF
--- a/src/model/error.ml
+++ b/src/model/error.ml
@@ -15,6 +15,7 @@ type filename_only_payload = {
 type t = [
   | `With_full_location of full_location_payload
   | `With_filename_only of filename_only_payload
+  | `Without_location of string
 ]
 
 let full message location =
@@ -22,6 +23,10 @@ let full message location =
 
 let filename_only message file =
   `With_filename_only {file; message}
+
+(** Only used internally *)
+let without_location message =
+  `Without_location message
 
 let make ?suggestion format =
   format |>
@@ -49,6 +54,9 @@ let to_string = function
 
   | `With_filename_only {file; message} ->
     Printf.sprintf "File \"%s\":\n%s" file message
+
+  | `Without_location message ->
+    message
 
 
 
@@ -90,7 +98,7 @@ let shed_warnings with_warnings =
   with_warnings.warnings
   |> List.iter (fun warning -> warning |> to_string |> prerr_endline);
   if !warn_error && with_warnings.warnings <> [] then
-    failwith "Warnings have been generated.";
+    raise_exception (without_location "Warnings have been generated.");
   with_warnings.value
 
 let set_warn_error b = warn_error := b
--- a/src/model/error.ml
+++ b/src/model/error.ml
@@ -82,9 +82,15 @@ let accumulate_warnings f =
 let warning accumulator error =
   accumulator := error::!accumulator
 
+let warn_error = ref false
+
 (* TODO This is a temporary measure until odoc is ported to handle warnings
    throughout. *)
 let shed_warnings with_warnings =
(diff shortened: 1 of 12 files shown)


jonludlam:
Looks like the failures are due to the expect tests expecting exact line numbers in the backtraces.

Julow:
This error message is very bad anyway, I wanted feedback before changing it.
`Failure` is raised in `odoc/*` when handling errors from `Odoc_loader.read_*` and are not catch.
I changed the code to properly handle that error but the output is still an unhandled exception.

aantron:
This looks fine. We can refactor it later, either passing a warning accumulator, or converting to monadic style.

How are you invoking odoc with `--warn-error`?

The test should be fixed to pass, which I guess will happen when you fix up the error message :)

Julow:
The option is added to `odoc compile`, `odoc html` and `odoc html-fragment`. Do other commands need it ?
I commited a small refactor that propagate the errors up to the main and print/exit there. This is only for the 3 subcommands, so it's not very consistent. I can work on doing that for all the subcommands in an other PR if you think it's a good approach ?

aantron:
> The option is added to odoc compile, odoc html and odoc html-fragment. Do other commands need it ?

What I meant is, how are you driving odoc? I am just curious for comparison. AFAIK when driving odoc through Dune, users still won't have easy access to the option.

> I commited a small refactor that propagate the errors up to the main and print/exit there. This is only for the 3 subcommands, so it's not very consistent. I can work on doing that for all the subcommands in an other PR if you think it's a good approach ?

The refactor looks okay at first glance, it is at least more consistent than what we had before :) I would accept a second PR for this. However, a lot of this code is probably in need of even more fundamental refactoring with respect to error propagation and reporting.

Julow:
> What I meant is, how are you driving odoc? I am just curious for comparison. AFAIK when driving odoc through Dune, users still won't have easy access to the option.

My first thought was that is Dune's job. Just like libraries, `dune build @doc` is for dev builds and `dune build -p my_package @doc` for release builds. Also, the `(documentation ..)` stanza could allow to enable or disable explicitly this flag (or allow to pass flags directly).
I don't known about other tools calling Odoc. Do you think a different API is needed ? (eg. an env variable)

> The refactor looks okay at first glance, it is at least more consistent than what we had before :) I would accept a second PR for this. However, a lot of this code is probably in need of even more fundamental refactoring with respect to error propagation and reporting.

I'd like to split this in an other PR and go a little bit further to make things a little cleaner.

aantron:
Ok, that all sounds good.

Julow:
Sorry for the delay and thanks for merging :)
Do you think I should open an issue on the limitations of this PR ?
Only a few warnings are printed before the error is raised terminating the program instead of collecting all the warnings.

aantron:
> Do you think I should open an issue on the limitations of this PR ?
> Only a few warnings are printed before the error is raised terminating the program instead of collecting all the warnings.

Yes, please :)

PR #400, opened 2019-12-03
Don't exit in library code

Julow:
Refactor error reporting to avoid exiting the program in library code.
Errors are propagated back to `odoc/bin/main.ml` by value and handled in an unified way.

Most of the changes are trivial but the patch is very invasive, exiting functions were very deep in the program.
Using exceptions can make the patch a little smaller but also harder as all exposed functions would need to catch them and there is a lot of them.

The exit status for some errors (marshalling) is now `1` instead of `2`. Is that an issue ?

I added the `Or_error` module that re-export `result` and define a bind operator. I didn't use `rresult` as it was removed in https://github.com/ocaml/odoc/pull/306.

Only the first commit is required for https://github.com/ocaml/odoc/pull/398

### DIFF
--- a/src/xref/odoc_xref.ml
+++ b/src/xref/odoc_xref.ml
@@ -13,6 +13,9 @@
  * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
+
+type 'a or_error = ('a, [ `Msg of string ]) Result.result
+
 type lookup_result_found = Component_table.lookup_result_found =
   { root : Odoc_model.Root.t; hidden : bool }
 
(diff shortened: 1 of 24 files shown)


src/xref/odoc_xref.ml
```diff
@@ -21,22 +22,49 @@ type lookup_result = Component_table.lookup_unit_result =
...
+    match fetch_page root with
+    | Ok page -> page
+    | Error e -> raise (Fetch_failed e)
+  in
+  Resolve.build_resolver ?equal ?hash lookup_unit fetch_unit lookup_page fetch_page
 
-let resolve = Resolve.resolve
+let resolve resolver unit =
+  try Ok (Resolve.resolve resolver unit)
+  with Fetch_failed (`Msg _ as e) -> Error e
```
Julow: I use an exception to propagate errors inside the resolver. This is to avoid a giant refactor and potentially making it slower.

aantron:
Thanks!

PR #408, opened 2020-01-13
4.10 compatibility

jonludlam:
This moves us in the direction of using 4.10's data types everywhere - though it doesn't go quite as far as making the module names optional.

### DIFF
--- a/src/loader/cmt.ml
+++ b/src/loader/cmt.ml
@@ -357,24 +357,27 @@ let rec read_module_expr env parent label_parent pos mexpr =
     | Tmod_ident _ ->
         Cmi.read_module_type env parent pos (Odoc_model.Compat.module_type mexpr.mod_type)
     | Tmod_structure str -> Signature (read_structure env parent str)
-    | Tmod_functor(id, _, arg, res) ->
-        let arg =
-          match arg with
-          | None -> None
-          | Some arg ->
-              let name = parenthesise (Ident.name id) in
-              let id = `Argument(parent, pos, ArgumentName.of_string name) in
-          let arg = Cmti.read_module_type env id label_parent 1 arg in
+    | Tmod_functor(parameter, res) ->
+        let parameter, env =
+          match parameter with
+          | Unit -> FunctorParameter.Unit, env
+          | Named (id_opt, _, arg) ->
+              let name, env =
+                match id_opt with
+                | Some id -> parenthesise (Ident.name id), Env.add_argument parent pos id (ArgumentName.of_ident id) env
+                | None -> "_", env
+              in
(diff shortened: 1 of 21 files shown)


aantron:
I think we should keep the test case, and make it conditional on the compiler version, if necessary. The original issue that it was added for was an assertion failure. We want to be sure not to reintroduce it when odoc is used with older versions.

jonludlam:
Good plan!

src/model/maps.mli
```diff
@@ -422,7 +422,7 @@ class virtual module_ : object
   method virtual signature : Signature.t -> Signature.t
 
   method virtual module_type_functor_arg :
-    FunctorArgument.t option -> FunctorArgument.t option
+    FunctorParameter.t -> FunctorParameter.t
```
lpw25: Should probably rename the method to use `param` instead of `arg` to match the type name change

src/loader/cmti.ml
```diff
@@ -529,8 +554,19 @@ and read_module_type_declaration env parent mtd =
...
-  let id = `Module(parent, Odoc_model.Names.ModuleName.of_string name) in
+  let id =
+#if OCAML_MAJOR = 4 && OCAML_MINOR >= 10
+    match md.md_id with
+    | Some id ->
+      let name = parenthesise (Ident.name id) in
+      `Module(parent, Odoc_model.Names.ModuleName.of_string name)
+    | None ->
+      `Module(parent, Odoc_model.Names.ModuleName.of_string "_")
+#else
```
lpw25: I don't think this is the right behaviour for the no id case. In such cases we should just not produce a module item. That probably means that `module_declaration` should return a list of items like some of the other similar functions (or an `option` but that is probably slightly more effort).

src/loader/cmt.ml
```diff
@@ -392,8 +416,19 @@ and unwrap_module_expr_desc = function
...
-  let id = `Module(parent, ModuleName.of_string name) in
+  let id =
+#if OCAML_MAJOR = 4 && OCAML_MINOR >= 10
+      match mb.mb_id with
+      | Some id ->
+        let name = parenthesise (Ident.name id) in
+        `Module(parent, ModuleName.of_string name)
+      | None ->
+        `Module(parent, ModuleName.of_string "_")
+#else
```
lpw25: As with `read_module_declaration`, `read_module_binding` should not produce an item when there is no identifier.

src/loader/cmti.ml
jonludlam: Done.

src/loader/cmt.ml
jonludlam: Done.

src/model/maps.mli
jonludlam: Done.

PR #411, opened 2020-02-20
adjust default `font-weight` for bold text

dhcmrlchtdj:
the original value (500) is too small, it will fallback to 400 and has no any visually different to normal text.

https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#Fallback_weights


aantron:
Confirmed, thanks!

PR #413, opened 2020-02-26
Add sections of inlined include to ToC

### LINKED ISSUE #412: headings from includes not listed in TOC
If in `a.mli` I have
```ocaml
(** some docstrings *)
val some_code : some typ

(** {1 some title} *)
val some_more_code : unit typ
```
and in `b.mli`
```
include A (** @inline *)

(** {1 This title} *)
module X : sig end
```
in the TOC of `B`, only `This title` will be listed.

trefis:
Also, heading levels from the include are shifted to stay nested under the current level. That seems like the right default but perhaps we'll want to add a flag to disable that behavior in the future.

Fixes #412.

This didn't have an impact on the testsuite, so I'll add fresh tests exhibiting the behavior change tomorrow morning.

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -733,12 +733,16 @@ type async
 
 module Top_level_markup :
 sig
+  type heading_level_shift
+
   val lay_out :
+    heading_level_shift option ->
     item_to_id:('item -> string option) ->
     item_to_spec:('item -> string option) ->
     render_leaf_item:('item -> rendered_item * Odoc_model.Comment.docs) ->
     render_nested_article:
-      ('item -> rendered_item * Odoc_model.Comment.docs * toc * Tree.t list) ->
+      (heading_level_shift -> 'item ->
+        rendered_item * Odoc_model.Comment.docs * toc * Tree.t list) ->
     ((_, 'item) tagged_item) list ->
       (Html_types.div_content Html.elt) list * toc * Tree.t list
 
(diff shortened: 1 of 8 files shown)


trefis:
Now with some tests.

src/html/generator.ml
```diff
@@ -1179,11 +1210,11 @@ module Class :
 sig
   val class_ :
     ?theme_uri:Tree.uri -> Lang.Signature.recursive -> Lang.Class.t ->
-      rendered_item * Odoc_model.Comment.docs * Tree.t list
+      rendered_item * Odoc_model.Comment.docs * toc * Tree.t list
```
jonludlam: Is there additional work to do here, as you've added the `toc` to the return type but they're hardcoded to `[]` below?

trefis: I don't believe so. I think right now it only makes sense for includes to add entries to the ToC.
Classes, modules, etc. are currently only displayed on separate pages.

jonludlam: OK, so it's just for uniformity of type?

trefis: :man_shrugging: 

I followed the current structure of the html generator, and went for minimal updates.
I considered adding a new kind of items : "inlined items", which would have been the only one returning a ToC.
But I think the result might actually have been slightly worse.

jonludlam: :+1:

jonludlam:
Thanks!

PR #418, opened 2020-04-01
Fix location of style URI symlinks in test/html

CraigFe:
The HTML expect tests contain symlinks to `odoc.css` and `highlight.pack.js` from elsewhere in the codebase. These symlinks were at relative position `../../../` from the generated HTML pages, which expect a relative position of `../../`.

This bug looks to have been introduced by https://github.com/ocaml/odoc/pull/193.


jonludlam:
Do the generated pages under _scratch still work?

CraigFe:
They do still work, although it took me a while to realise how they _ever_ worked (since they have the same issue of being three levels below the symlinks).

It seems that they work via the `odoc support-files` command in the tests, so are unrelated to this change.

Issue #416, opened 2020-03-30
Empty variant type definitions have no RHS in rendered output


CraigFe:
An empty variant type is rendered as `type t =`, which looks like a bug to me:

![image](https://user-images.githubusercontent.com/26125912/77892181-7734ab00-7272-11ea-9868-d5ba8aa81042.png)

I would expect the output to look something like:
```
type empty = |
```

jonludlam:
Thanks for the report!

PR #419, opened 2020-04-01
Fix HTML rendering of empty variant types

CraigFe:
Resolves https://github.com/ocaml/odoc/issues/416 with the solution suggested there.

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -391,21 +391,24 @@ struct
         in
         anchor, cell
     in
-    let rows =
-      cstrs |> List.map (fun cstr ->
-        let open Odoc_model.Lang.TypeDecl.Constructor in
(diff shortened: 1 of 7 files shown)


CraigFe:
Namely:
image

jonludlam:
Thanks @CraigFe !

CraigFe:
Aha, the tests fail because of empty variant types not being available pre-4.07.

Is there an existing way to restrict tests to a subset of compiler versions? (I can't find one.)

aantron:
@CraigFe only this:

https://github.com/ocaml/odoc/blob/738f4e5405ef11d2442ddd40b9d6f07db46e14c4/test/html/test.ml#L288-L303

jonludlam:
only a rather heavy-handed one - stick it in `recent.mli`, which is only used in 4.08 and above. We could do with a better way to handle these sorts of things.

CraigFe:
Done :slightly_smiling_face:

PR #421, opened 2020-04-02
Issue 420 nested `<em>` tags not rendering properly in browsers

### LINKED ISSUE #420: Distinction between `{i text}` and `{e text}`
I played a bit with italics and emphasis, but I wasn't able to infer the
difference between the two. (And the documentation is [not very explicit](http://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#ss:ocamldoc-formatting).)

As a LaTex user I was expecting `{e text}` to behave in a similar way than
`\emph{text}`. `emph` has the logical meaning "I want to emphasize
that" and this does not always translate to the physical task of writing in
italics. For example, adding emphasis to a word in the middle of an italic
sentence will put the word straight in LaTex.

Expression           | LaTex-based expectation | odoc 1.5.0 behaviour
---------------------| ----------------------- | ----------------
`{i foo}`            | _foo_                   | _foo_
`{e foo}`            | _foo_                   | _foo_
`{e bar {e foo} bar}`| _bar_ foo _bar_         | _bar foo bar_
`{i bar {e foo} bar}`| _bar_ foo _bar_         | _bar foo bar_

That lead to several questions:
- Is the actual behavior the expected behavior ?
- Is there a difference in how `i` and `e` are rendered by odoc ?

Edit: Actually I found #188 that seems to show that nested emphasis should behave as described in this issue, but they are not in my test setup.

voodoos:
This pr fixes #420 

## Issue
Browsers does not style nested emphasis as expected (ie: emphasis should toggle italics on/off, not always on, depending on the context).

## Changes
This PR add logic to `src/html/comment.ml` to annotate `<em>` at an odd level of nesting with the class `odd`. Nesting level is reset when crossing `<it>`tags. Then a few css rules are added to style adequately each case. 

## Notes 
Because this is an issue only for browser rendering, I chose to fix it at the renderer level.
Another choice may have been to fix it at the parser level, by adding an `int` argument to the ``` `Emphasis``` variant.

### DIFF
--- a/src/html/comment.ml
+++ b/src/html/comment.ml
@@ -176,11 +176,18 @@ let location_to_syntax (loc:Odoc_model.Location_.span) =
 let style_to_combinator = function
   | `Bold -> Html.b
   | `Italic -> Html.i
-  | `Emphasis -> Html.em
   | `Superscript -> Html.sup
   | `Subscript -> Html.sub
 
-
+let styled_element ~emph_level ~(next : ?emph_level:int -> 'a -> 'b) content = function
+  | `Emphasis ->
+    let a = if emph_level mod 2 = 0 then [] else [Html.a_class ["odd"]] in
+    let emph_level = emph_level + 1 in
+    Html.em ~a
+      (next ~emph_level content)
+  | #Comment.non_nest_aware_styles as style ->
+      (style_to_combinator style)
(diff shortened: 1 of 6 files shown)


aantron:
Thanks!

Issue #196, opened 2018-09-19
Stable structured output (a.k.a `odoc json ...`)


ostera:
Something I've been considering recently has been extending `odoc` with a `json` module that would export the complete cross-referenced environment as a single json blob that can be used for:

* Generating search indexes — throwing these at elasticsearch would already do miracles for a _hoogle_ style website, and could allow for quick experimentation on the concept
* Building a single-page application powered by this data — no need to modify the existing html generation 👌 
* Opening up the experimentation to a bigger audience!

I'd consider other formats but json seems to be ubiquitous and fast-enough for the use-case of documentation generation.

Opening this issue to gather comments and feedback 🗒

lpw25:
> Generating search indexes — throwing these at elasticsearch would already do miracles for a hoogle style website, and could allow for quick experimentation on the concept

I should probably mention that we have a branch here:

https://github.com/lpw25/odoc/tree/elasticsearch

which generates JSON that use with elasticsearch for a search index within Jane Street.

It needs tidying up and the interface should be changed, which I haven't had time to do, but it should probably be added upstream at some point.

I think it is better to have odoc produce some JSON specifically for search indices than just dump everything as JSON and work from that. For a start you shouldn't under-estimate the size of the data. The documentation of our tree is quite a few Gb and we would rather not have another even larger JSON copy just so we can extract the much smaller search data.

That's not to say you might not want to have a JSON dump for some other purpose -- I'd just rather support the search stuff without needing it.

ostera:
@lpw25 I understand the pragmatism behind the search-index only approach, it is however a tad limiting for exploration and could lead to multiple json-exports that are perhaps too specific to be a part of `odoc`. Although a far-fetched example, I can't imagine `odoc slackbot` being a thing.

I'm envisioning the following usage:

```sh
# Build a search index, internally may or may not use the complete json export
odoc search-index > elastic-search-dump.json

# Export the complete documentation tree as JSON
odoc json ... > ./data/docs.json

# Run a GraphQL server from it
json-graphql-server data/docs.json

# Explore it with `jq`: i.e, get all the nested modules types
echo ./data/docs.json | jq '.modules[.modules[types]]`
```

And naturally, you can use this data to power any sort of application.

The point is that `odoc` doesn't *necessarily* decide anymore.

rizo:
I must suggest that an alternative approach could be converting the `.odoc` files into JSON separately. This functionality doesn’t have to be embedded into odoc itself.

[...]

Issue #121, opened 2018-04-17
Interop with Docusaurus


aantron:
Basically, odoc would be parsing the docs and doing the cross-referencing, and [Docusaurus](https://docusaurus.io/) would be doing the displaying.

Docusaurus is already in use by [reasonml.github.io](https://reasonml.github.io/) and [bucklescript.github.io](https://bucklescript.github.io/), and seems to be doing a good job there.

cc @kennetpostigo @bikallem

bikallem:
This is cool. Docusaurus seems to require markdown files. Does odoc produce markdown?

bikallem:
Answering myself. It doesn't.

Khady:
It is a bit annoying to have to install one more package manager and use npm/yarn to generate documentation. But it guess we can live with it...

dbuenzli:
I'm not sure I understand. What does Docusaurus provide ? If it's just about the rendering then odoc can simply inspire itself for its markup structure/css.

bikallem:
From what I understand, https://docusaurus.io/ provides the facility to generate whole documentation website including blogs and API docs - similar to https://reasonml.github.io/ and https://bucklescript.github.io. This is probably useful for those without much website design/development skills.

aantron:
@Khady Docusaurus wouldn't be a requirement of odoc, just an option. Outputting Markdown for Docusaurus might be more maintainable and produce much higher quality in the short term than figuring out odoc's markup.

MoOx:
I would like to contribute to this discussion by bringing my 2 cents. Sorry if my story is long, but I prefer to give too much infos instead of not enough.

Context: I am coming on this issue because I wanted to improve [bs-react-native docs](https://reasonml-community.github.io/bs-react-native/BsReactNative/index.html) which is currently not really pretty and convenient to browse, and since I made a static site generator (which is friend with react & reasonml), I am looking if odoc could produce alternative outputs (like a low level format instead of html or markdown or a superset - more about this below).

---

I am currently developing [phenomic](https://github.com/phenomic/phenomic), a static site generator generic enough to not be tied to markdown (or anything actually), with react as a renderer (even if others solutions could be implemented - currently, since I use react daily, I did only this one).

When files are "send" to phenomic (which is optional, see later), the inputs (markdown, asciidoc, json) are converted to... just json (and injected into a "kindof" database so you can retrieve this data during the "website compilation" step).
[...]

PR #423, opened 2020-04-14
Add an agnostic output representation

Drup:
So I did it myself

For a while, odoc had several issues:
- The HTML generation code was, I'm going to be blunt, terrible. It had no separation of concern at all. Semantics and layout collided in a mess combining rendering concerns, tyxml typing, and modeling issues such as dozens types of paths. It was hard to understand for module experts and it was *impossible* to understand for layout experts.
- Odoc was locked to emit HTML. Even worse, only one kind of html
- The awfulness of the HTML generation also prevented doing any sort of formatting of the output, which was fairly problematic.

These problem put a huge impediment in odoc's ambition to be *the* documentation tool for ocaml and reason. So, this PR finally fixes it.

This set of commits introduce a new representation with the following properties:
1. **NO SEMANTICS!**
1. Agnostic layout representation
1. Code is already formatted
1. Simplified Urls (basically, either a path to a page, or inside a page)

The first half of the commit log is absolutely unreadable. I doubt there is any way to make it clean, sorry about that. You should start by reading the content of the `/document` and the `/html` directories directly. The later commits (and the one I'll add in the future) should be more self-contained.
Given the scope of this, I would prefer if it's merged then refined, rather than waiting for it to be unmeargeble. :)

----------

Now that I'm done with the very rant-y general presentation, let's discuss the specifics in a calm and civilized fashion :heart: :rainbow:

# Layout representation

The types can be found here: https://github.com/Drup/odoc/blob/internal_repr/src/document/types.ml

Agnostic layout representation are typicaly present in tools such as pandoc. They usually use two "construct": `inline` and `block`. This works nicely as it gives a clear mapping to many markup languages and a rather clear semantics. However, for the purpose of laying out documentation about code, I found that insufficient. I introduce two parallel constructs: "documented source", and "items". 

- DocumentedSrc are basically a bunch of code, but where some pieces have been annotated with ... documentation! At the moment, it's mostly used for type declarations with inline comments (for record fields, etc). It is analogous to `inline` constructs
- Items are the "block" equivalent: they are a whole paragraph of either text or (documented) code.

In both case, these can be nested via an explicit construct. This provides nice layout opportunities for the renderer later down the line. For instance, we could perfectly imagine a "one-page" renderer that inline the content of all modules. I plan to refine the type of document further, but the philosophy should remain.

Since we have an explicit document representation, we can do various things as separate passes, such as coalescing, computing Toc, etc. This is not done yet. Finally, and most importantly, the type for documents should be rather plain and easily accessible! Most of the semantics is gone, leaving very high level constructs. I hope it will be very easy to modify by anyone, especially people who are far better designers than me. I encourage people to look at the code in `/html` in this light.

# Code formatting

On the way, I introduce the `Codefmt` module which allow to use all `Format` capabilities *but still output some agnostic representation*. So far, I do not use these new facilities at all, but it should now be much easier to get started.

The "right" implementation requires 4.08, but I made some hack that will work for older versions.

# Change in HTML output

So far, the changes are very minor:
- Some spans inside links have been removed.
- Most instance of `</code><code>` are gone
- Some output ordering for functors has changed
- Class for type declarations have moved a bit (this probably requires changing the CSS, I will do that later).

# Laundry lists of things to do (preferably after this is merged)

- [x] Allow to derive the table of content from the agnostic representation instead of pre-computing it
- [x] Separate Coalescing into its own module, and get rid of `Item.Declarations`. 
- [ ] Improve the code creating Links. It's much better than before, but still messy.
- [ ] Figure out the title situation
- [ ] Purge the model layer from HTML-specific idiosyncrasies
- [x] Remove the remaining state in `Odoc_html.Tree`
- [ ] Finally forbid headers in documentation of data constructors and records.

# Future work (preferably with other people)

- Make a latex output
- Make a man page output
- Make a JSON/Sexp/somerserializationformat output
- Create a plugin system so that other people can write their own output
- Use odoc in the compiler
- Replace ocamldoc for good.


aantron:
Thanks @Drup! I've started giving this a pretty thorough review. Should be done in some days. I agree with merging this relatively quickly, as there is a considerable risk of conflicts accumulating otherwise.

We've discussed decoupling "core" odoc from the HTML generator on and off for years, so this should be a welcome change. The current code does indeed make it very difficult both to output anything other than HTML, and even to change the HTML output in non-trivial ways.

I also want to link to some existing related comments and issues:

- https://github.com/ocaml/odoc/issues/196 about JSON output for general usage, and @lpw25's [comment](https://github.com/ocaml/odoc/issues/196#issuecomment-423056894) about an existing JSON generator (it might be specifically optimized for indexing; I have not looked). There is another [comment](https://github.com/ocaml/odoc/issues/196#issuecomment-594655090) from @rizo on JSON work.
- https://github.com/ocaml/odoc/issues/121 about using Docusaurus for rendering, which is probably best solved by outputting JSON, and recommeniding some external script for converting to Markdown.

Drup:
I forgot to mention that I'm aware of the model rewrite in progress by @jonludlam . The new agnostic representation is much lower level and do not intent to conflict with the new model in any way. I've taken a quick look at the WIP implementation, and it seems to me the code is largely orthogonal and should be very easy to adapt.

Drup:
I added a new module to provide high level functions to manipulate the IR. In particular, it allows to implement the computation of the ToC and coalescing of items directly in the renderer without too much difficulties. I used it to simplify the IR further.

I can use the same process to remove the notion of `Section` from the IR, but we first need to discuss if we want to change the layout or not. If we do, I prefer to do that in another PR.

Once Section is gone, we are in a very good position to remove 90% of the content of the `Top_level_markup` module, which would significantly simplify this whole ordeal.

Drup:
I've made another patchset which removes sections (as discussed in https://github.com/ocaml/odoc/issues/327 and several other places) and take advantage of the changes to remove huge chunks of code. The commits can be seen here: https://github.com/Drup/odoc/compare/internal_repr...Drup:no_section

test/html/expect/test_package+ml/Recent/index.html
```diff
@@ -30,22 +30,22 @@ <h1>
...
     <dt class="spec type" id="type-variant">
-     <a href="#type-variant" class="anchor"></a><code><span class="keyword">type</span> variant</code><code> = </code>
-     <table class="variant">
+     <a href="#type-variant" class="anchor"></a><code><span class="keyword">type</span> variant = </code>
+     <table>
       <tbody>
        <tr id="type-variant.A" class="anchored">
-        <td class="def constructor">
-         <a href="#type-variant.A" class="anchor"></a><code>| </code><code><span class="constructor">A</span></code>
+        <td class="def variant constructor">
```
aantron, line 37: > Class for type declarations have moved a bit (this probably requires changing the CSS, I will do that later).

Could you comment on why this has happened?

src/model/comment.ml
```diff
@@ -5,18 +5,14 @@ module Identifier = Paths.Identifier
 type 'a with_location = 'a Location_.with_location
 
 
-type non_nest_aware_styles = [
+type style = [
   | `Bold
   | `Italic
+  | `Emphasis
```
aantron, line 11: This appears as a partial reversion of #421 (see https://github.com/ocaml/odoc/pull/421/files#diff-02e8c64dc888acbac7e60f11e7ee5efd).

test/html/expect/test_package+ml/Nested/F/argument-1-Arg1/index.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html><html xmlns="http://www.w3.org/1999/xhtml">
  <head>
   <title>
-   1-Arg1 (test_package+ml.Nested.F.1-Arg1)
+   Arg1 (test_package+ml.Nested.F.1-Arg1)
```
aantron, line 4: Ok, a nit. Or perhaps a question to be addressed in a future issue: should the path in the parentheses also end in `Arg1`, rather than `1-Arg1`?

test/html/expect/test_package+ml/Nested/F/index.html
```diff
@@ -27,6 +27,20 @@ <h1>
     <p>
      Some additional comments.
     </p>
+    <h3 id="heading">
```
aantron, line 30: > Some output ordering for functors has changed

I assume that refers to this code, so functor parameters are now documented ahead of the table of contents. Could you comment on why this is so?

src/model/comment.ml
Drup: Yes, it's on purpose, I reimplemented it in a different way in b6b59d1 . Imho, This is purely an HTML rendering issue, it should not affect the model or even the IR in any way. So now it doesn't.

test/html/expect/test_package+ml/Nested/F/index.html
Drup: The output for functor seemed inconsistent with the rest. Semantically, the functor parameters should be part of the "heading" of a page. Practically, there parameters were not in the header but yet outside of the main `<section>` where the rest of the content sits. Apart from functors, the content of a page is always "breadcrumb, heading, toc, content", so it felt natural to do it like that. Given that the toc doesn't reference the functor parameters and is stylistically pushed aside anyway, it shouldn't change anything, so I went ahead.

test/html/expect/test_package+ml/Nested/F/argument-1-Arg1/index.html
Drup: I ... felt it was better that way. The path, on the other hand should preferably be unique and can be richer/clunkier. It's certainly not difficult to change, one way or the other.

test/html/expect/test_package+ml/Recent/index.html
Drup: I think it's better explained by looking at the representation of typedecls in the IR:
a typedecl is a `DocumentedSrc.t` and we have

```
  type 'a documented = {
    attrs : Class.t ;
    anchor : Url.Anchor.t option ;
    code : 'a ;
    doc : Block.t ;
  }
  type t = one list
  and one =
    | Code of only_code
    | Documented of Inline.t documented
    | Nested of t documented
```

The shape of the IR made it fairly natural to put annotations/classes on each line, instead on the global thing. 

Now that I think about it, it wouldn't actually be much effort at all to fix this. I'll do that.

We can also discuss the shape of the IR. I converged towards this one because it helped me produce the cleanest code, but it's of course not set in stone.

Drup: Actually, it's more awkward than I though because of polymorphic variants. I propose to push that in a later PR. :)

aantron: Ok. We can also change the CSS if that's the easier way.

Drup: That was my plan.

Drup:
I've rebased. The history is now reasonably clean and each commit compiles and pass the tests.

aantron:
Thanks!

Issue #422, opened 2020-04-13
Darkmode


ulrikstrid:
bisect_ppx recently added [darkmode](https://github.com/aantron/bisect_ppx/pull/301).
The approach we took was to convert colors to css variables and then used the new media query to override the default variables with darker colors.

css variables are not supported in IE but every other modern browser supports them, would this be fine for odoc?
I might be able to do the work, at least something initial since I'm not a good designer and might pick bad colors, if there is interest/support for this change.

edit: We can actually keep IE compatibility by defining the colors twice.

ulrikstrid:
I did a initial "spike" to just see what it would look like.

![image](https://user-images.githubusercontent.com/1607770/79340239-50d96580-7f2a-11ea-9db1-6887a091a634.png)

PR #424, opened 2020-04-17
Initial darkmode implementation

ulrikstrid:
Initial implementation of what's suggested in #422.
This should keep IE compatibility by keeping the "old" colors in addition to the css variables.


aantron:
Thanks!

We should adjust the colors in follow-on work. For example, the blue color of ordinary links is very difficult to see on my notebook screen under glare, over the dark background. I'll open an issue about that. If we don't get around to it before release, we can temporarily emit the `light` class to avoid surprising anyone whose browser is set to dark mode.

FYI for others: the color scheme is chosen according to the browser's light or dark mode setting. We can also later add a command line flag or environment variable, to force light or dark mode, by emitting the `light` or `dark` class on the `<html>` element in the output.

PR #426, opened 2020-04-18
Remove sections and cleanup the code

Drup:
This patch removes the wrapping in `<section>` tags, as discussed in #327 and other various places.
The structure is the one proposed by @dbuenzli in https://github.com/ocaml/odoc/issues/327#issuecomment-467593801 . It can be changed later on fairly easily as well.

The generator initially contained a pretty complicated piece of code to rediscover sectioning while generating the markup. Removing sections allowed to simplify this piece of code dramatically, which is what is done in the rest of the commits. All in all, This removes around 600 lines of code (making the combination of this patch set + #423 a net negative in number of lines!). There are still many code simplification opportunities around, notably around the handling of links, but the IR is now in an (imho) acceptable shape.

I recommend reading commits by commits, and to ignore whitespace in diffs for the test fixes.

Now, Onto man pages! :rocket:

### DIFF
--- a/src/document/ML.ml
+++ b/src/document/ML.ml
@@ -88,4 +88,4 @@ module ML = Generator.Make (struct
   end
 end)
 
-include ML.Page
+include ML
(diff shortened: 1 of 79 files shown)


aantron:
Thanks!

Issue #417, opened 2020-04-01
odoc crashes while processing include with shadowing


fpottier:
Processing the following code causes `odoc` 1.5.0 to crash. (This is a simplified version of actual code.)

```
module M
  (S : sig
     module E : sig
       type 'a t
     end
  end)
  (D : sig type element end)
= struct
  include S
  module E = struct
    include E
    include struct
      open D
      type 'a t = element E.t
    end
  end
end
```

The baktrace is as follows:

```
master $ make doc
        odoc src/.sek.objs/byte/sek__SupplyDefault.odoc (exit 2)
(cd _build/default/src/.sek.objs/byte && /Users/fpottier/.opam/4.09.1/bin/odoc compile -I . -I ../../../_doc/_odoc/pkg/sek --pkg sek -o sek__SupplyDefault.odoc sek__SupplyDefault.cmt)
odoc: internal error, uncaught exception:
      CamlinternalLazy.Undefined
      Raised at file "camlinternalLazy.ml", line 22, characters 42-57
      Called from file "camlinternalLazy.ml", line 31, characters 17-27
      Re-raised at file "camlinternalLazy.ml", line 36, characters 4-11
      Called from file "camlinternalLazy.ml", line 31, characters 17-27
      Re-raised at file "camlinternalLazy.ml", line 36, characters 4-11
      Called from file "src/xref/component_table.ml", line 531, characters 44-58
      Called from file "camlinternalLazy.ml", line 31, characters 17-27
      Re-raised at file "camlinternalLazy.ml", line 36, characters 4-11
      Called from file "src/xref/resolve.ml", line 67, characters 18-31
      Called from file "src/model/maps.ml", line 2084, characters 19-33
      Called from file "src/model/maps.ml", line 20, characters 15-18
      Called from file "src/model/maps.ml", line 1548, characters 20-54
      Called from file "src/model/maps.ml", line 1571, characters 20-43
      Called from file "src/model/maps.ml", line 1368, characters 22-36
      Called from file "src/model/maps.ml", line 9, characters 15-18
      Called from file "src/model/maps.ml", line 12, characters 19-31
      Called from file "src/model/maps.ml", line 12, characters 19-31
      Called from file "src/model/maps.ml", line 12, characters 19-31
      Called from file "src/model/maps.ml", line 1277, characters 20-34
      Called from file "src/xref/resolve.ml", line 2240, characters 24-53
      Called from file "src/xref/resolve.ml", line 2169, characters 18-42
[...]

PR #439, opened 2020-06-14
New model

jonludlam:
This PR contains the new model that's been in development for a while.

Largely this follows the design doc on the wiki - https://github.com/ocaml/odoc/wiki/Model-design
although to ensure backward compatibility with existing tools (dune and odig) the link
phase is done by the `html` command.

This new version is quite memory hungry on larger projects (e.g. Core_kernel), which will require some tweaking before release.


jonludlam:
I thought it might be useful to have the results of this run on core_kernel to see what it looks like.
The results are currently hosted here: https://www.cl.cam.ac.uk/~jjl25/docs/

Highlights:
- No pages generated for hidden modules (with double underscores) - these are now unnecessary, as the links to
  canonical modules are now generated correctly. See e.g. the main page for Core_kernel where previously many
  of the module links were to hidden pages.
  Before: https://ocaml.janestreet.com/ocaml-core/latest/doc/core_kernel/Core_kernel/
  After: https://www.cl.cam.ac.uk/~jjl25/docs/core_kernel/Core_kernel/index.html

- Links to modules go into the body of the module rather than to the declaration point, unless the module is
  opaque, in which case we link to the declaration point. Compare clicking on the right-hand side of
  `Applicative` here:
  Before: https://ocaml.janestreet.com/ocaml-core/latest/doc/core_kernel/Core_kernel/index.html#module-Applicative
  After: https://www.cl.cam.ac.uk/~jjl25/docs/core_kernel/Core_kernel/index.html#module-Applicative

- Destructive type substitution works correctly - see for example the first 'include' in this module:
  Before: https://ocaml.janestreet.com/ocaml-core/latest/doc/core_kernel/Core_kernel/Array/index.html
  After: https://www.cl.cam.ac.uk/~jjl25/docs/core_kernel/Core_kernel/Array/index.html

- Adding type equalities to signatures works as expected. See e.g. type `key` in expansion of Table here:
  Before: https://ocaml.janestreet.com/ocaml-core/latest/doc/core_kernel/Core_kernel/Pid/index.html#module-Table
  After: https://www.cl.cam.ac.uk/~jjl25/docs/core_kernel/Core_kernel/Pid/index.html#module-Table

- Shadowing works better (not quite perfectly yet). See for example the output for the code posted in
  issue #417: https://www.cl.cam.ac.uk/~jjl25/shadow/shadow/Shadow/index.html

- Complex functors work as expected. See expansion of `Make_let_syntax` in `base`
  https://www.cl.cam.ac.uk/~jjl25/docs/base/Base/Applicative/index.html#module-Make_let_syntax
  note that in the expansion both `t` and `Open_on_rhs_intf` have been removed.


Still to resolve:
- Shadowed items introduced via `include module type of struct include X end` are not
  removed from the resulting signature - see for example Core_kernel.Float - duplication of type `t`

- Links to items that are hidden from the docs via `(**/**)` shouldn't resolve - they currently do,
  leading to broken links.

- Classes haven't really been looked at hard yet

- `module type of` still isn't quite right yet, and renders as `module type of sig...end` rather than
  `module type of struct ... end`

avsm:
Just making notes as I flick through the (new and vastly improved) generated docs:

- https://www.cl.cam.ac.uk/~jjl25/docs/ocaml/Stdlib/index.html - the rendering of the intro paragraph overflows onto the right column on Safari.

- in the Stdlib ocaml package, only some modules are there. ListLabels, List, etc are missing.  https://www.cl.cam.ac.uk/~jjl25/docs/ocaml/index.html

trefis:
> https://www.cl.cam.ac.uk/~jjl25/docs/ocaml/Stdlib/index.html - the rendering of the intro paragraph overflows onto the right column on Safari.

There are a few rendering issues in firefox as well.
I assume that's unrelated to this PR but rather due to some changes in the markup that happened recently, and that odig's CSS hasn't been updated yet (which makes sense given there was no release of odoc with these changes).

> in the Stdlib ocaml package, only some modules are there. ListLabels, List, etc are missing. https://www.cl.cam.ac.uk/~jjl25/docs/ocaml/index.html

That's generated by odig, not odoc.

PR #442, opened 2020-07-06
Various optimisations

jonludlam:


src/xref2/env.ml
```diff
@@ -655,22 +660,28 @@ let initial_env :
...
-          let m = module_of_unit unit in
-          let env = add_module (unit.id :> Identifier.Path.Module.t) m env in
+      | Import.Resolved (root, name) ->
+          let id = `Root(root, name) in
+          let m =
+            Component.Delayed.put (fun () ->
+              let unit = resolver.resolve_unit root in
+              module_of_unit unit)
+          in
+          let env = add_module id m env in
```
jonludlam, line 677: I'm not convinced this does anything useful right now aside from fill in the cache in `compilation_unit.ml` - I think root module lookup doesn't look in `env.elts` - @Julow would know for sure.

jonludlam: (There's a bit of an impedance mismatch here as `Env.t` is an immutable object, but the root module lookup is done on demand, which is why it's a bit odd)

Julow: I think we can remove this, the speed improvements is around 3%.
I tried to remove that code entirely (and the `imports` value) and got no diffs on Core's output. It seems this value is only used by the `depends` command ?

jonludlam: We certainly need to turn all the unresolved imports into resolved ones (or forward) for the `odoc html-deps` to work.

src/loader/cmti.ml
```diff
@@ -526,14 +526,20 @@ and read_module_type env parent label_parent mty =
...
-          | Tmod_ident(p, _) -> Alias (Env.Path.read_module env p)
+          | Tmod_ident(p, _) ->
+            Format.eprintf "typeof mpath\n%!";
+            TypeOf (MPath (Env.Path.read_module env p))
           | _ ->
-              let mty =
-                !read_module_expr env parent label_parent mexpr
-              in
-                ModuleType mty
+              match !read_module_expr env parent label_parent mexpr with
```
lpw25, line 533: I would have expected us to match on the original `mexpr` here, as with the `Tmod_ident` case above.

src/loader/cmti.ml
```diff
@@ -526,14 +526,20 @@ and read_module_type env parent label_parent mty =
         let decl =
           let open Module in
           match mexpr.mod_desc with
-          | Tmod_ident(p, _) -> Alias (Env.Path.read_module env p)
+          | Tmod_ident(p, _) ->
+            Format.eprintf "typeof mpath\n%!";
```
jonludlam, line 530: Oops, these should go.

src/loader/cmti.ml
```diff
@@ -526,14 +526,20 @@ and read_module_type env parent label_parent mty =
...
-              let mty =
-                !read_module_expr env parent label_parent mexpr
-              in
-                ModuleType mty
+              match !read_module_expr env parent label_parent mexpr with
+              | Signature [Odoc_model.Lang.Signature.Include {Include.decl = Alias p; _}] ->
+                Format.eprintf "typeof struct_include\n%!";
+                TypeOf (Struct_include p)
+              | Signature [Odoc_model.Lang.Signature.Include {Include.decl = ModuleType (Signature items); _}] ->
+                Signature items
```
lpw25, line 538: I'm not sure you need this special case. I think that:
```ocaml
module type S = sig include sig type t end end
```
and
```ocaml
module type S = module type of struct include struct type t end end
```
should be treated the same.

If we want to eliminate instances of `include sig ... end` then I think that should be done elsewhere, and you'll get the equivalent of this special case for free.

jonludlam:
Unfortunately @Julow's change to resolve references in signature items has made it tricky to use the Delayed module where this PR does. I think I'll try to fix it by pulling the doc item out of the module, or possibly by pushing the Delayed bit type inside.

src/loader/cmti.ml
jonludlam: I had a little go at matching the mexpr without converting it to a Lang first, and matching the case I care about seems a lot less legible to me:

```ocaml
match mexpr.mod_desc with
| Tmod_ident(p, _) ->
  TypeOf (MPath (Env.Path.read_module env p))
| Tmod_structure {str_items = [{str_desc = Tstr_include {incl_mod = {mod_desc = Tmod_ident (p, _); _}; _}; _}]; _}
| Tmod_structure {str_items = [{str_desc = Tstr_include {incl_mod = {mod_desc = Tmod_constraint ({ mod_desc = Tmod_ident (p, _); _}, _, Tmodtype_implicit, _) ; _}; _}; _}]; _} ->
    TypeOf (Struct_include (Env.Path.read_module env p))
| _  ->
  !read_module_expr env parent label_parent mexpr 
```

or by pulling in `unwrap_module_expr_desc` it becomes:

```ocaml
match mexpr.mod_desc with
| Tmod_ident(p, _) ->
  TypeOf (MPath (Env.Path.read_module env p))
| Tmod_structure {str_items = [{str_desc = Tstr_include {incl_mod = {mod_desc; _}; _}; _}]; _} -> begin
  match unwrap_module_expr_desc mod_desc with
  | Tmod_ident(p, _) ->
    TypeOf (Struct_include (Env.Path.read_module env p))
  | _ ->
    !read_module_expr env parent label_parent mexpr 
  end
| _  ->
  !read_module_expr env parent label_parent mexpr 
```

so I'm not sure we gain much.

jonludlam: (and pulling in unwrap_module_expr_desc means either doing another function reference like `read_module_expr`

```
let read_module_expr : (Ident_env.t -> Identifier.Signature.t -> Identifier.LabelParent.t -> Typedtree.module_expr -> ModuleType.expr) ref = ref (fun _ _ _ _ -> failwith "unset")
```

or cut-n-pasting the (small) function in.

lpw25: In terms of gain, I'm just much more confident that the above is doing the right thing. With the current code, I have to know all possible module expressions which could produce the given signature. In this case I think its fine, but I had to think about it quite a bit to be sure.

jonludlam: You're more familiar with `Typedtree` types than I am :-)

However, it does bring up the other point you raised with the special-case pattern. It came about because I was checking out what _didn't_ match the simple `Signature [Include (Alias p)]`, and trying to see where it came from. That one actually came up in [persistent_env.mli](https://github.com/ocaml/ocaml/blob/80326033cbedfe59c0664e3912f21017e968a1e5/typing/persistent_env.mli#L19-L21). It _is_ easier to see what's happening in that case (we get Tmod_apply, so `read_module_expr` just reads the signature out of `mod_type`), so let's go with the second of the suggestions above.

src/loader/cmti.ml
jonludlam: See comment above.

jonludlam:
Actually the change required following the references patch wasn't so bad, just passing in the doc explicitly to `Odoc_xref2.Env.add_module`

src/loader/cmti.ml
lpw25: Ah, that's what the other case was for! In which case, I suggest that instead of copying `unwrap_module_expr_desc` you copy [`path_of_module` from the compiler](https://github.com/ocaml/ocaml/blob/trunk/typing/typemod.ml#L1611).

jonludlam: Much nicer, and no copying required! It was only necessary for `unwrap_module_expr_desc` because of the circular dependency between `cmti.ml` and `cmt.ml`, which I think you had previously mentioned is a bit of an artificial split anyway.

Issue #388, opened 2019-09-24
Stack overflow on parsing a cmt


bluddy:
After getting past #387, I found that `ldap` encountered another issue with odoc. Trying to parse `ldap_matchingrules.cmt`, I get a stack overflow. This seems related to heavy object types.

To reproduce (my fork contains the correction for the issue in #387):

* `opam pin add ldap https://github.com/bluddy/ocamldap.git`
* `odig odoc`

I get the stack overflow:
```
'/home/yotam/.opam/4.08.1/bin/odoc' 'compile' '--pkg' 'ldap' '-o' '/home/yotam/.opam/4.08.1/var/cache/odig/odoc/ldap/ldap_matchingrules.odoc' '/home/yotam/.opam/4.08.1/lib/ldap/ldap_matchingrules.cmt' '-I' '/home/yotam/.opam/4.08.1/var/cache/odig/odoc/ldap/' '-I' '/home/yotam/.opam/4.08.1/var/cache/odig/odoc/ocaml/' '-I' '/home/yotam/.opam/4.08.1/var/cache/odig/odoc/pcre/'
odoc: internal error, uncaught exception:
      Stack overflow
      Raised by primitive operation at file "typing/ident.ml" (inlined), line 102, characters 20-40
      Called from file "typing/ident.ml", line 250, characters 14-46
      Called from file "src/model/ident_env.cppo.ml", line 262, characters 4-32
      Called from file "src/model/ident_env.cppo.ml", line 288, characters 29-47
      Called from file "src/loader/cmi.ml", line 381, characters 18-42
      Called from file "src/loader/cmi.ml", line 372, characters 17-39
      Called from file "src/loader/cmi.ml", line 375, characters 20-42
      Called from file "src/loader/cmi.ml", line 387, characters 27-49
      Called from file "src/loader/cmi.ml", line 493, characters 52-74
      Called from file "list.ml", line 92, characters 20-23
      Called from file "src/loader/cmi.ml", line 492, characters 8-117
      Called from file "src/loader/cmi.ml", line 385, characters 28-50
      Called from file "src/loader/cmi.ml", line 375, characters 20-42
      Called from file "src/loader/cmi.ml", line 375, characters 20-42
      Called from file "src/loader/cmi.ml", line 387, characters 27-49
      Called from file "src/loader/cmi.ml", line 493, characters 52-74
      ...
```

bluddy:
I'm getting these on more packages: links.0.9 and rotor.0.1

[...]

PR #443, opened 2020-07-06
xref2: Fix stack overflow on classes

Julow:
Fixes the small reproduction case in https://github.com/ocaml/odoc/issues/388

This fixes the crash but the output for classes is not great and should be
reworked.

The output for:

```ocaml
class foo =
  object(o)
    method test () = o
  end
```

is:

```ocaml
method test : < test : unit -> 'a; .. > -> unit -> 'a
```

### DIFF
--- a/src/loader/cmi.ml
+++ b/src/loader/cmi.ml
@@ -475,38 +475,44 @@ and read_row env _px row =
 and read_object env fi nm =
   let open TypeExpr in
   let open TypeExpr.Object in
-  match nm with
-  | None ->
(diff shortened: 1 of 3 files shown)


jonludlam:
Fixing the crash is definitely worth having, even if the output isn't what we'd like it to be. Also, it's better than you get if you put it in an mli file:

```ocaml
class foo : object ('a) method test : unit -> 'a end
```

ends up rendered as:

```
method test : unit -> 'a
```

where `` `a `` is not declared anywhere!

PR #444, opened 2020-07-06
xref2: Remove UnitName

Julow:
cc @jonludlam 

The first commit removes unecessary items from `Ref_tools.mli`


jonludlam:
That is smaller isn't it? :-) Thanks!

PR #445, opened 2020-07-06
xref2: Ambiguous reference warning

Julow:
Adds a warning when a reference is ambiguous:

```
Reference to 't' is ambiguous. Please specify its kind: val-t, type-t.
```

Adds the notion of "implicit warning" which may not be well defined/named and complicated error handling a bit in xref2.
The interface of `Find` needed to be changed.

### DIFF
--- a/src/xref2/link.ml
+++ b/src/xref2/link.ml
@@ -197,7 +197,11 @@ and with_location :
     Env.t ->
     a Location_.with_location ->
     a Location_.with_location =
- fun fn env x -> { x with Location_.value = fn env x.Location_.value }
+ fun fn env x ->
+  let value =
+    Lookup_failures.with_location x.location (fun () -> fn env x.value)
+  in
+  { x with value }
 
 and comment_docs env d = List.map (with_location comment_block_element env) d
 
--- a/src/xref2/link.ml
+++ b/src/xref2/link.ml
@@ -638,7 +638,7 @@ and type_decl : Env.t -> Id.Signature.t -> TypeDecl.t -> TypeDecl.t =
         in
         Format.eprintf "found hidden path: %a\n%!" Component.Fmt.resolved_type_path p';
         match Tools.lookup_type env p' with
-        | Ok (Found (`T t')) ->
+        | Ok (`T (_, t')) ->
             {
               default with
               equation =
@@ -650,7 +650,7 @@ and type_decl : Env.t -> Id.Signature.t -> TypeDecl.t -> TypeDecl.t =
                       params
                   with _ -> default.equation );
             }
-        | _ -> default )
+        | Ok (`C _ | `CT _ | `T_removed _) | Error _ -> default )
     | None -> default
   in
   (* Format.fprintf Format.err_formatter "type_decl result: %a\n%!"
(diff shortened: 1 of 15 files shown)


jonludlam:
This is failing on older OCamls because the exception printing is different:

```diff
# resolve_ref "X"
-Exception: Odoc_model.Error.Conveyed_by_exception _.
+Exception: Odoc_model__Error.Conveyed_by_exception _.
```

Julow:
I fixed this and rebased.

Julow:
Rebased again over the merge of https://github.com/ocaml/odoc/pull/442
There was a lot of conflicts, it may be useful to review again.

src/xref2/link.ml
```diff
@@ -205,6 +205,20 @@ and with_location :
     a Location_.with_location =
  fun fn env x -> { x with Location_.value = fn env x.Location_.value }
 
+(** Like [with_location] but allow to raise warnings. *)
+and with_location' :
```
jonludlam, line 209: Is there any reason to keep the non-warning-raising `with_location` above? Why can't everything use this function instead?

src/xref2/ref_tools.ml
```diff
@@ -589,51 +638,47 @@ and resolved2 (r, _) = resolved1 r
...
 
 let resolve_reference_dot_page env page name =
   L.in_page env page name >>= resolved1
 
 let resolve_reference_dot_type env ~parent_ref t name =
   Find.any_in_type t name >>= function
-  | `Constructor _ -> CS.of_component env parent_ref name >>= resolved1
-  | `Field _ -> F.of_component env parent_ref name >>= resolved1
+  | `Cs _ -> CS.of_component env parent_ref name >>= resolved1
+  | `Fd _ -> F.of_component env parent_ref name >>= resolved1
```
jonludlam, line 675: I do find these slightly _too_ terse - I don't mind abbreviations generally but I don't like mentally tripping over them when I don't find them really obvious. Perhaps we could use `Constr` and `Fld`, `Meth` and `Instv`? `Mv` as `InstanceVariable` I find particularly baffling.

Julow: I went a bit too far with abbreviations ! I used them to avoid confusion with paths/references/identifiers (and module name clash).
An alternative would be to use the full name with a prefix, eg. `F_constructor` in `Find` and `R_constructor` in `Ref_tools`. What do you think ?

src/xref2/link.ml
Julow: I preferred to use this function only when necessary to be able to change it easily later. It's adding yet an other "implicit" way of passing warnings and I'm not sure it's the right way.

jonludlam: One function is easier to change than two, and having only one is definitely less error-prone. If you strongly feel we should keep both, then we should distinuish between them with something better than `'`.

Julow: The part that must be able to change is `Lookup_failures` and "implicit" warnings added in `Error`, that use a global reference. This function is useful to mark where the side effects are done (instead of everywhere), to be changed easily later.
I changed its name.

src/xref2/ref_tools.ml
Julow: I changed to the `FConstructor` style.

PR #446, opened 2020-07-06
Update deps in opam file

Julow:
- `fmt` is required by `document`
- `mdx` should be a test-dependency
- I changed "dev" to "with-test" because they are necessary to run the tests with `opam install -t`


jonludlam:
Thanks!

PR #451, opened 2020-07-09
Improve xref2 tests

Julow:
Move `src/xref2/test` to the main `test` directory and change printf/expected file tests to use Mdx.
Tests still use printf with Mdx but future tests should be easier to write. `resolve` test looks better.


jonludlam:
Failing on 4.02.3: https://travis-ci.org/github/ocaml/odoc/jobs/706531068#L1490

dbuenzli:
Frankly I'd move on to remove support for 4.02.3, it had bugs in its handling of `cmt` files. 4.03 was the first sane release regarding. 4.02.3 was readded only because of bucklescript two years ago (see https://github.com/ocaml/odoc/pull/179), I hope they have managed to move on by now.

Julow:
I simply disabled that test on <4.03. Not much coverage is lost, especially if 4.02 is discouraged.

jonludlam:
We should probably let @ostera, @ryyppy, @rizo or other Reason folk let us know whether it's still necessary. Personally I'm finding it reasonably easy to continue supporting 4.02.3 right now, so I don't mind keeping it in - usually when Travis tells us it's broken there's been a relatively simple fix.

ryyppy:
Thanks for pinging! BuckleScript dropped 4.02.3 and upgraded to 4.06.1 for quite a while now, so I think we are fine here 👍

PR #453, opened 2020-07-16
Small set of patches for functors

Drup:
The output of functor in odoc is not terribly satisfying. A while ago, I moved the description of parameters to the header of the page, but that didn't make anything better, on the contrary. With the last few changes, we can improve things a little bit more.

This patch sets provides 2 improvements:
- Parameters are back inside the body of the page, inside a (single) header of level 1 (just below the main title). The `Signature` header is gone. As a result, parameters and their documentation are now in the sidebar outline.
- The output of functor declaration uses the short form when available. This applies to all backends and should preserve expansion/alternatives/etc.
Functors

### DIFF
--- a/src/document/ML.ml
+++ b/src/document/ML.ml
@@ -73,6 +73,7 @@ module ML = Generator.Make (struct
     let close_tag_semicolon = false
     let include_semicolon = false
     let functor_keyword = true
+    let functor_contraction = true
   end
 
   module Class =
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -1324,7 +1324,7 @@ struct
         let prelude =
           Item.Heading {
             label = Some "parameters" ;
-            level = 2 ;
+            level = 1 ;
             title = [inline @@ Text "Parameters"];
           }
           :: params
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -1244,8 +1244,7 @@ struct
           O.documentedSrc (
             O.keyword "module" ++ O.txt " " ++
             O.txt (Paths.Identifier.name arg.id) ++
-              O.txt Syntax.Type.annotation_separator ++
-              mty (arg.id :> Paths.Identifier.Signature.t) render_ty
+            mty_in_decl (arg.id :> Paths.Identifier.Signature.t) render_ty
           )
         | Some (prelude, items) ->
           let url = Url.Path.from_identifier arg.id in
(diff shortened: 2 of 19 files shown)


jonludlam:
In the expansion there's no longer a clear distinction between the parameters of the functor and the result, which I find confusing.

Before:
Screenshot 2020-07-23 at 13 40 17

After:
Screenshot 2020-07-23 at 13 40 44

jonludlam:
Using the short form looks nice though!

Drup:
Right, I'm still not sure what to do about that. My issue is that when people start adding sections for the signature part, the nesting is kind of a mess. Should the nesting be below the "Signature" heading or at the same level ? If people define their own structure afterwards, the rendering with a "Signature" heading isn't really fantastic.

We definitely need some separation here though. Maybe a good way would be to heuristically insert a heading when there is none ?

jonludlam:
Suggestion from @Drup:
> We should just decide that a document is segmented by an arbitrary number of named list of items
> in most signatures, it's "header" and "signature", but for functors, it would also have "parameters" (and for pages, we could > let users do whatever they like)

With this the structure of each section is independent and thus does not suffer the same problem as inserting a heading.

Drup:
So, I partially reverted some reorganization, and added back the "Signature" section. I didn't do the generalization of sectioning that I proposed. I think it's not urgent and we might want to wait a bit more to do it, when we really feel like we need it, and it's clearer what exactly we want to do.

Drup:
hmm, 1/2 CIs is happy, @jonludlam ? :)

jonludlam:
That ocaml-ci failure ought to have been fixed by @Julow's recently merged PR - if you rebase on that it ought to succeed I think!

jonludlam:
This needs a bit of a rebase after the recent 'expansion in module type expr' PR that was merged.

Drup:
Rebase is done! Wow, that was a huge pain. Test passes but I think should take a new look @jonludlam

jonludlam:
Sorry about that! Do the new types make more sense? IIRC this was what you were advocating for in discord.

Drup:
It doesn't change much for this patch. The patch is slightly simpler because you did some of the work that was done in the patch. I should take a shot at the "click on ... for expansion" feature to really exercise my change request. :) 

That being said, I have no clue whatsoever what umt/umty represents now.

jonludlam:
it was _supposed_ to be 'unexpanded module type', but I had to put the expansion on the `TypeOf` constructor anyway. Ho hum.

Issue #448, opened 2020-07-08
Signature comment ignored


Julow:
The top comment is not added to the env. The ref `L1` doesn't resolve but `L2` does.

```ocaml
(** {1:L1 Label 1} *)

(** {1:L2 Label 2} *)

(** {!L1} {!L2} *)
```

It seems the doc of signatures is lost when constructing `Lang`.
Maybe we could change `Lang.Signature.t` to `{ doc : Comment.docs; items : item list}` or add an artificial `Comment` item ?

PR #454, opened 2020-07-17
Don't ignore labels attached to signature items

Julow:
Fix https://github.com/ocaml/odoc/issues/448

This is especially useful for the top comments, which is attached to the compilation unit.

### DIFF
--- a/test/cases/labels.mli
+++ b/test/cases/labels.mli
@@ -1,0 +1,52 @@
+(** {1:L1 Attached to unit} *)
+
+(** {1:L2 Attached to nothing} *)
+
+(** {1:L3 Attached to module} *)
+module A : sig end
+
+(** {1:L4 Attached to type} *)
+type t
+
+(** {1:L5 Attached to value} *)
+val f : t
+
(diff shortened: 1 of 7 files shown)


test/cases/labels.mli
```diff
@@ -0,0 +1,52 @@
...
+(** {1:L7 Attached to class type} *)
+class type cs = object end
+
+(** {1:L8 Attached to exception} *)
+exception E
+
+type x = ..
+
+(** {1:L9 Attached to extension} *)
+type x += X
```
Julow, line 32: This case is not working yet.

test/cases/labels.mli
```diff
@@ -0,0 +1,52 @@
...
+
+(** {1:L10 Attached to module subst} *)
+module S := A
```
jonludlam, line 35: This is causing a failure in older ocamls:

https://travis-ci.org/github/ocaml/odoc/jobs/709270578#L1249

Julow: Fixed, running the test on `>4.08` only.

PR #455, opened 2020-07-17
Refactorize the CLI code

Drup:
This is an attempt at bringing sanity to the CLI code.

The cli code contains a staggering amount of duplication. On top of that, a surprisingly large amount of HTML logic is disseminated everywhere in the code. This set of patches attempts (!) to put back HTML logic where it belongs (in `html/`) and to factorize the maximum amount of logic between the backends.

On the way, I remove the `--index-for` option and split the `generate` and `target` commands into one command per backend. I'm fairly certain I broke some latex and `generate` bits, but since there are no tests, I'm not sure.

Because of the new model, the situation with `link`, `generate` and `process` still has some duplication. I hope @jonludlam can reduce everything further. For now, I push this out there so that people can look at it and provide feedback/contribute.


jonludlam:
This is already looking much nicer, thanks! I'll give it a whirl and see if I can spot anything broken.

Drup:
Something I forgot to mention: html fragments are still an exception to the structure. Ideal, they should receive the same treatement as all the other output modes, but I don't understand the code enough to craft proper link/generate/target functions.

jonludlam:
This has broken odig as `html-targets` no longer accepts '-I' as an argument. It hasn't been using the argument since the new model was merged -- which means, I think, that the command is not doing exactly what it's supposed to be doing -- but this hasn't caused any actual problems as far as I'm aware. I think we should at least accept the argument and throw it away to avoid unnecessarily breaking odig.

dbuenzli:
I don't think it should be hard to update `odig` to cope with this. But before I blindly start updating odig:

> I think, that the command is not doing exactly what it's supposed to be doing -- but this hasn't caused any actual problems as far as I'm aware.

it would still be good to know why this was necessary before (if it really was) and no longer is.

Drup:
I would also move towards removing all the cruft we can, and not hesitate to break the CLI to start on better foundations.

We already changed everything, it's as good as a time we will ever have.

jonludlam:
Odig will certainly need to change to support the new 3-phase process (compile, link, generate), but I want to be _absolutely sure_ that we've got it right before we start changing odig / dune / anything else - I don't want to be saying one thing now and then changing my mind and doing something else later. Just as an example of one possibility that the current performance issue I'm working on might suggest - we might want the linking done of dependent packages before we do the compilation step, which right now is unnecessary. So I totally agree that we should be aiming for better foundations, but let's make sure they're the correct foundations. Until we've got that sorted though, we can at least avoid totally breaking things.

jonludlam:
Just a little more on what specifically `html-targets` is doing.

The idea is to run this command on the odoc file and it will tell you what files it's going to produce. Since the new model landed, at the point this is run (after the `compile` step), we haven't finished processing the modules and can't give an exact answer - most importantly, for the top-level modules alias modules that are used for namespacing, the modules aliases are _not_ expanded at this stage, so the command will show far fewer output files than will actually be produced. We _could_ fix this by making `html-targets` actually do the link, like the `html` command itself does, which would require the `-I` arguments to work correctly. @dbuenzli - could you comment on what the effect would be of these two behaviours?

lpw25:
> we might want the linking done of dependent packages before we do the compilation step

Just wanted to point out that the reason for having the linking step is that references can create arbitrary cycles, so you need to be able to compile everything without linking anything, and linking needs to only depend on the results of compilation.

For `html-targets`, I think it has only ever really made sense in the world where linking is a separate step. i.e. the (hypothetical?) `generate-targets` command makes sense and `html-targets` is dubious. Broadly, we should aim to get rid of the old `html` and move everyone onto `generate`.

Drup:
I would rather suggest avoid having multiple overlapping commands, and remove/replace "not deprecated but you shouldn't use it" commands.

Personally, I'm still not quite sure why we need those "targets" commands (and the fact they are not documented doesn't help). I tried to reproduce their behavior in this PR, but that's about it. If we are to seriously discuss CLI design, I really would like to have a concrete example of a small project (2 modules) and the "ideal" succession of commands that is supposed to build its documentation, and the effect of each command (both semantically, and in term of file read/creation).

lpw25:
> I'm still not quite sure why we need those "targets" commands

Many build systems do not like commands where they doesn't know in advance which files it will produce. Since it is not possible to predict this information for `generate`, the tool provides a way of getting the information.

The fact that the targets command is broken for `html` forces the Jenga rules for odoc to wrap `odoc` in something which generates files in a tempory directory and then `tar`s them up, we then need a process outside of Jenga to un`tar` them for use -- its not great. If `targets` worked we could instead just tell jenga which files we were going to produce.

dbuenzli:
> Personally, I'm still not quite sure why we need those "targets" commands (and the fact they are not documented doesn't help)

The targets commands are here to tell you which files exactly are written by  an invocation of the tool. Your incremental build system needs to know that. 

There are a basically two ways to go around this:

1. The tool outputs are simple enough to be predictable from the cli itself. In that case you don't need such a command, only folk knowledge. For example if I compile `gcc -c a.c -o a.o `  I know `a.o` is going to be produced and that's all. Note that some invocations like `ocamlc -c` are only *seemingly* simple: what an invocation of `ocamlc` is going to produce depends on files that are in the `cwd`.
2. The tool outputs are not simple enough to be predictable from the cli itself.

In case 2. there are 3 ways to go around this: 

1. You provide another command that given an invocation of the tool tells you what is going to be produced. 
2. The command invocation also outputs the files that were actually written to a manifest file e.g. `--manifest FILE`.
3. You do nothing and let the build systems do a file system diff to understand what the tool actually produced (very annoying see @lpw25's response). 

Whether you want to go the 1. or 2. route depends on your build system (some may not support knowing produced files after the fact) and performance, e.g. if the work for 1. is expensive and needs to be redone when you invoke the real command then 2. is a better option.

Personally I'm fine with either 1. + 2. as long as it is *accurate*.

Drup:
As @jonludlam pointed out, I feel like given the way odoc works, I feel like the only way to precisely output all the files emitted, odoc has to do more than half the work. 

Maybe we should consider moving to solution 2. It's also much easier to implement, and will always be more accurate and less brittle.

jonludlam:
From the output of the link step, it's very easy to produce an exact list of the files that will be produced. This is definitely the way forward. The only question I think is how much effort is put into making older odig / any other tools work - we can make it twice as slow but correct, or quick but not showing the complete list. In my testing odig hasn't complained that we produce too many files (though it does if we don't produce enough, as does dune, hence we are creating 'stub' html files for hidden modules right now -- https://github.com/ocaml/odoc/pull/455/files#diff-f11cbd99d34fce8f31cd4275cdc3576aR28-R36)

dbuenzli:
> In my testing odig hasn't complained that we produce too many files

It won't complain but subsequent odig run will be problematic then, these files will be missing since they won't have been cached.

> The only question I think is how much effort is put into making older odig / any other tools work 

Speaking for `odig`, none. I don't mind moving `odig` in lockstep with `odoc`, just coordinate with me a bit upfront (and provide nice docs on how to operate `odoc` :-)

PR #459, opened 2020-07-24
Remove 402 support

ryyppy:
This will update the esy configuration to ocaml 4.06 (dropping 4.02.3 support).

@jonludlam can you verify if this unbreaks your CI tests?

Maybe @aantron / @ulrikstrid could have a quick look at this as well, since they have been involved in the esy setup afaikt?


ryyppy:
Relevant changes are in the `package.json` file, as stated [here](https://github.com/ocaml/odoc/pull/459/commits/10a01af17365801953b6ccdc356d6a7c2e608d67)

jonludlam:
Thanks @ryyppy - nothing's broken, I just want to make sure we're testing what people are using!

PR #463, opened 2020-07-30
xref2: Remove Tools.ResolvedMonad

Julow:
As @jonludlam noticed, the Unresolved case is not much used and always correspond to the input value.
Use `result` instead, which will be convenient to show better errors.

### DIFF
--- a/src/xref2/compile.ml
+++ b/src/xref2/compile.ml
@@ -19,7 +19,7 @@ let type_path : Env.t -> Paths.Path.Type.t -> Paths.Path.Type.t =
     let cp = Component.Of_Lang.(type_path empty p) in
     match Tools.resolve_type_path env cp with
     | Ok p' -> `Resolved (Cpath.resolved_type_path_of_cpath p')
-    | Error _ -> Cpath.type_path_of_cpath cp
+    | Error _ -> p
 
 and module_type_path :
     Env.t -> Paths.Path.ModuleType.t -> Paths.Path.ModuleType.t =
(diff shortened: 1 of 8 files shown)


src/xref2/compile.ml
```diff
@@ -29,8 +29,8 @@ and module_type_path :
   | _ ->
     let cp = Component.Of_Lang.(module_type_path empty p) in
     match Tools.resolve_module_type_path env cp with
-    | Resolved p' -> `Resolved (Cpath.resolved_module_type_path_of_cpath p')
-    | Unresolved p -> Cpath.module_type_path_of_cpath p
+    | Ok p' -> `Resolved (Cpath.resolved_module_type_path_of_cpath p')
+    | Error _ -> Cpath.module_type_path_of_cpath cp
```
jonludlam, line 33: I think this is just `p` right?

src/xref2/tools.ml
```diff
@@ -703,63 +673,62 @@ and resolve_module :
...
-        | Unresolved func_path', Resolved (arg_path', _) ->
-            Unresolved (`Apply (func_path', `Resolved arg_path'))
-        | Resolved (func_path', _), Unresolved arg_path' ->
-            Unresolved (`Apply (`Resolved func_path', arg_path'))
-        | Unresolved func_path', Unresolved arg_path' ->
-            Unresolved (`Apply (func_path', arg_path')) )
-    | `Identifier (i, hidden) as unresolved ->
-        of_option ~unresolved (Env.(lookup_by_id s_module) i env)
+            | Ok (p, m) -> Ok (p, Component.Delayed.put_val m)
+            (* | Error _ -> *)
```
jonludlam, line 706: Let's drop these lines rather than comment them out.

src/xref2/compile.ml
Julow: Fixed.

jonludlam:
Looks much better, thanks!

PR #466, opened 2020-07-31
xref2: Remove some Name.of_string

Julow:
This is on top of https://github.com/ocaml/odoc/pull/445

I think names shouldn't be created outside of `Lang`.
Some are still remaining and could be removed with some work. (eg. constructors, fields and extension constructors don't have names in `Lang`)

### DIFF
--- a/src/xref2/find.ml
+++ b/src/xref2/find.ml
@@ -81,7 +81,7 @@ let filter_in_sig sg f =
 
 let module_in_sig sg name =
   find_in_sig sg (function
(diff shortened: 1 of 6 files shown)


jonludlam:
This looks good! Are you planning to remove the remaining ones soon? If not I think we can merge this one now, as it's certainly an improvement.

Julow:
I think we can merge this now. The remaining ones needs a bit more work (eg. constructor and field names are not in signatures).

PR #468, opened 2020-08-04
Eliminate opaque warning with xref2 test

dra27:
Toploop is called, so this library can only be used in bytecode.


jonludlam:
Thanks!

PR #469, opened 2020-08-04
Add duniverse target

dra27:
`make -j duniverse` assembles the basic dependencies of odoc in the tree (including dune) and further invocations will use this duniverse.

Not all the targets work, since Dune's `@` targets propogate a bit too readily.

Ultimately, being able to compile the odoc binary with its dependencies is useful in the OCaml (manual) tree - but I'm parking this here for now in case its useful, and because odoc doesn't build with OCaml trunk at the moment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,15 @@ _esy
 
 # Generated docs
 docs/
+
+# Duniverse
+astring/
+cmdliner/
+cppo/
+dune-local/
+fmt/
+fpath/
+ocaml-re/
+result/
+tyxml/
+uutf/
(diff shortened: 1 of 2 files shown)


Drup:
That seems like a good idea! I'm not sure we need all those Makefile targets though, but I do not have any experience with duniverse. I'll let you (and @Octachron ?) work out the details and tell us when you are happy with it.

dra27:
Which targets would you like me to drop 😄 It was annoying not to be able to use a pattern rule for `%/dune-project: %`, but that only adds 4 extra lines!

jonludlam:
This seems fine to me - happy to merge unless @Octachron has any objections.

Octachron:
I don't have any specific objections. Note that I currently build the manual with a 4.12+trunk on a regular opam switch with some pinned package which works fine currently.

Issue #464, opened 2020-07-30
xref2: Comments attached to extension declaration are ignored


Julow:
In this example, `Lext` doesn't resolve:
```ocaml
(** {1:Lt} *)
type t = ..

(** {1:Lext} *)
type t += A (** {1:LA} *) | B (** {1:LB} *)

(** {!Lt} {!Lext} {!LA} {!LB} *)
```

PR #470, opened 2020-08-14
Check comment attached to extensions

Julow:
Fix https://github.com/ocaml/odoc/issues/464

### DIFF
--- a/src/xref2/env.ml
+++ b/src/xref2/env.ml
@@ -639,18 +639,13 @@ let rec open_signature : Odoc_model.Lang.Signature.t -> t -> t =
             add_module_type t.Odoc_model.Lang.ModuleType.id ty env
         | Odoc_model.Lang.Signature.Comment c -> add_comment c env
         | Odoc_model.Lang.Signature.TypExt te ->
-            let parent =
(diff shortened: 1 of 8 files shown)


src/xref2/env.ml
```diff
@@ -631,17 +631,26 @@ let rec open_signature : Odoc_model.Lang.Signature.t -> t -> t =
...
         | Odoc_model.Lang.Signature.ModuleType t ->
             let ty = module_type empty t in
             add_module_type t.Odoc_model.Lang.ModuleType.id ty env
         | Odoc_model.Lang.Signature.Comment c -> add_comment c env
         | Odoc_model.Lang.Signature.TypExt te ->
+            let parent =
+              match te.L.Extension.constructors with
+              | [] -> assert false
+              | { id = `Extension (parent, _); _ } :: _ -> parent
+            in
```
Julow, line 646: Parent should be the current signature. Unfortunately, the identifier is not available in this scope.

jonludlam: We could add the parent to `Lang.Extension.t` as we have with `Include.t`

jonludlam:
Looks better I think!

PR #472, opened 2020-08-18
document: support documentation and linking of type extensions

CraigFe:
This adds support for documenting type extensions, as in:

```ocaml
type extensible +=
   | Extension (** Documentation for {!Extension}. *)
   | Another_extension (** Documentation for {!Another_extension}. *)
```

which gets rid of [a `TODO` in the document generator](https://github.com/ocaml/odoc/blob/master/src/document/generator.ml#L563). The approach taken is symmetric with that of variant cases.

### DIFF
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -560,9 +560,18 @@ struct
 
 
   let extension_constructor (t : Odoc_model.Lang.Extension.Constructor.t) =
-    (* TODO doc *)
-    let doc = constructor (t.id :> Paths.Identifier.t) t.args t.res in
-    O.documentedSrc (O.txt "| ") @ doc
+    let id = (t.id :> Paths.Identifier.t) in
+    match Url.from_identifier ~stop_before:true id with
+    | Error e -> failwith (Url.Error.to_string e)
+    | Ok url ->
+       let anchor = Some url in
+       let attrs = ["def"; url.kind] in
+       let code =
+         O.documentedSrc (O.txt "| ")
+         @ constructor id t.args t.res
+       in
+       let doc = Comment.to_ir t.doc in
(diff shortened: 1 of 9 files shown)


CraigFe:
Here's an example of the HTML output:

image

and with one of the cases anchored:

image

<hr/>

Here's the troff output rendered with `man`:

image

src/document/generator.ml
```diff
@@ -560,21 +560,27 @@ struct
...
   let extension_constructor (t : Odoc_model.Lang.Extension.Constructor.t) =
-    (* TODO doc *)
-    let doc = constructor (t.id :> Paths.Identifier.t) t.args t.res in
-    O.documentedSrc (O.txt "| ") @ doc
+    let id = (t.id :> Paths.Identifier.t) in
+    match Url.from_identifier ~stop_before:true id with
+    | Error e -> failwith (Url.Error.to_string e)
+    | Ok url ->
+       let anchor = Some url in
+       let attrs = ["def"; url.kind] in
```
CraigFe, line 568: Not sure if we want other `attrs` here. The variant case uses `def variant`, but I'm not sure it affects any of the renderers.

Drup: Let's keep it so far, it's easy to change if we need more.

Drup:
Nice work! The code looks good. Let's merge as soon as CI is happy.

Drup:
> The CI failure seems to be because it doesn't install test dependencies.

Is that new ? That's a bit problematic ...

jonludlam:
Seems to have been a change in the environment - according to travis `master` passed the tests 14 days ago, and then failed `13` days ago with the same HEAD changeset.

jonludlam:
Travis says OK finally :-)

PR #473, opened 2020-08-18
CI: Install test dependencies

Julow:
It seems to be because of this change: 2025b1abb59fc40c878d6f34c8290cd22fe03e46


Julow:
This PR fixes the `opam install` command and the cyclic dependency on `mdx`.
The problem now is that Travis' cache is not invalidated when changing the Opam file. This is why this problem happens only now.
Trying to fix that.

PR #480, opened 2020-09-03
Minor fixes

jonludlam:
Several small fixes.


Julow (commented):
`src/xref2/scratch.md` is not uptodate. (it is checked by `dune runtest`, which the CI will use instead of the Makefile)

jonludlam:
Ah, thanks! I was only using 'make test'

Julow (approved):
Looks good to me.

PR #481, opened 2020-09-08
Disallow unresolved arguments to Apply

jonludlam:
As discussed in the previous dev meeting. This fixes the expansion of a few modules, though it's not entirely clear why. We may well want to relax this condition in the future as many functor bodies can be expanded even if the argument is unresolved, as it's usually the case that the result is independent of the argument.


Drup:
Looks good, at least for now as you say! :+1:

jonludlam:
https://github.com/ocaml/odoc/issues/483

PR #484, opened 2020-09-16
latex backend: minor rendering tweaks

Octachron:
This PR implements three small tweaks on the latex backend:

1. the description of signature items is now indented.
> val foo: int
>     The documentation is indented
>     Compared to the item itself      

2. Small tables uses now paragraph columns: overfull cells are rendered with a line break rather than overflow

3. Internal references use a monospace fonts.


Drup (approved):
Collection of minor changes, it looks good, let's merge!

PR #486, opened 2020-09-17
Travis: Fix Mdx dependency

Julow:
`opam install --deps-only -t` won't install it, probably because it would install the released version of Odoc.


jonludlam:
Looking better - just the esy builds to fix now!

Julow:
I suspected the `esy.lock` needed to be re-generated but I failed to do that yet. Do you have a working npm/esy to do that ?

PR #487, opened 2020-09-17
Module type expr expansion2

jonludlam:
Move expansions to ModuleType.expr instead of maintaining it separately.

This means less chance of duplicating effort (e.g. FunctorParameter appeared in expansions as well as ModuleType.expr), and simplifies some other types (e.g. no more AlreadyASig)

Also included in this PR are some bugs fixes and commits to tidy the output a bit.


Julow (commented):
There is some diffs in the generated HTML. For example, these values:

```ocaml
(* In Base.Array: *)

val sum : (module Base__Container_intf.Summable with type t = 'sum) -> 'a t -> f:('a -> 'sum) -> 'sum

(* In Base.Float: *)

module Parts : sig ... end with type Parts.outer := t
```

Are now rendered like this:

```ocaml
val sum : (module ... ) -> 'a t -> f:('a -> 'sum) -> 'sum 

module Parts : sig ... end
```

The `...` are not links.

jonludlam:
Yes, both of these are intentional. The second is justified because the `with` clause tells you nothing, and the first is justified because the link to the module is unavailable. Really we ought to emit a warning for the first, because there's no way to find out what the signature of the module is.

lpw25:
The second one makes total sense: that `with` really does tell the user nothing. However the first-class module one contained useful information: whilst odoc doesn't know where that `Summable` module type is the user might already know what it is.

Julow:
Indeed, `Summable` is not a link in the first example (it is a "unresolved" link) but I think it's still useful to show it, the type doesn't make sense without.

jonludlam:
Right - we need to make sure we emit a sensible warning. I'll revert that part.

jonludlam:
OK, rebased, removed the commit that hid the hidden package paths, and updated the Lang printer definitions

Julow (approved):
There is a diff like this in Base:
```diff
-val fold_until : 'a t -> init:'b -> f:('b -> 'a -> ('b, 'final) Base__Set_intf.Continue_or_stop.t) -> finish:('b -> 'final) -> 'final
+val fold_until : 'a t -> init:'b -> f:('b -> 'a -> ('b, 'final) Base__Container_intf.Export.Continue_or_stop.t) -> finish:('b -> 'final) -> 'final
```

I think this is expected (there is no canonical annotation, the same path was resolving the same way in the module that includes this).

PR #488, opened 2020-09-22
OCaml 4.12 support: injectivity annotation

### LINKED ISSUE #467: 4.12 injectivity annotations aren't (yet) supported
https://github.com/ocaml/ocaml/pull/9500 adds support for injectivity annotations on abstract types - at present odoc can't build with trunk OCaml, therefore, since `Asttypes.variance` changed.

Octachron:
This PR adds support to odoc for the new injectivity annotations in 4.12 (note that odoc requires a patched dune and topkg to build on 4.12) with some tests.

Close #467

### DIFF
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -687,16 +687,18 @@ struct
     : 'row. ?delim:[`parens | `brackets] -> Odoc_model.Lang.TypeDecl.param list
     -> text
   = fun ?(delim=`parens) params ->
-    let format_param (desc, variance_opt) =
-      let param_desc =
+    let format_param {Odoc_model.Lang.TypeDecl.desc; variance; injectivity} =
+      let desc =
         match desc with
-        | Odoc_model.Lang.TypeDecl.Any -> "_"
-        | Var s -> "'" ^ s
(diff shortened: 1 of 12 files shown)


jonludlam:
Great! thanks, @Octachron 

Unfortunately travis has failed on 4.09 and 4.10 - this is because you've added the tests to 'recent.mli', which will currently try to build on anything later than 4.08. @Julow is working on a better test suite that we'll integrate ASAP, so I suggest rather than adding a 'more_recent.mli' for >4.11, we just change the bounds on all the 'recent' tests so they are 4.11. 

https://github.com/ocaml/odoc/blob/master/test/html/test.ml#L304
https://github.com/ocaml/odoc/blob/master/test/man/test.ml#L242

Octachron:
Since the test can only works on 4.12, testing on 4.11 cannot work. If we want to merge this PR before the 4.12 release (which would slightly the process to build the OCaml's stdlib documentation on trunk), I propose to split the test in a separate PR to be merged after the 4.12 release.

jonludlam:
That would also work, yes - and I meant >4.11 in my previous comment of course :-)

PR #489, opened 2020-09-23
Fix `module type of` expansions

jonludlam:
OCaml expands `module type of` expressions early, and subsequent operations -- such as substitution, functor application and so on -- operate on the expansion rather than the `module type of` expression. Before this PR, odoc did not do this, which led to differences in expansions. For example, on the following code:

```ocaml
module type S = sig
  module X : sig type t end
  module type Y = module type of X
end
module X1 : sig type t type u end
module type T = S with module X := X1
```

OCaml produces

```ocaml
module type T = sig module type Y = sig type t end end
```

whereas odoc was producing

```ocaml
module type T = sig module type Y = sig type t type u end
```

This PR fixes this problem by expanding all `module type of` expressions present in the current cmt/cmti before doing anything else, in a kind of 'zeroth' pass. All subsequent operations use this expansion. The zeroth pass occasionally needs to be done more than once if a `module type of` expression includes another `module type of`, e.g.:

```ocaml
module type X = sig type t end
module X0 : sig
  type t = int
  type t2 = int
end

module type Y = functor ( X : X ) -> sig
  module type Y = module type of struct include X end
end

module Z : Y
module Z2 : Z(X0).Y
module type Z3 = sig include module type of Z2 end
```

in this case, we notice that the expansion of `module type Z3` hits an unexpanded `module type of` expression during the first pass, and so we do another pass, by which time `module type Y` has had its expansion calculated.

Additionally, this PR also attempts to remove `module type of` expressions when they are no longer correct - e.g. in the above first example the expansion of `T` will not include a `module type of` expression, since no expression of this form can accurately describe what the result is. Instead it is replaced with a `Signature`.

### DIFF
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -1434,8 +1434,8 @@ struct
     | With (_, umt) -> extract_path_from_umt ~default umt
     | Path (`Resolved r) ->
       (Paths.Path.Resolved.ModuleType.identifier r :> Paths.Identifier.Signature.t)
-    | TypeOf (ModPath (`Resolved r)) 
-    | TypeOf (StructInclude (`Resolved r)) ->
+    | TypeOf { t_desc = ModPath (`Resolved r); _ } 
+    | TypeOf { t_desc = StructInclude (`Resolved r); _ } ->
       (Paths.Path.Resolved.Module.identifier r :> Paths.Identifier.Signature.t)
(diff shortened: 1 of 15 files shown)


jonludlam:
The CI failure is some odd esy problem, seems unrelated to the PR. Merging!

PR #490, opened 2020-09-23
Test using ocaml-ci

jonludlam:
As far as possible this PR is to fix the build on ocaml-ci.

It breaks the circular dependency between mdx and odoc by vendoring mdx via a submodule. This is not necessary for dev work - you won't have to `git submodule init`, so existing dev workflow is unchanged.

At some point it might make sense to pull out the parser as a separate repo again (octavius style) which will break the dependency cycle in a better way.

Additionally this PR brings our dune files up to date and adds an (as yet unused) ocamlformat file.


src/xref2/dune
```diff
@@ -7,39 +7,39 @@
...
-(alias
- (name   runtest)
+(rule
+ (alias   runtest)
  (deps   (:x scratch.md)
          (package odoc))
  (enabled_if (> %{ocaml_version} 4.08))
  (action (progn
-           (run ocaml-mdx test %{x})
+           (run ocaml-mdx-test %{x})
```
Julow, line 43: I wonder if there is a way to fix this in Mdx ? The problem is that the binary `ocaml-mdx` depends at runtime on the binary `ocaml-mdx-test`. Is here a way to declare this to Dune ?

Julow (commented):
You included Mdx at master, shouldn't we use 1.7.0 ?

I can't reproduce the error on `lint-doc`. On my machine, Dune builds the current Odoc before building the doc.
It would still not work however because of the many warnings generated since the new model.

src/xref2/dune
jonludlam: not yet - see ocaml/dune#3499

jonludlam:
The doc-lint issue is this: https://github.com/ocurrent/ocaml-ci/issues/250

jonludlam:
I initially used mdx 1.7.0 but had issues with ocamlformat - let me try putting it back to 1.7.0 to see what happens.

jonludlam:
Looking good - I think the dune `vendored_dirs` thing fixed it.

Julow (approved):

PR #492, opened 2020-09-24
Printer for Lang values

Julow:
This PR adds a printer for values of `Lang` and `Paths`.

~~The part in `test/projects` is an attempt at writing a project-like test and to use this tool for printing out `.odoc` files~~
It can be called like this: `dune exec -- test/odoc_print/odoc_print.exe <.odoc file>`. It will be easier to call it from inside tests.

Every PPXes I tried couldn't work because of recursive modules, polymorphic variants and "re-exported" types. I wanted to only add test-dependencies, which didn't help here.

This is implemented with a very simple GADT, that describes a type and allow to traverse values of that type.
The advantages of this are:
- The printer is defined separately, is 30 lines of code and work consistently on every types.
- Type description are easy to write and to maintain.
  This is not true for some of our huge polymorphic variants, which required a lot of work and some headaches.

This tool can only be used to write printers (eg. sexp, json) but not to write parser.


Drup:
I don't think this should be in `test/`. It should be directly accessible, as a library `odoc-serialize` for instance, and a serialization format for the various bits. We could consider having it for document IR too.

jonludlam:
Agreed that it shouldn't be in test - obviously that's what has motivated this, but it's potentially more generally useful.

Julow:
I moved it to the library `model_desc`.

jonludlam:
OK, if we generalise it to the document IR we might rename it (or split it) later. Is there any other reason this is a draft PR, or can we merge it?

Julow:
I think we can merge it. The reason was that it's not used yet.

jonludlam:
Great! In it goes :-)

PR #493, opened 2020-09-25
Add missing test-dependency on Yojson

Julow:
Yojson is used to print Lang values to a format that can be post-processed.


jonludlam:
Not sure why this didn't fail in the first PR.

PR #494, opened 2020-09-25
Simplify/clarify the handling of raw markup and add support in manpages.

Drup:
With this change, the frontend of odoc doesn't know anything about raw markups anymore. Raw markup is just a pair of a tag designating the language, and some .... raw markup. Each backend will look at the tag and decide for itself what to do with it.

Later on, we can have more sophisticated handling (like HTML backend handling latex raw markup through mathjax, for instance).

So far, the following tags are handled:
- Html backend handles `html` tag
- Latex backend handles `latex` and `tex` tags
- manpage backend handles `manpage`,`roff` and `troff` tags.

### DIFF
--- a/src/html/dune
+++ b/src/html/dune
@@ -13,5 +13,5 @@ let () =
  (name odoc_html)
  (public_name odoc.html)
     |} ^ preprocess ^ {|
- (libraries odoc_model odoc_document tyxml))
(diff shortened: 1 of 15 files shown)


jonludlam:
Looks good thanks!

PR #495, opened 2020-09-25
Improve handling of warnings

Julow:
Fix an uncaught exception with `--warn-error` and improve the code:

- Cleaner way of handling `--warn-error`
- Remove exceptions from the loader API. Handling of errors and warnings is cleaner and fits better with code for the first point.
- Remove `shed_warnings`. This means that all warnings are accumulated before an error is raised.


jonludlam:
I don't think we're finished yet with the improvements to Error.ml - but this is an improvement so let's have it.

PR #496, opened 2020-09-30
Don't turn some lookup failures into warnings

Julow:
(on top of https://github.com/ocaml/odoc/pull/495)

With this PR, failures when looking up root modules are not turned into warnings.
This is to allow building the doc for Odoc with warnings turned into errors (the CI enables that).

There is now two kinds of "lookup failures": "Root" (which won't be turned into fatal warnings) and "internal" (which will).


jonludlam:
Once this passes `lint-doc` we'll merge.

Julow:
The remaining errors were from Link. I added a commit to share error handling between Compile and Link, repurposing the `Errors` module.

PR #502, opened 2020-10-14
Cram tests for the model

Julow:
The first commits improve `odoc_print` to correctly handle different kinds of files (page or compilation unit) and change the encoding of variants so it's easier to query.
This PR adds two tests for the new model using Dune's cram feature.

There is a new test-dependency: `jq`. It is used to filter the output of `odoc_print` and to show only the parts we are interested in.
Without it, a simple change to the model would break every tests.


jonludlam:
Ah! I hadn't noticed you had rebased! merging now.

Julow:
Sorry I forgot to notify you

Issue #505, opened 2020-10-16
Record fields and constructors are no longer indented


lubegasimon:
`Record` fields and `constructors` are no longer indented as of now (in `master`) as compared to the current release version (`1.5.1`).

 **_current state:_**

![Screenshot_2020-10-16 Renderer (odoc Odoc_document Renderer)](https://user-images.githubusercontent.com/34124177/96220985-6e978e00-0f92-11eb-8668-222775e6adf5.png)

**_former state:_**

![Screenshot_2020-10-16 Tree (odoc Odoc_html__ Tree)](https://user-images.githubusercontent.com/34124177/96220664-ceda0000-0f91-11eb-90c9-3d7b2948090f.png)

lubegasimon:
closed in favor of https://github.com/lubegasimon/odoc/commit/94d975f635227ec905298b337ce370a7b6389045

Issue #506, opened 2020-10-16
Record fields and constructor are linkable.


lubegasimon:
`Record` fields and `constructors` show `#` when hovering over them, thus become linkable.

<img width="448" alt="Screenshot 2020-10-14 at 22 41 46" src="https://user-images.githubusercontent.com/34124177/96221893-0053cb00-0f94-11eb-8814-6c924bd90f4c.png">

trefis:
I'm not sure I understand this issue (and I'm not fluent enough in CSS to understand its fix).
But fields and constructors are linkable on purpose, so I'm not sure what the issue is (apart that the rendering is a bit ugly).

lubegasimon:
> But fields and constructors are linkable on purpose

Ah! Probably I missed the rationale of them being linkable, could you please explain why?

trefis:
Err, nothing in particular, it's just useful to be able to refer to them, in the same way that it's useful to refer to any other item in the page.
Why shouldn't they be linkable?

lubegasimon:
> Why shouldn't they be linkable?

Because they're not referenced anywhere, except to themselves, however, to me, the need for them could be when sharing the link with another party, which I find reasonable to share the link to the type declaration in which they're difined, rather than individual fields or constructors.

Drup:
As @trefis said, this is intentional. In general, odoc's credo when it comes to these things is: "Everything that can possibly be linkified should be". This comes from a lot of experience with ocamldoc where people complained not being able to give links to every imaginable pieces of documentation. In any case, this doesn't hurt anyone. :)

lubegasimon:
Cool, I can revert that change.

PR #507, opened 2020-10-16
Fix record fields and constructors indentation

lubegasimon:
Addressing #505.

Screenshot_2020-10-16 Renderer (odoc Odoc_document Renderer)(1)

### DIFF
--- a/src/odoc/etc/odoc.css
+++ b/src/odoc/etc/odoc.css
@@ -543,10 +543,6 @@ div.def table {
 }
 
 td.def {
-  padding-right: 2ex;
-}
-
-.record td.def {
   padding-left: 2ex;
 }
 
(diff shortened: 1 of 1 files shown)


Drup:
The indentation fix is great :+1: . I'll merge as soon as you drop the linkification change (as we discussed in #506 ).

Drup:
Also, as a small aside: if we decide to remove pieces of markup, I prefer if to actually remove them, instead of just turning them invisible. ;)

lubegasimon:
> Also, as a small aside: if we decide to remove pieces of markup, I prefer to actually remove them, instead of just turning them invisible. ;)

You're right, for performance reasons.

Drup:
> You're right, for performance reasons.

Actually, no: for software engineering reason: code that is used to produce unneeded stuff is dead, and dead code is technical debt, which leads to unmaintainable messy software.

Drup (approved):

PR #509, opened 2020-10-16
Simplify paths_desc

jonludlam:

### DIFF
--- a/src/model/paths.ml
+++ b/src/model/paths.ml
@@ -69,13 +69,6 @@ module Identifier = struct
 
   let label_parent n = label_parent_aux (n :> t)
 
-  let split n =
-    let name = name n in
-    match (n :> t) with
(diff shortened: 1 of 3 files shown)


src/model/paths.ml
```diff
@@ -69,13 +69,6 @@ module Identifier = struct
 
   let label_parent n = label_parent_aux (n :> t)
 
-  let split n =
-    let name = name n in
-    match (n :> t) with
-    | `Root (root, _) -> (`Root (Some root), name)
-    | `CoreType _ | `CoreException _ -> (`Root None, name)
-    | n -> (`Parent (label_parent n :> Paths_types.Identifier.label_parent), name)
```
Julow, line 77: There was a potential stack overflow because the `Page` case is not handled as a "Root". `label_parent` of a page is its input.

PR #510, opened 2020-10-16
test/parser: Use Model_desc instead of Sexplib for expected output

Julow:
Replace the use of sexp in parser tests by yojson output.
Sexplib dependency pulls an other package (js-build-tools) that fails to build on Windows.

This PR removes the following test-dependencies: base, csexp, dune-configurator, easy-format, num, parsexp, sexplib, sexplib0

The diff is huge because because every expected test outputs have changed.


Julow:
The diff is too big, there is an other possible way to fix this: We only need `sexplib0`, which shouldn't pull the problematic dependency.

PR #516, opened 2020-10-21
Latex uses forward slash for paths on Windows

Julow:
It seems that Latex uses forward slash as file separator on every platforms. Source: https://en.wikibooks.org/wiki/LaTeX/Modular_Documents#Using_different_paths

This is interesting for running the same testsuite on every platforms.


Octachron:
You are right indeed.

PR #517, opened 2020-10-21
Handle CRLF line endings when computing indentation of code blocks

Julow:
The detected indentation of code blocks was not the same depending on line endings.
Here is a CI build on Windows that include this patch: https://github.com/Julow/odoc/runs/1285922231?check_suite_focus=true

### DIFF
--- a/src/parser/lexer.mll
+++ b/src/parser/lexer.mll
@@ -73,13 +73,22 @@ let trim_trailing_blank_lines : string -> string = fun s ->
 let trim_leading_whitespace : first_line_offset:int -> string -> string =
  fun ~first_line_offset s ->
   let count_leading_whitespace line =
-    let rec count_leading_whitespace' index =
-      match line.[index] with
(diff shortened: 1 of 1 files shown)


dra27:
With the caveat that I started at looking at odoc's lexer about 5 minutes ago, do you definitely need to preserve the `\r`? The ideal path for these things is to be highly liberal in what you accept (so `\r\n` or `\n` is a newline) but then then get rid of the `\r` as soon as possible, merely ensuring that they will be correctly translated back to `\r\n` on Windows when sent to a file (which, mostly, just means opening output files in text mode)

Julow:
You mean getting rid of the argument of `` `Blank_line `` ? I agree we should abstract away this as soon as possible. I'm not sure this value is used anywhere, I think we can remove it.
Here it's even more obvious, the result of this function is only used locally and this case is never used. I removed that part.

PR #520, opened 2020-10-24
More module type of fixes

### LINKED ISSUE #500: Uncaught exception Odoc_xref2.Tools.UnexpandedTypeOf
Trying to compile the following mli file
```ocaml
(* a.mli *)
module M: sig include module type of String end
```
with
```bash
ocamc -c -bin-annot a.mli
odoc compile --package test a.cmti
```
yields the following error:
> odoc: internal error, uncaught exception:
      Odoc_xref2.Tools.UnexpandedTypeOf

(when there is no `.odoc` file for `String` in the include path).

jonludlam:
Fixes #500 and a couple of other issues.


Julow (approved):
Looks good !

Is here remaining exceptions in `Tools` ? (eg. can exceptions from `Cpath` or `Subst` leaks from Tools ?)
Can we drop the remaining handling of exceptions ? `tools.ml:904`, `ref_tools:131`

jonludlam:
good question - might be worth looking over this in a follow-up PR.

PR #526, opened 2020-10-28
latex backend: avoid hyperref and longtable destructive interferences

Octachron:
While compiling the OCaml manual, it may happen (or not) that very large sum types make the hyperref package self-destruct with 
> \pdfendlink ended up in different nesting level than \pdfstartlink

due to a bad interaction between the hyperref and longtable package.

There is no clear fix beyond manual edition of the source, which is not reasonable for the generated latex source.

Thus this PR simply limits the number of rows of generated tables, which in turns removes the assumption that there is a reasonable latex package that knows how to handle multi-pages tables.

### DIFF
--- a/src/latex/generator.ml
+++ b/src/latex/generator.ml
@@ -148,6 +148,8 @@ let entity ~in_source ~verbatim x =
   else
     Txt [escape_entity x]
 
+let small_table_limit = 10
+
 let rec pp_elt ppf = function
   | Txt words ->
     Fmt.list Fmt.string ~sep:none ppf words
@@ -171,7 +173,14 @@ let rec pp_elt ppf = function
   | List {typ; items} -> list typ pp ppf items
   | Description items -> Raw.description pp ppf items
   | Table { row_size=Large|Huge as size; tbl } -> large_table size ppf tbl
-  | Table { row_size=Small|Empty; tbl } -> Raw.small_table pp ppf tbl
+  | Table { row_size=Small|Empty as size; tbl } ->
+    if List.length tbl <= small_table_limit then
+      Fmt.pf ppf "%a%a%a"
+        Raw.break Line
+        (Raw.small_table pp) tbl
+        Raw.break Line
+    else
+      large_table size ppf tbl
   | Label x -> Raw.label ppf x
   | Indented x ->  Raw.indent pp ppf x
   | Ligaturable s -> Fmt.string ppf s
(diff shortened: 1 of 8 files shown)


Drup:
> Thus this PR simply limits the number of rows of generated tables, which in turns removes the assumption that there is a reasonable latex package that knows how to handle multi-pages tables.

Welp.

Drup (approved):

src/latex/generator.ml
```diff
@@ -200,15 +209,18 @@ and href ppf (l,txt) =
...
           pp a
           Raw.break Aesthetic
           (Raw.indent row) q  in
-    let matrix ppf m =
-
-      List.iter (row ppf) m in
+    let matrix ppf m = List.iter (row ppf) m in
     match size with
     | Huge -> Raw.break ppf Line; matrix ppf tbl
     | Large | _ -> Raw.indent matrix ppf tbl
```
gasche, line 226: In the `Huge` case, the output now always begins and ends with a Line break. In the `Large|Small|Empty` case, your change to the output guarantees that we always have a Line break at the end, but not at the beginning. On the other hand, small-height small tables get a Line break at both the beginning and end. This sounds a bit inconsistent, but I don't know what the intent is.

src/latex/generator.ml
```diff
@@ -171,7 +173,14 @@ let rec pp_elt ppf = function
...
   | Table { row_size=Large|Huge as size; tbl } -> large_table size ppf tbl
-  | Table { row_size=Small|Empty; tbl } -> Raw.small_table pp ppf tbl
+  | Table { row_size=Small|Empty as size; tbl } ->
+    if List.length tbl <= small_table_limit then
+      Fmt.pf ppf "%a%a%a"
+        Raw.break Line
+        (Raw.small_table pp) tbl
+        Raw.break Line
+    else
+      large_table size ppf tbl
```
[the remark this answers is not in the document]
Octachron: In this context, `small_table` means essentially `any tables that the basic tabular latex environment may render correctly`.

There is indeed two notions of sizes colliding here: the height of the tables, and the maximal "width" of cells. I have renamed `small_table_limit` to `small_table_height_limit`, does this help?  

I have moved the newlines to the raw latex side, with a comment that explains that we don't want the table to be inlined in-between delimiters.

src/latex/generator.ml
Octachron: The Huge case is in fact unnecessary, this was probably some scoria from prior implementation of recursive tables.

gasche: With the new implementation I'm still confused by line breaks in the code: if I understand correctly, `row` produces a final line break on lists of size 0 or 1, but not 2 and above, and it's not clear to me why.

Octachron: There is probably too much hardwired knowledge about the imbrication of different environments, but here the indentation environment provides the line break for rows with more than one elements.

gasche:
Thanks, this is all clearer with the new changes. Please feel free to go along.

jonludlam:
Thanks everyone! In it goes.

PR #528, opened 2020-10-29
html backend: fix linking to mld pages

Octachron:
Currently the link generated for `{!page:mld}` in the html backend refers to `mld` rather than `mld.html`.
This PR fixes this issue.

### DIFF
--- a/src/html/link.ml
+++ b/src/html/link.ml
@@ -73,6 +73,7 @@ let href ~resolve { Url.Anchor. page; anchor; kind } =
       @ target_from_common_ancestor
     in
     let page = String.concat "/" relative_target in
+    let page = if kind="page" then page ^ ".html" else page in
     begin match anchor with
     | "" -> page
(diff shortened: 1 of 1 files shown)


src/html/link.ml
```diff
@@ -73,6 +73,7 @@ let href ~resolve { Url.Anchor. page; anchor; kind } =
       @ target_from_common_ancestor
     in
     let page = String.concat "/" relative_target in
+    let page = if kind="page" then page ^ ".html" else page in
```
Drup, line 76: That's not the right place, you should fix `Path.for_linking` above. Looking at it, the implementation should probably merge code with `as_filename` as much as possible...

Drup:
Oups, I'm surprised I didn't realized it in the tests .... The fix is not in the right place though.

jonludlam:
I have a patch incoming that contains a fix for this, so I'm closing this PR. Thanks though!

Issue #523, opened 2020-10-28
References to classes


Julow:
Should `type:` references resolve to classes and class types ?

Currently, there is an inconsistency: `type:cl` resolves to a class but `M.type-cl` and `type:M.cl` don't.

Drup:
Both class and class types automatically define an associated `type`, so yes, it should.

PR #529, opened 2020-10-29
 Allow type references to resolve to classes 

Julow:
Fixes https://github.com/ocaml/odoc/issues/523

References with explicit `type:` or `type-` would not resolve classes or class types (except in one case).

This brings a new inconsistency: `type-cl.method-m` is not a valid reference due to the type of method references. Should we relax this or is it expected ? (classes are types but types are not classes)

### DIFF
--- a/test/xref2/refs/refs.md
+++ b/test/xref2/refs/refs.md
@@ -588,6 +588,41 @@ Exception: Failure "resolve_reference".
    t)
 ```
 
+Classes as type references:
+
+```ocaml
+let resolve_ref = resolve_ref_of_mli {|
+  module M : sig
+    class cl : object method m : int end
+    class type clt = object method m : int end
+  end
+
+  class cl : object method m : int end
+  class type clt = object method m : int end
+|}
+```
+
+```ocaml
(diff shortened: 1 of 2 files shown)


lpw25:
> Should we relax this or is it expected ? (classes are types but types are not classes)

We should not relax it for exactly the reason you suggest

test/xref2/refs/refs.md
```diff
@@ -588,6 +588,69 @@ Exception: Failure "resolve_reference".
...
+It is not possible to reference to methods through type references:
+
+```ocaml
+# let resolve_ref = resolve_ref_of_mli {|
+    class c : object method m : int end
+    type t = < m : int >
+  |}
+val resolve_ref : string -> ref = <fun>
+# resolve_ref "type-c.m"
+- : ref = `Method (`Identifier (`Class (`Root (Common.root, Root), c)), m)
```
Julow, line 641: This is unexpected. Is this a bug ?

Julow: (issue: https://github.com/ocaml/odoc/issues/530)

PR #531, opened 2020-11-10
Backport 4.12

jonludlam:
Backport @Octachron's support for the new injectivity annotations in OCaml 4.12


jonludlam:
CI failure is due to mdx fixed in trunk via vendoring mdx - I don't think it's necessary to do that for this.

PR #532, opened 2020-11-17
Parent child initial

jonludlam:


jonludlam:
This doesn't contain the code to render a side-bar tree yet - just the infrastructure to manage parents and children.

.vscode/tasks.json
```diff
@@ -19,7 +19,8 @@
             "group": {
                 "kind": "test",
                 "isDefault": true
-            }
+            },
+            "problemMatcher": []
```
Julow, line 23: Did you intent to commit that file ? (it is already on `master`).

src/odoc/bin/main.ml
```diff
@@ -137,18 +128,28 @@ end = struct
     in
     Arg.(value & opt_all string default & info ~docv:"MODULE" ~doc ["open"])
 
+  let children =
+    let doc = "Specify the odoc file as a child. Can be used multiple times. Only applies to mld files" in
+    let default = [] in
+    Arg.(value & opt_all string default & info ~docv:"CHILD" ~doc ["child"])
```
Julow, line 134: Cmdliner has a `list` option type that takes several values separated by a `,`, would it be more readable/easier to type on the CLI ? (`value & list string & ...`)

src/odoc/bin/main.ml
```diff
@@ -137,18 +128,28 @@ end = struct
...
+    let package_opt =
+      let doc = "Package the input is part of (deprecated - use '--parent')" in
+      Arg.(value & opt (some string) None &
            info ~docs ~docv:"PKG" ~doc ["package"; "pkg"])
     in
+    let parent_opt =
+      let doc = "Parent page or subpage" in
+      Arg.(value & opt (some string) None &
+           info ~docs ~docv:"PARENT" ~doc ["parent"])
+    in
```
Julow, line 146: Both options don't work at the same time, `--package` would be ignored if `--parent` is passed. Maybe we should resolve this here so `Compile` don't have to decide what to do with these at all:
```ocaml
let parent_opt =
  let package_opt = ...
  let parent_opt = ...
  let one_of pkg parent =
    match pkg, parent with
    | Some _, Some _ -> (* An error here *)
    | Some pkg, None -> Some (`Package pkg)
    | None, Some parent -> Some (`Parent parent)
    | None, None -> None
  in
  Term.(const one_of $ package_opt $ parent_opt)
in
```

Julow (approved):
Why is it necessary to specify both `{!child:...}` references in pages and the `--child` argument ?
  I'd expect the first one to work by itself and the second to be equivalent to inserting a hidden `{!child:...}` reference.

.vscode/tasks.json
jonludlam: I saw it was on master - Not really sure whether we want it or not tbh. Do they normally get committed?

Julow: I think they don't. I think they don't belong here, especially if there is nothing specific to the project, they should be in a plugin (so it can be maintained and shared with every OCaml projects) or a personal configuration (so it can be modified according to preferences).

src/odoc/bin/main.ml
jonludlam: I thought it would be better to be consistent with the other CLI arguments.

src/odoc/bin/main.ml
jonludlam: Yeah that's a good idea!

src/odoc/bin/main.ml
Julow: Make sense. Maybe we should have a single letter option then ? I don't know any tool that have "long" options that are expected to be repeated.
The options will be seen in build systems logs and maybe wrote by hand sometimes, so readability matters.

jonludlam: Ah, good point! I had forgotten that.

.vscode/tasks.json
jonludlam: OK, I'll pull the whole file out. Ta!

dbuenzli:
@jonludlam I reread the wiki about the document tree module, and I still find the whole thing quite confusing. Notably it's still unclear to me on how the parent/child hierarchy is going to be discovered by `odoc` so that it can be appropriately `odoc`ed. 

More precisely, given a set of installed `.mld` and `.cmti` files, how do I infer the `--child` and `--parent` options I need to give to the various elements ?

jonludlam:
We'll need some conventions and/or extra metadata I think. Do you have some suggestions? What we've got in this PR is really just a basis to build the rest on.

dbuenzli:
> We'll need some conventions and/or extra metadata I think. Do you have some suggestions? What we've got in this PR is really just a basis to build the rest on.

I think it would be good to *first* design that. It can have quite a few implications on the model and the way `odoc` should be run and what it outputs.

It would be good to avoid the need for extra metadata and make the document tree self-described by the documents themselves. That means nodes should either declare their children or children their parents (the latter having the advantage of parenting unicity). Once we have that it should be possible for `odoc` to spit out the dependencies (e.g. on `link-deps`).

jonludlam:
The plan is to merge the explicit command-line based one first and then work out how to make it more convenient in the hand-written manuals case.

dbuenzli:
Well do as you think, but somehow it's rarely a good idea to design the user experience as an after thought.

lpw25:
> More precisely, given a set of installed .mld and .cmti files, how do I infer the --child and --parent options I need to give to the various elements ?

To a certain extent this is kind of up to you, as the designer of odig and its expected packaging standards. Similarly, the mechanism for how the relationship is specified in the source is up to the designers of dune or whatever other build system people use. Of course, much like with the standards for where OCaml artifacts should be installed, we would hope that there is some agreement between e.g. dune and odig about how things should be laid out -- and perhaps we should try to help form those standards here.

Personally, I am expecting the rules to be extensions of the current ones. So a module is contained in its package, which we would now always require to have a page (although it would often I expect be generated). I would also expect mld files to be contained in their packages. If people wish to create additional layers of hierarchy e.g. for nested pages, then I would expect some simple rules for that: probably something based around using directories to represent the nesting.

lpw25:
I think that maybe we could explain these changes differently, and that might address @dbuenzli's concerns. There are essentially two changes each with a separate motivation:
1. There was some requests for features that required knowing what all the modules in a package were. The way we accommodate these features involves semantically identify the "package" -- previously a totally abstract notion within odoc -- with its corresponding mld file. Now we specify all the modules when building that mld file, and so there is a place to go and get the list of all modules in a package.
2. There were some use cases (e.g. docs.ocaml.org, maybe dune with both libraries and projects) where people wanted a hierarchy with more than 2 levels. We achieve that by replacing "package" with "parent", and allowing a parent to itself have a parent.

1 requires odig to update how it calls odoc, but doesn't change the relationship between the different files, so I presume that you have no objection to that part.

2 potentially allows hierarchical documentation, which would require some way to specify that hierarchy to odig. However, the aim is not to force heirarchy on odig's use case. You can perfectly reasonably not support people making nested pages. The nesting is needed for other use cases, and since its there it might be nice to expose it to users. But that is really your call to make: odoc does not itself take a position on the matter.

dbuenzli:
> 2 potentially allows hierarchical documentation, which would require some way to specify that hierarchy to odig. However, the aim is not to force heirarchy on odig's use case. You can perfectly reasonably not support people making nested pages.

Well I mean the aim should still be that when you hit the actual package root, the docs do not end up being built and presented differently whether they are created by `odig`, `docs.ocaml.org` or your build system. I think the best of doing that is to actually solve the problem in `odoc` itself and not rely on third-party conventions and/or metadata – if only because from an *authoring* point of view that's a sub-par experience.

jonludlam:
I should have been more explicit. While we'd like to be able to allow hierarchical docs to be created by end users, this PR does _not_ address that, and indeed we've not agreed on exactly how this will be done. What this PR _does_ do is allow odoc to have more structured approach to laying out the docs of packages - that is, it will allow tools like odig and dune and whatever ends up building docs.ocaml.org to organise the docs for each package in a more flexible way. As such, the process
for figuring out parent/child relationships for a bunch of cmti and mlds is up to the people designing those tools, and, as @lpw25 points out, this is unchanged.

For example to recreate the layout odig has today would be a matter of creating a package-level mld file and making this the parent of all of the cmti and mld files. We can also definitely organise the docs into a hierarchy that supports the documentation of multiple versions of many packages at once. We could probably make use of the _existing_ extra metadata that we have already to organise them into packages and subpackages, though I'm slightly less convinced that this can be done without requiring changes to what's installed by the packages, but only because I haven't tried it yet.

So, this is making the organisation of the existing package documentation more flexible rather than allowing the package documentation itself to be more flexible - though it gives us the ability to do that later if we wish.

src/document/url.ml
```diff
@@ -87,13 +87,22 @@ module Path = struct
...
+      let kind = "cpage" in
+      let page = PageName.to_string page_name in
+      mk kind page
+    | `Page (parent, page_name) ->
+      let parent = from_identifier (parent :> source) in
+      let kind = "cpage" in
+      let page = PageName.to_string page_name in
+      mk ~parent kind page
+    | `LeafPage (parent, page_name) ->
+      let parent = from_identifier (parent :> source) in
```
Drup, line 105: What's this "cpage" kind ? You can use something more longformed (the content is pretty arbitrary).

src/html/link.ml
```diff
@@ -46,34 +50,51 @@ let rec drop_shared_prefix l1 l2 =
     drop_shared_prefix l1s l2s
   | _, _ -> l1, l2
 
-let href ~resolve { Url.Anchor. page; anchor; kind } =
-  let leaf = if !semantic_uris || kind = "page" then [] else ["index.html"] in
-  let target = Path.for_linking page @ leaf in
-  match resolve with
```
Drup, line 52: If we are not using the `semantic_uris` option, you should deprecate/remove it.

src/document/url.ml
jonludlam: 'container page' - it was originally that, but it appeared everywhere in the latex docs so I cut it down to this. 'container pages' are those that can contain others (Page and RootPage), and are output in the html as <pagename>/index.html, in contrast to the 'page' types that are output as <pagename>.html

Drup (commented):
The changes in the Document IR look good! In fact, I wanted to do some of those and never got around to it. The URL code is still a bit crufty, but I think it's inherent to the task.

We'll be able to reuse and extend the ToC stuff nicely, which I'm very happy about. :)

src/document/url.ml
Drup: I see, and we don't know if they have children or not just by looking at their urls.

I would suggest just using `container-page`. We already have `module-type`, and it's fine.

Julow (commented):
I hit an `assert false` with this script:
```sh
echo "{0 A}" > a.mld
echo "{0 B}" > b.mld
odoc compile --package a -c page-b -c b a.mld -o page-a.odoc
odoc compile --parent a b.mld -I . -o page-b.odoc
odoc link page-a.odoc -o page-a.odocl -I .
```

In the first `compile`, only `-c b` is necessary, `-c page-b` is accepted but not used.
This cause the error below but also it is not possible to have both a `b.mld` and `b.ml`.

Backtrace:
```
      File "src/odoc/env.ml", line 138, characters 16-22: Assertion failed
      Raised at Odoc_odoc__Env.lookup_unit.handle_root in file "src/odoc/env.ml", line 138, characters 16-28
      Called from Odoc_xref2__Env.lookup_root_module in file "src/xref2/env.ml", line 356, characters 14-32
      Called from Odoc_odoc__Odoc_link.from_odoc.(fun) in file "src/odoc/odoc_link.ml", line 16, characters 12-55
      Called from Odoc_odoc__Odoc_link.from_odoc.(fun) in file "src/odoc/odoc_link.ml", line 11, characters 14-376
```

PR #533, opened 2020-11-17
Better handling of errors in 'type_of.ml'

jonludlam:
The code for 'module type of' expansion runs before any other resolution
happens, and it needs to run until a fixed point. Since the change to
use 'Result.t' over simply raising exceptions when we notice we'll need
to recurse, a bug was introduced when the specific error case we're
looking for was swallowed in another layer. This PR attempts to make
sure we don't ever throw away an error raised during signature expansion
and adds some code to check in the combined error whether the root cause
was in fact the case we're looking for; `` `UnexpandedTypeOf `` in this
case.

### DIFF
--- a/src/xref2/errors.ml
+++ b/src/xref2/errors.ml
@@ -13,12 +13,13 @@ module Tools_error = struct
     | `UnresolvedForwardPath
       (** The module signature depends upon a forward path *)
     | `UnresolvedPath of
-      [ `Module of Cpath.module_ | `ModuleType of Cpath.module_type ]
+      [ `Module of Cpath.module_ * simple_module_lookup_error
+      | `ModuleType of Cpath.module_type * simple_module_type_lookup_error ]
       (** The path to the module or module type could not be resolved *)
     | `UnexpandedTypeOf of Component.ModuleType.type_of_desc
       (** The `module type of` expression could not be expanded *) ]
 
-  type simple_module_lookup_error =
+  and simple_module_lookup_error =
     [ `Local of Env.t * Ident.path_module
       (** Internal error: Found local path during lookup *)
     | `Unresolved_apply  (** [`Apply] argument is not [`Resolved] *)
@@ -37,7 +38,8 @@ module Tools_error = struct
     | `UnresolvedForwardPath
       (** The module signature depends upon a forward path *)
     | `UnresolvedPath of
-      [ `Module of Cpath.module_ | `ModuleType of Cpath.module_type ]
+      [ `Module of Cpath.module_ * simple_module_lookup_error
+      | `ModuleType of Cpath.module_type * simple_module_type_lookup_error ]
     | `Parent of parent_lookup_error ]
 
   and simple_module_type_lookup_error =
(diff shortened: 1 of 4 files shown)


src/xref2/errors.ml
```diff
@@ -129,6 +126,35 @@ module Tools_error = struct
     | `Fragment_root -> Format.fprintf fmt "Fragment root"
 end
 
+ (* Ugh. we need to determine whether this was down to an unexpanded module type error. This is horrendous. *)
+ let is_unexpanded_module_type_of =
```
Julow, line 130: This only need to work on `signature_of_module_error` and to recurse into `parent_lookup_error` and `` [> `Parent of parent_lookup_error ] `` (the only interesting part of other `*_error`). For example:
```ocaml
let is_unexpanded_module_type_of =
  let rec signature_of_module = function
    | `UnexpandedTypeOf _ -> true
    | `OpaqueModule | `UnresolvedForwardPath -> false
    | `UnresolvedPath (`Module (_, e)) -> rec_parent e
    | `UnresolvedPath (`ModuleType (_, e)) -> rec_parent e
  and rec_parent = function
    | `Parent p -> parent p
    | _ -> false
  and parent = function
    | `Parent_sig e -> signature_of_module e
    | `Parent_module_type p -> rec_parent p
    | `Parent_expr p -> rec_parent p
    | `Parent_module p -> rec_parent p
    | `Fragment_root -> false
    | `Parent p -> parent p
  in
  signature_of_module
```

Also, I don't think `parent_lookup_error` should be part of `any`.

src/xref2/errors.ml
```diff
@@ -13,12 +13,13 @@ module Tools_error = struct
     | `UnresolvedForwardPath
       (** The module signature depends upon a forward path *)
     | `UnresolvedPath of
-      [ `Module of Cpath.module_ | `ModuleType of Cpath.module_type ]
+      [ `Module of Cpath.module_ * simple_module_lookup_error
+      | `ModuleType of Cpath.module_type * simple_module_type_lookup_error ]
```
Julow, line 17: Allowing these will makes traversing errors more complicated than just looking at `` `Parent ``.

Julow (approved):

src/xref2/errors.ml
jonludlam: aren't they all errors?

src/xref2/errors.ml
jonludlam: The problem is that they are always as a result of another error, which we were always throwing away - in particular here:

https://github.com/ocaml/odoc/blob/8b0a6323d80345f1c60332375417b40886395454/src/xref2/expand_tools.ml#L86

If we do that we risk forgetting that a problem was down to `UnexpandedTypeOf` somewhere. In that case above, the error is a `simple_module_lookup_error` which was `` `Parent_sig (`UnexpandedTypeOf) `` - so we'd have to check everywhere we wanted to return a `` `Module `` or `` `ModuleType `` and return an `` `UnexpandedTypeOf ``. It seemed a bit more error-prone than just making sure we never throw away any Error info.

src/xref2/errors.ml
Julow: `parent_lookup_error` are not errors I think, they're just an indirection on top of `` `Parent `` (present in every kinds of error).

src/xref2/errors.ml
Julow: It's fine to go with this. We'll know better next time we need to traverse errors again :)

Issue #538, opened 2020-11-25
Wrong type after substitution


Julow:
In this example:

```ocaml
module type S' = sig
  type 'a t

  val map : 'a t -> ('a -> 'b) -> 'b t

  val join : 'a t t -> 'a t
end

module type S = sig
  type 'a t

  include S' with type 'a t := 'a t
end
```

The module type `S` renders like this:

```txt
type 'a t
> include S' with type 'a t := 'a t
val map : 'a t -> ('a -> 'b) -> 'a t
val join : 'a t -> 'a t
```

The types `'a t t` and `'b t` are turned into `'a t` during the compile phase (not link).
This seems to be here since a long time, I stopped bisecting when the `link` command was added.

CraigFe:
May be related to https://github.com/ocaml/odoc/issues/462.

Julow:
Indeed but it's worth keeping both issues open because your example may not be solved with the fix I'm implementing.

Issue #462, opened 2020-07-29
Destructive substitution results in type variable name collision


CraigFe:
Using current Odoc master (`d70a0fc`), inlining of destructive type substitution can cause distinct type variables to be assigned the same name in generated documentation. For instance, in the following example:

```ocaml
module type Generic = sig
  type 'a t
  type 'a elt

  val map : ('a elt -> 'b elt) -> 'a t -> 'b t
end

module type Parametric = Generic with type 'a elt := 'a
```

the documentation of module type `Parametric` is as follows:

![image](https://user-images.githubusercontent.com/26125912/88847923-097c9980-d1e8-11ea-9787-c116ed727736.png)

where `map` gets given the wrong type (presumably due to an incorrect substitution `'b elt` → `'a`).

jonludlam:
Thanks @CraigFe - that looks like it's going to be fun...!

Julow:
This is fixed now.

PR #542, opened 2020-12-04
Fix substitution of type params

Julow:
Fix https://github.com/ocaml/odoc/issues/538 and https://github.com/ocaml/odoc/issues/462

This substitutes recursively type variables when encountering a "replaced" type.

### DIFF
--- a/src/xref2/subst.ml
+++ b/src/xref2/subst.ml
@@ -1,7 +1,12 @@
-exception TypeReplacement of Component.TypeExpr.t
-
 exception Invalidated
+
 exception MTOInvalidated
+
+type 'a or_replaced = Not_replaced of 'a | Replaced of Component.TypeExpr.t
+
+let map_replaced f = function
+  | Not_replaced p -> Not_replaced (f p)
+  | Replaced _ as r -> r
 
 open Component
 open Substitution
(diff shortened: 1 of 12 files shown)


test/xref2/gen.ml
```diff
@@ -0,0 +1,31 @@
...
+
+let generate input =
+  run "odoc man-generate -o %S %s" out_dir (Filename.quote input)
+
+let () =
+  for i = 1 to Array.length Sys.argv - 1 do
+    generate Sys.argv.(i)
+  done;
+  list_dir_rec [] out_dir
+  |> List.iter (fun p -> run "COLUMNS=80 man -l %s | col -b" p)
```
Julow, line 31: These tests may be very fragile because they depend on `man` and `col`, which are very complicated programs. I'm replacing that with Mdx tests instead.

src/xref2/subst.ml
```diff
@@ -571,11 +636,13 @@ and extension_constructor s c =
...
-  {
-    e with
-    type_path = type_path s e.type_path;
-    constructors = List.map (extension_constructor s) e.constructors;
-  }
+  let type_path =
+    match type_path s e.type_path with
+    | Not_replaced p -> p
+    | Replaced (TypeExpr.Constr (p, _), _) -> p
+    | Replaced _ -> (* What else is possible ? *) assert false
```
Julow, line 643: I'm not sure what we should do here. More tests are required

test/xref2/subst/test.md
```diff
@@ -76,10 +76,141 @@ type vv/5 = local(SubstituteMe/2,false).v
...
+  |}
+- : Component.Signature.t =
+module type S/69 = sig
+  type t/70
+  val map/71 : ([a] t/70) -> ((a) -> b) -> [b] t/70
+   (removed=[])end
+module M/68 : sig
+  type t/72
+  include : S/69 with [root(S/69).t = [(alias (poly_var [ `A of (a * b) ]) b)] t/72] (sig =
+    val map/73 : ([(alias (poly_var [ `A of (a * b) ]) b)] t/72) -> ((a) -> b) -> [(alias (poly_var [ `A of (b * b) ]) b)] t/72
```
Julow, line 212: I've found an edge case, the last type is wrong: `(alias (poly_var [ `A of (b * b) ]) b)`.
We should either rename `b` to `c` in this type or choose a free variable before substituting.
What do you think ?

PR #544, opened 2020-12-09
Fix the steps in CONTRIBUTING.md to get a working dev setup

daimrod:
- fetch odoc with its git submodule mdx
- install jq(1) as required per `conf-jq`
- install test dependencies with opam

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,7 +46,7 @@ To make edits to the HTML generation, run the following commands:
     git clone https://github.com/ocaml/odoc.git
     cd odoc
     opam pin add --no-action odoc .
-    opam install --deps-only odoc
+    opam install --with-test --deps-only odoc
     ```
 
     Also make sure to install a recent version of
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -40,27 +40,42 @@ of OCaml language item that needs to be displayed in docs.
...
+2. Set up for development:
 
     ```
-    git clone https://github.com/ocaml/odoc.git
+    git clone --recurse-submodules https://github.com/ocaml/odoc.git
     cd odoc
     opam pin add --no-action odoc .
-    opam install --deps-only odoc
+    opam install --with-test --deps-only odoc
+    opam install --deps-only mdx
```
Julow, line 72: The `mdx` submodule is meant to fix a dependency cycle between odoc and mdx, which was problematic in OCaml-CI. It's temporary.
I think it's fine to do `opam install mdx` and to not init the submodule.

Julow (approved):

CONTRIBUTING.md
daimrod: Ok.  I can update the PR to just install mdx.  What do you prefer?

jonludlam: Let's leave it as-is. Once we've got rid of the submodule we can remove this instruction.

jonludlam:
Thanks!

PR #545, opened 2020-12-10
Enable CI for OCaml 4.02.3

Julow:
OCaml-CI doesn't test below 4.04.1.
Parser tests are disabled below 4.04.1 because `sexplib0` is not available.


jonludlam:
something's up with ocaml-ci's dependency analysis here

Issue #543, opened 2020-12-06
Local opens in class type declarations cause an assertion failure


lpw25:
I think that the following mli will cause odoc to hit an assertion failure in `Cmti.read_class_signature`:

```ocaml
class type ct =
  let open Float in
  object
    method x : t
  end
```

(I've based this on a larger reproduction case, but haven't actually checked the smaller version. Let me know if it doesn't fail and I'll try to make a small case again).

Julow:
This example does reproduce the issue. It seems to be not implemented, I'll investigate.

PR #546, opened 2020-12-11
Fix assert failure on local open in class types

Julow:
Fix https://github.com/ocaml/odoc/issues/543

This was a bug in the loader. The loader for `cmt` files have the same logic but I couldn't reproduce the issue with a cmt.

### DIFF
--- a/src/loader/cmti.ml
+++ b/src/loader/cmti.ml
@@ -341,6 +341,15 @@ let read_exception env parent (ext : extension_constructor) =
     let res = opt_map (read_core_type env label_container) res in
         {id; doc; args; res}
 
+let read_self_type env container typ =
+  if typ.ctyp_desc = Ttyp_any then None
+  else Some (read_core_type env container typ)
+
+let read_class_type_constr env label_parent p params =
+  let p = Env.Path.read_class_type env p in
+  let params = List.map (read_core_type env label_parent) params in
+  ClassType.Constr (p, params)
+
 let rec read_class_type_field env parent ctf =
   let open ClassSignature in
   let open Odoc_model.Names in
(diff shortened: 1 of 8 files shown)


src/loader/cmti.ml
```diff
@@ -369,49 +378,61 @@ let rec read_class_type_field env parent ctf =
...
+#if OCAML_MAJOR = 4 && OCAML_MINOR >= 06 && OCAML_MINOR < 08
+  | Tcty_open (_, _, _, _, cty) -> read_class_type env parent label_parent cty
+#elif OCAML_MAJOR = 4 && OCAML_MINOR >= 08
+  | Tcty_open (_, cty) -> read_class_type env parent label_parent cty
 #endif
 
+and read_class_signature_of_class_type env parent label_parent cty =
+  match read_class_type env parent label_parent cty with
+  | Class.ClassType ct -> ct
+  | Arrow _ -> assert false
```
Julow, line 422: The asserts are still here but now they are tested only when explicitly asked (this function).
This is because the compiler uses the `class_type` on the rhs of `inherit` and of `class type` but Odoc wants a more constrained type.

jonludlam:
We should probably change the tests into dune cram tests.

lpw25:
Personally, I think the code was better how it was previously. I think that functions like `read_class_signature_of_class_type` which translate an item and then inspect the result of the translation are a code smell.

The fix for the bug should just be a single line: recursing in the `Tcty_open` case in `read_class_signature`.

Julow:
The rhs of `inherit` and of `class type foo =` are `class_type` in the typed tree. This type was handled in two different ways before this PR with asserts in reachable paths. Now, there is only one function for converting `class_type`, the fact that Odoc wants a smaller type is dealt separately and explicitly.
Note that `read_class_type` is used elsewhere and must cover every values of `class_type` (for example in its implementation).

lpw25:
The use of `class_type` for both types in the type tree is a mistake in the representation of typed tree. The [description in the manual](http://caml.inria.fr/pub/docs/manual-ocaml/classes.html#ss:classes:class-types) shows the true underlying structure -- with `class-body-type` separate from `class-type`.

The previous code reflected this underlying structure and used assertions in the places where the typed tree's use of an inaccurate type meant that there were impossible states.

The new code instead perpetuates to inaccurate representation through the translation and then inspects the result and asserts on the other side -- it is harder to understand than the earlier code because now you need to understand the translation as well as the original representation to see that the assertion is genuinely unreachable.

The only problem with the previous code was that someone messed up the assertions when local open in classes was added to the language. That could easily be fixed without restructuring the code.

lpw25:
I feel I should some context for my point. Your proposed code is basically fine, but it is an instance of something that is a particular pet peeve of mine. It is a common thing people do when writing compilers and it is very often a source of bugs and hard to maintain code, to the point where I think that it should just be prohibited in such code bases even in small, mostly harmless, cases.

The essence of the issue is that when you write a transformation function from `a` to `b` you should be deconstructing `a`s and constructing `b`s. You should never find yourself constructing `a`s or deconstructing `b`s. Whenever you do you have made the code harder to reason about because one part of the transformation depends non-obviously on another part of the transformation. In the long run, someone will inevitably make a change to the depended-upon part of the transformation and introduce a bug.

In summary, your code is basically ok but it breaks a rule I like to enforce on such code, so I'd prefer it remain as it was (minus the bug obviously).

Julow:
Thanks for the explanations, I reverted to a much smaller patch, like you suggested.

PR #553, opened 2020-12-29
weed out unnecessary css properties

lubegasimon:


ulrikstrid:
Great, thank you for cleaning up my mess. We need to do the same changes on lines 79-124 as well, you should be able to just copy the other changes down there.

lubegasimon:
> We need to do the same changes on lines 79-124 as well

Ah, this is not exactly right!

PR #561, opened 2021-01-13
Auto-format the whole project using OCamlformat

Julow:
To reproduce this PR:

```bash
git checkout -b $RANDOM master
git cherry-pick 8c5f4b527f5ccafe8bd24984617480f1554b810d # Some setup
until dune build @fmt; do dune promote; done
git commit -a -m format
git diff @ ocamlformat # 'ocamlformat' is this PR
```

Fixing `git blame`:

I added a `.git-blame-ignore-revs` file that specifies which commits need to be ignored by `git blame`.
This doesn't work out of the box, it must be passed to `git`:
```sh
git blame --ignore-revs-file .git-blame-ignore-revs
```

It is also possible to add a permanent option but I don't recommend it because git will complain if this file is not present (eg. when checked out to before it was added).

Thanks to https://akrabat.com/ignoring-revisions-with-git-blame/

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Auto-format the whole project using OCamlformat
+75efc08df487fa1a89fb86e446aadc8f6273b5bb
--- a/.ocamlformat-ignore
+++ b/.ocamlformat-ignore
@@ -1,0 +1,4 @@
+src/compat/*
+src/loader/*
+src/model/*.cppo.ml
+test/xref2/lib/*
--- a/dune
+++ b/dune
@@ -1,10 +1,12 @@
 ; Note: We disable warning 18 here to support compiler versions < 4.03
 ; See https://caml.inria.fr/mantis/view.php?id=7135 for details
+
 (env
  (dev
-  (flags (:standard -g -w -18-53)))
+  (flags
+   (:standard -g -w -18-53)))
  (release
-  (flags (:standard -g -w -18-53))))
+  (flags
+   (:standard -g -w -18-53))))
 
 (vendored_dirs mdx)
-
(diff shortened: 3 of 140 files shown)


Julow:
It is currently not possible to install OCamlformat (`> 0.15.1`) in our dev environment because its dependency constraints are not compatible with Mdx.
To work around that, either:
- Install OCamlformat in a separate Opam switch and run formatting like this: `opam exec --switch=ocamlformat-0.16.0 -- dune build @fmt`
- Don't install OCamlformat with Opam. I use Nix but that version of ocamlformat is not yet available. It is possible to copy and rename OCamlformat binaries and manage them without Opam.

jonludlam:
Is there a compelling reason to use ocamlformat > 0.15.0? If we used that version we can just `opam install` it and there are no further hoops to jump through?

Julow:
I amended this PR to use ocamlformat 0.15.0. There will be quite a lot of diffs (+476 -476) like this when switching to 0.16.0:

```diff
     if Stack.is_empty state.context then List.rev state.current
     else (
       leave state;
-      flush state )
+      flush state)
 end
 
 (** Modern implementation using semantic tags, Only for 4.08+ *)
```

and like this:
```diff
       | With { w_expr; _ } -> extract_path_from_umt ~default w_expr
       | TypeOf { t_desc = ModPath (`Resolved r); _ }
       | TypeOf { t_desc = StructInclude (`Resolved r); _ } ->
-          ( Paths.Path.Resolved.Module.identifier r
-            :> Paths.Identifier.Signature.t )
+          (Paths.Path.Resolved.Module.identifier r
+            :> Paths.Identifier.Signature.t)
       | _ -> default
 
     and mdexpr :
```

jonludlam:
The lint failure looks like it's due to this: https://github.com/ocurrent/ocaml-ci/pull/299

PR #568, opened 2021-01-26
Html indent

jonludlam:
Add a parameter to `html-generate` to format the output nicely - useful for diffing

Also this no longer causes exceptions when processing 'packed' compilation units.

### DIFF
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -377,6 +377,10 @@ module Odoc_html = Make_renderer (struct
     in
     Arg.(value & flag (info ~doc [ "closed-details" ]))
 
+  let indent =
+    let doc = "Format the output HTML files with indentation" in
+    Arg.(value & flag (info ~doc [ "indent" ]))
+
   (* Very basic validation and normalization for URI paths. *)
   let convert_uri : Odoc_html.Tree.uri Arg.converter =
     let parser str =
@@ -409,10 +413,10 @@ module Odoc_html = Make_renderer (struct
       value & opt convert_uri default & info ~docv:"URI" ~doc [ "theme-uri" ])
 
   let extra_args =
-    let f semantic_uris closed_details theme_uri =
-      { Html_page.semantic_uris; closed_details; theme_uri }
+    let f semantic_uris closed_details indent theme_uri =
+      { Html_page.semantic_uris; closed_details; theme_uri; indent }
     in
-    Term.(const f $ semantic_uris $ closed_details $ theme_uri)
+    Term.(const f $ semantic_uris $ closed_details $ indent $ theme_uri)
 end)
 
 module Html_fragment : sig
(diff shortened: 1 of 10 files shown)


src/model/paths_types.ml
```diff
@@ -259,6 +262,8 @@ and Resolved_fragment : sig
     | `Class of signature * ClassName.t
     | `ClassType of signature * ClassTypeName.t ]
 
+  type leaf = [ module_ | type_ ]
```
Julow, line 265: What about classes, substs and others ?
(Would help move the `assert false` from `render_resolved_fragment`)

src/odoc/bin/main.ml
```diff
@@ -377,6 +377,10 @@ module Odoc_html = Make_renderer (struct
     in
     Arg.(value & flag (info ~doc [ "closed-details" ]))
 
+  let indent =
+    let doc = "Format the output HTML files with indentation" in
+    Arg.(value & flag (info ~doc [ "indent" ]))
```
Julow, line 382: What about enabling it by default without exposing the option ? Whitespaces compress well and indented output is always nicer to look at.

src/xref2/compile.ml
```diff
@@ -64,7 +64,7 @@ and content env id =
   | Module m ->
       let sg = Type_of.signature env m in
       Module (signature env (id :> Id.Signature.t) sg)
-  | Pack _ -> failwith "Unhandled content"
+  | Pack p -> Pack p
```
Julow, line 67: How hard would it be to actually compile packs ?

Julow (approved):
Looks good.

src/model/paths_types.ml
jonludlam: this refers to the type of the substitution - ie, `X with module Y=Z` or `X with type t = u` -- so only module or type required.

src/odoc/bin/main.ml
jonludlam: could do - I've got no strong views on this. I've added it because my previous script was just running things through `html tidy`, and the larger my test set got the slower the tidying got!

src/xref2/compile.ml
jonludlam: I've not even looked into it to be honest - not many packages use packing right now, but the one I have installed was causing my tests to fail!

src/model/paths_types.ml
Julow: Thanks, I got confused by the `type t` below. Why does it include everything ?

jonludlam: that's the union one, helpful for printers and the like. it's quite handy to have just the leaf one so we don't have to deal with the Root constructor unnecessarily

Drup:
As @Julow said, It should probably be enable it by default. The performance cost is probably negligible, as is the size cost after compression.

PR #569, opened 2021-01-26
Include fixes

jonludlam:
Bit of a mammoth PR. This one is focused on improving the performance of compilation when there are includes involved. On files that nest includes or contain particularly large included signatures the performance gains can be big:

before:
```
/usr/bin/time -l odoc compile --parent v0_14_0 -I ... Async_unix__Require_explicit_time_source.odoc
        70.93 real        69.61 user         0.90 sys
```

after:
```
/usr/bin/time -l odoc compile --parent v0_14_0 -I ... Async_unix__Require_explicit_time_source.odoc
        3.90 real         3.59 user         0.29 sys
```

and the space usage is also improved (though not quite as dramatically) - from 2.2 gigs down to 700 megs.

There are a number of additional correctness patches in here too - a number of bugs became apparent when comparing the results of the optimized odoc with the original.


src/xref2/expand_tools.ml
```diff
@@ -61,7 +61,7 @@ and aux_expansion_of_module_alias env ~strengthen path =
          where we're aliasing e.g. Stdlib.List which is itself an alias for
          Stdlib__list - we want to strengthen with Stdlib.List rather than
          with Stdlib__list. *)
-      match (aux_expansion_of_module env ~strengthen:false m, m.doc) with
+      match (aux_expansion_of_module env ~strengthen:true m, m.doc) with
```
Julow, line 64: The comment needs to be updated.

src/xref2/subst.ml
```diff
@@ -862,15 +875,24 @@ and rename_bound_idents s sg =
       in
       rename_bound_idents s
         ( Include
-            { i with Component.Include.expansion_ = { items; removed = [] } }
+            {
+              i with
+              Component.Include.expansion_ =
+                { items; removed = []; compiled = false };
```
Julow, line 881: Shouldn't be copying `compiled` ? (this probably doesn't matter here because it's handled in `apply_sig_map`)

src/xref2/subst.ml
```diff
@@ -199,7 +202,9 @@ let rec resolved_module_path :
   | `Hidden p1 -> `Hidden (resolved_module_path s p1)
   | `Canonical (p1, p2) ->
       `Canonical (resolved_module_path s p1, module_path s p2)
-  | `OpaqueModule m -> `OpaqueModule (resolved_module_path s m)
+  | `OpaqueModule m ->
+      if s.unresolve_opaque_paths then raise Invalidated
```
Julow, line 206: This exception is raised and caught in a lot of places, what about turning it into a value to ensures the new raises are handled ? (like `or_replaced`)

src/xref2/tools.ml
```diff
@@ -1179,7 +1189,8 @@ and fragmap :
               if handled' then
                 map_include_decl i.decl sub >>= fun decl ->
                 let expansion_ =
-                  Component.Signature.{ items = items'; removed = removed' }
+                  Component.Signature.
+                    { items = items'; removed = removed'; compiled = false }
```
Julow, line 1193: Should it be `expansion_.compiled` ?

test/xref2/with.t/run.t
```diff
@@ -0,0 +1,48 @@
...
+be able to link anywhere because we don't render the inner signature anywhere.
+
+  $ cat test.ml
+  module type S = sig
+    type t
+    type u
+  end with type t = int with type u = int
+  
+
+  $ cat compile.sh
```
Julow, line 12: It is possible to write that file from here, so it is written only once:
```
  $ cat > compile.sh <<EOF
  #!/bin/bash
  
  ocamlc -c -bin-annot test.ml
  odoc compile page.mld --child module-test
  odoc compile -I . --parent page test.cmt
  odoc link -I . test.odoc
  odoc html-generate --indent -o html test.odocl
  odoc support-files -o html
  EOF
```

test/xref2/with.t/run.t
```diff
@@ -0,0 +1,48 @@
...
+  
+  ocamlc -c -bin-annot test.ml
+  odoc compile page.mld --child module-test
+  odoc compile -I . --parent page test.cmt
+  odoc link -I . test.odoc
+  odoc html-generate --indent -o html test.odocl
+  odoc support-files -o html
+  
+
+  $ ./compile.sh
```
Julow, line 23:
```suggestion
  $ bash ./compile.sh
```
Absolute paths to `bash` can't work in general, a less wrong line would be `#!/usr/bin/env bash`, but it still won't work in some distros. It's generally better to not use them.

src/xref2/tools.ml
```diff
@@ -1270,28 +1281,35 @@ and fragmap :
...
-  let items = map_items Subst.add_module_substitution items in
+  let items = map_items items in
 
   let res =
     Subst.signature sub
-      { Component.Signature.items; removed = removed @ sg.removed }
+      {
+        Component.Signature.items;
+        removed = removed @ sg.removed;
+        compiled = false;
```
Julow, line 1311: `sg.compiled` ?

src/xref2/subst.ml
```diff
@@ -698,12 +706,17 @@ and include_ s i =
   {
     i with
     decl = include_decl s i.decl;
-    expansion_ = apply_sig_map s i.expansion_.items i.expansion_.removed;
+    expansion_ =
+      apply_sig_map s i.expansion_.items i.expansion_.removed
+        i.expansion_.compiled;
```
Julow, line 711: There could be a `apply_sig_map'` function taking care of unpacking the signature fields

src/xref2/expand_tools.ml
jonludlam: true!

src/xref2/subst.ml
jonludlam: Yes, we should do this. The mechanism pre-exists though, so I think we can do it in a follow-up PR.

test/xref2/with.t/run.t
jonludlam: true - though I find it convenient to have the script handy for testing - I often seem to find it tricky to locate the dir in which dune has executed the test!

src/xref2/tools.ml
jonludlam: not without a bit more logic - there may be some circumstances under which we don't need to recompile (if there are only type equalities added?) but most we will do.

src/xref2/subst.ml
jonludlam: Yes I think you're right - we should preserve it here, then we can override it with the logic in `apply_sig_map` later.

test/xref2/with.t/run.t
jonludlam: It doesn't require bash, I'll change the #! line to /bin/sh - that should be sufficient.

src/xref2/tools.ml
jonludlam: no, same reason as the other one in fragmap.

src/xref2/subst.ml
jonludlam: done!

PR #570, opened 2021-01-27
Fix `html-targets` when operating on odoc files

jonludlam:
For backward compatibility, `odoc html-targets` can work on odoc files
as well as odocl files. When working on odoc files, it does the link
step (without writing the odocl file) then proceeds as usual on the
result. This is obviously inefficient, but the result is accurate.


Drup:
So that's what this code is for! :D 

If I'm reading the code correctly, it applies to all the `targets` command, right ? I think this is nice.

Issue #125, opened 2018-04-17
Make sure in-page search works, despite any markup


aantron:
Ideally, we could even somehow test for this automatically. Ideas welcome.

Noted by @rleonid in https://discuss.ocaml.org/t/1841/10.

rizo:
Is this just about the `->` markup using a regular dash sign? If so, I believe it's already fixed in master. If not, could you please give other examples?

dbuenzli:
It may be larger than that see https://github.com/ocaml/odoc/issues/19#issuecomment-307637452

aantron:
That was the only specific example, to my knowledge.

yawaramin:
In page search is not working. E.g. go to https://b0-system.github.io/odig/doc/dolog/Log/index.html#val-info . Then search in page for `val info`. No results.

hcarty:
@yawaramin I get `val info` highlighted + found when searching that page in Firefox.  Do you potentially have a browser plugin altering the page?

yawaramin:
@hcarty you're right, it's working in FF. I also checked and confirmed it's working in Safari. However it is not working in Chrome incognito with no extensions. Strange. Could be a Chrome bug...

hcarty:
Ah - yep, confirmed that it's not working on Chrome here either.

dbuenzli:
I recently noticed this aswell it seems if you set `display: inline-block` on `span`'s it inhibits in-page search across these spans. 

If someone knows where this can be filed against `chrome` here's a repro case (I submited an issue via chrome's UI but the experience felt like writing to  `/dev/null`) :
```
<!DOCTYPE html>
<html>
    <head>
      <meta charset="utf-8">
      <title>inline-block search bug</title>
      <style type="text/css">
        span { display: inline-block }
      </style>
    </head>
    <body>
      <p>
        <span>Is</span> <span>this</span> <span>searchable?</span>
      </p>
      <p>
        Is this searchable?
      </p>
    </body>
</html>
```

yawaramin:
@dbuenzli Chrome bugs are filed against the Chromium project, so https://bugs.chromium.org/p/chromium/issues/list . I tried a couple of different searches there but not finding anything more relevant than https://bugs.chromium.org/p/chromium/issues/detail?id=809886 , which talks about text-selection weirdness for inline-block spans, which also seems to present on odoc pages. May be related.

dbuenzli:
[...]

PR #579, opened 2021-01-28
Attempt at getting better layout by adding some spans

Drup:
This is not perfect by a long shot, but it's marginally better than before.
The diff is ridiculously large because it adds lot's of extra spans.

@lpw25 If you could test that on your problematic rendering, that would be good.


test/html/expect/test_package+custom_theme,ml/Include/index.html
```diff
@@ -19,68 +19,68 @@
...
     <div class="spec include">
      <div class="doc">
       <details open="open">
        <summary>
-        <span class="def"><code><span class="keyword">include</span> <a href="module-type-Not_inlined/index.html">Not_inlined</a></code></span>
+        <span class="def"><code><span><span class="keyword">include</span> <a href="module-type-Not_inlined/index.html">Not_inlined</a></span></code></span>
        </summary>
        <div class="spec type" id="type-t">
-        <a href="#type-t" class="anchor"></a><code><span class="keyword">type</span> t</code>
+        <a href="#type-t" class="anchor"></a><code><span><span class="keyword">type</span> t</span><span></span><span></span><span></span></code>
```
Octachron, line 37: Is it possible to elide those empty spans?

Drup: Should be done now. maybe even too aggressively (it removes empty code blocks too...)

dbuenzli:
@Drup could you give a rough idea of what you did ? I was thinking we should maybe try to find a new rendering strategy (separation of content and styling is a myth).

We tried [two](https://github.com/ocaml/odoc/issues/19) so far. I think the last one with `inline-block` `span`s generally does a good job (I don't think `odoc`'s stylesheet does it) but sadly it [breaks](https://github.com/ocaml/odoc/issues/125#issuecomment-504468736) in-page search on Chrome, the search doesn't work across the `inline-block`ed spans.

Drup:
@dbuenzli Ah, I see I was not precise enough: I just added back some of the spans that disappeared in the turmoil. Nothing else. This is only about browser-indentation, to make it less horrible than before.

dbuenzli:
> We tried [two](https://github.com/ocaml/odoc/issues/19) so far. I think the last one with `inline-block` `span`s generally does a good job (I don't think `odoc`'s stylesheet does it) but sadly it [breaks](https://github.com/ocaml/odoc/issues/125#issuecomment-504468736) in-page search on Chrome, the search doesn't work across the `inline-block`ed spans.

Just FTR I found a way to fix that see [here](https://github.com/ocaml/odoc/issues/125#issuecomment-770105925).

jonludlam:
I've rebased and fixed the tests on old OCaml versions in #604 so closing this one.

dbuenzli:
@Drup honestly I dont really see what this brought to the party (and FWIW, it broke, the search wise but visual wise excellent `display: inline-block` approach). This rather seem to have derived into odd `span`itis. For example having functions arrows are inside a span with their left argument which seems quite odd.

One thing I now realized is that `&nbsp` does seem c&p as space. So that's something we should play with to avoid odd cuts. E.g. between type parameters and type constructors.

Issue #508, opened 2020-10-16
overridden values of modules included in another module still get rendered


lubegasimon:
Overridden values of modules included in another module get rendered, thus causing duplicates.
For instance, considering `Module Odoc_compat.Char`, `lowercase_ascii` is duplicated.

![Screenshot_2020-10-16 Char (odoc Odoc_compat Char)](https://user-images.githubusercontent.com/34124177/96234411-1072a700-0fa2-11eb-94b9-3fabbd6e3a2a.png)

PR #580, opened 2021-02-03
stop rendering overridden module vals

lubegasimon:
Fixes https://github.com/ocaml/odoc/issues/508

*_after fix_*:

Screenshot 2021-02-03 at 14 30 54


src/model/ident_env.cppo.ml
```diff
@@ -165,6 +176,8 @@ let extract_signature_tree_item item =
             | Some id -> `Module (id, false) :: items
             | None -> items)
           mds []
+    | Tsig_value {val_id; _} ->
+        [`Value (val_id, false)]
```
Julow, line 180: This is in the `#if OCAML_MAJOR = 4 && OCAML_MINOR >= 10` branch, this will be removed on lower versions and the pattern will not be exhaustive.
I think you can move it to after the `#endif`.

src/model/ident_env.cppo.ml
```diff
@@ -220,8 +250,15 @@ let extract_structure_tree_item item =
 #else
     | Tstr_type (_rec_flag, decls) -> (* TODO: handle rec_flag *)
 #endif
-        List.map
-          (fun decl -> `Type (decl.typ_id, false)) decls
+        List.map (fun decl -> `Type (decl.typ_id, false)) decls
+
+#if OCAML_MAJOR = 4 && OCAML_MINOR = 02
```
Julow, line 255: This doesn't match 4.02.1 - 4.02.3. 
```suggestion
#if OCAML_MAJOR = 4 && OCAML_MINOR < 03
```

src/model/names.ml
```diff
@@ -31,12 +48,12 @@ module Name : Name = struct
...
   let to_string = function
     | Std s -> s
-    | Internal (s, i) -> Printf.sprintf "$%s$%d" s i
+    | Internal (s, i) -> Printf.sprintf "{%s{%d" s i
 
   let to_string_unsafe = function Std s -> s | Internal (s, _i) -> s
 
   let of_string s =
-    if String.length s > 0 && s.[0] = '$' then raise (Invalid_argument s)
+    if String.length s > 0 && s.[0] = '{' then raise (Invalid_argument s)
```
Julow, line 56: This is checking that we don't do `of_string (to_string name)` ? Perhaps we can remove this check and rename the function `std_of_string` (there is `internal_of_ident`).

Julow (commented):
Most of the CI failures seems to come from Mdx, can you try rebasing on top of https://github.com/ocaml/odoc/pull/591 ?

src/model/names.ml
lubegasimon: >  Perhaps we can remove this check

Hmm! I still find this right as is, could you please clarify.

lubegasimon:
> Most of the CI failures seems to come from Mdx, can you try rebasing on top of #591 ?

It happens, that this has to be merged first.

src/model/names.ml
Julow: I mean that this function must not be called from the result of `to_string` so instead of checking that, we could rename the function to show that it has nothing to do with `to_string`. I suggested using the same convention as `internal_of_string` and `internal_of_ident` but `make_internal` and `make_std` would be better.

Julow:
You can import these changes in this PR like this:
```
git fetch -f origin pull/591/head
git rebase FETCH_HEAD
```
Git will be smart enough to remove them from your branch when #591 is merged and you rebase again on top of master

src/model/names.ml
lubegasimon: I don’t mean to be adamant, but I think `of_string` and `of_ident` agree to the convention that I’ve seen being used in OCaml modules, for example `of_string` taking string as argument and returns t (for example).

> I mean that this function must not be called from the result of to_string

It’s also called from the result of `Ident.name x`.   `let of_ident id = of_string (Ident.name id)`

Julow: `of_string` is supposed to make a `t` from the result of `to_string`. Here that's not what we want, we want a "constructor" kind of function, create a `t` from an input that mean something specific.

lubegasimon: Ah, I see!

lubegasimon: I think this should be addressed in a separate PR since it’s irrelevant to the PR and issue title

src/model/ident_env.cppo.ml
```diff
@@ -19,6 +19,7 @@ open Names
 
 module Id = Paths.Identifier
 module P = Paths.Path
+open Typedtree
```
Julow, line 22: This is only used for the new `read_pattern` function. I suggest to add a `let open Typedtree in` there and to remove this `open` for consistency.

src/model/ident_env.cppo.ml
```diff
@@ -275,7 +313,7 @@ let extract_structure_tree_item item =
     | Tstr_open o ->
       ((extract_extended_open o) :> extracted_items list)
 #endif
-    | Tstr_eval _ | Tstr_value _
+    | Tstr_eval _
     | Tstr_primitive _ | Tstr_typext _
```
Julow, line 317: `Tstr_primitive` binds a value too.

src/model/names.ml
Julow: Now that you made the change to the name of the function, I think this check is no longer needed.

Julow (approved):
Looks good !

src/model/ident_env.cppo.ml
lubegasimon: This statement isn’t clear

Julow: `Tstr_primitive` correspond to:
```ocaml
external f : unit -> unit = "%primitive_name"
```
It should be treated as a value (as `val f : unit -> unit`)

src/model/names.ml
jonludlam: I'm a little nervous of removing the check - the original aim was to spot bugs, and the sort of bug it was looking for is certainly still possible. However, thinking a little more, I wonder whether we could incorporate this with a more general check - we could ensure at construction time that the names conform to OCaml's lexical conventions. This would include adding parentheses when necessary. This could catch bad references, such as `{!value-Foo}`.

Julow: This function is no longer called `of_string` so at least the bug it was looking for is now less probable. I don't like adding such checks because it is checking for something that changes (OCaml's rules) and won't be correctly tested (because it's too complicated). I think this part of the code (and everything that is not the parser) shouldn't care for which characters are in these strings.

jonludlam: To be clear - if we _do_ go along this path though, it should be in a different PR.

lubegasimon: Noted!

PR #583, opened 2021-02-08
Stop rendering dead links

### LINKED ISSUE #560: Dead links to hidden items appearing in documentation
In the `Field` module in base we have the following code:
```ocaml
(**/**)
module For_generated_code : sig
  (*_ don't use this by hand, it is only meant for ppx_fields_conv *)

  type ('perm, 'record, 'field) t =
    { force_variance : 'perm -> unit
    ; name : string
    ; setter : ('record -> 'field -> unit) option
    ; getter : 'record -> 'field
    ; fset : 'record -> 'field -> 'record
    }

  val opaque_identity : 'a -> 'a
end
(**/**)

(** ['record] is the type of the record.  ['field] is the type of the
    values stored in the record field with name [name]. ['perm] is a way
    of restricting the operations that can be used. *)
type ('perm, 'record, 'field) t_with_perm =
  | Field of ('perm, 'record, 'field) For_generated_code.t
[@@unboxed]
```

The `For_generated_code` is correctly left out of the documentation, however the `t_with_perm` type is documented as:
```
type ('perm, 'record, 'field) t_with_perm =
  | Field of ('perm, 'record, 'field) For_generated_code.t
```
with `For_generated_code.t` a dead link. It should instead be documented as:
```
type ('perm, 'record, 'field) t_with_perm
```

lubegasimon:
fixes #560

### DIFF
--- a/src/compat/odoc_compat.ml
+++ b/src/compat/odoc_compat.ml
@@ -1,7 +1,7 @@
 (**
-   Compatibility module reexporting ~equivalent functions based on the current
-   OCaml version
- *)
+  Compatibility module reexporting ~equivalent functions based on the current
+  OCaml version
+*)
 module String =
 struct
   include String
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -459,6 +459,27 @@ module Make (Syntax : SYNTAX) = struct
             O.documentedSrc (cstr ++ O.txt " " ++ O.keyword "of" ++ O.txt " ")
             @ record fields
 
+    let rec read_typ_exp typ_expr =
+      let open Lang.TypeExpr in
+      let open Paths.Path in
+      match typ_expr with
+      | Constr (p, ts) ->
+          is_hidden (p :> Paths.Path.t)
+          || List.exists (fun t -> read_typ_exp t) ts
+      | Poly (_, t) | Alias (t, _) -> read_typ_exp t
+      | Arrow (_, t, t2) -> read_typ_exp t || read_typ_exp t2
+      | Tuple ts | Class (_, ts) -> List.exists (fun t -> read_typ_exp t) ts
+      | _ -> false
+
+    let internal_cstr_arg t =
+      let open Lang.TypeDecl.Constructor in
+      let open Lang.TypeDecl.Field in
+      match t.args with
+      | Tuple type_exprs ->
+          List.exists (fun type_expr -> read_typ_exp type_expr) type_exprs
+      | Record fields ->
+          List.exists (fun field -> read_typ_exp field.type_) fields
+
     let variant cstrs : DocumentedSrc.t =
       let constructor id args res =
         match Url.from_identifier ~stop_before:true id with
@@ -476,6 +497,7 @@ module Make (Syntax : SYNTAX) = struct
       | _ :: _ ->
           let rows =
             cstrs
+            |> List.filter (fun cstr -> not (internal_cstr_arg cstr))
             |> List.map (fun cstr ->
                    let open Odoc_model.Lang.TypeDecl.Constructor in
                    let url, attrs, code =
@@ -693,13 +715,15 @@ module Make (Syntax : SYNTAX) = struct
               | Variant cstrs -> variant cstrs
               | Record fields -> record fields
             in
-            O.documentedSrc
-              ( O.txt " = "
-              ++
-              if need_private then
-                O.keyword Syntax.Type.private_keyword ++ O.txt " "
-              else O.noop )
-            @ content
+            if List.length content > 0 then
+              O.documentedSrc
+                ( O.txt " = "
+                ++
+                if need_private then
+                  O.keyword Syntax.Type.private_keyword ++ O.txt " "
+                else O.noop )
+              @ content
+            else []
       in
       let tconstr =
         match t.equation.params with
--- a/src/loader/doc_attr.ml
+++ b/src/loader/doc_attr.ml
@@ -28,7 +28,7 @@ let empty : Odoc_model.Comment.docs = empty_body
 
 
 
-let load_payload : Parsetree.payload -> (string * Location.t) option = function
+let load_payload : Parsetree.payload -> string * Location.t = function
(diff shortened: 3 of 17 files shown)


test/html/test.ml
```diff
@@ -289,6 +289,9 @@ let source_files_post408 =
 
 let source_files_pre410 = [ ("bugs_pre_410.ml", [ "Bugs_pre_410/index.html" ]) ]
 
+let source_files_stop_dead_link_doc =
```
Julow, line 292: The convention would be to call this `source_files_post404`. This way more tests can be added to this list.

src/model/names.ml
```diff
@@ -31,12 +48,12 @@ module Name : Name = struct
 
   let to_string = function
     | Std s -> s
-    | Internal (s, i) -> Printf.sprintf "$%s$%d" s i
+    | Internal (s, i) -> Printf.sprintf "{%s{%d" s i
```
jonludlam, line 51: Can we change this to `{%s}%d` please? It just looks a bit wrong as it is!

jonludlam:
This'll need a quick rebase now the previous PR has been merged.

lubegasimon:
> This'll need a quick rebase now the previous PR has been merged.

Done

src/loader/ident_env.cppo.ml
```diff
@@ -152,6 +155,7 @@ let extract_extended_open o =
       extract_extended_open_items o.open_bound_items
 #endif
 
+let doc_off_mode_on = ref false
```
jonludlam, line 158: I don't really like this global var - we could instead simply have this as a parameter to `extract_signature_tree_item`

jonludlam: Particularly since the scope of the stop comment is only supposed to be until the end of the current module/module type/class/class type (https://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#ss:ocamldoc-stop)

jonludlam (changes requested):
Good so far, just a couple of changes needed.

src/compat/odoc_compat.ml
```diff
@@ -2,30 +2,30 @@
...
-  include Char
-
-#if OCAML_MAJOR = 4 && OCAML_MINOR = 02
-  let lowercase_ascii = lowercase
-#else
-  let lowercase_ascii = lowercase_ascii
-#endif
-
-end
+ module String =
```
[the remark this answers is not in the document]
lubegasimon: I think that was a result of running `dune build @fmt`, and then `dune promote`

src/document/generator.ml
```diff
@@ -472,6 +487,7 @@ module Make (Syntax : SYNTAX) = struct
       | _ :: _ ->
           let rows =
             cstrs
+            |> List.filter (fun cstr -> not (internal_cstr_arg cstr))
```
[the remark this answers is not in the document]
lubegasimon: We remove a constructor if it has only one hidden argument, if it has more than that, then we remove the hidden argument itself.

src/document/generator.ml
```diff
@@ -689,13 +705,15 @@ module Make (Syntax : SYNTAX) = struct
...
-            @ content
+            if List.length content > 0 then
+              O.documentedSrc
+                ( O.txt " = "
+                ++
+                if need_private then
+                  O.keyword Syntax.Type.private_keyword ++ O.txt " "
+                else O.noop )
+              @ content
+            else O.documentedSrc (O.txt "")
```
Julow, line 716:
```suggestion
            else []
```

src/loader/doc_attr.ml
```diff
@@ -28,7 +28,7 @@ let empty : Odoc_model.Comment.docs = empty_body
 
 
 
-let load_payload : Parsetree.payload -> (string * Location.t) option = function
+let load_payload : Parsetree.payload -> string * Location.t = function
```
Julow, line 31: Why this change ? It's unrelated and move an `assert false` deeper in the code.

src/compat/odoc_compat.ml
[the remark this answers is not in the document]
Julow: `dune build @fmt` shouldn't modify this file, I think you can safely revert this.

src/document/generator.ml
```diff
@@ -455,6 +455,21 @@ module Make (Syntax : SYNTAX) = struct
...
+      | Tuple type_exprs ->
+          List.exists
+            (fun type_expr ->
+              match type_expr with
+              | Constr (p, _) ->
+                  let open Paths.Path in
+                  is_hidden (p :> Paths.Path.t)
+              | _ -> false)
+            type_exprs
+      | Record _ -> false
```
Julow, line 471: We should handle this case too:
```
type t =
  Field of { field : For_generated_code.t }
```

src/document/generator.ml
```diff
@@ -455,6 +455,21 @@ module Make (Syntax : SYNTAX) = struct
...
+      let open Lang.TypeExpr in
+      match t.args with
+      | Tuple type_exprs ->
+          List.exists
+            (fun type_expr ->
+              match type_expr with
+              | Constr (p, _) ->
+                  let open Paths.Path in
+                  is_hidden (p :> Paths.Path.t)
+              | _ -> false)
```
Julow, line 469: What about the other cases that recursively contain an `TypeExpr.t` ? `Alias`, `Arrow`, `Tuple`, `Constr`/`Class` arguments.

For example:
```ocaml
type t = Field of (int * For_generated_code.t) * int (* Tuple *)
type t = Field of For_generated_code.t option (* Constr argument *)
```

src/document/generator.ml
lubegasimon: good catch

src/loader/doc_attr.ml
lubegasimon: The motive that led to this change was so as to not parse the content doc comments directly, rather check for `/*`

jonludlam (approved):

Issue #374, opened 2019-08-09
Allow to generate page fragments


hhugo:
In particular, it would be useful to integrate odoc a bit into existing designs such as   http://ocsigen.org/js_of_ocaml/dev/manual/overview

dbuenzli:
The css support is already here, see `odoc support-files --whithout-theme` and `odoc html --theme-uri`. I thought we already had an issue tracking generating a page fragment rather than a full page but I can't find it.

hhugo:
It is https://github.com/ocaml/odoc/issues/327 ?

dbuenzli:
Not really. If you look at the sources of current pages like [this one](https://b0-system.github.io/odig/doc@odig.dark/alcotest/Alcotest/index.html) everything it already cleanly in its own `div`, #327 is just a request to have a better page structure. 

One current slight problem though is the toplevel class named `content` on the `div` that needs fixing and be replaced by `odoc` for better stylesheet isolation/composition this is discussed in https://github.com/ocaml/odoc/issues/298.

I think the simplest here would be to simply add a `--fragment` option to `odoc html`. The effect of that option would be the following: 

1. Rather than generating the files with the `.html` suffix it generates the files with the suffix `.htmlf` (or `.inc` or `.txt` or `.frag.html`). 
2. These files are exactly those that are generated now (in particular links) except only the current toplevel `div` is in there.

Now the build system (or the `php` script :-) of your website can simply read each of the `$PATH.htmlf` files and generate a corresponding `$PATH.html` file with your website chrome added.

Note that are other proposals floating around in the treasure trove of *ideas* that this issue tracker is... E.g. [`json` output](https://github.com/ocaml/odoc/issues/196) or this https://github.com/ocaml/odoc/issues/121 issue. I personally think as a first good step all that should be closed and simply the above be implemented, I'm sure it will already go a long way for most "website chrome adding" needs.

dbuenzli:
@hhugo what do you think about the proposed scheme ? This option could then also be added to `odig odoc`. I suspect it should not be too hard so I could try to squeeze this in along with #298 for the next release.

trefis:
The `--fragment` proposal sounds very reasonable (and useful) to me.

I have no opinion on the CSS `content →odoc` renaming.

dbuenzli:
> I have no opinion on the CSS `content →odoc` renaming.

It's a class so we can have both, I'll just PR to add the `odoc` class along side the `content`.

[...]

PR #589, opened 2021-02-12
Add classes to the odoc HTML page structure.

### LINKED ISSUE #298: Replace the `content` class by `odoc`
Currently the main body in the output is classified as `content` which is first pretty meaningless and second, quite bad the day you want to compose the outputs and the css with other documents. I suggest this should be `odoc`.

dbuenzli:
/cc @drup I'm being told by @jonludlam

1. Classify the root odoc tree (currently `body`) as `odoc`. This closes #298 and will help stylesheet management/composition if #374 ever becomes a reality.
2. Classify toplevel `nav` as `odoc-nav` (new class)
3. Classify module preamble `header` as `odoc-preamble` (new class)
4. Reclassify the toc `nav` from `toc` to `odoc-toc`.
5. Reclassify the content `div` rom `.content` to `.odoc-content`.

Altogether this allows to replace complicated CSS selectors like:

```
body > nav:first-child { /* toplevel nav */ }
```
By: 
```
.odoc-nav { /* toplevel-nav */ }
```

Besides 1. allows to specify `odoc` page geometry without having to refer to the root tree kind o element (currently `body` which you won't control if you get embedded).

# Pro-tip

To deal with the utterly user-friendly expect test system when you change the markup you want:

```
CMD=$(make test 2>&1 | grep 'cp '); \
while [ -n "$CMD" ]; do CMD=$(sh -c "$CMD" 2>&1 | grep 'cp '); done
```

It's slow but at least you don't have to c&p a cli invocation 50 times. Someone ought to fix that madness (I tried various forms of `--auto-promote`, `dune promote` without any success).

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -350,7 +350,7 @@ module Toc = struct
     in
     match toc with
     | [] -> []
-    | _ -> [ Html.nav ~a:[ Html.a_class [ "toc" ] ] [ sections toc ] ]
+    | _ -> [ Html.nav ~a:[ Html.a_class [ "odoc-toc" ] ] [ sections toc ] ]
 
   let on_sub : Subpage.status -> bool = function
     | `Closed | `Open | `Default -> false
--- a/src/html/tree.ml
+++ b/src/html/tree.ml
@@ -104,15 +104,16 @@ let page_creator ?(theme_uri = Relative "./") ~url name header toc content =
                else [ [ Html.txt lbl ] ])
         |> List.flatten
       in
-      [ Html.nav l ]
+      [ Html.nav ~a:[ Html.a_class [ "odoc-nav" ] ] l ]
     else []
   in
 
   let body =
-    breadcrumbs @ [ Html.header header ] @ toc
-    @ [ Html.div ~a:[ Html.a_class [ "content" ] ] content ]
+    breadcrumbs @ [ Html.header ~a:[ Html.a_class [ "odoc-preamble" ] ] header ]
+    @ toc
+    @ [ Html.div ~a:[ Html.a_class [ "odoc-content" ] ] content ]
   in
-  Html.html head (Html.body body)
+  Html.html head (Html.body ~a:[ Html.a_class [ "odoc" ] ] body)
 
 let make ?theme_uri ~indent ~url ~header ~toc title content children =
   let filename = Link.Path.as_filename url in
--- a/src/odoc/etc/odoc.css
+++ b/src/odoc/etc/odoc.css
@@ -689,7 +689,7 @@ h1+.modules, h1+.sel {
 
 /* Sidebar and TOC */
 
-.toc:before {
+.odoc-toc:before {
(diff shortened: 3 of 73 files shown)


dbuenzli:
The `ocaml-ci` failures seem to be due to `mdx` which I'm not sure exactly why `odoc` depends on nowadays. I think it would be good to review a bit the testing structure to simplify it and trim a few dependencies. The fact that there seem to be a `mdx` git module in `odoc` should raise some flags.

jonludlam:
The mdx submodule is temporary. The problem is that we're using mdx for some tests, but mdx depends on the odoc comment parser. The solution is to undo the merge of octavius into odoc and distribute the comment parser as a separate project. Mdx can then depend upon that, and odoc can depend upon mdx, and the dependency cycle will be removed. It's a few days work though as the parser is tangled up with the model and it needs a bit of unravelling.

dbuenzli:
> The problem is that we're using mdx for some tests, but mdx depends on the odoc comment parser.

Dropping the dependency seems a simpler course of action… When I'm testing `odoc` I don't want to test `mdx`.

jonludlam:
yes, but we want to use mdx to test odoc :-)

jonludlam:
But this has been a constant source of problems. I'm going to nuke the tests until we can remove the parser.

Drup (approved):
Sounds good to me. :+1:

As for the tests .... I fully agree, I already apply your "pro tip", and it's still a real pain. I don't even actually understand the scripting used for testing and why it's so bad. :/

dbuenzli:
> yes, but we ~want to~ do use mdx to test odoc :-)

Well I'll let you judge. But frankly migrate them to something simpler. I heard `dune` has built-in support for [cram testing](https://dune.readthedocs.io/en/stable/tests.html#cram-tests). That sounds good enough. I'm not sure I see what `mdx` brings to the party.

The leaner the tooling, the easiest it is to contribute and maintain. I mean I had to install 2 non-opam packages on my system, a bunch of opam packages including one that has a cyclic dependency on `odoc` and come up with a shell snippet just to be able to run and fix the test suite for a trivial PR.

> I'm going to nuke the tests until we can remove the parser.

The parser feels right in `odoc`. When we had octavius in another repo we had issues flying everywhere; it's not worth the admin overhead, keep it simple.

lubegasimon:
> I heard `dune` has built-in support for [cram testing](https://dune.readthedocs.io/en/stable/tests.html#cram-tests). That sounds good enough.

Right, it’s something I plan on doing in the coming days, just that there seems to be pressing tasks that Jon suggested to be addressed before the next `odoc` release. cc @jonludlam

Drup:
@jonludlam Are we waiting for the new testing for this ? We could merge it as-is (modulo rebase, ofc).

jonludlam:
I'm going through the PRs now, with some manual testing.

jonludlam:
Thanks @dbuenzli

PR #591, opened 2021-02-15
Remove mdx

jonludlam:
mdx currently requires the comment parser that is part of odoc, and until this commit, odoc required mdx as a test dependeny. This PR is to remove odoc's dependency on mdx until the parser can be extracted into a separate repository. The intention is to re-add the mdx requirement as soon as the circular dependency problem is resolved, hence I have left disabled dune files in place where that made sense.


jonludlam:
It'd be really useful to get someone more familiar with esy to check this over - in particular [this](https://github.com/ocaml/odoc/pull/591/commits/58c4c2bfe317b9137a88f2c9bdc840bbaf470283#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L97) seemed suspect to me.

.github/workflows/build.yml
```diff
@@ -104,4 +94,4 @@ jobs:
         run: esy install --verbose
 
       - name: Build odoc
-        run: esy odoc --version
+        run: esy x dune exec src/odoc/bin/main.exe -- --version
```
Julow, line 97: This doesn't seem right. I managed to make it work with this: https://github.com/Julow/odoc/commit/cd8fb9b28fe1e5fe7febb22266731c7b05507e74 ([esy build result](https://github.com/Julow/odoc/runs/1911160080))

jonludlam: That looks much better - thanks!

Issue #585, opened 2021-02-11
References to aren't working as expected


jonludlam:
For example, use odig and look at the `Astring` documentation - the first link is broken `{{!strf}one value}`

dbuenzli:
I'll take the opportunity to mention https://github.com/ocaml/odoc/issues/332 and https://github.com/ocaml/odoc/issues/332. 

In general the resolution strategy for {!id} for needs to be clarified an properly documented, since it differs from the OCaml language. 

AFAIR `odoc`'s one always differed from `ocamldoc`'s one. With the years I gradually fixed my links to match `odoc`'s notion but it can become slightly verbose since there are cases where you need to qualify more (of course now I can't find an example). 

OTOH I vaguely remember discussing that with @trefis which considered `ocamldoc`'s to be too ambiguous and/or broken.

dbuenzli:
I just fell on the case that always bothered me with `odoc`. Basically the problem is that `{!id}` is oblivious of `open`s. 

Suppose I have: 

```ocaml
module Ns : sig
  module A : sig 
  end
end
```

Now in the following module `{!A}` does not get resolved: 

```ocaml
module B sig 
   open Ns

   (** see {!A} *)

end
```

You have to write `{!Ns.A}`. I would like to be able to write {!A} that it renders as `A` but links to `Ns.A`.

It has been a long time since I haven't used `ocamldoc` but I believe it worked that way since I had to "fix" a lot of my mlis.

(I did not test with 2.0.0 yet)

Julow:
I opened a new issue for your last comment (https://github.com/ocaml/odoc/issues/594) so this issue can be closed by https://github.com/ocaml/odoc/pull/592

PR #592, opened 2021-02-17
Fix unresolved references from the first comment of a file

Julow:
This fixes https://github.com/ocaml/odoc/issues/585 but doesn't fix the issue reported by this comment https://github.com/ocaml/odoc/issues/585#issuecomment-777599121

Resolution failed because doc comment of a file was resolved before the rest of the module is added to the environment.
The first commit is unrelated, it removes the location from `odoc_print`'s output to make the output more readable.

The second issue seems to be in the loader, I'm still investigating.

### DIFF
--- a/src/model_desc/comment_desc.ml
+++ b/src/model_desc/comment_desc.ml
@@ -3,13 +3,7 @@ open Odoc_model
 open Comment
 open Paths_desc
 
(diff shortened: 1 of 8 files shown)


jonludlam:
Looks good, thanks!

PR #593, opened 2021-02-18
Improve and classify page content

dbuenzli:
This PR is built on top of #589. The relevant commits are from the 15 of february.

It does the following (see the indiviual commits for more details).

1. Regularizes the markup for declarations (no special casing when there is no docstring).
2. Classifies the docstring div as `.spec-doc` (using `.doc` here was non trivial w.r.t. to the current usage of this class by the stylesheet maybe we can clarify/simplify later).
3. Classifies all the direct children of `.odoc-content`.

Regarding 3. I'm not happy about the `.odoc-unattached` `aside`s since they make markup structure differ between modules and mld pages. E.g. you can have `h1 (aside  p)` in modules and you will have `h1 p` in mlds, which makes more cases to treat than needed. But I'll try to unnest them in another PR.

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -323,7 +323,9 @@ and items ~resolve l : item Html.elt list =
         let elts =
           let doc = match doc with
           | [] -> []
-          | docs -> [ div (flow_to_item @@ block ~resolve docs) ]
+          | docs ->
+            let a = [ Html.a_class [ "spec-doc" ] ] in
+            [ div ~a (flow_to_item @@ block ~resolve docs) ]
           in
           [ div (div ~a content :: doc) ]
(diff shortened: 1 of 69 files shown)


Drup:
Looks good to me. I think we could make the code a little bit regular by adding a few "smart div constructors" and only use those everywhere. This would ensure changes are consistent, and that we cover everything.

jonludlam:
Format has failed - could you `dune build @fmt && dune promote` please?

dbuenzli:
Would you rather consider merging #600 that I just rebased @drup vouched both of them and it has these commits with a final format commit.

jonludlam:
that's fine!

Issue #478, opened 2020-08-31
Top-level doc-comments of module types are discarded


CraigFe:
With the following set of files:

```ocaml
(* --- main_intf.ml --------------------------------------------- *)
module type S = sig
  (** Top-level comment *)

  val t : unit
  (** Item comment *)
end

(* --- main.ml -------------------------------------------------- *)
let t = ()

(* --- main.mli ------------------------------------------------- *)
include Main_intf.S (** @inline *)
```

the generated documentation for `Main` looks like the following:

![image](https://user-images.githubusercontent.com/26125912/91742129-0b09fa80-ebb6-11ea-8150-9b8747fa53d6.png)

i.e. the top-level doc-comment has been lost. As a workaround, it's possible to move top-level doc comments to the `.mli` file, but this splits the documentation across two files unnecessarily.

Is this rendering behaviour intentional?

jonludlam:
Thanks @CraigFe - good timing, I'm just working on this right now - I'll get back with findings!

lpw25:
This behaviour is deliberate. The top-level comment in a module type or module is treated as a description of the module type or module itself. This is how you can have descriptions of top-level modules. In other words:
```ocaml
module M : sig
  (** module [M] *)
  ...
end
```
and
```ocaml
(** module [M] *)
module M : sig
  ...
end
```
are equivalent. This behaviour is possibly unnecessary for non-top-level modules, so perhaps it should be removed. Or perhaps we should remove it for module types and keep it for modules.

Even if we removed that behaviour, you still wouldn't get the behaviour that you want with an `_intf` module type: you don't just 
want the comment to be included in the resulting signature you want it to become the description comment for the module. For that I have previously suggested that we want a something stronger than `@inline` for `_intf` includes that doesn't only include the module contents inline but also treats the top-most comment as the description comment.

lpw25:
I propose the following solution to this:
1. The first comment of a signature should be considered to be part of the signature itself, it should be stored separately from the other comments. By extension, expansions of a module type would have first comments.
2. A module's main documentation should be any comment attached to its definition followed by the first comment of its expansion.
3. A module type's main documentation should just be any comment attached to its definition.
4. A signature that starts with an `@inline` include before any doc comments gets the first comment of that include as its first comment.
5. All the above should apply similarly to class types and classes with respect to the first comment in a class signature.

This is mostly a change of behaviour

Julow:
[...]

PR #595, opened 2021-02-22
Don't remove the doc of inline includes

Julow:
Fixes the initial issue of https://github.com/ocaml/odoc/issues/478 but doesn't implement @lpw25's suggestion.

The first doc-comment of the expansion of an include was always removed. This PR avoids this for `@inline` includes.

The first two commits are unrelated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,3 +159,9 @@ uutf/dune-project : uutf
 .PHONY : distclean
 distclean :
 	rm -rf $(DUNIVERSE_DEPS) dune-local
+
+.PHONY : promote-html
+promote-html:
+	EXPECTED=`cat _build/default/test/html/_scratch/expected`; \
+	ACTUAL=`cat _build/default/test/html/_scratch/actual`; \
+	mkdir -p "`dirname "$$EXPECTED"`" && cp "$$ACTUAL" "$$EXPECTED"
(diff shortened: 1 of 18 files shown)


jonludlam:
Are you intending to implement the rest of the suggested fix?

Julow:
Not in this PR.
@lpw25 Do you think we should change the way documentation in signatures work before 2.0.0 ? (like you suggested here: https://github.com/ocaml/odoc/issues/478#issuecomment-762744498)

lpw25:
I'm not entirely clear on the status after this PR, but I do think that we should have the rules for where to put comments settled before 2.0.0. We're about to ask people to update their docs to match the new rules to get good docs.ocaml.org output, and I think we should try to avoid having to ask them twice.

I think the important thing for this bit is that you should be able to add a preamble comment to a module like:

https://github.com/janestreet/base/blob/master/src/hash.mli

by adding a first comment to the associated module type here:

https://github.com/janestreet/base/blob/master/src/hash_intf.ml#L144

jonludlam:
It's likely we'll revisit this code with the other module doc changes, but we'll merge this now for the tests.

Julow:
> I think the important thing for this bit is that you should be able to add a preamble comment to a module like:

This should work fine since https://github.com/ocaml/odoc/pull/627

PR #596, opened 2021-02-22
Canonical module type and type

jonludlam:
Allows @canonical tags to be put on module type and types.


src/model/error.mli
```diff
@@ -1,5 +1,9 @@
 type t
 
+type 'a with_warnings = { value : 'a; warnings : t list }
+
+type warning_accumulator
+
```
Julow, line 6: Why these changes ?

Julow (approved):
I think we should name the new constructors `` `CanonicalType `` and `` `CanonicalModuleType `` instead. (also `T` and `Ty` are not consistent with other names we have, they should be `MT` and `T`)

src/model/error.mli
jonludlam: Oops, I think I meandered off into 'separate out the parser library' territory while doing this work, and didn't quite revert everything!

jonludlam:
Mmm, I think you're right about the names of the new constructors. Thanks for the feedback!

Issue #329, opened 2019-02-26
Improve markup for @-tags


dbuenzli:
The markup for @-tags is poor. No class labelling for the list of tags, no spans to distinguish the different tag constituents, no class to target/distinguish the different tags

Contrast the markup here:

https://docs.mirage.io/fmt/Fmt/index.html#val-set_utf_8
https://b0-system.github.io/odig/doc/fmt/Fmt/index.html#val-set_utf_8

or here with a few different tags like `@`since, `@`deprecated  (note the `@`raise is not used there though)

https://docs.mirage.io/ocaml/String/index.html
https://b0-system.github.io/odig/doc/ocaml/Stdlib/String/index.html

Also In general I'd move away from `dl` for *lists* of elements: they are harder to style than it should be, because list elements are made of two consecutive blocks of elements rather than a single element. It's much easier to style `li` elements with properly `span`ned and classified elements.

dbuenzli:
No progress here either. [This kind](https://github.com/c-cube/ocaml-containers/blob/b5ecb273ef0cf16351bb8295b79863c562f1218c/src/core/CCArray.mli#L62-L63) of @-tag in CCArray namely:

```
@since 1.2, but only
@since 2.1 with labels
```

ends up in the following markup:

```html
<dl>
  <dt>since</dt>
  <dd>1.2, but only</dd>
</dl>
<dl>
  <dt>since</dt>
  <dd>2.1 with labels</dd>
</dl>
```

The fact that there are two lists actually make it easier to style. But if we'd put that in a single list it would be annoying to style. 

But in any case this markup does not allow to use in page search in chrome to find e.g. "since 2.1" which is annoying (In general it seems that in page search in chrome only works across `display: inline` elements).

A tentative better markup would be to turn that into: 

```html
<ul class="odoc-tags">
  <li><span class="odoc-tag">since</span> , 1.2 but only</li> 
  <li><span class="odoc-tag">since</span> 2.1 with labels</li> 
</ul>
```

Drup:
It's a bit sad, since that is actually a good use case for `dl`, but I don't have any opposition to changing the markup to that.

dbuenzli:
> It's a bit sad, since that is actually a good use case for `dl`

Indeed but somehow `dl`'s model has always been weird from a structural point of view: sequences of *pairs* of elements. The lack of a surrounding box for each definition makes it harder to style than it can be (though a bit less a problem nowadays that CSS went level 42++)

[...]

PR #597, opened 2021-02-23
Render modules synopsis for `{!modules:...}`

Julow:
Fixes https://github.com/ocaml/odoc/issues/297

Resolve and render the synopsis of modules referenced with the `{!modules:A B C}` markup.

The first sentence of the first paragraph of the modules' documentation is taken. Headings, tags and other elements that are not paragraphs and lists are skipped, I think that's what ocamldoc does: https://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#sss:ocamldoc-preamble

### DIFF
--- a/src/model/comment.ml
+++ b/src/model/comment.ml
@@ -29,11 +29,20 @@ type inline_element =
   | `Reference of Reference.t * link_content
   | `Link of string * link_content ]
 
+type paragraph = inline_element with_location list
+
+type module_reference = {
+  module_reference : Reference.Module.t;
+  module_synopsis : paragraph option;
+}
+(** The [{!modules: ...}] markup. [module_synopsis] is initially [None], it is
+    resolved during linking. *)
+
 type nestable_block_element =
-  [ `Paragraph of inline_element with_location list
+  [ `Paragraph of paragraph
   | `Code_block of string
   | `Verbatim of string
-  | `Modules of Reference.Module.t list
+  | `Modules of module_reference list
   | `List of
     [ `Unordered | `Ordered ] * nestable_block_element with_location list list
   ]
--- a/src/document/comment.ml
+++ b/src/document/comment.ml
@@ -213,7 +213,11 @@ let module_references ms =
         (m.module_reference :> Odoc_model.Paths.Reference.t)
     and synopsis =
       match m.module_synopsis with
-      | Some synopsis -> [ block @@ Paragraph (inline_element_list synopsis) ]
+      | Some synopsis ->
+          [
+            block ~attr:[ "synopsis" ]
+            @@ Paragraph (inline_element_list synopsis);
+          ]
       | None -> []
     in
     (reference, synopsis)
--- a/src/document/comment.ml
+++ b/src/document/comment.ml
@@ -206,6 +206,18 @@ and inline_element_list elements =
        (fun elt -> inline_element elt.Odoc_model.Location_.value)
        elements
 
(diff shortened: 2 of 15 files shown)


Drup:
Looks good on my side :+1:

@dbuenzli An opinion on the resulting HTML ?

src/model/comment.ml
```diff
@@ -29,11 +29,20 @@ type inline_element =
   | `Reference of Reference.t * link_content
   | `Link of string * link_content ]
 
+type paragraph = inline_element with_location list
+
+type module_reference = {
+  module_reference : Reference.Module.t;
+  module_synopsis : paragraph option;
+}
```
Julow, line 37: This module is used for two things: the output of the parser and the representation of comments inside Odoc. Here I intent to change the latter but it's a breaking change for users of the parser (eg. OCamlformat and Mdx).

A fix to that would be to duplicate the previous version of this module into the parser library and adding a conversion step.
We ought to extract the parser into its own package, that would be a breaking change too and would require that extra step anyway.

Julow:
I added the `synopsis` class too.

src/model/comment.ml
jonludlam: Don't mdx and ocamlformat use `Ast.docs` rather than `Comment.docs`?

dbuenzli:
According to what I see this uses definition lists, please don't use definitions list, they [needlessly hard to style](https://github.com/ocaml/odoc/issues/329#issuecomment-781271998). Use unordered lists as they were before. 

Also there's no need to put the synopsis in its own paragraph as it should be only one paragraph; just `span` it after the module name (don't forget to add a space between the module name and the synopsis `span`).

Basically what you want is the same markup as [odig's lists of packages](https://b0-system.github.io/odig/doc/). *Mutatis mutandis* that would be: 

```html
<ul class="modules">
  <li id="module-M">
    <a href="#module-M" class="anchor"></a>
    <a href="M/index.html"><code>M</code></a>
    <span class="synopsis">M's synopsis</span>
  </li>
</ul>
```

Drup:
If we should never use definition lists in HTML, I suggest we change the HTML renderer rather. The structure should be preserved internally (for the Latex output, for instance).

dbuenzli:
I guess so, I was actually confused not to see anything in the HTML renderer so I eventually went to the samples (I'm not up-to-date with the intermediate representation business).

dbuenzli:
> The first sentence of the first paragraph of the modules' documentation is taken.

Detecting the first sentence is very unreliable. I think we had mostly agreed with @lpw25 that `odoc` would use the first paragraph. Here's a [discussion](https://github.com/ocaml/ocaml/pull/784) about it.

gasche:
Thanks @Julow for the implementation work!

I am fine with the "first paragraph" specification, I agree it is easier and it also sounds more convenient for users anyway -- what if I want several sentences in my synopsis? In practice I see people breaking the paragraph after the synopsis, so I think that most libraries would not notice any difference compared to ocamldoc's behavior.

jonludlam:
Thanks for the reference, @dbuenzli - I'm _much_ happier to use the first paragraph than try to figure out where the first sentence ends!

We should probably add this difference to our `interfaces` doc so it's clear we're behaving a little differently than ocamldoc here.

src/model/comment.ml
Julow: You are right ! I invented the problem.

Julow:
I've changed to "first paragraph".

PR #600, opened 2021-02-24
Unnest unattached

dbuenzli:
This PR is built on top of #589 and #593. The relevant commits are the three last ones. 

It does the unnesting mentioned in [that comment](https://github.com/ocaml/odoc/pull/593#issue-575508300). This unifies the markup structure for .mld and module pages and drops the `.odoc-unattached` class added in #589.

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -323,7 +323,9 @@ and items ~resolve l : item Html.elt list =
         let elts =
           let doc = match doc with
           | [] -> []
-          | docs -> [ div (flow_to_item @@ block ~resolve docs) ]
+          | docs ->
+            let a = [ Html.a_class [ "spec-doc" ] ] in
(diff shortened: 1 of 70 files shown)


Drup:
Initially, as was of the opinion than the simplified markup was good, but the inconsistency is problematic. If you think it makes the styling easier, I'm okay with it. Since we coalesce text elements, this shouldn't grow the markup all that much.

dbuenzli:
Personally this is the model I always had in the head for `.mli`s, namely a literate text interspersed with definitions. 

(Also `aside` what semantically very doubtful)

> Since we coalesce text elements, this shouldn't grow the markup all that much.

This can't grow the markup, this reduces the markup (by one `div`).

jonludlam:
sorry, caused conflicts again! I've rebased myself (it was only promoted test results that were conflicting)

PR #603, opened 2021-02-25
Fix #587

### LINKED ISSUE #587: Failure during include
Repro (due to @lukemaurer)

a_intf.ml
```ocaml
module type S = sig
  module Foo : sig end
end
```

b_intf.ml
```ocaml
module Bar = struct end

module type S = sig
  include A_intf.S
  module Foo : sig end
end

module type B = sig
  module type S = S

  include S
end
```

b.mli
```ocaml
include B_intf.B
```

c.mli
```ocaml
include B.S
```

Causes a failure in `Subst`

jonludlam:
The error was due to a bug in the shadowing code, which manifests due to `Subst's` recent tightening in the PR to improve speed.

### DIFF
--- a/src/xref2/lang_of.ml
+++ b/src/xref2/lang_of.ml
@@ -248,10 +248,7 @@ module ExtractIDs = struct
     let name = Ident.Name.type_ id in
     let identifier =
       if List.mem name map.shadowed.s_types then
-        `Type
-          ( parent,
-            Odoc_model.Names.TypeName.internal_of_string (Ident.Name.type_ id)
-          )
+        `Type (parent, Odoc_model.Names.TypeName.internal_of_string name)
       else `Type (parent, Ident.Name.typed_type id)
     in
     {
--- a/src/xref2/lang_of.ml
+++ b/src/xref2/lang_of.ml
@@ -247,8 +247,11 @@ module ExtractIDs = struct
   let rec type_decl parent map id =
     let name = Ident.Name.type_ id in
     let identifier =
-      if List.mem_assoc name map.shadowed.s_types then
-        List.assoc name map.shadowed.s_types
+      if List.mem name map.shadowed.s_types then
+        `Type
+          ( parent,
+            Odoc_model.Names.TypeName.internal_of_string (Ident.Name.type_ id)
+          )
       else `Type (parent, Ident.Name.typed_type id)
     in
     {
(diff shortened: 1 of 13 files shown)


test/xref2/v408_and_above/github_issue_587.t/build.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
```
Julow, line 1:
```suggestion
#!/usr/bin/env sh
```
Otherwise, this script can be inlined into `run.t`.

Julow (approved):

test/xref2/v408_and_above/github_issue_587.t/build.sh
jonludlam: I find it quite helpful to have the script external as it's easier to run and find the output - particularly for those that output html that I'd like to look at (not this one, as it happens). Is there a better way to do this?

src/xref2/lang_of.ml
```diff
@@ -247,8 +247,11 @@ module ExtractIDs = struct
   let rec type_decl parent map id =
     let name = Ident.Name.type_ id in
     let identifier =
-      if List.mem_assoc name map.shadowed.s_types then
-        List.assoc name map.shadowed.s_types
+      if List.mem name map.shadowed.s_types then
+        `Type
+          ( parent,
+            Odoc_model.Names.TypeName.internal_of_string (Ident.Name.type_ id)
```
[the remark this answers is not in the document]
jonludlam: ah good catch :-)

test/xref2/v408_and_above/github_issue_587.t/build.sh
Julow: My preferred way is to add commands to the `run.t` file, I usually add `$ find .` then cat some files. I do `dune runtest --auto-promote test/xref2` to increase speed (I have a binding in my editor for that).
That way I'm sure everything is run from scratch and I avoid inconsistent states.

jonludlam: OK, that's fine for just seeing the content of the files but I often want to see the HTML output in the browser. I suppose that's really only when developing the test, so I can always transcribe it into the `run.t` afterwards, as I agree it's better to have it inline.

Issue #562, opened 2021-01-14
Remove Stdlib prefix


ghilesZ:
Hi,

I'm using odoc with dune to generate documentation for a library and it generates signatures like:

````OCaml
val print : Stdlib.Format.formatter -> t -> unit
````
While what i would like is:

````OCaml
val print : Format.formatter -> t -> unit
````
OCamldoc used to provide the **-hide** option which could do this. Does odoc have something similar? If yes, how to use it from dune?

Optionally, it would be nice to have a link to the Format module of the Stdlib but this is maybe asking too much.

Thanks in advance,
Ghiles.

PR #604, opened 2021-02-25
Rebase/fixup of #579

jonludlam:
Rebased on master and fixed formatting and tests.

For the record: Older OCamls (before 4.08 ish IIRC) render paths from Stdlib -- for example, Int64.t -- as just `Int64.t` whereas newer ones prefix `Stdlib` on them. This is rather unavoidable unless we start compiling odoc files for Stdlib modules as part of the test suite. For this reason, we try to keep clear of these types in the tests.

Note also, #562 is relevant.


Drup:
Thanks @jonludlam !

PR #607, opened 2021-02-26
No dl in HTML

dbuenzli:
This gets rids of `dl` in HTML.

In order to do so: 

1. We enrich the IR's description lists to be able to have classes on description pairs.
2. We tweak the IR representation for @-tags along the line mentioned in #329 
3. We tweak the IR representation for module synopses (no spurious `p`)
4. We tweak and cleanup the CSS file a bit. 

Btw. it seems the `ocp-indent` file is not consistent with what `ocamlformat` wants, e.g. w.r.t. identation. It's quite annoying.


dbuenzli:
This is best reviewed per commit.

Drup:
This looks pretty good to me! Do you intend to add the collapse in another PR ? I'm surprised there is no impact on the man/latex tests, but I guess that's good. :)

src/document/types.ml
```diff
@@ -61,11 +61,13 @@ and Block : sig
...
   and one = { attr : Class.t; desc : desc }
 
+  and description_item = { item_attr : Class.t; term : Inline.t; def : t }
+
   and desc =
     | Inline of Inline.t
     | Paragraph of Inline.t
     | List of list_type * t list
-    | Description of (Inline.t * t) list
+    | Description of description_item list
```
Drup, line 70: I would suggest adding a `description` module, with types `one` and `t`, and then you can use `attr` as label again. Also, I would prefer,`key` and `definition` for the other field names.

src/document/comment.ml
```diff
@@ -253,59 +253,59 @@ and nestable_block_element_list elements =
...
+    let tag_name = inline ~attr:["at-tag"] (Text tag) in
+    let tag_value = match value with
+      | None -> []
+      | Some t -> [sp; inline ~attr:["value"] t]
+    in
+    let term = tag_name :: tag_value in
+    Some { Block.item_attr = [tag]; term; def }
+  in
+  let text_def s =
+    [{ Block.attr = []; desc = Block.Inline [ inline @@ Text s ] }]
```
Drup: Pretty sure you can use `block` here.

dbuenzli: Done.

dbuenzli:
> Do you intend to add the collapse in another PR ?

If I don't manage to do it by the end of the week-end, I'll let that to someone else.

lubegasimon:
> This gets rids of dl in HTML.

Why do you want to do that? The reason I’m asking is that I want to do some reviews, which requires me to know the reason behind your implementation.

dbuenzli:
> Why do you want to do that?

See the dicussion here #329.

lubegasimon:
> > Why do you want to do that?
> 
> See the dicussion here #329.

Thanks!

src/document/types.ml
```diff
@@ -51,6 +51,13 @@ and Inline : sig
 end =
   Inline
 
+and Description : sig
+  type one = { attr : Class.t; key : Inline.t; definition : Block.t }
```
lubegasimon, line 55: Just a minor suggestion: I wouldn’t use `key`, rather `term` because it’s a convention (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl)

lubegasimon:
LGTM!

src/document/types.ml
dbuenzli: That's what I had before. I used key on @Drup's [request](https://github.com/ocaml/odoc/pull/607#discussion_r583824544).

Drup: The technical name used by the `dl` specification doesn't matter, since we don't use DL to represent those anymore.

Drup (approved):
@jonludlam the CI failure is spurious, please fire away. :)

jonludlam:
In it goes!

PR #612, opened 2021-02-27
Refine includes

### LINKED ISSUE #586: CSS changes lost the 'include' visual cue
The visual cue that elements came from an include has been lost in commit c1747d35570a11205526f0cc1168c589aa006105 along with some other nice bits and pieces.

Before:

Screenshot 2021-02-11 at 12 46 21

After: 

Screenshot 2021-02-11 at 12 46 32

dbuenzli:
This PR is build on top of #607 the relevant commits are the 4 last ones (sorry for always proceeding that way but changing the markup bit-by-bit in different branches is too unconvenient).

The current structure of includes is quite involved and messy. For one non-inlined include we have:

```
div.odoc-include
  div."spec include"
    div.doc 
      content of doc string (may be multiple paragraphs)
      details 
        summary
          span.def
            code
      include defs
```

This PR proposes to change the structure to 

```
div.odoc-include 
  div.spec-doc
    content of doc string
  details
    summary."spec include"
      code
    include defs
```

Except for the docstring of the include coming *before* the specification (it would have been nice to put in the `summary`, but the content model  disallows it). This more regular with other structure items which have the following markup: 
```
  div.odoc-spec 
    div."spec $(kind)"
      code
    div."spec-doc"
      content of doc string
``` 
In particular structure item doc strings are always in a `spec-doc` div and the `spec` class is always on elements spanning directly the source of structure items.

I have adapted the `odoc.css` and checked that stuff like double includes like [this one work](https://b0-system.github.io/odig/doc@odoc.default/containers/CCArray/index.html#arrays). I took the oportunity to fix #586 along the way.

### DIFF
--- a/src/odoc/etc/odoc.css
+++ b/src/odoc/etc/odoc.css
@@ -20,6 +20,12 @@
   --target-background: rgba(187, 239, 253, 0.3);
   --target-shadow: rgba(187, 239, 253, 0.8);
   --pre-border-color: #eee;
+  --code-background: #f6f8fa;
+  --spec-summary-border-color: #5c9cf5;
+  --spec-summary-background: var(--code-background);
+  --spec-summary-hover-background: #ebeff2;
+  --spec-details-after-background: rgba(0, 4, 15, 0.05);
+  --spec-details-after-shadow: rgba(204, 204, 204, 0.53);
 }
 
 .dark:root {
@@ -84,9 +90,7 @@
     --toc-color: #777;
     --toc-background: #252525;
     --toc-list-border: #ccc;
-    --spec-summary-background: var(--code-background);
     --spec-summary-hover-background: #ebeff2;
-    --spec-summary-border-color: #5c9cf5;
     --spec-details-after-background: rgba(0, 4, 15, 0.05);
     --spec-details-after-shadow: rgba(204, 204, 204, 0.53);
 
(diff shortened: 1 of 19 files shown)


src/odoc/etc/odoc.css
```diff
@@ -392,7 +392,7 @@ pre code {
 
 /* Module member specification */
 
-.spec:not(.include), .spec.include details summary {
+.spec {
```
dbuenzli, line 399: For people devising markup this is a good example of why classes should not be used for vastly different contexts.

Previous to this commit the `spec include` classes were used on a div wrapping the whole include and its markup content. But otherwise the `spec` class was always used only on `div`s simply wrapping the source of structure items. This entailed this incredibly complex selector for styling structure items sources which, with the new structure, now boils down to a simle `.spec`.

Drup (approved):
The markup for includes was definitely clunky, this is much better.

@dbuenzli It would be super helpful if you could write some quick guidelines on top of the HTML rendered detailing the general philosophy guiding the placement of the classes and the shape of the markup (you explained it in your various PRs, but I think it should be in the code).

dbuenzli:
> quick guidelines on top of the HTML rendered detailing the general philosophy guiding the placement of the classes and the shape of the markup

Not sure I can establish a general philosophy from gut feelings :-) But I can try to add something at some point another markup PR. 

> (you explained it in your various PRs, but I think it should be in the code).

Do you have particular comments in mind ? 

*** 

The PR has been rebased.

jonludlam:
Thanks @dbuenzli!

Issue #601, opened 2021-02-24
Type extensions are no anchorable and should be classified as `type`


dbuenzli:
The individual cases are anchorable but the toplevel `type t +=` isn't which is annoying when you want refer to the extension, not a particular case. I guess something must be solved to generate unique `id`s in that case.

Also it would be nice to classify them as `spec type extension` rather `spec extension` Since most of the rendering is going to be shared with `spec type`, this simplifies stylesheets. It seems to stem from [this line](https://github.com/ocaml/odoc/blob/3d40f62ebcfc72b13114770de690c2346775ebde/src/document/generator.ml#L514) but I'm a bit usure what will happen for generators if we replace that `"extension"` with `"type extension"` (e.g. [here](https://github.com/ocaml/odoc/blob/3d40f62ebcfc72b13114770de690c2346775ebde/src/html/generator.ml#L209) in the HTML generator). @Drup knows maybe ?

Drup:
hmm, if we want to have several "kinds", we should make this a `string list`, instead of a `string option`. I think it's a good idea.

Drup:
As choice of anchor, I suggest we use the constructors, either a complete concatenation, or just the first one.

dbuenzli:
> As choice of anchor, I suggest we use the data constructors that are being added, either a complete concatenation, or just the first one.

That seems a good idea. Something like `type-$(type)-$(first-constructor)`. Also is there a syntax to refer to them with the `{!ref}` syntax ?

> hmm, if we want to have several "kinds", we should make this a `string list`, instead of a `string option`. I think it's a good idea.

Yes that's the reason why I opened an issue rather than doing it directly. That being said I'm not super happy with the markup classification in general in the content of `type` definitions.  

Unless you want to go wild, it's likely that for most kind of type definitions the styling rules are the same. That is the classification on `td` (`field`, `contstructor`, `extension`, etc.)  is not really what you want. You want classification of the `kind` on the container, i.e. on the `table` (e.g. to play with indentation of all the `td`s). CSS being CSS you can always workaround, but I'm always interested in keeping it as simple as possible. 

Also using `.doc` for the docstring in types cases and `.doc` on a `div` on `include` is not such a good idea. (In general using the same class for vastly different contexts leads to CSS side effect pains).

[...]

Issue #614, opened 2021-02-28
HTML markup, move away from tables


dbuenzli:
Tables are being used for documenting type cases and record fields. 

This doesn't work in practice for the same reasons mentioned [here](https://github.com/ocaml/odoc/issues/297#issuecomment-462446614) for module synopses; basically don't use table/alignement if you don't know how long things are going to be beforehand.

Here's an example with two stylesheets that show the problem and how ugly it is no matter the page width:

  https://b0-system.github.io/odig/doc/biniou/Bi_inbuf/index.html
  https://b0-system.github.io/odig/doc@odoc.default/biniou/Bi_inbuf/index.html

I propose to move to ordered lists `ol`, with one `li` per definition and a twist (if the logic is easy to implement):

* If the doc string of the definition is only made of inline content we `span` it after the code.
* If the doc string of the definition is made of multiple paragraphs we `div` it. 

Basically we move from:

```
table
  tbody
    tr
      td.def
        code
      td.doc
```
to
```
ol
  li.def 
    code
    (span | div).def-doc
```
With appropriate styling this allows HTML renderers to

1. Have the code and a one-liner doc string on a single line and break and indent whenever either of these no longer fits the page width.
2. Have the code on its own break and indent line and a multi-paragraph doc string separately rendered below like top level specifications do.

The scheme should be able to accomodate short and long styles of documenation:
```ocaml
{ incr : int; (** The position increment. *)
  f : ...;
  (** This field has the function for bla bla. It takes the following arguments. 
      {ul ... } *)
  decr : int; (** The position decrement *) }
```
Rendering the thing as: 

```
{ 
   incr : int;    The position increment.
   f : ...;
      This field has the function for bla bla. It takes the following arguments:
       - ...
   decr : int;    The position decrement.
}
```

Besides by using: [`user-select:none`](https://caniuse.com/?search=user-select) on `.def-doc`it should be possible to reasonably c&p the definitions; but unfortunately not the content of the doc strings themselves becomes no longer selectable, but from an ergonomic point of view I think you are interested in selecting the definitions most of the time (e.g. because you want to c&p to pattern match on all the cases of a value). 

Note that applying this to the example mentioned above this will result in 1. which will be ugly. But we argue that the authors should introduce paragraphs. In fact [the source](https://github.com/ocaml-community/biniou/blob/master/src/bi_inbuf.mli) already has them albeit not with the blank line that this supposed to separate them. 

/cc @Drup

Drup:
[...]

PR #615, opened 2021-02-28
HTML backend: refine specification rendering

### LINKED ISSUE #296: Record field and variant case rendering breaks cut and paste
Please don't do this. Either do it like `ocamldoc` did by putting the field doc string in a regular ocaml comment or find a markup strategy that allows to select only the type definition. 

For reference: 

* The [ocamldoc way](https://erratique.ch/software/vg/doc/Vg.P.html#TYPEoutline)
*  The current odoc way [odoc way](https://b0-system.github.io/odig/doc/vg/Vg/P/index.html#type-outline). 
* An [odoc variant](https://b0-system.github.io/odig/doc/uuidm/Uuidm/index.html#type-version) doc that suffers from the same problem.

dbuenzli:
This PR is built on top of #612, the relevant commits start at e7d6fd3f839f7701e1f

It does the following:

* Classifies the function type `->` so that browsers can be instructed
  not to break it.
* IR model: as discussed in #601 adds several kinds for `Item`. We reuse the `Class.t` type usually named `attr` (sic).  A few items get reclassified, mostly in a backward compatible way. See the commit message for details.
* Renames class `.doc` to `.def-doc`.
* Accomodates `odoc.css` with the changes. 
* Tweak `odoc.css` for better c&p behaviour (closes #296).


dbuenzli:
While doing this PR I realized the rendering by the default stylesheet of variants with a doc string is not very nice, I just want to stress that this is not a side effect of this PR. I don't think it's worth fixing before #614 is.

src/document/generator.ml
```diff
@@ -1027,10 +1027,10 @@ module Make (Syntax : SYNTAX) = struct
           ++ virtual_ ++ params ++ O.txt " " )
         @ cname @ expr
       in
-      let kind = Some "class-type" in
+      let attr = [ "class-type" ] in
```
Drup, line 1030: `["class"; "type"]` ?

src/document/generator.ml
```diff
@@ -1157,10 +1157,10 @@ module Make (Syntax : SYNTAX) = struct
         O.documentedSrc
           (O.keyword "module" ++ O.txt " " ++ O.txt name ++ O.txt " := " ++ path)
       in
-      let kind = Some "module-substitution" in
+      let attr = [ "module-substitution" ] in
```
Drup, line 1160: `["module"; "substitution"]` (similar to types) ?

src/odoc/etc/odoc.css
```diff
@@ -406,6 +406,9 @@ pre code {
   padding: 0.35em 0.5em;
 }
 
+.spec .def-doc { user-select:none; webkit-user-select:none }
+.spec .def-doc:hover { user-select:auto; webkit-user-select:auto }
+
```
Drup, line 411: I highly dislike this. In my experience, websites that mess with copy paste and selection are the leading cause of internet rage and deserve to burn in hell. You are free to use it in your CSS, but I really don't want that in the default one.

However, I got an idea: what about invisible `(*`/`*)` delimiters, which would no take visual space, but would get selected during copy pasting ? I'm pretty sure we can cook something up with some careful styling.

Drup (approved):
I'm generally in favor of these changes, except for the copy-paste one. Your solution is, imho, not acceptable, but I'm pretty sure we can do better. The rest is good to go (as soon as the previous PRs are merged).

src/odoc/etc/odoc.css
dbuenzli: > I highly dislike this. In my experience, websites that mess with copy paste and selection are the leading cause of internet rage and deserve to burn in hell.

That seems a little bit excessive. Did you actually [try it](https://b0-system.github.io/odig/doc/biniou/Bi_inbuf/index.html) ? 

Until something better is found I think we should leave it, it certainly provides a better user experience for copy pasting type definitions. 

I don't mind removing that commit if other people agree with you, but honestly the result provided by default stylesheet is both unreadable and ergonomically lacking, I'm offering improvements but I'm not interested in fighting for them.

> However, I got an idea: what about invisible `(*`/`*)` delimiters, which would no take visual space,

You can try to do this. Another thing that could be tried is to add the comment delimiters with CSS pseudo selectors.

src/document/generator.ml
dbuenzli: I don't always have a good grasp on what all these definitions corresponds to visually (In the good old days `odoc` installed the `ocamlary` in `odoc`'s libdir which was nice to develop stylesheets). 

But it's not necessarily smart to split that as you suggest, e.g. if class types are to be treated quite differently from other things labelled `type`.

In any case I didn't review these classifications, but they should certainly be.

src/odoc/etc/odoc.css
Drup: > You can try to do this. Another thing that could be tried is to add the comment delimiters with CSS pseudo selectors.

I just did! See https://github.com/ocaml/odoc/pull/626, I think it works quite well.

dbuenzli:
> I'm generally in favor of these changes, except for the copy-paste one.

This has been removed.

> In any case I didn't review these classifications, but they should certainly be.

I suggest we do that in a separate PR once #630 has been resolved. 

---

The PR has been rebased

jonludlam:
Thanks!

PR #618, opened 2021-03-01
Remove hand-written compare and hash functions on paths

Julow:
No data need to be elided so `Stdlib.compare` and `Hashtbl.hash` can be used.


src/model/names.ml
```diff
@@ -68,12 +68,7 @@ module Name : Name = struct
 
   let equal (x : t) (y : t) = x = y
 
-  let compare x y =
-    match (x, y) with
-    | Internal (x, _), Internal (y, _) -> String.compare x y
```
Julow, line 73: Eliding this field was probably the reason of all this but I think this is not right because we use this function for data-structures, where this would be hazardous, and never intentionally for checking that two `Internal` have the same name.
Also, `equal` above uses `( = )`.

jonludlam:
Needs another rebase!

PR #619, opened 2021-03-02
add center, left, and right alignment syntax

lubegasimon:


src/latex/generator.ml
```diff
@@ -310,9 +316,12 @@ let rec block ~in_source (l : Block.t) =
   let one (t : Block.one) =
     match t.desc with
     | Inline i -> inline ~verbatim:false ~in_source:false i
-    | Paragraph i ->
-        inline ~in_source:false ~verbatim:false i
+    | Paragraph (Some p_style, i) ->
+        [(Paragraph_style (p_style, inline ~in_source:false ~verbatim:false i))]
         @ if in_source then [] else [ Break Paragraph ]
+    | Paragraph (None, i) ->
+      [(Paragraph_style (`Left, inline ~in_source:false ~verbatim:false i))] (*FIXME: you may want to handle me well*)
```
Julow, line 323:
```suggestion
      inline ~in_source:false ~verbatim:false i
```
This should be calling the previous code.
Also, I think the paragraph logic shouldn't be duplicated, I suggest moving the optional logic to an other function:
```ocaml
  maybe_paragraph_style p_style (inline ~in_source:false ~verbatim:false i)
  @ ...
```

src/latex/raw.ml
```diff
@@ -111,6 +111,12 @@ let emph pp = create "emph" pp
 
 let bold pp = create "bold" pp
 
+let left pp = create "left" pp
+
+let center pp = create "center" pp
+
+let right pp = create "right" pp
```
Julow, line 118: I don't thing `left` and `right` exist. (I'm not a latex expert)
I think you'll need to use `flushleft`, `center` and `flushright` and to call them with `env` (see `verbatim` above)

src/model/comment.ml
```diff
@@ -39,7 +41,7 @@ type module_reference = {
     resolved during linking. *)
 
 type nestable_block_element =
-  [ `Paragraph of paragraph
+  [ `Paragraph of paragraph_style option * inline_element with_location list
```
Julow, line 44: Here the style argument is placed first but it's last for `Heading`, for consistency I suggest picking one side and changing the other.

src/model/comment.ml
```diff
@@ -79,7 +81,7 @@ type attached_block_element = [ nestable_block_element | `Tag of tag ]
 
 type block_element =
   [ nestable_block_element
-  | `Heading of heading_level * Identifier.Label.t * link_content
+  | `Heading of heading_level * Identifier.Label.t * link_content * paragraph_style option
```
Julow, line 84: Do we really want to allow custom alignment on headings ?

Julow (changes requested):

src/model/comment.ml
lubegasimon: Indeed not!
@jonludlam, what’s do you say?

src/latex/raw.ml
lubegasimon: Yep, `flushleft` and `flushright` are right. thanks 🙂!

src/model/comment.ml
jonludlam: the intent here is to preserve compatibility with ocamldoc - so does ocamldoc allow this sort of styling on headings?

Julow: ocamldoc allows alignment on every elements, including code blocks and headings. I don't think we need to do the same thing as ocamldoc here.
I find 4 `.mli` files containing these when grepping the source of every packages on opam (excluding odoc and ocaml) and only one has more than a paragraph in it (https://github.com/savonet/ocaml-duppy/blob/master/src/duppy.mli#L26, which won't work anyway due to the `{i`, which is more restricted).

dbuenzli: Just kill the feature, this feels like a waste of time. Not only does give a way to hurt UX (which relies on rendering consistency) but there are tons of other small issues in the markup worth solving which can drastically improve UX.

The particular example given by @Julow is not even about alignement it's about quoting (e.g. like markdown provides).

FWIW I never bothered supporting alignment on the stylesheets I made for odoc and ocamldoc for the past 10 years and no one ever complained despite these being reused in the wild and consumed via `odig`.

lubegasimon: Fortunately / unfortunately, the fix is almost done! 

> there are tons of other small issues in the markup worth solving which can drastically improve UX.

However, I do agree with you on that.

lubegasimon: So we leave out customising heading alignment, right?

dbuenzli: > Fortunately / unfortunately, the fix is almost done!

I understand you have put hard work into this and I'll let the official maintainer @jonludlam judge what is best. 

But I just note the fix adds 160 loc to the code base for a dubious feature that @Julow has proved to be publicly almost universally unused (and that I pledge I won't support in `odig`'s themes). Personally I think we should exercice a bit of good jugement on the "feature parity" with `ocamldoc`. There are (very few) things `ocamldoc` didn't get right and there's no need to perpetuate them.

Julow: If this is not merged, I think we need to accept these elements in the parser for compatibility (which would print a warning and won't apply the styling).

jonludlam: Agreed with adding them to the parser - we can at least then remove the markup from the output. @lubegasimon can you please close this PR and open a new one with just the parser changes -- hopefully the rebase after the recent parser changes won't be too annoying. Let's ask for opinions at the next dev meeting and we can come to a collective decision -- I have to admit I wasn't aware it was quite so infrequently used, so that's very helpful data @Julow.

lubegasimon:
closed in response to https://github.com/ocaml/odoc/pull/619#discussion_r587734198

PR #621, opened 2021-03-03
Parser work

jonludlam:
Initial work so the parser can be extracted into a separate library.


src/parser/error.mli
```diff
@@ -0,0 +1,14 @@
+type t = { location : Location.span; message : string }
+
+val make :
+  ?suggestion:string ->
+  ('a, Format.formatter, unit, Location.span -> t) format4 ->
+  'a
+
+type 'a with_warnings = { value : 'a; warnings : t list }
+
+type warning_accumulator = t list ref
```
Julow, line 10: This type should be abstract.

Julow (approved):
Looks good. The parser for references should be splitted out of model too.

src/parser/error.mli
jonludlam: Good catch, thanks!

Issue #590, opened 2021-02-12
Enable references to refer to pages with '.' or '-'


jonludlam:

lubegasimon:
Could someone add some pointers to this. cc @Drup @jonludlam @aantron

Julow:
It is currently not possible to write a reference to a page containing a `-` or a `.` in its name:
- `{!page-cohttp-lwt}` the parser think that `page-cohttp` is the "kind" and `lwt` the name.
  This is due to `module-type` or `class-type` kinds containing a `-`. A fix to this might be to disallow kinds containing `-` or to add a way to escape it.
- `{!page-1.7.3}` parses as `` (`Dot (`Dot (`Root (`Page, "1"), "7"), "3") ``. We need a way to escape the `.` character too.

It was suggested during the last meeting to allow escaping with quotes: `{!page-"cohttp-lwt"}`, `{!page-"1.7.3"}`

dbuenzli:
Note that when @trefis implemented the `.mld` support one of the constraints was that the name of mld pages had to follow the convention of module names. I don't know if lifting that constraint may have further implications.

> It was suggested during the last meeting to allow escaping with quotes: `{!page-"cohttp-lwt"}`, `{!page-"1.7.3"}`

The `page-` to reference pages always felt weird to me. I suspect @trefis had his reasons but rather than introduce new notations and escapes wouldn't it be possible to reframe that in the context of [cross reference disambiguation](https://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#sss%3Aocamldoc-crossref). 

That is introduce `page:` as a "syntactic class" for making references:

```
{!page:cohttp-lwt}
{!page:1.7.3} 
```

That would make a two things less to learn: the weird `page-` references and how to escape them.

jonludlam:
The hyphen notation is consistent with the 'new-style' cross-reference disambiguation. The 'colon-style' behaviour doesn't allow us to distinguish between the two types `t` in the following:

```
module X : sig type t end
module type X = sig type t end
```

whereas the newer 'hyphen-style' does: `{!module-X.t}` vs `{!module-type-X.t}`. So the `page-` notation is consistent with that. Additionally, we'd still need to be able to refer to labels in the pages, so we'd either have to introduce a new separator for the page and label, or we'd need escaping, or the quoting that's suggested above.

trefis:
[...]

PR #622, opened 2021-03-05
Tweak syntax for page name refs

lubegasimon:
Fixes https://github.com/ocaml/odoc/issues/590

### DIFF
--- a/src/model/reference.ml
+++ b/src/model/reference.ml
@@ -128,11 +128,21 @@ let tokenize location s =
         scan_identifier started_at
           (open_parenthesis_count - 1)
           (index - 1) tokens
+    | '"' when open_parenthesis_count = 0 -> (
+        try
+          scan_identifier started_at 0
+            (String.rindex_from s (index - 1) '"' - 1)
+            tokens
+        with _ ->
+          Error.raise_exception (Error.make "Unmatched quotation!" location) )
     | _ -> scan_identifier started_at open_parenthesis_count (index - 1) tokens
   and identifier_ended started_at index =
     let offset = index + 1 in
     let length = started_at - offset in
     let identifier = String.trim (String.sub s offset length) in
+    let identifier =
+      Astring.String.cuts ~sep:"\"" identifier |> String.concat ""
+    in
     let location = Location_.in_string s ~offset ~length location in
 
     if identifier = "" then
--- a/src/model/reference.ml
+++ b/src/model/reference.ml
@@ -128,7 +128,7 @@ let tokenize location s =
         scan_identifier started_at
           (open_parenthesis_count - 1)
           (index - 1) tokens
-    | '"' when open_parenthesis_count = 0 -> (
+    | '"' -> (
         try
           scan_identifier started_at 0
             (String.rindex_from s (index - 1) '"' - 1)
--- a/test/parser/expect/reference-component-kind/dash-in-page-name.txt
+++ b/test/parser/expect/reference-component-kind/dash-in-page-name.txt
@@ -1,0 +1,4 @@
+((output
+  (((f.ml (1 0) (1 17))
+    (paragraph (((f.ml (1 0) (1 17)) (reference (root foo-bar page) ())))))))
+ (warnings ()))
--- a/test/parser/expect/reference-component-kind/dot-and-dash-in-page-name.txt
+++ b/test/parser/expect/reference-component-kind/dot-and-dash-in-page-name.txt
@@ -1,0 +1,5 @@
+((output
+  (((f.ml (1 0) (1 24))
+    (paragraph
+     (((f.ml (1 0) (1 24)) (reference (root foo-bar.v0.0.1 page) ())))))))
+ (warnings ()))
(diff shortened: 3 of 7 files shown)


src/model/reference.ml
```diff
@@ -128,6 +128,31 @@ let tokenize location s =
...
+           because of the scan_kind and scan_identifier calls that are attached to the parser.
+        *)
+        let f_occurence = String.index s '"' in
+        let l_occurence = String.rindex_from s f_occurence '"' in
+        let off_set = f_occurence + 1 in
+        let length = started_at - off_set - 1 in
+        let identifier = String.trim (String.sub s off_set length) in
+        (* The if ... then conditions specially handles cases like "foo"."bar",and
+           it should be checked first, otherwise, and exception will be thrown*)
+        if s.[0] = '"' && s.[f_occurence] = '"' then
```
lubegasimon, line 149: I’m still failing to handle a case like `{!”Foo”. “Bar”}` whose test is at https://github.com/ocaml/odoc/pull/622/commits/e4a86e69850fb694aa11fffadfae6030ecbd5a7f#diff-f59d447ba8018121335d4cfbebc893d052260724d49e9c187d9f4d7e2f7685ff

src/parser/token.ml
```diff
@@ -66,7 +66,8 @@ type t =
   | `Begin_list_item of [ `Li | `Dash ]
   | `Minus
   | `Plus
-  | section_heading
+  | (* | `Quote *)
```
jonludlam, line 69: These should be removed

jonludlam:
It'd be good to add some test cases to the parser tests - test/parser/test.ml, since when we remove the parser these tests will go with it.

src/model/reference.ml
lubegasimon: Done!

src/parser/token.ml
lubegasimon: Done!

src/model/reference.ml
```diff
@@ -128,6 +128,41 @@ let tokenize location s =
...
+        let length = started_at - off_set - 1 in
+        let identifier = String.trim (String.sub s off_set length) in
+
+        (* The if ... then conditions specially handles cases like "foo"."bar",and
+            it should be checked first, otherwise, and exception will be thrown
+
+           However, I guarantee that this won't work expectedly for a case like
+           {!"Foo"."Bar".section-bar}
+        *)
+        if s.[0] = '"' && s.[f_occurence] = '"' then
```
Julow, line 154: This is wrong. (eg. `{!"Foo.1.2.3"."Bar.1.2.3"."Baz"}`).
The main mistake is `f_occurence`, which should be `index`. The parser works from right to left, `index` is the right quote and `l_occurence` is the corresponding left quote, nothing more needed.
The current value for `f_occurence` is the first quote of the string, we don't want that.

src/model/reference.ml
```diff
@@ -128,6 +128,41 @@ let tokenize location s =
         scan_identifier started_at
           (open_parenthesis_count - 1)
           (index - 1) tokens
+    | '"' ->
```
Julow, line 131:
```suggestion
    | '"' when open_parenthesis_count = 0 ->
```

src/model/reference.ml
```diff
@@ -128,6 +128,41 @@ let tokenize location s =
...
+           left to right. We do this specially to keep the identifier (page name in this case) intact.
+
+           However, the if ... else block operations, we return to then concept on how the parser works,
+           because of the scan_kind and scan_identifier calls that are attached to the parser.
+        *)
+        let f_occurence = String.index s '"' in
+        let l_occurence = String.rindex_from s f_occurence '"' in
+        let off_set = f_occurence + 1 in
+        let length = started_at - off_set - 1 in
+        let identifier = String.trim (String.sub s off_set length) in
```
Julow, line 146: `identifier_ended` is also checking that the identifier isn't `""`, `identifier_ended` should be called from here. (so you have the `location`, which will need adjustment to include the quotes)

src/model/reference.ml
```diff
@@ -128,6 +128,41 @@ let tokenize location s =
...
+          let identifiers = String.split_on_char '.' s in
+          List.map
+            (fun ident ->
+              let ident =
+                String.trim (String.sub ident 1 (String.length ident - 2))
+              in
+              (None, ident, location))
+            identifiers
+        else
+          scan_kind identifier location (f_occurence - 1) (l_occurence - 1)
```
Julow, line 164: The `location` is the wrong one, see the comment about `identifier_ended` above.

We cannot call `scan_kind` without checking the next character first. Here we expect three things:
- `-`, to call `scan_kind`
- `.`, an identifier with no kind, recurse
- beginning of the string, an identifier with no kind, return

Julow (changes requested):

lubegasimon:
> It'd be good to add some test cases to the parser tests - test/parser/test.ml, since when we remove the parser these tests will go with it.

Indeed, I’ll surely address this in an independent PR

src/model/reference.ml
```diff
@@ -128,11 +128,21 @@ let tokenize location s =
         scan_identifier started_at
           (open_parenthesis_count - 1)
           (index - 1) tokens
+    | '"' when open_parenthesis_count = 0 -> (
```
jonludlam, line 131: I don't believe we need the `when` clause here.

jonludlam:
Thanks!

PR #626, opened 2021-03-09
Make type definition copy-pastable with hidden comment delimiters.

Drup:
This is an alternative take on the copy paste problem mentioned in https://github.com/ocaml/odoc/issues/296. It inserts comment delimiters (in ocaml, `(*` `*)`) around documentation, but hide them while keeping them in the text selection with some CSS trickery. The trickery in question seems rather standard (it's used in bootstrap as a way to have text only for screen-readers, which is rather consistent with what we want). It also has the advantage that people can style it to keep the delimiters in their own CSS, if they like that.

cc @dbuenzli

### DIFF
--- a/src/odoc/etc/odoc.css
+++ b/src/odoc/etc/odoc.css
@@ -351,6 +351,22 @@ h4 {
   font-size: 1.12em;
 }
 
+/* Comment delimiters, hidden but accessible to screen readers and 
+   selected for copy/pasting */
+
+/* Taken from bootstrap */
+/* See also https://stackoverflow.com/a/27769435/4220738 */
+.comment-delim {
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border: 0;
+}
 
 /* Preformatted and code */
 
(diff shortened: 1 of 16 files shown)


dbuenzli:
I don't understand the CSS trickery but if it does not bring rendering problems and will continue to work with #614, it is certainly a nicer solution.

(Sorry I don't have the time to look into this further right now)

Drup:
My understanding is that the CSS use absolute positioning so that they don't take space in the normal layout, and then put them as 0 size/margin/padding with no overflow to hide it. All those are ignored by text selection.

src/odoc/etc/odoc.css
```diff
@@ -351,6 +351,22 @@ h4 {
...
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border: 0;
+}
```
dbuenzli, line 369: Just FTR I personally dislike cargo-culting CSS, that usually leads to pains down the road :-) But based on @Drup [understanding](https://github.com/ocaml/odoc/pull/626#issuecomment-793969920) it seems that in `odig`'s stylesheets you can boil that down to:

```css
.spec .def-doc .comment-delim /* make them invisible yet copy-pastable */
{ position: absolute; width: 1px; height: 1px; overflow: hidden; }
```
(Tested in all browsers major browsers)

Drup: Yeah, I also think we can shorten it, but there is a "belt and braces": we make it invisible *and* put it somewhere else. :)

PR #627, opened 2021-03-09
Pass documentation along with signatures

Julow:
Before this PR, the top-comment of a signature was handled unevenly and sometimes dropped.
This PR improves how the top-comment is passed and implement [@lpw25's suggestion](https://github.com/ocaml/odoc/issues/478#issuecomment-762744498).

An issue remain in <https://github.com/ocaml/odoc/issues/478>, the documentation from an `@inline` include is not in the preamble and is not used for the synopsis due to being handled too late. This will be fixed in a follow-up PR.

### DIFF
--- a/test/html/expect/test_package+re/Nested/module-type-Y/index.html
+++ b/test/html/expect/test_package+re/Nested/module-type-Y/index.html
@@ -21,27 +21,18 @@
    <h1>
     Module type <code><span>Nested.Y</span></code>
    </h1>
-   <p>
-    This is module type Y.
-   </p>
-   <p>
-    Some additional comments.
-   </p>
+   <h2 id="type">
+    <a href="#type" class="anchor"></a>Type
+   </h2>
   </header>
   <nav class="odoc-toc">
    <ul>
-    <li>
-     <a href="#type">Type</a>
-    </li>
     <li>
      <a href="#values">Values</a>
     </li>
    </ul>
   </nav>
   <div class="odoc-content">
-   <h2 id="type">
-    <a href="#type" class="anchor"></a>Type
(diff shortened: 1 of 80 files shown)


jonludlam (approved):
Looking good!

Julow:
The tests failures were caused by the comment not represented by a floating attribute with OCaml `<= 4.06` in
```ocaml
module M : sig
  (** Doc. *)
end
```

however it is in
```ocaml
module M : sig

  (** Doc. *)
end
```

It is present in `cmti` files but I couldn't find a way to retrieve it, any idea ?

jonludlam:
Using current odoc on

```ocaml
module M : sig
  (** Doc. *)
end
```

I don't get the documentation in module M, so it's not a regression. Also, I get

```
File "m.mli", line 2, characters 8-18:
Error (warning 50): unattached documentation comment (ignored)
```

Julow:
Indeed not a regression. I missed the warning.

test/html/expect/test_package+re/Nested/module-type-Y/index.html
```diff
@@ -27,21 +27,18 @@ <h1>
    <p>
     Some additional comments.
    </p>
+   <h2 id="type">
+    <a href="#type" class="anchor"></a>Type
+   </h2>
```
dbuenzli, line 32: This merge broke the HTML structure. This `h2` has nothing to do here, it should still be in `odoc-content`.

Julow: Issue: https://github.com/ocaml/odoc/issues/631

Issue #235, opened 2018-11-05
First heading not included in TOC


rizo:
If the interface documentation starts with a heading (like `{1 ...}`) and that heading is the first doc item in the interface the heading is not included in the TOC.

This was reported by @c-cube and observed in Containers where modules start with a heading (_e.g._ [CCArray](https://github.com/c-cube/ocaml-containers/blob/master/src/core/CCArray.mli)).

The general recommendation is to start an interface documentation with a short sentence/paragraph, but this bug should be fixed anyway.

c-cube:
for the context, back when I used ocamldoc the same headers gave this kind of TOC: https://simon.cedeela.fr/ocaml-containers/1.0/

aantron:
I remembered why section headings in the top comment don't go into the table of contents.

The top comment is supposed to be introductory/summary material, and the table of contents is displayed below it when it is displayed inline (for example, at narrow screen widths). For comparison, see a [typical Wikipedia article](https://en.wikipedia.org/wiki/OCaml).

```
(** Module summary, first sentence will eventually go into the module index.

    {1 This section does not go into the in-page TOC}

    In fact, the heading shouldn't be here, because this whole comment is supposed
    to be brief introductory material. *)

(** {1 This section does go into the in-page TOC

    The TOC is inserted immediately above the rendering of this comment. *)
```

Perhaps we should emit a warning about this.

dbuenzli:
@aantron In the wikipedia article the top heading is at a different heading level. In our case this is the module name which is heading 0 which should not not go in the toc. 

I don't think there's any need for warnings or opinions here. It's better to have a simple and predictable system, simply include any heading level 1 in the toc wherever these happen to occur.

aantron:
I'm indeed going to make the change to put all the headings (other than the title) into the TOC, but first a couple more points of discussion about the warning/position of the TOC :)

- What I meant is that the text immediately under the top heading never has any additional headings in it on Wikipedia (AFAIU).
- On the matter of being predictable, there is a different way in which not discouraging headings in that top comment is unpredictable. If the typical author is reviewing their docs with in a wide window (the typical case), they may not realize where the TOC appears when it is displayed inline (after the top comment). So, they might not realize that there is an important presentation difference between putting a heading into the top comment, and putting it into a separate comment.

It seems useful to warn authors about headings in order to make them aware of this.

[...]

PR #628, opened 2021-03-10
Include headings in synopsis

Julow:
Heading were skipped while computing the synopsis for a module. Now the heading's text is used when it's the first element.

In the examples from https://github.com/ocaml/odoc/issues/235, the `CCArray` module wouldn't have a synopsis in `{!modules:...}` lists. This PR doesn't answer the rest of the discussion.

### DIFF
--- a/src/xref2/link.ml
+++ b/src/xref2/link.ml
@@ -25,13 +25,17 @@ let synopsis_from_comment docs =
     match elem.value with
     | `Paragraph p -> Some p
     | `List (_, items) -> list_find_map (list_find_map from_element) items
-    | _ -> None
+    | `Verbatim _ | `Code_block _ | `Modules _ ->
+        (* Skip some elements *)
+        None
   in
   list_find_map
     (function
       | { value = #Comment.nestable_block_element; _ } as elem ->
           from_element elem
-      | _ -> None)
+      | { value = `Heading (_, _, p); _ } ->
+          Some (p : Comment.link_content :> Comment.paragraph)
+      | { value = `Tag _; _ } -> None)
     docs
 
 exception Loop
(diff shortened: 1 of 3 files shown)


dbuenzli:
What is this PR supposed to solve ? The issue reported in #235 is not about giving a synopsis to `CCArray` it's about it's first heading not showing up in the toc and showing up in `.odoc-preamble` rather than`.odoc-contents`.

More precisely this PR changes the definition of synopses from take the *first paragraph* from take the *first paragraph or the contents of the first heading*.  I'm against this, this has never been an established practice and I don't think it makes sense (heading in modules are used to section the module's *contents*).

Julow:
This PR is not solving any issue, I'm just proposing to change the definition of synopsis.
I found that ocamldoc does this. I'll close.

dbuenzli:
> I found that ocamldoc does this. I'll close.

Just for the record `ocamldoc` does *not* do this. Namely no synopses are to be seen with the following invocations:

```
> cat a.mli
module B : sig
  (** {1:bla First header } *)
end
> ocamlc a.mli 
> ocamldoc -I . -html a.mli
> open index.html 
``` 

Using this definition would mean that 1) a lot of modules would get bogus synopses 2) What is a synopsis for a module and what is a header for a module become messy concepts.

Julow:
I can reproduce with:
```ocaml
(** {1 Heading}

    Doc for A. *)
module A = struct

end

(** {!modules:Ocamldoc_example.A} *)
```
The synopsis is what goes into the `{!modules:}` list.

dbuenzli:
Ah interesting! So basically it's totally confusing. It's better to settle on what was agreed on #597 for `odoc`.

PR #638, opened 2021-03-12
IR model: stop trying to play with word joiners (U+2060)

### LINKED ISSUE #620: Unintentional escape characters in odoc output (html)
In this odoc output for [Unix](https://b0-system.github.io/odig/doc/ocaml/Unix/index.html), look at the function `val link`.
Here's what it looks like:
```
val link : ?⁠follow:bool -> string -> string -> unit
```
but copy and paste it into your clipboard, and assign it into a string in JavaScript, and look at the character position where `f` in "follow" _should_ be.

```javascript
var text = "val link : ?⁠follow:bool -> string -> string -> unit"
alert(text.charCodeAt(12));
```

It prints the char code `8288`. This means there's some non printable char codes which make copy/pasting items from the page into utop or a file frustratingly fail to compile (you'll get errors about weird characters but can't see with your eyes where they are until you inspect byte by byte).

I'm assuming there was some reason why these were in there? Does anyone know the back story there?

dbuenzli:
It breaks cut and paste in the HTML renderer and other renderers have to ignore it.

Closes #620.

### DIFF
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -44,7 +44,7 @@ let make_name_from_path { Url.Path.name; parent; _ } =
 let label t ppf =
   match t with
   | Odoc_model.Lang.TypeExpr.Label s -> O.pf ppf "%s" s
-  | Optional s -> O.pf ppf "?%t%s" (O.entity "#8288") s
+  | Optional s -> O.pf ppf "?%s" s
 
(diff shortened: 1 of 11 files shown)


Drup (approved):

jonludlam:
Thanks!

PR #639, opened 2021-03-12
Lib ocamlary

### LINKED ISSUE #630: Install the ocamlary
Would it be possible to install the `ocamlary` back ? 

It [used](https://github.com/ocaml/odoc/blob/fd4605f6c94d61a9dd3cffabc76c712ba246ae4c/test/inactive/core/dune) to be installed by `odoc`. It's quite useful to be able to inspect `odoc`'s outputs on darker corners of the OCaml language.

dbuenzli:
On popular request, brings back the `ocamlary` library and installs it as `odoc.ocamlary`.

~~It seems that the source that was in the test/inactive/core inactive library had slightly diverged from the one in `test/cases`. I kept the former since a manual inspection seemed to indicate it had a few more test cases (also the former had the implementation)~~ See commit comment.

Closes #630.


dbuenzli:
> I kept the former since a manual inspection seemed to indicate

Mmh according to all this fancy CI that was maybe not a got idea… I'll try to converse to see how it goes.

dbuenzli:
I don't understand this `dune build @doc` CI failure, if I do this on my machine: 
```
ODOC_WARN_ERROR=true opam exec -- dune build @doc 
```
That exits with `0`.

dbuenzli:
Ah ! It does error after a `dune clean`, good old `make clean && make` :-).

So I tweaked the `ocamlary` to compile without odoc warnings. See the commit for details.

PR #642, opened 2021-03-16
Use expect test for parser tests

lubegasimon:


src/parser/test/test.ml
```diff
@@ -0,0 +1,9262 @@
...
+           \n'-' (bulleted list item) should begin on its own line."
+            "File \"f.ml\", line 1, characters 8-9:\
+           \n'-' (bulleted list item) is not allowed in the tags section.\
+           \nSuggestion: move '-' (bulleted list item) before any tags."))) |}]
+  end
+  [@ocaml.warning "-32"] in
+  ()
+
+let%expect_test _ =
+  let module Reference_component_kind = struct
```
Julow, line 5595: This section is meaningless, for the parser, references are just strings. This is meant to test `Odoc_model.Reference.parse`.
I suggest:
- Copy the tests related to quotes and a few tests related to the `kind-` syntax to the `Simple_reference` section
  (the `kind-` syntax is not tested there, only `kind:` is)
- Move all these tests to the main Odoc project (in a new `test/model` directory)

src/parser/test/dune
```diff
@@ -0,0 +1,7 @@
+(library
+ (name odoc_parser_test)
+ (inline_tests)
+ (optional)
```
Julow, line 4: I suggest removing that line. With this, tests won't run with no warning if ppx_expect isn't installed.

odoc.opam
```diff
@@ -47,6 +47,7 @@ depends: [
   "conf-jq" {with-test}
 
   "bisect_ppx" {with-test & = "2.5.0"}
+  "ppx_expect" {with-test}
```
Julow, line 50: On line 46, `sexplib0` won't be installed on ocaml version below `4.04.1`. Either this need to change or, if it really can't work below that version, we need a `enabled_if` condition on the parser tests.

Julow (changes requested):
`Odoc_model.Semantics` and `Odoc_model.Reference.parse` are no longer tested. (which were in the previous version)
They need to be tested too but outside of the "parser" library, maybe in a new `test/model` directory. A lot of these tests need to move there (the `Reference_component_kind` section, every tests that contain occurences of warnings defined in `src/model/semantics.ml`).

odoc.opam
jonludlam: that's what (optional) above is for.

Julow: The problem with `(optional)` is that we won't know if the tests are run. `enabled_if` seems more robust. I'm also suggesting to remove this constraint and see.

src/parser/test/test.ml
```diff
@@ -0,0 +1,9262 @@
...
+
+    let blank_line =
+      test "\n\n";
+      [%expect "((output ()) (warnings ()))"]
+
+    let cf_lf =
+      test "\r\n";
+      [%expect "((output ()) (warnings ()))"]
+  end
+  [@ocaml.warning "-32"] in
```
Julow, line 174: Apparently ocaml `<= 4.05` doesn't handle these attributes. I suggest removing that warning at the toplevel, just before the first test suite:
```ocaml
[@@@ocaml.warning "-32"]
```

odoc.opam
jonludlam: I'm not sure I understand why it's more robust - wouldn't we have the same problem of not knowing if the tests are run with both constructs? Is it just that it's more explicitly about the ocaml version?

Julow: If some versions can't run some tests, that's fine but that should be done with `enabled_if` to ensure it's correct and that we really want that (eg. one pure condition instead of implicit assumption about the environment).
`(optional)` dependencies will implicitly disable some tests depending on the current state of my dev environment (may be out of sync from the changing opam description). I almost wrote that tests don't run in my review.
Also, I may have `sexplib0` installed for an other reason, triggering tests that will fails (`enabled_if` is needed).

lubegasimon: I find https://github.com/ocaml/odoc/pull/642#discussion_r595298675 sensible, albeit, ` (enabled_if false)` still won’t warn if `ppx_expect` is not installed, and ` (enabled_if true)`, warns that `sexplib0` requires `>=4.04.1`

lubegasimon: > Is it just that it's more explicitly about the ocaml version?

I think so

Julow: Removing the `(optional)` is a good thing but now we have broken tests. You need to add `(enabled_if (>= %{ocaml_version} 4.04.1))`

jonludlam:
I think this is looking good now!

Issue #632, opened 2021-03-11
No synopsis rendered for modules for `Stdlib` modules


Julow:
```ocaml
module M : sig
  (** Doc for [M]. *)
end
```

In this example, the page for the top-level module doesn't contain the synopsis for the module `M` near its definition.

Ocamldoc does this and the documentation for the Stdlib may rely on this.

cc @dbuenzli

dbuenzli:
I'm afraid but stdlib modules still lack synopses is it maybe due to  https://github.com/ocaml/odoc/pull/649 ? I'm not familiar with the namespacing.

In general it would be nice to actually check issues are resolved before closing them. It seems I keep on reopening issues. 

It's not hard: 

```sh
opam pin add odoc .
opam pin add b0 --dev
opam pin add odig --dev
odig cache clear
odig doc -u 
```

dbuenzli:
> I'm afraid but stdlib modules still lack synopses is it maybe due to #649 ?

I tested with master now that it's merged. Still no synopses in the `Stdlib` module with `odig`  :-( 

Could this maybe due to the way `odig` drives `odoc` ? I'll try to investigate a bit.

dbuenzli:
So I think something is wrong in the way the stdlib does namespacing, since AFAIK that trick has never been properly documented, I'm not sure exactly *what* is wrong. 

But in any case the dependencies of the `Stdlib` module seem off because: 

```
> odoc compile-deps $(opam var lib)/ocaml/stdlib.cmti
CamlinternalFormatBasics b6c6694955e10001aed267571104a961
Stdlib 4b04b4eda19aa722df365141895fb347
```

Now I contrasted that with `Base` since its submodules have the synopses (when there is one). It seems base has no `base.cmti` so odig falls back on looking up `base.cmt`. This leads to: 

```
> odoc compile-deps $(opam var lib)/base/base.cmt
…
Base__List 5d64f423db3f4984be8e1eaf1922b201
Base__Lazy 0cbde1c360bc2c98496746406b8c2cbd
…
```

So the references to the namespaced modules are there in `Base`.

I suspect `odoc` is correct now but it would be nice to be able to sort out what needs to be done so that the stdlib gets its synopses. I changed the scope of the issue and we can close it whenever an appropriate issue has been opened where the work needs to be done. 

Finally I just checked  `stdlib.cmt` it doesn't seem it would help in this case: 

```
> odoc compile-deps $(opam var lib)/ocaml/stdlib.cmt 
[...]

PR #643, opened 2021-03-16
Take expansion's doc when computing a declaration's synopsis

Julow:
Fixes https://github.com/ocaml/odoc/issues/632

Before this patch, no synopsis was attached to declarations like this: (also work on classes, class types and module types)

```ocaml
module M : sig
  (** Expansion doc. *)
end
```

This is consistent with what ocamldoc does.

The first commit refactors the function generating the synopsis for `{!modules:}` lists so it can be shared with the generator. Before that, a simpler rule was used that didn't skip headings and other elements:

```ocaml
(** {1 Heading}

    @canonical M

    This is the synopsis. *)
```

```ocaml
(** - This is the synopsis. *)
```

These are rules I made up in https://github.com/ocaml/odoc/pull/597, it's still time to change them and to document them too.

### DIFF
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -994,16 +994,18 @@ module Make (Syntax : SYNTAX) = struct
         if t.virtual_ then O.keyword "virtual" ++ O.txt " " else O.noop
       in
 
-      let cname, expansion =
+      let cname, expansion, expansion_doc =
         match t.expansion with
-        | None -> (O.documentedSrc @@ O.txt name, None)
+        | None -> (O.documentedSrc @@ O.txt name, None, None)
         | Some csig ->
             let expansion_doc, items = class_signature csig in
             let url = Url.Path.from_identifier t.id in
             let page =
               make_expansion_page name `Class url [ t.doc; expansion_doc ] items
             in
-            (O.documentedSrc @@ path url [ inline @@ Text name ], Some page)
+            ( O.documentedSrc @@ path url [ inline @@ Text name ],
+              Some page,
+              Some expansion_doc )
       in
       let summary =
         O.txt Syntax.Type.annotation_separator ++ class_decl t.type_
@@ -1020,7 +1022,7 @@ module Make (Syntax : SYNTAX) = struct
       in
       let attr = [ "class" ] in
       let anchor = path_to_id t.id in
-      let doc = Comment.first_to_ir t.doc in
+      let doc = Comment.synopsis ~decl_doc:t.doc ~expansion_doc in
       Item.Declaration { attr; anchor; doc; content }
 
     let class_type (t : Odoc_model.Lang.ClassType.t) =
@@ -1029,16 +1031,18 @@ module Make (Syntax : SYNTAX) = struct
       let virtual_ =
         if t.virtual_ then O.keyword "virtual" ++ O.txt " " else O.noop
       in
-      let cname, expansion =
+      let cname, expansion, expansion_doc =
         match t.expansion with
-        | None -> (O.documentedSrc @@ O.txt name, None)
+        | None -> (O.documentedSrc @@ O.txt name, None, None)
         | Some csig ->
             let url = Url.Path.from_identifier t.id in
             let expansion_doc, items = class_signature csig in
             let page =
               make_expansion_page name `Cty url [ t.doc; expansion_doc ] items
             in
-            (O.documentedSrc @@ path url [ inline @@ Text name ], Some page)
+            ( O.documentedSrc @@ path url [ inline @@ Text name ],
+              Some page,
+              Some expansion_doc )
       in
       let summary = O.txt " = " ++ class_type_expr t.expr in
       let expr = attach_expansion (" = ", "object", "end") expansion summary in
@@ -1050,7 +1054,7 @@ module Make (Syntax : SYNTAX) = struct
       in
       let attr = [ "class-type" ] in
       let anchor = path_to_id t.id in
-      let doc = Comment.first_to_ir t.doc in
+      let doc = Comment.synopsis ~decl_doc:t.doc ~expansion_doc in
       Item.Declaration { attr; anchor; doc; content }
   end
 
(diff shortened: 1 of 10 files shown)


src/document/generator.ml
```diff
@@ -1533,7 +1537,7 @@ module Make (Syntax : SYNTAX) = struct
       let content = { Include.content; status; summary } in
       let attr = [ "include" ] in
       let anchor = None in
-      let doc = Comment.first_to_ir sg_doc in
+      let doc = Comment.synopsis ~decl_doc:[] ~expansion_doc:(Some sg_doc) in
```
Julow, line 1540: Unrelated to this PR:
This ignores `t.doc` and takes only the first paragraph of the expansion doc, the rest is removed. I think we should at least show `t.doc` in its entirety.
I'll propose something is an other PR.

dbuenzli:
> it's still time to change them and to document them too.

Could you please spell them out ? Because I'm not sure what you mean by skipping.

More precisely, nothing should be *skipped*. You simply take the first paragraph of the comment attached to module. If there is no such thing there is no synopsis. Do not try to peek past headings this will create a lot of bogus synopses.

For example here there should be no synopsis:

> ```ocaml
> (** {1 Heading}
> 
>     @canonical M
> 
>     This is the synopsis. *)
> ```

dbuenzli:
Here are [example](https://b0-system.github.io/odig/doc/decompress/index.html) of bogus synopses.

dbuenzli:
If we agree on the current state of discussion between me and @lpw25 in #640 I would suggest the following definitions:

1. The *preamble* of a module are all the paragraphs in the comment attached to the module until the first heading.
2. The *synopsis* of a module is the first paragraph of the preamble.

This not exactly what `ocamldoc` did but it's what most documented modules out there do (some might need fixing but not that much) and these definitions are easy to understand and remember (for their good properties see the discussion in #640).

Julow:
> Could you please spell them out ?

The temporary rule is: The synopsis is the first paragraph, from top to bottom, ignoring headings and tags but considering list items.

I agree it is surprising and produces bogus synopses. That's why I warned about it.

> The synopsis of a module is the first paragraph of the preamble.

I think you actually suggest: the first element of the preamble, *if* it is a paragraph. (I guess also if it is a list and the first item has a paragraph as first element)
By "element" I mean "block element", which is defined in code as paragraphs, headings, tags, lists, verbatim/code blocks and module lists.

I suggest adding to this rule: Heading's text if the first element is a heading.
I think it's common to start with a heading and I'd find it confusing to get an empty synopsis.

dbuenzli:
> The temporary rule is: The synopsis is the first paragraph, from top to bottom, ignoring headings and tags but considering list items.

I don't think it's a good idea.

> I think you actually suggest: the first element of the preamble, _if_ it is a paragraph.

Yes. Note that the simple new rule I propose will not produce bogus synopses (except for those that relied on the ill-defined *sentence* extract). It will however suppress some, namely those who relied on synopses being extracted from headings, like [those](https://b0-system.github.io/odig/doc/decompress/index.html) I suspect. 

> I suggest adding to this rule: Heading's text if the first element is a heading.

I'd rather not because then it all becomes confusing. The heading is not part of the preamble so we loose the simple and obvious definitions.

Besides it also makes the purpose of headings shady. Headings are here to chunk the module contents. I do not like to overload or change their functionality when they happen to appear at the beginning of a module.

Finally a synopsis is supposed to be a small sentence terminated by a dot. Headings are not supposed to be terminated by a dot.

> I think it's common to start with a heading and I'd find it confusing to get an empty synopsis.

I'm not sure it's common. In any case in general you should rather start with a preamble. There's already a heading starting the page which is `module M`. 

The pattern supported by the rules I propose has been well-established by the stdlib for as long as I have been programming in OCaml and as I mentioned [here](https://github.com/ocaml/odoc/pull/640#issuecomment-797792130) corresponds to what you'll find most of the time out there.

dbuenzli:
One thing I forgot to mention is that by having less ways on how synopses are defined we enforce more page structure regularity across APIs which overall improves the OCaml documentation user experience.

Julow:
I changed the rule according to our discussion.

dbuenzli:
> An exception is made if it is a list, the first element of the first
> list item is considered.

I'm not sure why you made that exception, is there any documentation pattern you have in mind ? Lists are usually cohesive blocks, extracting out the first paragraph of their first element seems a bit weird.

Julow:
Lists contain paragraphs. I don't have a strong reason, that's how I'd expect it to work. Ocamldoc was taking the first sentence of anything (heading, paragraph, verbatim block), taking more seems more consistent than otherwise.

dbuenzli:
> I don't have a strong reason, that's how I'd expect it to work.

Ah ! I don't :-) Especially from a writing perspective I don't see the first paragraph of the first element of a list as being a particularly self-contained and self-identified unit. I'd rather see that as a bug.

Beyond that writing the documentation features, i.e. explaining features for humans, is always useful to judge them (if you can't explain it simply, iterate…) – I did my best to write something for what you propose, maybe you can make it simpler but I still prefer the first definition.

1. The _synopsis_ of a module is the first block of the preamble, if it is a paragraph.
2. The _synopsis_ of a module is defined by the first block of the preamble. If the first block is a paragraph, it is the paragraph itself. If the first block is a list, it is the first paragraph of the first item of the list. Otherwise the synopsis is empty.

src/document/generator.ml
Julow: Issue: https://github.com/ocaml/odoc/issues/645

jonludlam:
Rebase needed please!

dbuenzli:
Also this still implements definition 2. the docs of #644 use definition 1.

Julow:
Does anyone have an opinion on this? (whether we should use the first item of a list if it is the first element)

jonludlam:
I think I'm with @dbuenzli on this one, I wouldn't have expected to pick out the first item in a list.

Julow:
I've removed the exception on lists. I've also fixed a bug where the synopsis could be computed for the expansion doc even if the decl doc was present but didn't contain a synopsis.

jonludlam:
AIUI this is now ready to merge, correct?

Julow:
I believe it is.

PR #644, opened 2021-03-17
Document how to write documentation for modules

Julow:
The `preamble` and `synopsis` parts describe the current state of https://github.com/ocaml/odoc/pull/640 and https://github.com/ocaml/odoc/pull/643

### DIFF
--- a/src/odoc/using-odoc.mld
+++ b/src/odoc/using-odoc.mld
@@ -24,7 +24,49 @@ complementary notes.
 
 {1:interfaces Documenting your interfaces}
 
-...
+Odoc is built to produce documentation for your {e libraries}, and the unit of
+organisation is the {e module}. Documentation is written by putting special
+{{!comments} comments} into the source of the {e module} or {e module
+interface}, and correctly placing these is critical to producing good output.
+Odoc expects the documentation to be structured in a logical way, and will work
+best if the following conventions are applied.
+
+The overall structure is that modules start with a {{!preamble} preamble} or
+'Lead Section' that serves as an overview of the most important information
+about module. This is followed by the content of the module, organised into {e
+sections} and {e subsections}, the structure of which will be used to populate
+a {e table of contents} which will be structurally placed immediately after the
+preamble.
+
+The first paragraph of the preamble will be treated as the module {{!synopsis}
+synopsis}, and will be used as a short description of the module when it
+appears in a list of modules elsewhere in the documentation of the library.
+
+{2 Comments}
+
+Documentation comments are delimited with [(** ] (exactly two [*]) and [*)] and
+can be attached to a declaration or be floating in a signature.
+{{:https://caml.inria.fr/pub/docs/manual-ocaml/doccomments.html}This is defined
+by OCaml.} It is also possible to attach documentation comments to individual
+record fields, constructors, function arguments, class items, etc..
+
+{[
+(** This is the top-comment. It documents the whole module and will be used to
+    derive the preamble and the synopsis. *)
+
+type t
+(** This comment is attached to [t]. *)
+
+(** This comment is "floating", it can be used to define sections.
+
+    {1 This is a heading} *)
+
+(* ... *)
+]}
+
+The documentation can be formatted, Odoc accepts the same markup language as
+{{:https://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#s%3Aocamldoc-comments}
+ocamldoc} with some exceptions, see {!page-interface.Changes}.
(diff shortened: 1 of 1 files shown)


src/odoc/using-odoc.mld
```diff
@@ -24,8 +24,168 @@ complementary notes.
...
+
+{[
+(** This is {e not} the top-comment because it's attached to [t]. *)
+type t
+]}
+
+{2 Preamble}
+
+The preamble is composed of the comment attached to a declaration and the
+top-comment of the expansion.
```
dbuenzli, line 122: The notion of *expansion* is not explained before this occurence. I'm personally unclear to what it refers to. Could you maybe expand on that ?

src/odoc/using-odoc.mld
```diff
@@ -24,8 +24,168 @@ complementary notes.
...
+It is special only because it will be placed in the [header] part of the page,
+just before the TOC (if any), and is used to compute the {e synopsis}.
+
+{[
+(** This is the comment attached to the declaration. *)
+module M : sig
+  (** This is the top-comment of the expansion. *)
```
dbuenzli, line 129: Just to make things clear. What is the synopsis and preamble here ? My expectation is that the outer comment gets merged with the second with a blank line between them (i.e. there are two paragraphs here).

src/odoc/using-odoc.mld
```diff
@@ -24,8 +24,168 @@ complementary notes.
...
+module M : sig
+  (** Preamble. *)
+
+  (** {1 Heading}
+
+      This paragraph is not part of the preamble. *)
+end
+]}
+
+The comment attached to the declaration shouldn't contain any heading.
```
dbuenzli, line 159: I understand what you mean here, but I'm not sure it's very clear. Maybe this would be better said around an example that has a comment attached to the declaration.

src/odoc/using-odoc.mld
```diff
@@ -24,8 +24,168 @@ complementary notes.
...
+
+{2 Synopsis}
+
+The synopsis of a module (a module type, a class, etc..) is the first
+paragraph of the {!preamble}, {e if} the preamble starts with a paragraph.
+
+It is rendered near declarations and in [{!modules:...}] lists.
```
dbuenzli, line 166: near declarations -> after module declarations

src/odoc/using-odoc.mld
```diff
@@ -24,8 +24,168 @@ complementary notes.
...
+    current page near the declaration of [M]. This paragraph will be part of
+    [M]'s preamble. *)
+module M : sig
+  (* ... *)
+end
+]}
+
+{[
+module M : sig
+  (** This paragraph is the synopsis of the module [M]. *)
```
dbuenzli, line 184: Here again might be worth indicating what happens when you have: 

```ocaml
(** Bli bla *)
module M : sig 
  (** Hey ho *)
end

**Review by dbuenzli — COMMENTED** [review]
Nice docs @Julow ! I have left a few comments.

File: src/odoc/using-odoc.mld (hunk 5 above)
Comment (Julow) [reply src/odoc/using-odoc.mld:184]: There is an other before that covering this case. This example comes in pair, maybe I should merge them in a single block and call my modules `M` and `N` ?

File: src/odoc/using-odoc.mld (hunk 3 above)
Comment (Julow) [reply src/odoc/using-odoc.mld:159]: Maybe I can drop this sentence ? It's not related to the example, it's an aside saying the thing described above doesn't happen with decl comments.

File: src/odoc/using-odoc.mld (hunk 1 above)
Comment (Julow) [reply src/odoc/using-odoc.mld:122]: I think I've managed to remove that word, it isn't necessary in the documentation. I meant the `sig ... end` part when there is one.

File: src/odoc/using-odoc.mld [hunk 6]
```diff
@@ -24,8 +24,171 @@ complementary notes.
...
+(** This comment is "floating", it can be used to define sections.
+
+    {1 This is a heading} *)
+
+(* ... *)
+]}
+
+The documentation can be formatted, Odoc accepts the same markup language as
+{{:https://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#s%3Aocamldoc-comments}
+ocamldoc} with some exceptions, see {!page-interface.Changes}.
```
Drup, line 69: This is not the topic of this PR, but I think we should just include the markup language documentation in our documentation, and only just say "this is based on the ocamldoc one".

Drup (commented):
This is great! I don't really have comments, you were very thorough :)

PR #646, opened 2021-03-18
Preview: upgrade to ocamlformat.0.18.0

gpetiot:
This is a preview of the not-yet-released ocamlformat.0.18.0, please wait until the package is published in opam to merge this PR.
The diff is composed of:
 -   a break was added between polytype quantification and arrow-type body (spurious spaces are removed), causing the arrow-type to properly break

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,1 +1,1 @@
-version=0.17.0
+version=0.18.0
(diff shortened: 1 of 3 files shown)


jonludlam:
Thanks for the preview! Good to know what will change.

Issue #559, opened 2021-01-07
Canonical does not work on compilation units


lpw25:
If you make an mli file with an `@canonical` in its first comment, it does not work as expected. The canonical seems to just be ignored.

PR #649, opened 2021-03-24
Handle @canonical tags on compilation unit

Julow:
Fix https://github.com/ocaml/odoc/issues/559#event-4382610148

Before this, `@canonical` tags in the top-comments of a file were ignored.

### DIFF
--- a/src/xref2/env.ml
+++ b/src/xref2/env.ml
@@ -11,10 +11,7 @@ type lookup_unit_result =
 
 type root =
   | Resolved of
-      (Digest.t
-      * Odoc_model.Paths.Identifier.Module.t
-      * bool
-      * Component.Module.t Component.Delayed.t)
+      (Digest.t * Odoc_model.Paths.Identifier.Module.t * Component.Module.t)
   | Forward
 
 type resolver = {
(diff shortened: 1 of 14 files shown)


src/xref2/env.ml
```diff
@@ -372,18 +369,15 @@ let lookup_root_module name env =
...
             | `Root _ as id ->
-                let m =
-                  Component.Delayed.put (fun () ->
-                      let unit = r.resolve_unit u.root in
-                      module_of_unit unit)
-                in
-                Some (Resolved (u.root.digest, id, u.hidden, m))
+                let unit = r.resolve_unit u.root in
+                let m = module_of_unit unit in
+                Some (Resolved (u.root.digest, id, m))
```
Julow, line 374: Resolving root modules is no longer lazy because we need to check for `canonical` everytime.
Is here a case where this would cause a slowdown ? (no change on base)

jonludlam: Could be actually - could you do a test run on core_kernel or similar? I'd be interested to see the difference in memory usage too.

Julow: I find no regression on core_kernel (29s in both cases). Maximum memory doesn't seem to change too.

PR #653, opened 2021-03-26
Alias more when strengthening

jonludlam:


jonludlam:
This patch reduces the produced HTML in accessor_core from 1.8 megs to 570k

jonludlam:
This needs some extra work - it's leading to broken links. For example, the first line in `ppx_optcomp.0.14.1` is a link to `Caml.Filename` and this gives a 404.

src/xref2/strengthen.ml
```diff
@@ -20,21 +20,30 @@ open Delayed
...
             let canonical =
               match canonical with
               | Some p -> Some (`Dot (p, name))
               | None -> None
             in
-            match module_ ?canonical (`Dot (prefix, name)) (get m) with
-            | None -> (item :: items, s)
-            | Some m' -> (Module (id, r, put (fun () -> m')) :: items, id :: s))
+            let m' = module_ ?canonical (`Dot (prefix, name)) (get m) in
+            (Module (id, r, put (fun () -> m')) :: items, id :: s)
```
Julow, line 46: The whole computation of `m'` could be delayed (it's not used otherwise)

src/xref2/strengthen.ml
```diff
@@ -20,21 +20,30 @@ open Delayed
...
     Cpath.module_ -> ?canonical:Cpath.module_ -> Signature.t -> Signature.t =
  fun prefix ?canonical sg ->
+  let sg', strengthened_modules = sig_items prefix ?canonical sg in
+  (* Format.eprintf "Invalidating modules: %a\n%!" (Format.pp_print_list Ident.fmt) strengthened_modules; *)
+  let substs =
+    List.fold_left
+      (fun s mid -> Subst.path_invalidate_module (mid :> Ident.path_module) s)
+      Subst.identity strengthened_modules
+  in
+  Subst.signature substs sg'
```
Julow, line 30: We can apply this only if `strengthened_modules` is not empty.

Julow (approved):
I'm not familiar with the strengthen logic but this sounds good to me.

src/xref2/strengthen.ml
Julow: I couldn't measure a performance improvement (it is smaller than the noise).

jonludlam: I'm slightly hesitant to do this because it would mean the identifiers don't get fresh names even though they're changed.

src/xref2/strengthen.ml
jonludlam: Good spot!

src/xref2/strengthen.ml
Julow: I haven't thought the identity subst would still have an effect. It makes sense here.

PR #654, opened 2021-03-30
Inherit top-comment from @inline includes 

Julow:
Finally closes https://github.com/ocaml/odoc/issues/478 (this is the last remaining issue).

A signature starting with an `include` with the `@inline` tag that doesn't have a top-comment inherits the top-comment from the included module type.

On top of https://github.com/ocaml/odoc/pull/647. It's changing unrelated code but the expanded documentation would be rendered twice otherwise.


Julow:
Rebased and ready for review.

Julow:
Just rebased :)

Issue #616, opened 2021-02-28
HTML page for functors is hard to comprehend.


dbuenzli:
Since the page rendering adds non-user defined headings and the resulting signature is inlined it makes it for a confusing heading structure: 

https://b0-system.github.io/odig/doc/asetmap/Asetmap/Set/Make/index.html
([`src`](https://github.com/dbuenzli/asetmap/blob/master/src/asetmap.mli))

Here are a few ideas: 

1. Use heading `h1` for `Parameter` and `Signature` (`h1` is out of reach for `ocamldoc` users).  
2. Wrap the current `Parameters` section in a `div.odoc-parameters` and follow with the inlined signature without adding the `Signature` heading; that is let the inlined sig define the rest of the page. 
3. Increment headings up by one in the inlined signature (this is asking for trouble both rendering an UX wise).
4. Keep headings as is now but include the signature in an `<article class="odoc-signature">`

Both 4 and 3 are asking for stylesheet and rendering troubles. 1, 3 and 4 also ask a few questions for the page toc and consistency w.r.t. other pages.

Altogether I have the impression that the easiest and best would be 2.

Drup:
I was also troubled by the rendering for functor page. I think I have tried or considered all of those at various points. Removing the `Signature` heading is actually really bad for readability, because the limit between parameters and signature becomes unclear.
Maybe the best is to insert the `Signature` heading only if there is no such heading in the documentation.

dbuenzli:
> Removing the `Signature` heading is actually really bad for readability, because the limit between parameters and signature becomes unclear.

That's the point of: 

> 2. Wrap the current `Parameters` section in a `div.odoc-parameters`

It then becomes possible to do something stylewise with either `.odoc-parameters + *` or `.odoc-parameters` itself to properly delineate the parameters block

PR #655, opened 2021-04-01
Add a plain text output.

Drup:
@lpw25 has regularly requested a "plain text" output for odoc, which allow to browse the document at a leisure pace, from the comfort of a text editor.

This patch adds a new output (yay), which does exactly that. The format chosen doesn't require any tooling except a text editor. Furthermore, if standard tooling from ocaml is available, such as merlin or ocp-index, jump to-definition should work immediately, along with nice formatting and colouration.

Here is an extract from `ocamlary`:

```ocaml
(** {3:advanced-type-stuff Advanced Type Stuff}
*)

(**
  This comment is for record.
  This comment is also for record.
*)
type record = {
  field1 : int;
  (**
    This comment is for field1.
  *)
  field2 : int;
  (**
    This comment is for field2.
  *)}

(**
  This comment is for variant.
  This comment is also for variant.
*)
type variant = 
  | TagA
  (**
    This comment is for TagA.
  *)
  | ConstrB of int
  (**
    This comment is for ConstrB.
  *)
  | ConstrC of int * int
  (**
    This comment is for binary ConstrC.
  *)
  | ConstrD of int * int
  (**
    This comment is for unary ConstrD of binary tuple.
  *)
```

The behaviour of includes is governed by the attributes, and unfolded by default. Submodules are also unfolded by default.

Since these files can be considered as the simplest representation of ML interfaces, I simply called them `.mli`. 

This output is largely functional, but not yet completely polished. In particular, it occasional produces less-than-ideal indentation. I plan to improve that aspect if people like this new output format. Functors are also a bit problematic, for reasons already mentioned in #616

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@ test : build
 	$(DUNE) build $(DUNE_ARGS) @test/model/runtest --no-buffer -j 1
 	$(DUNE) build $(DUNE_ARGS) @test/html/runtest --no-buffer -j 1
 	$(DUNE) build $(DUNE_ARGS) @test/man/runtest --no-buffer -j 1
+	$(DUNE) build $(DUNE_ARGS) @test/mlioutput/runtest --no-buffer -j 1
 	$(DUNE) build $(DUNE_ARGS) @test/latex/runtest --no-buffer -j 1
 	$(DUNE) build $(DUNE_ARGS) @test/xref2/runtest --no-buffer -j 1
 
--- a/src/mlioutput/dune
+++ b/src/mlioutput/dune
@@ -1,0 +1,6 @@
+(library
+ (name odoc_mli)
+ (public_name odoc.mli)
+ (instrumentation
+  (backend bisect_ppx))
+ (libraries odoc_model odoc_document))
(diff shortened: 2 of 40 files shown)


lpw25:
This looks like a good approach for the solving the "reviewable output" use case. It's less clear this is what you want for the "read docs from the editor" use case, since you might want something that editors know how to show with working links e.g. markdown, but this is still pretty good for that use case too.

It's not entirely clear from the example whether the output is showing references or formatting in the comments. I'd be tempted to show formatting but not references. Or maybe show that there is a resolved reference but not where it links to.

Drup:
> It's not entirely clear from the example whether the output is showing references or formatting in the comments. I'd be tempted to show formatting but not references. Or maybe show that there is a resolved reference but not where it links to.

The made-up markup uses a form of command `{X ..}`. This is used for bold and other things. References are rendered using the syntax: `{!Foo.Bar.baz}`. Currently, the actual kind is not inserted, but that would be very easy to add.

Drup:
There are approximately 2 billions ways that are strictly better to do what we want. :wink: 
The date has passed. Let's close! :tropical_fish:

dbuenzli:
Coquin :-)

Drup:
;)

(That being said, If someone is interested to implement a markdown/wikicreole/etc output, the code can probably be used as inspiration!)

PR #658, opened 2021-04-06
Resolve some references in synopses in `{!modules}` lists

Julow:
These references need to be resolved from their containing module's environment, which is only available when linking that module.
This PR adds a warning in case a reference is found and doesn't try to resolve it.

It should be possible to resolve _some_ fully-qualified references using the available environment as a best-effort and to resolve references local to their parent module. That PR doesn't try that.

Rebased on top of https://github.com/ocaml/odoc/pull/654


test/xref2/references_to_pages.t/run.t
```diff
@@ -21,5 +25,5 @@ Every references in `Bad_references` should not:
 Every references in `P` should resolve:
 
   $ odoc_print page-p.odocl | jq_scan_references
-  {"`Root":["M","`TUnknown"]}
-  {"`Dot":[{"`Root":["M","`TUnknown"]},"t"]}
+  {"`Resolved":{"`Identifier":{"`Root":[{"`RootPage":"test"},"M"]}}}
+  {"`Resolved":{"`Type":[{"`Identifier":{"`Root":[{"`RootPage":"test"},"M"]}},"t"]}}
```
Julow, line 29: This changed because I've added back the missing `M` module. I think that's how the test was intended to be.

test/xref2/module_list.t/main.mli
```diff
@@ -56,3 +57,10 @@ module Inline_include : sig
   include T
   (** @inline *)
 end
+
+module Resolve_synopsis : sig
+  (** This should be resolved when included: {!Main.Resolve_synopsis.t}. These
+      shouldn't: {!t} {!Resolve_synopsis.t} *)
```
Julow, line 63: It should be easy to resolve these two references:
- Add the target module to the new env
- Open it

How sound it is ?

Julow:
I implemented best-effort resolving of fully-qualified references encountered in the synopsis.

There were no error reported when a reference failed to resolve so I've added one and attached some information to it when it happens while resolving a synopsis. The result is quite verbose but don't trigger fatal warnings, it'll need to be improved later. (errors need to be propagated while resolving the references)

test/xref2/module_list.t/main.mli
jonludlam: Let's leave this as an issue to resolve later - we need to discuss the various options available to us.

Julow:
I've removed the warning code. (https://github.com/ocaml/odoc/issues/664)

PR #661, opened 2021-04-13
Separate handling of internal tags

Julow:
Fixes https://github.com/ocaml/odoc/issues/632#issuecomment-808082774

Internal tags are used to change Odoc's behavior and are never rendered. This caused Stdlib modules to have no synopsis because the `@canonical` tag was found before any synopsis could be extracted.

This PR handles these tags (`@canonical`, `@inline`, `@open` and `@closed`) earlier in the pipeline and removes them from the model. 

This isn't the only solution to this problem but I think it's a nice refactoring to have. I believe regular tags should also be handled ([previous discussion](https://github.com/ocaml/odoc/pull/647#discussion_r597879716)).


test/xref2/canonical_unit.t/run.t
```diff
@@ -4,6 +4,8 @@ top-comment.
 The module Test__X is expected to be referenced through Test.X.
 
   $ compile test__x.mli test.ml
+  File "test.ml", line 15, characters 6-24:
+  Unexpected tag '@canonical' at this location.
```
Julow, line 8: This is expected. This `@canonical` tag is indeed unhandled.
This PR implements the required refactoring to fix this easily, though.

src/model/semantics.ml
```diff
@@ -195,24 +246,14 @@ and nestable_block_elements status elements =
...
-      ok tag
-  | `Canonical { value = s; location = r_location } -> (
-      let path = Reference.read_path_longident r_location s in
-      match path with
-      | Result.Ok path -> ok (`Canonical path)
-      | Result.Error e ->
-          Error.warning status.warnings e;
-          let placeholder = [ `Word "@canonical"; `Space " "; `Code_span s ] in
-          let placeholder = List.map (Location.at location) placeholder in
-          Error (Location.at location (`Paragraph placeholder)))
```
Julow, line 215: I've removed this placeholder logic. This tag isn't intended to be rendered. If it contains a syntax error, only the programmer need to be informed (through a warning) and not users reading the documentation.

Julow (commented):
I've added a warning in case an internal tag is unhandled (eg. `@inline` on an item that isn't an `include` or `@canonical` on a value), for the same reason as for removing the placeholder in case of syntax error.

src/loader/cmi.ml
```diff
@@ -332,12 +332,7 @@ let mark_class_declaration cld =
   List.iter mark_type_parameter cld.cty_params;
   mark_class_type cld.cty_params cld.cty_type
 
-let canonical doc : [`Dot of Path.Module.t * string] option =
-  let doc = List.map Odoc_model.Location_.value doc in
-  match List.find (function `Tag (`Canonical _) -> true | _ -> false) doc with
-  | exception Not_found -> None
-  | `Tag (`Canonical (`Dot (p, n))) -> Some (`Dot (p, n) )
-  | _ -> None
+(* Handle top-comments *)
```
jonludlam, line 335: I don't think this comment should be here

Julow: Indeed, removed.

src/loader/cmi.ml
```diff
@@ -512,7 +505,7 @@ let read_value_description env parent id vd =
   let open Signature in
   let id = Env.find_value_identifier env id in
   let container = (parent : Identifier.Signature.t :> Identifier.LabelParent.t) in
-  let doc = Doc_attr.attached container vd.val_attributes in
+  let doc, () = Doc_attr.attached Odoc_model.Semantics.Expect_none container vd.val_attributes in
```
jonludlam, line 508: would it be useful to have a helper function do this? Since _most_ of the elements don't have internal tags.

jonludlam:
Other than that minor point, I think this is a really nice change.

test/xref2/module_preamble.t/run.t
```diff
@@ -13,6 +13,8 @@ and that "hidden" modules (eg. `A__b`, rendered to `html/A__b`) are not rendered
 
   $ odoc compile --pkg test -o a__b.odoc -I . a__b.cmti
   $ odoc compile --pkg test -o a.odoc -I . a.cmti
+  File "a.mli", line 4, characters 4-17:
+  Unexpected tag '@canonical' at this location.
```
Julow, line 17: This isn't expected however. This is caused by `` `Root `` paths to not be valid canonical paths, I'll add a different warning for this case.

src/loader/cmi.ml
Julow: Done.

test/xref2/module_preamble.t/run.t
Julow: I've just added a specific warning for this case.

PR #662, opened 2021-04-15
Handle @canonical tags in the top-comment of signatures

Julow:
Previously, `@canonical` tags were allowed to be attached to module, module type and type declarations and in the top-comment of compilation units.

This handles `@canonical` tags in the top-comment of signatures and structures, for example:

```ocaml
module M : sig
  (** @canonical Lib.M *)

  ...
end
```


test/xref2/canonical_unit.t/run.t
```diff
@@ -3,9 +3,7 @@ top-comment.
 
 The module Test__X is expected to be referenced through Test.X.
 
-  $ compile test__x.mli test.ml
-  File "test.ml", line 15, characters 6-24:
-  Unexpected tag '@canonical' at this location.
+  $ compile test__x.mli test__y.ml test.ml
```
Julow, line 6: The scope of this test increases and will conflict with https://github.com/ocaml/odoc/pull/653. Perhaps at this point, we could test every canonical features in one larger test case ?

jonludlam: could do, though I'd still prefer to try to test canonical on it's own rather than in conjunction with hidden, as there is some overlap in how they both work. Obviously we want to test them together as well, just in a different test -- perhaps a 'namespaces' test.

Julow: Alright, I'll remove the "hidden" part of this test to turn it into a general canonical test. The combination with hidden modules is tested in the `ocaml_stdlib.t`, we might want to have a bigger "wrapped_library" test too.

Julow: Done.

jonludlam:
This looks good, just needs a rebase.

Issue #651, opened 2021-03-25
Improve generators' tests


lubegasimon:
The state of the current tests for generators (html, man, and latex) is not neat. This issue proposes an improvement to expect-test style.

jonludlam:
Merged!

PR #665, opened 2021-04-20
improve generator tests

lubegasimon:
This PR is intended to fix https://github.com/ocaml/odoc/issues/651.

**Todo:**
- [ ] Currently (in this PR), to generate the output files, one has to run the test command three times. The first one for generating `link.dune.inc.gen` and promoting it to `link.dune.inc`, the second for generating `<html&latex&man>.dune.inc.gen`, and promoting them to `<html&latex&man>.dune.inc`, and the third for generating and promoting output files (latex, html, and man-pages). In that sense, it makes more sense to wrap the series of test command under one command


Makefile
```diff
@@ -40,6 +40,7 @@ serve :
 test : build
 	$(DUNE) runtest $(DUNE_ARGS) src/parser/test
 	$(DUNE) build $(DUNE_ARGS) @test/model/runtest --no-buffer -j 1
+	$(DUNE) build $(DUNE_ARGS) @runtest --auto-promote --no-buffer -j 1
```
lubegasimon, line 43: this should be `$(DUNE) build $(DUNE_ARGS) @test/generators/runtest --no-buffer -j 1`

test/generators/gen_link/gen_link.ml
```diff
@@ -0,0 +1,214 @@
...
+
+let die s =
+  prerr_endline s;
+  exit 1
+
+let path () f = Filename.quote (Fpath.to_string f)
+
+let ext' () f = Filename.quote (Fpath.get_ext f)
+
+let gen_rule_for_source_file source_file f =
```
Julow, line 135: This function shouldn't take the full path as a string and the file name as an `Fpath.t`. It should only the full path as an `Fpath.t`. `basename` below should be `Fpath.rem_ext (Fpath.basename path)`.

test/generators/html/gen_html/gen_html.ml
```diff
@@ -0,0 +1,92 @@
...
+        (fun t ->
+          List
+            [
+              Atom "with-stdout-to";
+              Atom (tweak_html_target t ^ ".gen");
+              List [ Atom "progn"; List [ Atom "system"; Atom ("cat " ^ t) ] ];
+            ])
+        targets
+  (* TODO: enhance error handling *)
+  | _ -> []
```
Julow, line 44: You can call `die` here, like in `gen_link`.

test/generators/html/gen_html/gen_html.ml
```diff
@@ -0,0 +1,92 @@
...
+    | Unix.WEXITED n -> `Exit n
+    | _ -> `Signal
+  in
+  (lines, status)
+
+let expected_html_targets odocl =
+  lines_of_command "odoc" [ "html-targets"; "-o"; "html.gen"; odocl ]
+
+let tweak_html_target target =
+  String.(concat "." (split_on_char '/' (sub target 14 (length target - 14))))
```
Julow, line 28: What is this constant "14" ? If you need a specific occurrence, I suggest you use pattern matching:
```ocaml
match split_on_char '/' target with
| _ :: _ :: rest -> concat ...
| _ -> assert false
```

test/generators/html/gen_html/gen_html.ml
```diff
@@ -0,0 +1,92 @@
...
+let diff_rules targets =
+  List.iter
+    (fun t -> Printf.printf "%s\n\n" (to_string_hum (diff_rule t)))
+    targets
+
+let gen_html_rule filenames =
+  filenames
+  |> List.iter (fun odocl ->
+         let targets, _ = expected_html_targets odocl in
+         Printf.printf "%s\n\n" (to_string_hum (target_rule odocl));
```
Julow, line 87: `target_rule` will call again `expected_html_targets` indirectly, which isn't good because it's a slow operation.
You should find a way to pass `targets` through this function.

test/generators/gen_link/gen_link.ml
```diff
@@ -0,0 +1,214 @@
...
+            ];
+        ];
+    ]
+
+let gen_html_rule filenames =
+  printf "%s\n\n" (to_string_hum (gen_rule_html filenames));
+  printf "%s\n\n" (to_string_hum diff_rule_html)
+
+let read_file_from_dir dir =
+  let filenames = Sys.readdir dir |> Array.to_list in
```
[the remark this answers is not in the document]
lubegasimon: done

test/generators/html/gen_html/dune
```diff
@@ -0,0 +1,3 @@
+(executable
+ (name gen_html)
+ (libraries stdio sexplib0 unix fpath))
```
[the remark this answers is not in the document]
lubegasimon: done

test/generators/gen_link/gen_link.ml
```diff
@@ -0,0 +1,214 @@
...
+  List.iter
+    (fun f ->
+      let open Fpath in
+      let source_file = append (to_dir_path (v dir)) (v f) in
+      gen_rule_for_source_file (to_string source_file) (v f))
+    filenames;
+  let filenames = List.map (fun f -> Fpath.v f) filenames in
```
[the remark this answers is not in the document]
lubegasimon: I’m not sure I understand clearly why you suggest the changes. I think my intuition of what’s already going on is right; but I’m not looking for reasons to preserve it. I’m happy to make changes having understood clearly what you mean

test/generators/gen_link/gen_link.ml
lubegasimon: Good catch

lubegasimon: Done!

test/generators/html/gen_html/gen_html.ml
lubegasimon: Done!

test/generators/html/gen_html/gen_html.ml
lubegasimon: good catch!

test/generators/gen_link/gen_link.ml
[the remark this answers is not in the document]
Julow: With the change I suggested, paths will be converted to `Fpath` sooner and the raw string arguments will never be used. Also, I suggest changing the `List.iter` code above to take fpaths instead of strings.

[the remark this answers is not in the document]
lubegasimon: Done!

test/generators/html/gen_html/gen_html.ml
lubegasimon: done!

test/generators/gen_link/gen_link.ml
```diff
@@ -0,0 +1,235 @@
...
+              Atom "%{deps}";
+            ];
+        ];
+    ]
+
+let file_rule path ext =
+  let open Printf in
+  printf "%s\n\n" (to_string_hum (cu_target_rule path ext));
+  printf "%s\n\n" (to_string_hum (odoc_target_rule path ext));
+  printf "%s\n\n" (to_string_hum (odocl_target_rule path))
```
Julow, line 82: Instead of using `printf` and `to_string_hum`, I suggest returning a `sexp list` and propagate it down to the main (`let ()`), where everything can be written at once using:
```ocaml
let () =
  let paths = read_file_from_dir "cases" in
  let stanzas = gen_rules paths in
  List.iter (Sexplib0.Sexp.pp Format.std_formatter) stanzas
```

test/generators/gen_link/gen_link.ml
```diff
@@ -0,0 +1,235 @@
...
+              Atom "html/html.dune.inc.gen";
+            ];
+        ];
+    ]
+
+let gen_html_rule paths =
+  printf "%s\n\n" (to_string_hum (gen_rule_html paths));
+  printf "%s\n\n" (to_string_hum diff_rule_html)
+
+let read_file_from_dir dir =
```
Julow, line 211: This function should be splitted into "read_file_from_dir" (returns the `paths` list) and "gen_rules" (taking the `paths` list).

test/generators/html/gen_html/gen_html.ml
```diff
@@ -0,0 +1,103 @@
...
+let gen_html_rule filenames =
+  filenames
+  |> List.iter (fun odocl ->
+         let targets, status = expected_html_targets odocl in
+         Printf.printf "%s\n\n"
+           (to_string_hum (target_rule odocl targets status));
+         diff_rules (List.map tweak_html_target targets))
+
+let () =
+  let files = List.tl (Array.to_list Sys.argv) in
```
Julow, line 102: I think we should use fpath in this program too. `files` should be converted to fpath as soon as possible. `tweak_html_target` will use `Fpath.segs` instead of `String.split_on_char`.

test/generators/gen_link/gen_link.ml
```diff
@@ -0,0 +1,235 @@
...
+      die
+        (sprintf
+           "Don't know what to do with %a because of unrecognized %a extension."
+           path' path ext' path)
+
+let odocls paths =
+  paths
+  |> List.map (fun p ->
+         let open Fpath in
+         let path = relativize ~root:(v "html") p in
```
Julow, line 161: You can put `v "html"` into a global variable and use it in `gen_rule_html` so the `"html"` constant isn't repeated.

test/generators/html/gen_html/gen_html.ml
```diff
@@ -0,0 +1,103 @@
...
+            [
+              Atom "with-stdout-to";
+              Atom (tweak_html_target t ^ ".gen");
+              List [ Atom "progn"; List [ Atom "system"; Atom ("cat " ^ t) ] ];
+            ])
+        targets
+  | _ ->
+      [
+        die
+          (Filename.quote "odoc html-generate"
```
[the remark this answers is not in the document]
lubegasimon: Done!

test/generators/html/gen_html/gen_html.ml
```diff
@@ -0,0 +1,103 @@
...
+  match String.split_on_char '/' target with
+  | _ :: _ :: rest -> String.concat "." rest
+  | _ -> assert false
+
+let die s =
+  prerr_endline s;
+  exit 1
+
+let html_gen_targets targets status =
+  match status with
```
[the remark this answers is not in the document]
lubegasimon: Done!

test/generators/gen_link/gen_link.ml
```diff
@@ -0,0 +1,235 @@
...
+    ]
+
+let odoc_target_rule path ext =
+  let open Fpath in
+  let basename = basename (rem_ext path) in
+  List
+    [
+      Atom "rule";
+      List [ Atom "target"; Atom (basename ^ ".odoc") ];
+      List [ Atom "deps"; Atom (basename ^ ext) ];
```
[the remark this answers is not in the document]
lubegasimon: done!

test/generators/gen_link/gen_link.ml
lubegasimon: done

test/generators/gen_link/gen_link.ml
lubegasimon: done

test/generators/html/gen_html/gen_html.ml
lubegasimon: done

Julow (commented):
Looks good !

Latex tests are very noisy because of a debug print that should be removed:

```patch
diff --git a/src/latex/generator.ml b/src/latex/generator.ml
index 3abb962e..9626b055 100644
--- a/src/latex/generator.ml
+++ b/src/latex/generator.ml
@@ -62,7 +62,6 @@ module Link = struct
   let filename url =
     let dir = dir url in
     let file = file url in
-    Format.eprintf "dir=%a file=%s\n%!" Fpath.pp dir file;
     if file = "" then Fpath.add_ext "tex" dir
     else Fpath.(add_ext "tex" (dir / file))
 end
```

Julow:
- `test/generators/cases/ocamlary.mli` is a copy of `src/ocamlary/ocamlary.mli`, I suggest making it a symlink instead (`ln -sf ../../../ocamlary/ocamlary.mli test/generators/cases/ocamlary.mli`)

- `Bugs_pre_410` shouldn't run on OCaml `< 4.10`, that's what is causing the CI failures.
  It needs `(enabled_if (>= %{ocaml_version} 4.10.0))` on the corresponding "alias runtest, action diff" rule (the other rules doesn't need to change).
  It's fine if this is hard-coded into `Gen_backend`, we might need more of these in the future:

  ```ocaml
  val case_enabled_if : Fpath.t -> sexp option
  ```

- A rebase is needed but might be annoying to do. I suggest squashing some of the commits together to avoid fixing the same conflict 14 times.

lubegasimon:
Ready for another review

jonludlam:
The v2 PR #697 looks much nicer, so closing this one!

Issue #557, opened 2021-01-06
Command to extract errors from odocl files


lpw25:
Currently problems, such as unresolved references, are reported as warnings when building documentation. Unfortunately, warnings just get lost in people's build logs and never seen. We can convert them to errors, but that prevents you from building the docs to see where the problem is.

I would like a command that would give me a list of issues when given an `.odocl` file. (I can't remember if these files exist yet, but I mean the files that represent a module after linking but before generating html).

At the moment I am using a hacked up python script that read's the html to do this. It is very useful for finding issues, but a proper command would be much better.

lubegasimon:
IIUC, we wouldn’t expect to have unresolved references after linking/generating `.odocl` files.

Octachron:
Errors happen, and I agree with @lpw25  that it would be interesting to get a clear error report while still building the documentation. Typically, building the standard library documentation requires to tolerate a handful of unresolved references, because ocamldoc and odoc does not agree on how to link mld files. I would be interested in a way to monitor that no other errors creep in the documentation.

jonludlam:
We generate warnings when we can't resolve paths, and there's an option to turn those warnings into errors. We don't do the same for references right now as unresolved references don't necessarily end up in the output - the document layer will replace them, which is why we're emitting the warnings when we actually end up rendering unresolved links in the html. I think the places where this happens are a little ad-hoc at the moment, and could do with formalising, at which point it'll be easier to point them out earlier - either at the point we generate the odocl or as a post-processing step on the generated file. We could even then turn on warnings-as-errors and break builds when we notice them. Some projects will be able to do this sooner than others, I imagine :-)

PR #667, opened 2021-04-29
Add a command to extract errors from odoc files

Julow:
Fixes https://github.com/ocaml/odoc/issues/557

Errors are saved into `.odoc` and `.odocl` files and can be printed back with `odoc errors file.odoc`.
The output format is exactly the same as during compiling and linking.

The first few commits are refactoring.


Julow:
I've rebased this PR on top of https://github.com/ocaml/odoc/pull/672 to take advantage of the refactoring.

I've added an option to turn warnings off: (of course, it can be changed)

```
       --print-warnings=VAL (absent=true or ODOC_PRINT_WARNINGS env)
           Whether warnings should be printed to stderr. See the errors
           command.
```

Julow:
#672 is merged, this PR is rebased and ready for review.

Julow:
The CI failure was due to the compile command is now respecting the "warn error" mode, which the CI enables. This found that the `Ocamlary` library was generating `OpaqueModule` warnings.
To fix this, `OpaqueModule` warnings are no longer turned into errors. I think that these warnings are expected and that the documentation is still correct. Perhaps this case can be removed entirely ?

PR #672, opened 2021-05-06
Refactor resolver

Julow:
This PR does some code improvements:
- Simplify the loading of odoc files. There is now a single module taking care of this with a simpler API.
- Stop looking up imports. The result was not used, the only observable side-effect was eventual warnings (which would also occur later).
- Rename `Odoc_odoc.Env` to `Odoc_odoc.Resolver` and simplify its API (remove the `builder` part).
- Load `odoc` files entirely. Remove the intermediate step of loading only the "root" header, which was only taken advantage of when looking up imports. (5% speed up on `dune build @doc` in core_kernel)


src/odoc/compilation_unit.ml
```diff
@@ -14,34 +14,54 @@
...
+let save_unit file t =
   Fs.Directory.mkdir_p (Fs.File.dirname file);
   let oc = open_out_bin (Fs.File.to_string file) in
-  Root.save oc t.Odoc_model.Lang.Compilation_unit.root;
+  output_string oc magic;
   Marshal.to_channel oc t [];
   close_out oc
 
-let units_cache = Hashtbl.create 23 (* because. *)
+let save_page file page =
```
jonludlam, line 35: It looks wrong that `Compilation_unit` is saving/loading pages.

src/odoc/resolver.ml
```diff
@@ -0,0 +1,227 @@
...
+              ambiguous target_name
+          in
+          prerr_endline (Odoc_model.Error.to_string warning));
+      Odoc_xref2.Env.Found m
+  | None -> Not_found
+
+(** Lookup a module. First looks into [imports_map] then searches into the paths. *)
+let lookup_unit ~important_digests ~imports_map ap target_name =
+  match StringMap.find target_name imports_map with
+  | Odoc_model.Lang.Compilation_unit.Import.Unresolved (_, Some digest) ->
```
jonludlam, line 172: Are the imports every anything other than `Unresolved` after this patch? If not, can we get rid of `Compilation_unit.Import.t`?

src/xref2/compile.ml
```diff
@@ -741,36 +740,6 @@ and type_expression : Env.t -> Id.Parent.t -> _ -> _ =
   | Poly (strs, t) -> Poly (strs, type_expression env parent t)
   | Package p -> Package (type_expression_package env parent p)
 
-type msg = [ `Msg of string ]
-
-exception Fetch_failed of msg
-
-let build_resolver :
```
jonludlam, line 748: I'm very glad this thing has gone :-)

src/odoc/compilation_unit.ml
Julow: This module is now saving and loading `.odoc` files. How should we call it ?
I used the name "compilation_unit" for `Lang.Page | Lang.Compilation_unit` in the `Resolver` module, the second case is named `Module`. This will need to be changed too.

src/odoc/resolver.ml
Julow: Looking at the `loader` code, imports are indeed always `Unresolved`. I'll simplify the type.

src/odoc/compilation_unit.ml
Drup: `Documentation_unit` (and `t = Page | Module`) ?

src/odoc/resolver.ml
jonludlam: I'm not convinced `link-deps` can _ever_ produce the correct set of dependencies. In any case, I'm also not convinced the logic you're talking about is sound. It's currently doing this:

```
List.iter odoctree.Odoc_model.Lang.Compilation_unit.imports
        ~f:(fun import ->
          match import with
          | Odoc_model.Lang.Compilation_unit.Import.Unresolved _ -> ()
          | Odoc_model.Lang.Compilation_unit.Import.Resolved (root, _) ->
              Hash_set.add deps root);
```

and iterating this function over all odoc files in a directory. The only thing this gains you over using simply unioning the output of `compile-deps` is that it won't mention modules that couldn't be resolved at `compile` time - ie, modules for which we're missing the `cm{t,ti,i}` file. There's probably one minor difference - if we're making a module that's entirely aliases to other modules, and nothing else within the same directory dereferences those aliases, then we might resolve that at compile time and thus miss it if we don't make a note of resolved imports. This is pretty rare though, and I only know of one instance of this (ocaml-compiler-libs).

The second point is that this entirely ignores all the references in the odoc files, so if there are any references to modules that aren't a compilation dependency then this logic will miss those dependencies. Even if we were to resolve the first part of the reference at compile time (as was suggested in a recent meeting), we'd need to look recursively at the referenced module - e.g. if we had the reference `Stdlib.List.t` and we resolved the `Stdlib` part, we'd need to have a link dependency not just on `stdlib.odoc` but also `stdlib__list.odoc`.

Julow: Indeed. Ideally, the link-dependencies of a unit is everything else. The smallest reasonable set would be the transitive compile-deps, which the `link-deps` command doesn't even compute.
Do you suggest removing this command ?

> it won't mention modules that couldn't be resolved at compile time

With my proposal it will also remove modules that weren't used at compile time, which may not be what we want.

jonludlam: I think that's likely worse, as it's much more likely we'll hit the problem of references above. At least with the 'union of compile-deps' scheme we're not producing _fewer_ link deps than before.

Julow: I've re-implemented the efficient lookup for imports, the performance regression is gone. The `link-deps` command is still here and has a testcase.

src/odoc/compilation_unit.ml
Julow: I've changed to unambiguous names. This module is now `Odoc_file` and this record is `Page_content ... | Unit_content ...`.

Drup: "Unit" alone doesn't really mean much (except maybe the type). Either you put the whole thing ("Compilation unit") or choose another word.

Julow: We commonly use the word "unit" for "compilation unit" already. `Compilation_unit_content of Lang.Compilation_unit.t` sounds unnecessary to me.

PR #673, opened 2021-05-09
Improve and simplify the layout for datatypes

Drup:
This implements #614 : It uses lists instead of tables to layout datatypes, and add some heuristics to simplify the layout when the documentation payload is a single paragraph.

In practice, the layout for a sum datatype with a single-paragraph comment now looks like this:
```html
<ol>
  <li id="type-variant.TagA" class="anchored">
    <a href="#type-variant.TagA" class="anchor"></a>
    <code><span>| </span><span><span class="constructor">TagA</span></span></code>
    <span class="def-doc">
      <span class="comment-delim">(*</span>This comment is for <code>TagA</code>.<span class="comment-delim">*)</span>
    </span>
  </li>
  ......
</ol>
```

The span def-doc is changed for a div for longer documentation comments.
The CSS has not been adapted yet (cue @dbuenzli !).

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -33,11 +33,12 @@ let mk_anchor_link id =
 
 let mk_anchor anchor =
   match anchor with
-  | None -> ([], [])
+  | None -> ([], [], [])
   | Some { Odoc_document.Url.Anchor.anchor; _ } ->
       let link = mk_anchor_link anchor in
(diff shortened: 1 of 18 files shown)


jonludlam:
we're waiting on the CSS for this, right?

leviroth:
With the merge of #830, can this be closed?

panglesd:
Yes!

Issue #659, opened 2021-04-06
Resolution bug for sections


dbuenzli:
There seem to be a resolution bug for sections on master: 

Consider: 

https://b0-system.github.io/odig/doc/brr/Brr/El/index.html#el_lookups

[`.mli` source](https://github.com/dbuenzli/brr/blob/83b2a3e3944378ea29dee62dda88a627239cd15a/src/brr.mli#L1901-L1911)

In this section the first reference gets linked on `Brr/index.html` instead of `Brr/Document/index.html`. Note that the references to the value `Document.root` just below are correctly linked.

jonludlam:
Thanks @dbuenzli - we'll take a look.

PR #676, opened 2021-05-12
Fix reference urls to sections 

Julow:
Fix https://github.com/ocaml/odoc/issues/659

It was pointing to the wrong parent page.


src/document/url.ml
```diff
@@ -232,16 +234,15 @@ module Anchor = struct
...
-     Error (Unexpected_anchor ("label " ^ anchor)) *)
+    | `Label (parent, anchor) -> (
+        let str_name = LabelName.to_string anchor in
+        (* [Identifier.LabelParent.t] contains datatypes. [`CoreType] can't
+           happen, [`Type] may not happen either but just in case, use the
+           grand-parent. *)
+        match parent with
+        | #Path.source as parent -> mk ~kind:"section" parent str_name
+        | `CoreType _ -> Error (Unexpected_anchor "core_type label parent")
+        | `Type (gp, _) -> mk ~kind:"section" gp str_name)
```
Julow, line 245: `Identifier.LabelParent.t` should probably not contain types.

test/xref2/labels.t/run.t
```diff
@@ -0,0 +1,62 @@
...
+
+References to the labels:
+We expect resolved references and the heading text filled in.
+
+  $ odoc_print test.odocl | jq -c '.. | .["`Reference"]? | select(.)'
+  [{"`Resolved":{"`Identifier":{"`Label":[{"`Module":[{"`Root":[{"`RootPage":"test"},"Test"]},"N"]},"B"]}}},[{"`Word":"An"},"`Space",{"`Word":"other"},"`Space",{"`Word":"conflicting"},"`Space",{"`Word":"label"}]]
+  [{"`Resolved":{"`Label":[{"`Identifier":{"`Module":[{"`Root":[{"`RootPage":"test"},"Test"]},"M"]}},"B"]}},[]]
+  [{"`Resolved":{"`Identifier":{"`Label":[{"`Root":[{"`RootPage":"test"},"Test"]},"A"]}}},[{"`Word":"First"},"`Space",{"`Word":"label"}]]
+  [{"`Resolved":{"`Identifier":{"`Label":[{"`Root":[{"`RootPage":"test"},"Test"]},"B"]}}},[{"`Word":"Floating"},"`Space",{"`Word":"label"}]]
+  [{"`Dot":[{"`Root":["M","`TUnknown"]},"C"]},[]]
```
Julow, line 25: That's an other bug, references to labels in top-comments don't resolve.

src/document/url.ml
Drup: What about inlined records ?

Julow: Label are section names, not fields. I don't expect a heading (sections are only defined by headings) attached to anything inside a type (classes are covered in the `Path.source` case).

test/xref2/dune
```diff
@@ -14,8 +14,16 @@
 (cram
  (deps %{bin:odoc} %{bin:odoc_print} %{bin:compile}))
 
+(subdir
+ v407_and_above
+ (cram
+  (deps %{bin:odoc} %{bin:odoc_print} %{bin:compile})
+  (enabled_if
+   (>= %{ocaml_version} 4.07.0))))
```
Julow, line 22: Before 4.07, labels in top-comments were ignored by the loader. This is probably a bug.

PR #677, opened 2021-05-19
Open stdlib by default

jonludlam:
This PR implements two ideas.

1. If there is a module, module type or type path like `X.Y.t`, it checks to see if `Y` is itself in scope, or indeed `t`, and replaces the path with the shortest valid one.
2. Adds the contents of `Stdlib` (or any other module specified on the command line, but by default just Stdlib) to the environment, activing the above code.

The upshot is that were we would previously see, e.g. `Stdlib.Format.formatter` we now see `Format.formatter`

Fixes: #562


src/xref2/tools.ml
```diff
@@ -398,7 +454,7 @@ and process_module_path env ~add_canonical m p =
 and handle_module_lookup env ~add_canonical id parent sg sub =
   match Find.careful_module_in_sig sg id with
   | Some (`FModule (name, m)) ->
-      let p' = `Module (parent, name) in
+      let p' = simplify_module env (`Module (parent, name)) in
```
Julow, line 457: I think this should be applied to the whole output of every `handle_` function so that adding a new case into the simplify function will immediately work and we won't miss any path.

Julow (approved):
Looks good to me!

Octachron:
Question: is it possible to have no open modules at all with the current cli? (which might be needed for building the stdlib itself without too much acrobatic)

Julow:
It is possible by passing `--open` with no value: `odoc compile --open= foo.cmti`.
To pass something else than the stdlib it's necessary to do `odoc compile --open=Stdlib --open=Foo bar.cmti`

PR #678, opened 2021-05-19
Octavius 2

jonludlam:
Rebrand the odoc parser as 'octavius'.

Also tidy the interface quite a lot.


src/parser/ast.ml
```diff
@@ -13,21 +23,33 @@ type inline_element =
...
+    {{:https://ocaml.org/releases/4.12/htmlman/ocamldoc.html#sss:ocamldoc-list}manual}).
+    *)
 
 type internal_tag =
   [ `Canonical of string with_location | `Inline | `Open | `Closed ]
+(** Internal tags are used to exercise fine control over the output of odoc. They
+    are never rendered in the output *)
 
-type external_tag =
+type ocamldoc_tag =
```
Julow, line 52: I think this reference to `ocamldoc` isn't necessary, we can call this `tag` instead.

Julow (commented):
If the parser stays in this repository, I think it should be called `odoc-parser` so it's easier to relate packages in Opam and to guess the library name.

src/parser/warning.mli
```diff
@@ -0,0 +1,8 @@
+type t = { location : Location.span; message : string }
```
aantron: The way this file has been moved with changes loses its git history and easy access to its authorship information.

octavius.opam
```diff
@@ -0,0 +1,32 @@
+opam-version: "2.0"
+maintainer: ["Jon Ludlam <jon@recoil.org>"]
+authors: [
+  "Anton Bachin <antonbachin@yahoo.com>"
+  "Leo White <leo@lpw25.net>"
+  "Jules Aguillon <juloo.dsi@gmail.com>"
+  "Jon Ludlam <jon@recoil.org>"]
```
aantron, line 7: @jonludlam Per our Discord discussion, can you justify this authorship metadata? I could be wrong about this, but I will detail what I see:

- While Leo has given important feedback that affected and advanced the development of the parser, looking at the [commit history](https://github.com/jonludlam/odoc/commits/odoc-parser-octavius/src/parser) of the entire `parser` directory in the branch this PR is from, I don't see any commits from Leo, nor is the parser in any other way based on any code authored by Leo. Both the code and design are my original work (I don't claim that the overall approach is in any way original with respect to the "state of the art" of parsers in general, just this particular parser is not originally based on the work of anyone else listed here except me).

- The next part is more debatable. Looking at the same commit history, I see a large number of important contributions from you and Jules. However, though I did not check absolutely every commit, and skipped many commits that, from their message, seemed to touch the parser only incidentally, I doubt that these contributions amount to authorship. They appear to be maintenance, which is what the `maintainer` field is for. Of course, this is a judgment call, but I saw many commits that added features locally, or were refactorings. While contributing important pieces of original code, these don't amount to authorship at large, in my opinion. For comparison, to illustrate (still just) my opinion in operation, even though I maintained Lwt for four years, and added many features, did a lot of clean-up, rewrote `lwt.ml`, `lwt.mli`, and several other files completely, and wrote almost all the current tests, I am still [not an author of Lwt](https://github.com/ocsigen/lwt/blob/da8704414d416b06c50026b28d64060d280e6e02/lwt.opam#L11-L18):

    ```
    authors: [
      "Jérôme Vouillon"
      "Jérémie Dimino"
    ]
    maintainer: [
      "Raphaël Proust <code@bnwr.net>"
      "Anton Bachin <antonbachin@yahoo.com>"
    ]
    ```

    I am a contributor and a former maintainer. I did not set the original structure of Lwt, nor write any major new modules (though I assigned authorship credit for sublibraries to some people that did). I didn't perform any fundamental rework that could have consituted authorship on the scale of the library (as you have done with odoc's new model, in my understanding).

    I don't see that any such thing has happened (yet?) with odoc's parser.

I'm less concerned about the second point than the first, however, which I truly don't understand. My first worry is that it is simply a holdover field from Octavius v1. If that's so, I think it's clear that we can't assign authorship on the basis of renaming a library so that it shares a name with another one :)

avsm: From an opam-repository perspective, we don't take a strong position on how projects ascribe metadata such as authorship _vs_ maintainers.  Thinking about it more now, we could take advantage of the multi-versioning packaging to achieve two goals:

- precise author information for each published version of the package (as suggested by @aantron above), so that the version-specific page on the docs site reflects who the primary author(s) of the package in question are.
- aggregated author information from _all_ versions of the opam package on the main package homepage, which shows the full historical set of authors associated with that package name. This would be attribute those who were associated with the project in the past, but do not have current code.

aantron: > This would be attribute those who were associated with the project in the past, but do not have current code.

In this particular case, however, the previous authors are not associated with v2 at all (as authors). It's just a completely separate project. It's not even a question of current code &mdash; it's not that their code was deleted inside the project, and then replaced by new code. The current parser, as a parser, is a completely separate, fresh project, that is now being renamed to match the name of another project. How are the previous authors associated (1) with the present project (2) in the past? They are associated with a separate, past project, which, confusingly, has the same name.

aantron: > * aggregated author information from _all_ versions of the opam package on the main package homepage, which shows the full historical set of authors associated with that package name.

This might make sense for almost all packages, but, for the reasons above, it does not make sense if we will practice dropping in one project into the name slot of another, as is happening in this PR. The package page, especially if it tends to display the latest version's other metadata, will be misleading if it shows authorship metadata of earlier, completely unrelated versions of the supposedly "same" package, mixed in with current authorship metadata.

The per-version metadata should still display the authors of that version, however, I do agree with that.

Julow: I have no problem to being described as a maintainer.
I have no strong opinion about the rest but I feel that both project descriptions should contain the same metadata. I do not see `octavius` as a different project but just as a way to fit into Opam's rules (that's why I suggest to name it `odoc-parser`).

avsm: > The current parser, compared to the original Octavius, is a fresh design, fresh code, fresh metadata, fresh tests, etc., strarted completely from scratch, with no inspiration from Octavius, and, indeed, in a sort of "opposition" (in design terms) to it.

It has the same name, and so we have no way of distinguishing the subtlety of a rewrite in opam. Another example of this is the `dns` package which was completely rewritten from scratch and replaced, but we can still recognise that the previous dns package had a set of contributors who did something called `dns`.

If you want to distinguish your parser from the earlier thing called `octavius` and be completely disassociated from it, then don't call this new one `octavius`.

aantron: > If you want to distinguish your parser from the earlier thing called `octavius` and be completely disassociated from it, then don't call this new one `octavius`.

I'm completely in favor of not calling it `octavius` &mdash; this is not my idea, and I see no point in shadowing Leo's or anyone else's work, or else muddling authorship metadata because of it. `odoc-parser` sounds good to me @Julow.

> we can still recognise that the previous dns package had a set of contributors who did something called `dns`.

I am talking about authors, not contributors. We don't list all contributors in any package's metadata, at all. If you mean authors, then I still don't see how it's necessary to add authors of an earlier package someone decided to shadow to the later new project,. The authorship metadata of the old version is still present on the older versions.

Authorship is not transmitted by name collisions, but by actual work done, and it doesn't seem reasonable that if, say, Async was renamed to Lwt, I would suddenly, not having worked on either for years, and never on Async, become a "former author or maintainer of Async a.k.a. the new Lwt" in its current metadata, when I had nothing to do with the codebase. This is just an artifact of some administrative decisions.

aantron: > some administrative decisions.

and by third parties to all the actual authors of all the projects, I should add!

aantron: There is yet another way in which this sequence of authors and versions misrepresents the history of the project, which is that it appears as if Leo was the original author of some kind of continuous Octavius v1-v2, and I became a co-author later, maybe by adding just only a big chunk of new code that "turned" v1 into v2. That's completely inaccurate, because v2 is just a separate project and the authors of v1 and the authors of v2 are not each others' co-authors, and merely a big chunk was not added &mdash; it's just a whole new project.

aantron: > we have no way of distinguishing the subtlety of a rewrite in opam

Sorry to spam, but re-reading this... Octavius v2 is not a "rewrite" of Octavius v1. It's just a different project. It's not API-compatible, similar in design, etc. There is no subtlety about it. I understand that opam can't express that it's a new project, but, as already mentioned by me and several people, that seems like a good reason not to shadow Octavius v1 if you're not willing to also be accurate in the project's metadata about the actual origin of the so-called "Octavius v2" (a name I never had in mind).

aantron: To give some detail on the API incompatibility, it's not just some minor breaking change, but a completely new API with new data structures (though there is inevitably some similarity due to them representing the same thing &mdash; on the other hand, there are fundamental differences like the representation of spacing). When I integrated the new parser into odoc, I replaced the comment markup generator, and made invasive changes throughout odoc. So it's not a matter of some "unfortunate minor breakage" and maybe a few internal cleanups causing v1 to become v2. It's literally a different unrelated project with different concepts deliberately chosen even at the type level.

avsm: Your objection is noted, and your work won't be renamed to octavius v2. We'll have to figure out then how to ensure that octavius v1 ceases to be used by anyone in a future odoc developer meeting (which you are, of course, most welcome to join if you want).

aantron: Thank you. I don't have a specific goal of this not being called `octavius`, by the way. Just not if it's in combination with the above.

As another thought, it just occurred to me that perhaps nobody working here on this issue has looked in detail at the code of Octavius v1 (could be wrong), and so isn't immediately aware of how substantially different the two parsers are.

avsm: I'm entirely aware of the history of both octavius v1 and this parser, having been around since day 0 of this project :-) My example of `dns` was deliberate: it is a total rewrite in exactly the same style as the odoc parsers being discussed here; see https://github.com/mirage/ocaml-dns/pull/159 for more (no common provenances between the codebases).

The point of keeping the same name from an opam perspective is that both the old and new libraries are fulfilling the same _purpose_. A user who comes along and was using octavius v1 would note there is a new version with a major bump, and follow a migration guide, and eventually be on v2 (with its new interface).  The goal is to hide the internal details of how it came to be from users and make a statement that "the successor to octavius v1 is v2". opam has upper bounds available for older consumers of the library.

However, this only really works if there's consensus among the developers of the old and the new. In this case, there isn't (which is just fine), and the fastest way to move on is just to provide clarity to our users about which ocamldoc parser they are supposed to be using as part of the release of the new one.  For instance, a quick search reveals [Imandra](https://docs.imandra.ai/imandra-docs/odoc/imandra-base/Imandra_surface/Document/index.html) uses Octavius currently.

PR #679, opened 2021-05-20
Fix js_of_ocaml compilation

jonludlam:
Without this I get a warning and runtime error from js_of_ocaml
complaining about the missing primitive `caml_alloc_dummy_infix`. Not
quite sure what this nor why this patch fixes it!


Julow (approved):
Looks good.

Issue #648, opened 2021-03-23
Warnings about ambigous references are bogus


jonludlam:
example:

```
Reference to 'enqueue' is ambiguous. Please specify its kind: val-enqueue, val-enqueue.
```

These are seen in many packages so should be easy to repro. I wonder if it's because we have the value `enqueue` defined in different modules, both in scope?

dbuenzli:
One occurence where that happens is in the following case: 

```ocaml
val the_fun : …

module Special : sig

   val the_fun : …
   (** [the_fun] is like {!the_fun} but … *)

end
```
In this case you get:

```
Reference to 'the_fun' is ambiguous. Please specify its kind: val-the_fun, val-the_fun.
```

In this case the warning is *not* bogus, but the suggestion does not fix anything.

dbuenzli:
Note also that nowadays it also warns if in the above the `the_fun` comes after the `Special` module. I think it did not before so that may be one of the reasons why you see a lot of these ambiguous references.

In general I think it would be time for 2.0.0 to properly define the notion of scope for `odoc`. See #594, #332, #42.

It should also be noted that in `ocamldoc` if you had ambiguities with say a section and a value it would simply default to link to the value. That difference should be documented. I don't think `ocamldoc`'s behaviour was good because it meant that links could change without you noticing (e.g. when you had an `id` section to which you referred and then later you add an `id` value to your module).

PR #682, opened 2021-05-27
Fix references to shadowed items

Julow:
Fixes https://github.com/ocaml/odoc/issues/648

This PR implements shadowing when resolving submodules. Before, several items with the same name were added to the environment, which caused the warnings.

The notion of scope didn't change: the whole signature is added to the environment before any reference is resolved (including opens and includes).

### DIFF
--- a/test/xref2/v408_and_above/github_issue_587.t/run.t
+++ b/test/xref2/v408_and_above/github_issue_587.t/run.t
@@ -1,5 +1,25 @@
 A quick test to repro the issue found in #587
 
   $ ./build.sh
+  File "odoc_bug__a_intf.cmt":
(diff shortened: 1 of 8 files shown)


test/xref2/v408_and_above/github_issue_587.t/run.t
```diff
@@ -1,5 +1,25 @@
 A quick test to repro the issue found in #587
 
   $ ./build.sh
+  File "odoc_bug__a_intf.cmt":
```
jonludlam, line 4: What's going on here? I don't understand why this would start failing now?

jonludlam:
The fastly outage is causing the test failures IFAICT

test/xref2/v408_and_above/github_issue_587.t/run.t
Julow: Fixed. I misunderstood lookup by identifier in the environment. The previous code was removing shadowed items from the environment but we need to keep them in the case of aliases to the same name. We only need to remove them from "lookup by name" results.

Julow:
Most CI jobs passed :)

PR #684, opened 2021-06-08
Add anchors to type extensions

Julow:
Fixes https://github.com/ocaml/odoc/issues/601

This PR adds anchors to type extensions but doesn't allow them to be referenced.
The anchor looks like `extension-decl-FirstConstructor`. This is enough because it's not possible to have two constructors with the same name in a signature.

The other part of the issue is fixed since https://github.com/ocaml/odoc/pull/615

This is the simplest fix for the issue. An alternative is to add an internal name and a new syntax for references to extensions.


Drup:
> The anchor looks like `extension-decl-FirstConstructor`. This is enough because it's not possible to have two constructors with the same name in a signature.

Huh, that's actually probably right, I somehow didn't realized that.

I think it's the last one! Can you try to get rid of the option type for anchors ? :)

Julow:
There are some other `None` in class fields (`inherit`, `constraint`).

Drup:
Right, we should anchor those eventually too.

jonludlam:
I don't think this completely fixes #601 as there's no way to refer to these things from within the doc itself (ie, there's no `{!type-extension-Constructor}` syntax yet.

PR #685, opened 2021-06-09
Odoc parser opam package

jonludlam:


jonludlam:
Need to figure out how to fix `opam-dune-lint`

src/parser/ast.ml
```diff
@@ -13,21 +23,33 @@ type inline_element =
...
+(** Inline elements are equivalent to what would be found in a [span] in HTML.
+    Mostly these are straightforward. The [`Reference] constructor takes a triple
+    whose second element is the reference itself, and the third the replacement
+    text. Similarly the [`Link] constructor has the link itself as first parameter
+    and the second is the replacement text. *)
 
 type nestable_block_element =
   [ `Paragraph of inline_element with_location list
-  | `Code_block of string
+  | `Code_block of string option * string
```
Julow, line 34: Some comments from https://github.com/ocaml/odoc/pull/678#discussion_r636941740

- Should we parse the "language" field ? (the first word)
  We might want to use it for syntax highlighting. This feature is targetted at Mdx, which uses this field.
- Both arguments should have a location attached.

src/parser/lexer.mll
```diff
@@ -262,6 +262,8 @@ let raw_markup =
   ([^ '%'] | '%'+ [^ '%' '}'])* '%'*
 let raw_markup_target =
   ([^ ':' '%'] | '%'+ [^ ':' '%' '}'])* '%'*
+let code_block_meta =
+  ([^ '['])*
```
Julow, line 266: This parser doesn't allow arbitrary strings. I propose this syntax:

Optionally wrap with curly braces: `{{@ocaml foo=[]}[ ... ]}`.
Closing brace can be escaped `\}`, no balancing because it would clash with the other escape.
Not using quotes because it wouldn't be clear where they are allowed, this is consistent with other syntaxes (eg. links), I also thing quotes would be more common in the payload.

src/parser/test/test.ml
```diff
@@ -2495,6 +2496,46 @@ let%expect_test _ =
...
            \nUnpaired '}' (end of markup).\
            \nSuggestion: try '\\}'."))) |}]
+
+    let code_block_with_meta =
+      test "{@ocaml env=f1 version>=4.06 [code goes here]}";
+      [%expect
+        {|
+        ((output
+          (((f.ml (1 0) (1 46))
+            (code_block "ocaml env=f1 version>=4.06 " "code goes here"))))
```
Julow, line 2506: Trailing whitespaces should be removed.

Julow (commented):
I've been able to fix `opam-dune-lint` at the cost of increase Dune's version to 2.8: https://github.com/jonludlam/odoc/pull/36

odoc-parser.opam
```diff
@@ -0,0 +1,39 @@
...
+bug-reports: "https://github.com/ocaml/odoc/issues"
+license: "ISC"
+
+authors: [
+  "Anton Bachin <antonbachin@yahoo.com>"
+]
+maintainer: "Jon Ludlam <jon@recoil.org>"
+dev-repo: "git+https://github.com/ocaml/odoc.git"
+
+synopsis: "Parser for OCamldoc"
```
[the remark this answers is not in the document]
jonludlam: I think of `ocamldoc` as referring to the language and the tool. The sense I'm trying to convey here is that it's the parser for the language - though you're right, there are extensions beyond what the original ocamldoc language specified.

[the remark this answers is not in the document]
Drup: I would argue we should transition to "the odoc markup language" (or a more catchy name). We already broke compat in various ways, and the distance is only going to grow with time.

src/parser/loc.ml
```diff
@@ -28,3 +28,24 @@ let span spans =
...
+  let loc_end =
+    Lexing.
+      {
+        pos_fname = span.file;
+        pos_lnum = span.end_.line;
+        pos_bol = 0;
+        pos_cnum = span.end_.column;
+      }
+  in
+  Location.{ loc_start; loc_end; loc_ghost = false }
```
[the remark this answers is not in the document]
jonludlam: The intention here is a convenience for users of the library, though it needs a bit more work. When we first call `Odoc_parser.parse_comment` we pass in a `Lexing.position` to indicate the start position of the comment. The intent here is to be able to turn our internal `Loc.t` types back into the original type, more-or-less. The use of `Location.t` as a return type is because it conveniently represents a span between two `Lexing.position` values. While it does bring in a dependency on `compiler-libs`, I think it's pretty unlikely anyone using this library isn't _already_ using `compiler-libs`, though we could always return a pair of `Lexing.positions` instead. However, in order to do this correctly we need to stash the original `Lexing.position` away and keep it for this operation. I'll give this a try and see how it looks. 

As an example of where this function would be useful, see here: https://github.com/janestreet/ppx_js_style/blob/master/src/ppx_js_style.ml#L393-L432

src/parser/ast.ml
jonludlam: Good question. I've currently done the minimum possible here in the parser, in the spirit of how references are parsed, though this and the stripping of whitespace both seem reasonable to me.

jonludlam: and I agree with the location comment, fill fix, thanks!

src/parser/lexer.mll
jonludlam: This PR just does what we [agreed in the dev meeting](https://github.com/ocaml/odoc/wiki/Dev-meeting-2021-02-11). Since the first user of this field will be mdx, do you happen to know whether this is going to be an immediate problem? That is, whether there are any square brackets in any of the current uses of this field in md files?

Julow: I think mdx only need to be able to read file names, ["relation"](https://github.com/realworldocaml/mdx/blob/master/lib/label.ml#L39), package names, commas, version numbers and pre-defined strings. Escaping isn't needed by Mdx for now.

IMO, we should always think about escaping when writing a language that accepts arbitrary strings unless we are fine with writing "accept any string that doesn't contain '['" in the doc.

jonludlam:
I think we've spent enough time struggling with the "parser library and odoc in the same repo" problem. I propose to put the parser into `github.com/ocaml-doc/odoc-parser` to sidestep the issue. We've discussed this possibility before on its own merits and it's never been obvious whether it's the right thing to do or not, but we didn't really take into account the problems we're encountering. While the problems are temporary and I'm sure they can be worked around in due course, they either stops us merging useful changes while we figure out some new fix or we end up turning off testing. By separating the repositories we fix these issues in the 'common' way and thus are much more likely to stay on the beaten path and thus spend less time searching for fixes.

We can always remerge them later if we like!

PR #687, opened 2021-06-17
Minimal 4.13 support

Octachron:
This PR makes odoc build on OCaml 4.13 but does not handle the new `with module type` feature.
This PR mostly adapts the loaders code to the few ASTs change in 4.13 .
There are also few fixed related to the change of typing for `|>`.

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -292,9 +292,9 @@ and items ~resolve l : item Html.elt list =
             | _ -> Stop_and_keep)
         in
         let content = flow_to_item @@ block ~resolve text in
-        content |> (continue_with [@tailcall]) rest
+        (continue_with [@tailcall]) rest content
     | Heading h :: rest ->
-        [ heading ~resolve h ] |> (continue_with [@tailcall]) rest
+        (continue_with [@tailcall]) rest [ heading ~resolve h ]
     | Include { attr; anchor; doc; content = { summary; status; content } }
       :: rest ->
         let doc = spec_doc_div ~resolve doc in
--- a/src/loader/cmt.ml
+++ b/src/loader/cmt.ml
@@ -48,7 +48,11 @@ let rec read_pattern env parent doc pat =
     | Tpat_constant _ -> []
     | Tpat_tuple pats ->
         List.concat (List.map (read_pattern env parent doc) pats)
+#if OCAML_VERSION < (4, 13, 0)
     | Tpat_construct(_, _, pats) ->
+#else
+    | Tpat_construct(_,_,pats,_) ->
+#endif
         List.concat (List.map (read_pattern env parent doc) pats)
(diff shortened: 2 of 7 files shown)


src/xref2/env.ml
```diff
@@ -272,35 +272,35 @@ let add_type identifier t env =
   |> add_cdocs identifier t.doc
   |> List.fold_right (add_cdocs identifier) docs
 
-let add_module_type identifier t env =
+let add_module_type identifier (t:Component.ModuleType.t) env =
```
Julow, line 275: Why are these needed ? There's an mli

src/html/generator.ml
```diff
@@ -292,9 +292,9 @@ and items ~resolve l : item Html.elt list =
             | _ -> Stop_and_keep)
         in
         let content = flow_to_item @@ block ~resolve text in
-        content |> (continue_with [@tailcall]) rest
+        (continue_with [@tailcall]) rest content
     | Heading h :: rest ->
-        [ heading ~resolve h ] |> (continue_with [@tailcall]) rest
+         (continue_with [@tailcall]) rest [ heading ~resolve h ]
```
Julow, line 297: Why is this needed ?

src/loader/cmti.ml
```diff
@@ -470,6 +470,10 @@ let rec read_with_constraint env parent (_, frag, constr) =
         let frag = Env.Fragment.read_module frag.Location.txt in
         let p = Env.Path.read_module env p in
           ModuleSubst(frag, p)
+#if OCAML_VERSION >= (4,13,0)
+   | Twith_modtype _ -> failwith "with module type not yet implemented"
```
Julow, line 474: We need to do this before 2.0: https://github.com/ocaml/odoc/issues/688

src/loader/cmt.ml
```diff
@@ -48,7 +48,11 @@ let rec read_pattern env parent doc pat =
     | Tpat_constant _ -> []
     | Tpat_tuple pats ->
         List.concat (List.map (read_pattern env parent doc) pats)
+#if OCAML_VERSION < (4, 13, 0)
```
Julow, line 51: :+1:  for the tuple

jonludlam:
Looks like just the format test failed.

Julow (approved):

src/html/generator.ml
Julow: You answered that already.

src/xref2/env.ml
Octachron: The existence of a mli is irrelevant? The `doc` field is not in scope, the previous code was relying on the specific order of type propagation to resolve the field, this was brittle. Adding a type annotation on `t` fixes the issue.

Julow: Indeed.

PR #689, opened 2021-06-18
Support for "with module type" constraints

Octachron:
This PR adds support for `with module type` constraint:
```ocaml
module type s = sig module type T module M:T end
module type u = s with module type T := s
```
and local module type substitution:
```ocaml
module type s:= sig
  type t
  type r
end
module M: s
```
This change adds a new class of fragments for module types which explains the size of the new code.

There is some cases in `xref2/subst.ml` that would require more attention during review: there are few assertions about which module type paths can be substituted or not. 

One limitation of this PR is that deep destructive substitution is not supported. For instance, 
```ocaml
module type s = sig
  module M: sig
     module type r
  end
  module N : M.r
end
with module type M.r := sig end
```
creates a dead link `M.r` in the constrained signature. However, this is a global issue with all deep destructive substitutions. It seems better to fix this issue independently.


src/xref2/subst.ml
```diff
@@ -377,12 +435,26 @@ and resolved_module_fragment :
  fun t r ->
   match r with
   | `Subst (mty, f) ->
-      `Subst (resolved_module_type_path t mty, resolved_module_fragment t f)
+      let p =
+        match resolved_module_type_path t mty with
+        | Not_replaced p -> p
+        | Replaced _ ->
+            (* the left hand side of subst is a named module type inside a module,
+               it cannot be substituted *)
```
jonludlam, line 443: So we're saying here that at the point we're substituting `A := B` then `A` has to exist? And this is a constraint that comes from the compiler, so while the types in odoc allow it we should never be in this situation?

jonludlam:
This all looks very sensible indeed. Many thanks!

Issue #635, opened 2021-03-12
odoc too noisy on partial docsets


dbuenzli:
Odoc became *again* incredibly noisy on docsets where you do not `odoc` all dependencies, an example is attached at the end of file, and that's for a *single* `.odoc` file.

A few things:

1. Do not report the same error twice. 
2. Once you have determined that a root is missing it's not necessary to report references that are rooted in it.
2. (Possibly) provide a mode that does not complain on external file resolutions. 

```
> brzo log --id 152
[152:spawn 13.8ms doc-build r:ae5601f96f24dfd8] ['/Users/dbuenzli/.opam/4.09.1/bin/odoc' 'html' '--theme-uri' '_odoc-theme' '-o' '/Users/dbuenzli/sync/repos/webs/_b0/brzo/ocaml-doc/html' '/Users/dbuenzli/sync/repos/webs/_b0/brzo/ocaml-doc/odoc/webs/webs.odoc' '-I' '/Users/dbuenzli/sync/repos/webs/_b0/brzo/ocaml-doc/odoc/webs/'][0]
File "/Users/dbuenzli/sync/repos/webs/_b0/brzo/ocaml-doc/odoc/webs/webs.odoc":
Failed to lookup type unresolvedroot(Stdlib).result Parent_module: Lookup failure (root module): Stdlib
File "/Users/dbuenzli/sync/repos/webs/_b0/brzo/ocaml-doc/odoc/webs/webs.odoc":
Failed to lookup type unresolvedroot(Stdlib).result Parent_module: Lookup failure (root module): Stdlib
File "/Users/dbuenzli/sync/repos/webs/_b0/brzo/ocaml-doc/odoc/webs/webs.odoc":
Failed to lookup type unresolvedroot(Stdlib).Format.formatter Parent_module: Parent_module: Lookup failure (root module): Stdlib
File "/Users/dbuenzli/sync/repos/webs/_b0/brzo/ocaml-doc/odoc/webs/webs.odoc":
Failed to lookup type unresolvedroot(Stdlib).result Parent_module: Lookup failure (root module): Stdlib
File "/Users/dbuenzli/sync/repos/webs/_b0/brzo/ocaml-doc/odoc/webs/webs.odoc":
Failed to lookup type unresolvedroot(Stdlib).Format.formatter Parent_module: Parent_module: Lookup failure (root module): Stdlib
[...]

PR #692, opened 2021-06-23
Aggregate "root" lookup failures

Julow:
Fixes https://github.com/ocaml/odoc/issues/635

These errors are common and are usually repeated several time. With this commit, they are merged into one message with only the necessary informations:

```
File "<test>":
Couldn't find the following modules:
  Stdlib
```

I initially tried to handle these errors while resolving imports but it was showing mangled names (eg. `Stdlib__bytes Stdlib__string ...` instead of `Stdlib`) and unused modules (eg. `CamlinternalFormatBasics`) and decreased performances a bit.

### DIFF
--- a/src/xref2/errors.ml
+++ b/src/xref2/errors.ml
@@ -169,9 +169,8 @@ let is_unexpanded_module_type_of =
   in
   inner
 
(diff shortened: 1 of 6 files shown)


jonludlam:
Looking good!

PR #693, opened 2021-06-25
Fix shadowing of values and externals

Julow:
Fix shadowing of values when resolving references, which resulted in warnings like this:
```
Reference to 'x' is ambiguous. Please specify its kind: val-x, val-x.
```

I noticed that the distinction between values and externals is not necessary (there is no identifier for externals) so I merged them.

The third commit fixes shadowing of values in signatures, which was implemented for externals but not for values.


Julow:
Rebased.

jonludlam:
Thanks! looks good.

PR #695, opened 2021-06-28
Flat files

lubegasimon:
This PR is not from a tracked issue rather is built from @jonludlam’s idea of outputting the output files in one dir, which is a plus for the ‘improving generator tests v2’ PR that is a work in progress. (https://github.com/ocaml/odoc/pull/665 being v1).

**TODO:** 
- [ ] add tests

### DIFF
--- a/commits-to-take
+++ b/commits-to-take
@@ -1,0 +1,2 @@
+bea122919d799 d6ef4ca0f0a2cf6d74
+59d171ed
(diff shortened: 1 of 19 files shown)


jonludlam:
The idea here is to do a small change primarily to the HTML backend to ensure all output files end up in the same directory, and the underlying reason is for testing with dune, which has trouble when you have targets in multiple directories.

I'd like to think of an alternative though that doesn't require changing the man/latex generators.

jonludlam:
Superseded by #698

PR #697, opened 2021-06-30
Improve generators tests v2

lubegasimon:
This PR is an enhancement of https://github.com/ocaml/odoc/pull/665 due to the complexity in the latter.


jonludlam:
This is looking great now!

Thanks very much for all your hard work on this, @lubegasimon - in it goes!

PR #711, opened 2021-08-03
Update the representation of code blocks

Julow:
The representation of code blocks will change slightly: https://github.com/ocaml-doc/odoc-parser/pull/2
This doesn't make use of the language tag yet.

### DIFF
--- a/src/model/comment.ml
+++ b/src/model/comment.ml
@@ -38,9 +38,14 @@ type module_reference = {
 (** The [{!modules: ...}] markup. [module_synopsis] is initially [None], it is
     resolved during linking. *)
 
+type code_block = {
+  code_block_lang : string with_location option;
+  code_block_content : string with_location;
+}
+
 type nestable_block_element =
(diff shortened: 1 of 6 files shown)


src/model/comment.ml
```diff
@@ -40,7 +40,9 @@ type module_reference = {
 
 type nestable_block_element =
   [ `Paragraph of paragraph
-  | `Code_block of string with_location option * string with_location
+  | `Code_block of
+    string with_location option * string with_location
```
Julow, line 44: The meaning of the first argument changed (from the whole metadata to just the language tag) but it wasn't used anywhere.

Drup: I suggest turning it into a (new) record type.

Julow: Good idea :)

Julow:
We don't need to interpret that field yet. We don't need this for compatibility either: https://github.com/ocaml/odoc/pull/742

PR #716, opened 2021-08-13
correct usage example

lubegasimon:
correct usage example in `test/xref2/lib/common.cppo.ml`

### DIFF
--- a/test/xref2/lib/common.cppo.ml
+++ b/test/xref2/lib/common.cppo.ml
@@ -10,7 +10,6 @@ utop # let id, sg, docs = Common.model_of_string test_data;;
 utop # let env = Env.open_signature sg Env.empty;;
 utop # let unit = Common.my_compilation_unit id sg;
 utop # Common.resolve unit
-utop # Resolve.signature Env.empty sg
 
 *)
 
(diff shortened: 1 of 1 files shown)


test/xref2/lib/common.cppo.ml
```diff
@@ -2,15 +2,14 @@ open Result
 
 (* Example usage of these:
 
-$ dune utop src/xref2/test/lib
+$ dune utop test/xref2/lib
 utop # open Odoc_xref2;;
 utop # open Odoc_xref_test;;
 utop # let test_data = "module type M = sig type t end module N : M type u = N.t";;
-utop # let id, docs, sg = Common.model_of_string test_data;;
+utop # let id, sg, docs = Common.model_of_string test_data;;
```
jonludlam, line 9: I think the last value in this triple is now the optional 'canonical' value - `model_of_string` ends up calling `read_interface`:

```ocaml
let read_interface root name intf =
  let id = `Root (root, Odoc_model.Names.ModuleName.make_std name) in
  let sg, canonical =
    read_signature Odoc_model.Semantics.Expect_canonical Env.empty id intf
  in
  (id, sg, (canonical :> Odoc_model.Paths.Path.Module.t option))
```

lubegasimon: Indeed, so we probably need to rename the variable from `docs` to `canonical`. (Sorry for the delay to respond)

Issue #664, opened 2021-04-19
Failures when resolving references are not reported


Julow:
When a reference is failing to resolve, no warning is generated.

PR #717, opened 2021-08-23
Report failures when resolving references

Julow:
Closes https://github.com/ocaml/odoc/issues/664

Add a warning when a reference failed to resolve. Before, this was not reported at all.

### DIFF
--- a/src/xref2/ref_tools.ml
+++ b/src/xref2/ref_tools.ml
@@ -1,7 +1,7 @@
 open Odoc_model.Paths
 open Odoc_model.Names
 open Reference
-open Utils.OptionMonad
+open Utils.ResultMonad
 
 type module_lookup_result =
   Resolved.Module.t * Cpath.Resolved.module_ * Component.Module.t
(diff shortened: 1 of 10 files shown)


src/xref2/ref_tools.ml
```diff
@@ -23,30 +23,40 @@ type type_lookup_result =
   | `C of class_lookup_result
   | `CT of class_type_lookup_result ]
 
-type value_lookup_result = Resolved.Value.t
```
jonludlam, line 26: If we don't need these let's get rid of them rather than just commenting them out.

src/xref2/ref_tools.ml
```diff
@@ -232,44 +246,42 @@ module CL = struct
...
+    env_lookup_by_name Env.s_class name env >>= fun e -> Ok (of_element env e)
 
-  let of_component _env c ~parent_ref name : t option =
-    Some (`Class (parent_ref, name), c)
+  let of_component _env c ~parent_ref name = Ok (`Class (parent_ref, name), c)
 end
 
 module CT = struct
-  type t = class_type_lookup_result
+  (* type t = class_type_lookup_result *)
```
jonludlam, line 255: For these I'm quite tempted to keep the type here as documentation (internal right now, since it's not actually exposed in the interface) - though the alias is not terribly helpful. How about:

```ocaml
type t = Resolved.ClassType.t * Component.ClassType.t
```

It also might be useful to declare the error type too - again, mainly for documentation purposes. What do you think?

Julow: I'd too prefer to define this type fully and define `class_type_lookup_result` as alias to it but I prefer to not duplicate the definition.
I've added back all the types and added an instance of the result type to the first function of every modules (for documentation).

jonludlam: I don't think we need `class_type_lookup_result` - so no duplication needed. I think the only one actually exposed is `module_lookup_result`

jonludlam: (beyond the fact that they are used in subsequent result types, which we can duplicate in the modules I think)

Julow: It's not exposed but it's used to define `type_lookup_result` and that's the duplication I would like to avoid. I think it's also useful to show that `type_lookup_result` is an aggregate of existing return types rather than a new thing.

jonludlam: OK, let's keep this as-is for now, we might still be able to tidy up but it's not a big issue :-)

PR #718, opened 2021-09-01
Remove the prefix when rendering a label reference

jonludlam:
Fixes ocaml/v3.ocaml.org-server#83

### DIFF
--- a/src/document/comment.ml
+++ b/src/document/comment.ml
@@ -55,7 +55,7 @@ module Reference = struct
     | `InstanceVariable (r, s) ->
         (* CR trefis: the following makes no sense to me... *)
         render_resolved (r :> t) ^ "." ^ InstanceVariableName.to_string s
-    | `Label (r, s) -> render_resolved (r :> t) ^ ":" ^ LabelName.to_string s
+    | `Label (_, s) -> LabelName.to_string s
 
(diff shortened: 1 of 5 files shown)


Julow (commented):
Some tests need to be promoted. Otherwise, looks good :)

jonludlam:
Thanks!

gadmm:
Just got a notification for this. IIUC in the original example this would give "global_memory_limits let you bound..." instead of "Global memory limits let you bound...", right? In the original version, the string is taken from the section title being pointed to.

jonludlam:
Ah, you're right! I entirely missed this feature! Hrm, this will be a bit more work.

gadmm:
Thanks for looking into it (an alternative which is entirely fine to me would be to be able to define the text of the link, like you can already do for external links).

jonludlam:
You should already be able to do this: `{{!section:Memprof_limits.global_memory_limits}Global memory limits}` or, using the better syntax, `{{!Memprof_limits.section-global_memory_limits}Global memory limits}`

jonludlam:
But I'd like to fix this issue anyway as it's a regression from 1.5!

gadmm:
I'm pretty sure I already tried something like this, I'll try again.

PR #719, opened 2021-09-02
All double underscored modules are hidden

jonludlam:
This makes all modules with a double underscore in their names
hidden - previously it was only those with a filename with a double
underscore.


Julow (approved):
Looks good :)

jonludlam:
I'm not particularly happy with the various mechanisms of hiding right now. I'd quite like to rationalise it all, but I think that can wait a bit.

Issue #701, opened 2021-07-06
Wrong link in table of contents due to duplicate generated name


gadmm:
Consider the following example:
```
{1 Awesome feature}
{2 Example}
{2 Use-cases}
{1 More awesome feature}
{2 Example}
{2 Use-cases}
```
The table of contents will generate erroneous links to the second "Example" and "Use-cases" sections, because the generated html names are the same as the first ones (`example` and `use-cases`).

A workaround is to name the sections manually but the problem is not immediately noticable, it is tedious to fix by hand, and it keeps happening.

Drup:
Hmm, you are absolutely correct.

Should we generate unique ids for all sections ? It makes manual linking quite more annoying. Another solution is to assume that top level sections are at least unique, and use that to disambiguate.

bluddy:
Another option is to detect duplicates and append a number to them (_1 etc).

jonludlam:
We discussed this issue in the last odoc dev meeting. The current plan is:

1) Make the table of contents work -  possibly, as @bluddy suggests, by appending something to the anchors to make them unique. This will only happen on the second and subsequent instances of an anchor, so the first will always have the 'standard' id.
2) Issue a warning if a reference resolves to an ambiguous label, and the reference will end up pointing at the first instance of the ambiguous label (because subsequent ones will be renamed).

gadmm:
Nice for the quick response!
I don't know how odoc is implemented so I cannot comment at a technical level and I do not know if you already take this into account, but it would be nice to not break existing links:
- anchors specified by hand are never altered (so if the generated duplicate comes before, the one before should have something appended).
- if there is currently no ambiguity for a generated anchor name, then it is not altered.

PR #720, opened 2021-09-02
Handle ambiguous labels

Julow:
Fixes https://github.com/ocaml/odoc/issues/701

This PR ensures that heading anchors are unique internally to keep the TOC working. Also, the warning raised by references to ambiguous labels is improved.

If an explicit label is ambiguous, a warning is generated but no effort is made to avoid renaming the anchor or to allow references to it.
I think the suggested special case is not necessary because a warning is generated, which mean the input is considered wrong.


jonludlam:
To confirm, you're planning on moving the logic to the document IR layer?

Julow:
I'm working on it :)

Julow:
I moved the disambiguation logic into the `document` library. That way, the code no longer relies on comparing locations and on the `Env` module and the disambiguation applies to automatic headings (eg. functor parameters/result).
I've had some difficulty with subpages but I think I found something sound.

Julow:
Noticed a problem with sub-subpages in my previous code, fixed now. This is ready for review.

src/xref2/env.ml
```diff
@@ -68,7 +68,7 @@ let ident_of_element = function
   | `ModuleType (id, _) -> (id :> Identifier.t)
   | `Type (id, _) -> (id :> Identifier.t)
   | `Value (id, _) -> (id :> Identifier.t)
-  | `Label id -> (id :> Identifier.t)
+  | `Label (id, _) -> (id :> Identifier.t)
```
jonludlam, line 71: Looks much more consistent :-)

src/xref2/env.ml
```diff
@@ -123,7 +133,8 @@ end = struct
         let mark_shadow e =
           if e.kind = kind then { e with shadowed = true } else e
         in
-        if List.exists has_shadow tl then List.map mark_shadow tl else tl
+        if shadow && List.exists has_shadow tl then List.map mark_shadow tl
+        else tl
```
jonludlam, line 137: This seems a bit awkward. I wonder if it would be cleaner to handle the shadowing on the lookup path? It seems that the 'shadowed' bool is a bit redundant - we could recalculate it even if it wasn't being stored.

src/xref2/component.ml
```diff
@@ -455,6 +455,12 @@ and CComment : sig
 end =
   CComment
 
+and Label : sig
+  type t =
+    [ `Heading of Odoc_model.Comment.heading ] Odoc_model.Comment.with_location
```
jonludlam, line 457: I'm not convinced by this. In all of the other Components we use `Ident.t` rather than identifier (e.g. in the above definition of `CComment`). This is because we use these components for doing expansions and often end up placing them in different contexts than they were originally found - this usually works because we end up turning the local `Ident.t` back into a global `Identifier.t` at the right place, but if you're preserving the `Identifier` it'll be incorrect. I think what you actually want is to pick out the `Heading` from the `CComment` module above as a separate type and use that instead.

src/xref2/link.ml
```diff
@@ -28,6 +28,40 @@ let synopsis_of_module env (m : Component.Module.t) =
...
+    (* If ambiguous again, insert an extra '_' before the index. *)
+    match Env.lookup_by_name Env.s_label new_name env with
+    | Ok _ | Error (`Ambiguous _) -> disambiguate_label index (name ^ "_")
+    | Error `Not_found -> new_name
+  in
+  let location_equal a b = Location_.location a = Location_.location b in
+  (* Returns [0] if [heading] is the first occurence. The first occurence is at
+     the end of the list stored in [env]. Compare only the location as it is
+     unique to each heading and unlikely to be updated. *)
+  let rec dup_index = function
```
jonludlam: Given that a lot of this code is removed in a subsequent commit, can we squash them?

src/xref2/env.ml
Julow: The problem with recomputing it in `find` is that it depends on the kind, which might be too heavy for the fast path.
I've changed the code to remove the shadowing bit entirely. There's now two maps, one where shadowed elements are removed and an other for identifiers.

src/xref2/link.ml
Julow: I've squashed the related commits.

src/xref2/component.ml
Julow: This is fixed. I've also made `Label.t` a record to make sure we catch every places where the conversion is needed.

PR #723, opened 2021-09-09
Add markdown generator

### LINKED ISSUE #709: Markdown output
We would like to be able to render to a markdown document.

lubegasimon:
Fixes #709 

**Todo**
~- [] Don’t render OCaml syntax, that is `sig`,`end` etc. This might require tweaking the `document` library slightly. (**Edit**: _Disclaimer_, no tweaking of the document library is needed!)~
- [ ] Add options to render markdown in different flavours. That is, [pandoc](https://pandoc.org/getting-started.html) and [github’s markdown](https://github.github.com/gfm/). (Stretch goal)
- [x] Add option to enable and disable links (suggested by @lpw25 in `odoc`-dev meeting 9/9/21)
- [x] Fix `end` keyword that is poorly aligned for submodules.
- [x] handling of `labels`
- [x] use unicode character rather than `&nbsp` (by @lpw25).
- [x] tie anchor tags to `—generate-links` options (@lpw25)
~- [ ] Treat doc comments in modules and records a little different from those outside.~
- [ ] updating the docs to include the markdown generator


lubegasimon:
@Drup, this is ready for review 🙂.

src/markdown/generator.ml
```diff
@@ -0,0 +1,310 @@
...
+      | Linebreak -> break ++ inline rest
+      | Link (href, content) ->
+          (let rec f (content : Inline.t) =
+             match content with
+             | [] -> noop
+             | i :: rest -> (
+                 match i.desc with
+                 | Text s ->
+                     command "" (str "[%s]" s ++ str "(%s)" href) ++ f rest
+                 | _ -> noop ++ f rest)
```
Drup: That whole function doesn't seem right. Rich content should be allowed in Links and results of something of the form `[<the content>](link)`. Here you seem to map the link over the texts in the content ... ?

Why not share the link creation with `InternalLink` below (which seems more correct).

src/markdown/generator.ml
```diff
@@ -0,0 +1,310 @@
...
+              | _ -> Stop_and_keep)
+          in
+          let f i =
+            let key = inline i.Description.key in
+            let def = block i.Description.definition in
+            break ++ str "@" ++ key ++ str " : " ++ def ++ break ++ break
+          in
+          list ~sep:break (List.map f descrs) ++ continue rest
+      | Source content -> source_code content ++ continue rest
+      | Verbatim content -> str "%s" content ++ continue rest
```
Drup, line 186: That's not marked up as verbatim (it needs indentation or backticks).

src/markdown/generator.ml
```diff
@@ -0,0 +1,310 @@
...
+      let continue r = if r = [] then noop else break ++ block r in
+      match b.desc with
+      | Inline i -> inline i ++ continue rest
+      | Paragraph i -> inline i ++ continue rest
+      | List (list_typ, l) ->
+          let f n b =
+            let bullet =
+              match list_typ with
+              | Unordered -> escaped "- "
+              | Ordered -> str "%d) " (n + 1)
```
Drup, line 168: Ordered lists in the standard are `%d.`, might as well use that.

Also, you'll need to be pretty careful about breaks inside list items (it requires some indentation to preserve list-ness).

src/markdown/generator.ml
```diff
@@ -0,0 +1,310 @@
...
+    | 3 -> "###"
+    | 4 -> "####"
+    | 5 -> "#####"
+    | 6 -> "######"
+    | _ -> ""
+    (* We can be sure that h6 will never be exceded! *)
+  in
+  match label with
+  | Some _ -> command level (str " " ++ inline title)
+  | None -> command (level ^ " ") (inline title)
```
Drup, line 216: Not sure what's the difference between those two, but in any case, the label is missing. There is a pandoc extension for labels.

src/markdown/generator.ml
```diff
@@ -0,0 +1,310 @@
...
+            Take.until l ~classify:(function
+              | { Block.desc = Description l; _ } -> Accum l
+              | _ -> Stop_and_keep)
+          in
+          let f i =
+            let key = inline i.Description.key in
+            let def = block i.Description.definition in
+            break ++ str "@" ++ key ++ str " : " ++ def ++ break ++ break
+          in
+          list ~sep:break (List.map f descrs) ++ continue rest
```
Drup, line 184: Why not use the `definition_list` extension from pandoc ? It'll be non-standard anyway.

Drup (commented):
Here's a first set of comments. Could you post the result for various files so that people can look at the results.

Maybe we can try a version that inline submodules, like in the latex version. Not sure It'll give good results, but it might be more readable in a text editor than 1 file per module.

src/markdown/generator.ml
Julow: This extension is [header_attributes](https://pandoc.org/MANUAL.html#extension-header_attributes).
We might need to add a "target" option (with values either `pandoc` or `gh` let's say) and to use the syntax only if it makes sense.

Anyway, for now, we need to at least use the `<a id="...">` trick to make the label work.

src/markdown/generator.ml
```diff
@@ -0,0 +1,310 @@
...
+      | Styled (sty, content) -> style sty (inline content) ++ inline rest
+      | Linebreak -> break ++ inline rest
+      | Link (href, content) ->
+          (let rec f (content : Inline.t) =
+             match content with
+             | [] -> noop
+             | i :: rest -> (
+                 match i.desc with
+                 | Text s ->
+                     command "" (str "[%s]" s ++ str "(%s)" href) ++ f rest
```
Julow:
```suggestion
                     command "" (str "[%s](%s)" s href) ++ f rest
```

src/markdown/generator.ml
```diff
@@ -0,0 +1,310 @@
...
+  let append t1 t2 =
+    match (t1, t2) with
+    | Concat l1, Concat l2 -> Concat (l1 @ l2)
+    | Concat l1, e2 -> Concat (l1 @ [ e2 ])
+    | e1, Concat l2 -> Concat (e1 :: l2)
+    | e1, e2 -> Concat [ e1; e2 ]
+
+  let ( ++ ) = append
+
+  let concat = List.fold_left ( ++ ) (Concat [])
```
Julow, line 41:
```suggestion
  let concat ts = Concat ts
```

src/markdown/generator.ml
```diff
@@ -0,0 +1,310 @@
...
+  | i :: rest -> (
+      match i.desc with
+      | Text "" -> inline rest
+      | Text _ ->
+          let l, _, rest =
+            Doctree.Take.until l ~classify:(function
+              | { Inline.desc = Text s; _ } -> Accum [ s ]
+              | _ -> Stop_and_keep)
+          in
+          str {|%s|} (String.concat "" l) ++ inline rest
```
Julow, line 120:
```suggestion
              | { Inline.desc = Text s; _ } -> Accum [ str "%s" s ]
              | _ -> Stop_and_keep)
          in
          concat l ++ inline rest
```

src/markdown/generator.ml
```diff
@@ -0,0 +1,310 @@
+open Odoc_document
+open Types
+open Doctree
+
+module Markup = struct
```
Julow, line 5: What about making `t` abstract ? Writing down the types of all the functions signatures would make it more readable. Also, it would allow internal functions to the `Markup` module that are inaccessible outside of it.
You'll need a function `let string s = String s` (or an other name) because there's some uses of `String` later. (same for `Anchor`).

src/markdown/generator.ml
```diff
@@ -0,0 +1,327 @@
...
+            | Some _ -> inline content ++ inline rest
+            | None ->
+                open_sq_bracket ++ inline content ++ close_sq_bracket
+                ++ open_parenthesis
+                ++ String ("#" ^ link.anchor)
+                ++ close_parenthesis ++ inline rest
+          else inline content ++ inline rest
+      | InternalLink (Unresolved content) -> inline content ++ inline rest
+      | Source content ->
+          env Any "`` " "`` " (source_code content) ++ inline rest
```
Julow, line 152: Should be `"`` " " ``"` (eg. spaces towards the inside) but most importantly, it should be just `` "`" "`" `` if `content` doesn't contain any `` ` ``. Also, shouldn't it be `command` instead of `env` ? (not sure what `command` does)

Perhaps there should be a `Inline_source` constructor in `Markup` for handling this ? (eg. let that module handle the `` ` ``).

src/markdown/generator.ml
Julow: I think `Env` is supposed to implement indentation ? (seems planned in the types but there's no implementation)

src/markdown/generator.ml
```diff
@@ -0,0 +1,327 @@
...
+          | [] -> ()
+          | [ x ] -> Format.fprintf fmt "%a" pp x
+          | x :: xs -> Format.fprintf fmt "%a@\n%a" pp x pp (Block xs)
+        in
+        inner b
+    | Concat l -> pp_many fmt l
+    | Break -> Format.fprintf fmt "@\n"
+    | Anchor s -> Format.fprintf fmt "<a id=\"%s\"></a>" s
+    | String s -> Format.fprintf fmt "%s" s
+    | Env (indent, s, t, s') ->
```
Julow, line 70: Perhaps there should be newlines before and after `t` in case it's a `Block` ? The `Concat` constructor makes it harder.

Thought about the `Concat` constructor: Remove it, use `Block` and `Inline` instead. Change `Inline` into `Inline of t list`. Change `++` to be smart and use `Block` if a direct children is a `Block _` or `Env (_, _, Block _, _)`, use `Inline` if it's anything else.

src/markdown/generator.ml
```diff
@@ -0,0 +1,327 @@
...
+  match l with
+  | [] -> noop
+  | i :: rest -> (
+      let continue r = if r = [] then noop else break ++ break ++ item nbsp r in
+      match i with
+      | Text b -> block b ++ continue rest
+      | Heading h -> break ++ heading h ++ break ++ continue rest
+      | Declaration { attr = _; anchor; content; doc } ->
+          let nbsp' = "&nbsp; &nbsp; &nbsp;" in
+          let decl = documented_src content (nbsp ^ nbsp') in
```
Julow, line 281: Instead of a string, `nbsp` could be a `Markup.t` or a `Markup.t list`. (if it makes the code clearer)

src/markdown/generator.ml
```diff
@@ -0,0 +1,327 @@
...
+and subpages generate_links i =
+  Utils.flatmap ~f:(calc_subpages generate_links) @@ Doctree.Subpages.compute i
+
+and page generate_links ({ Page.header; items; url; _ } as p) =
+  let header = Shift.compute ~on_sub header in
+  let items = Shift.compute ~on_sub items in
+  let subpages = subpages generate_links p in
+  Block
+    ([ Inline (Link.for_printing url) ]
+    @ [ item "&nbsp; " header ++ item "&nbsp; " items ]
```
Julow, line 314: Is this initial `&nbsp;` wanted ? I'm not sure. I feel like it should start at nothing and increment 4 at a time.

src/markdown/generator.ml
```diff
@@ -0,0 +1,327 @@
...
+    @ subpages)
+
+let rec subpage subp =
+  let p = subp.Subpage.content in
+  if Link.should_inline p.url then [] else [ render p ]
+
+and render (p : Page.t) =
+  let content fmt =
+    Format.fprintf fmt "%a" Markup.pp (page !generate_links p)
```
Julow, line 323: I think we should pass around a record instead of a bare boolean:
- If it's named `args` or `conf`, it'll be easier to see its purpose. This is a common pattern.
- It'll be used by future options.

This record is already defined in src/odoc/markdown.ml, I think it should be defined here and never passed as a global variable or constructed more than once.
This global variable is not necessary at all.

src/markdown/generator.ml
Julow: Random suggestion: Insert a `---` line between the header and the items, it'll make a horizontal bar.

src/markdown/generator.ml
```diff
@@ -0,0 +1,310 @@
...
+  match e with "#45" -> escaped "-" | "gt" -> str ">" | s -> str "&%s;" s
+
+let raw_markup (_ : Raw_markup.t) = noop
+
+let style (style : style) content =
+  match style with
+  | `Bold -> command "**" (content ++ str "**")
+  | `Italic | `Emphasis -> command "_" (content ++ str "_")
+  | `Superscript -> command "<sup>" content
+  | `Subscript -> command "<sub>" content
```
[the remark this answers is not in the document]
lubegasimon: >Aren't the super/sub cases missing a closing tag here ?

The existing one should work well.

>Do you intend to try to do something about emphasis inside emphasis ?

No, because it would zero down to an emphasised statement. ( _Forexample _this_._)

src/markdown/generator.ml
lubegasimon: Indeed, indentation will be better

src/markdown/generator.ml
```diff
@@ -0,0 +1,310 @@
...
+module Markup = struct
+  type t =
+    | Inline of string list
+    | Block of t list
+    | Concat of t list
+    | Break
+    | Anchor of string
+    | String of string
+    | Env of indent * string * t * string
+    | Command of string * t
```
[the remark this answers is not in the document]
lubegasimon: Indeed, I’ve gotten rid of it in the most recent commit. Thanks!

src/markdown/generator.ml
lubegasimon: Thanks, however, I wasn’t able to abstract the link creation because of the difference in types of arguments to `Link` and `InternalLink`, rather improved the `Link` branch and it now looks a lot like `InternalLink`’s.

src/markdown/generator.ml
lubegasimon: This must be resolved in the most recent commit using `space ++ space`  — > `(“  “)`

src/markdown/generator.ml
lubegasimon: This is resolved in the most recent commit, thanks!

src/markdown/generator.ml
lubegasimon: same as https://github.com/ocaml/odoc/pull/723#discussion_r712841985

src/markdown/generator.ml
lubegasimon: same as https://github.com/ocaml/odoc/pull/723#discussion_r712841985

src/markdown/generator.ml
lubegasimon: same as https://github.com/ocaml/odoc/pull/723#discussion_r712841985

src/markdown/generator.ml
lubegasimon: same as https://github.com/ocaml/odoc/pull/723#discussion_r712841985

src/markdown/generator.ml
lubegasimon: instance of https://github.com/ocaml/odoc/pull/723#discussion_r712841985

src/markdown/generator.ml
lubegasimon: > Is this initial &nbsp; wanted ? I'm not sure. I feel like it should start at nothing and increment 4 at a time.

Yes, because it’s to that that we increment `&nbsp` whenever there is a recursive call of `item`

src/markdown/generator.ml
lubegasimon: > What about making t abstract ?
 

I’m not sure about your suggestion!

src/markdown/generator.ml
lubegasimon: I don’t clearly understand why we need to perform this change!

src/markdown/generator.ml
Julow: > Yes, because it’s to that that we increment &nbsp whenever there is a recursive call of item

You didn't understand my comment. Top-level items have a single nbsp, nested items have 3 more nbsps for each level. 1, 4, 7, 10.
I can't see a reason for the first one, I think this should be 0, 4, 8, 12.

src/markdown/generator.ml
Julow: I think I misunderstood how inline and block works in your code.

Can you explain what the `Env` constructor does and what is expected when `t` is a `Block _` or a `Inline _` ?

src/markdown/generator.ml
Julow: Making the type `t` abstract will help split the concern between `Markup` and the rest of the module and also improve readability. Like this:

```ocaml
module Markup : sig

  type t

  (* The signature of other functions *)

end = struct
```

That way it's easier to understand the rest of the code by looking only at that interface but also it's easier to modify and document `Markup`.

src/markdown/generator.ml
lubegasimon: I was being used for formatting text descriptions, for example setting `some text` in source code style by wrapping it in ` [ ]`. However, the `Env` constructor and its operation are no longer being used.

src/markdown/generator.ml
lubegasimon: Being that we now have a `—md-flavour` option, more of the renderer<`Pandoc` and `GFM`>-specific syntaxes will be added. Thanks!

lubegasimon:
This PR is ready for another review 🙂.

src/odoc/bin/main.ml
```diff
@@ -572,6 +572,30 @@ module Odoc_latex = Make_renderer (struct
...
+
+  let md_flavour =
+    let doc =
+      "The markdown renderer you are targeting, though only\n\
+      \    Pandoc and Github Flavoured Markdown are supported, the default \
+       being GFM."
+    in
+    Arg.(
+      value & opt string ""
+      & info ~docv:"MARKDOWN_FLAVOUR" ~doc [ "md-flavour" ])
```
Julow, line 592: This option should be renamed `flavour`.

src/markdown/generator.ml
```diff
@@ -0,0 +1,382 @@
...
+    match level with
+    | 1 -> make_hashes 1
+    | 2 -> make_hashes 2
+    | 3 -> make_hashes 3
+    | 4 -> make_hashes 4
+    | 5 -> make_hashes 5
+    | 6 -> make_hashes 6
+    | _ -> ""
+    (* We can be sure that h6 will never be exceded! *)
+  in
```
Julow, line 264:
```suggestion
  let level = make_hashes level in
```

src/markdown/generator.ml
```diff
@@ -0,0 +1,382 @@
...
+  | [] -> noop
+  | i :: rest -> (
+      let continue i = if i = [] then noop else inline i nbsp in
+      let make_link c s =
+        open_sq_bracket ++ continue c ++ close_sq_bracket ++ open_parenthesis
+        ++ string s ++ close_parenthesis ++ continue rest
+      in
+      let cond then_clause else_clause =
+        if !(args.generate_links) then then_clause else else_clause
+      in
```
Julow, line 175: This is very hard to read, please repeat the `if` everytime.

src/markdown/generator.ml
```diff
@@ -0,0 +1,382 @@
...
+          cond
+            (match link.page.parent with
+            | Some _ -> continue content ++ continue rest
+            | None -> make_link content (make_hashes 1 ^ link.anchor))
+            (continue content ++ continue rest)
+      | InternalLink (Unresolved content) -> continue content ++ continue rest
+      | Source content ->
+          cond
+            (source_code content nbsp ++ continue rest)
+            (backticks ++ source_code content nbsp ++ backticks ++ continue rest)
```
Julow, line 214: Why does this change formatting depending on generate-links ? I see no reason.

src/markdown/generator.ml
```diff
@@ -0,0 +1,382 @@
...
+        ++ string s ++ close_parenthesis ++ continue rest
+      in
+      let cond then_clause else_clause =
+        if !(args.generate_links) then then_clause else else_clause
+      in
+      match i.desc with
+      | Text "" -> continue rest
+      | Text s -> (
+          match s with
+          | "end" | "}" | "]" ->
```
Julow, line 180: Documentation containing these words is broken:
```
(** end } ] *)
```

The keywords must not be matched into any text, only when nested into a `source_code` with the right tag.

src/markdown/generator.ml
```diff
@@ -0,0 +1,382 @@
...
+    | Space -> Format.fprintf fmt " "
+    | Anchor s -> Format.fprintf fmt "<a id=\"%s\"></a>" s
+    | String s -> Format.fprintf fmt "%s" s
+    (* We use double backticks to take care of polymorphic variants or content
+       within backtick, and the spaces before and after the backticks for
+       clarity on what should be enclosed in backticks. For example,
+       "type nums = [ | `One | `Two ]" would be rendered as "``|`````Monday`` "
+       if the spaces were missing.
+    *)
+    | Backticks -> Format.fprintf fmt " `` "
```
Julow, line 124: Should use a single backtick when possible (when there is no backticks in the enclosed code).

src/markdown/generator.ml
```diff
@@ -0,0 +1,382 @@
...
+and item nbsp' (l : Item.t list) : Markup.t =
+  match l with
+  | [] -> noop
+  | i :: rest -> (
+      let continue r = if r = [] then noop else item nbsp' r in
+      match i with
+      | Text b -> block b nbsp' ++ continue rest
+      | Heading h -> break ++ heading h nbsp' ++ break ++ break ++ continue rest
+      | Declaration { attr = _; anchor; content; doc } ->
+          let nbsp'' = nbsp ++ nbsp ++ nbsp ++ nbsp in
```
Julow, line 343: Move that code into `documented_src` because it needs both values.

src/markdown/generator.ml
```diff
@@ -0,0 +1,382 @@
...
+  | `Bold -> string "**" ++ (content ++ str "**")
+  | `Italic | `Emphasis -> string "_" ++ (content ++ str "_")
+  | `Superscript -> string "<sup>" ++ content
+  | `Subscript -> string "<sub>" ++ content
+
+let make_hashes n = String.make n '#'
+
+type args = { generate_links : bool ref; md_flavour : string ref }
+
+let args = { generate_links = ref true; md_flavour = ref "" }
```
Julow, line 152: Please don't use a global variable. The `args` value should be passed to every functions.

Julow (changes requested):
Spacing between block elements is not right:
```ocaml
(** p1 *)

(** p2

   p3

   {ul
     {- a}
     {- b}
    } *)
```
renders to:
```
p1
p2
p3
- a

- b
```
We expect an empty line between two blocks.

src/markdown/generator.ml
Julow: This is not fixed. Indentation is not applied to every lines:
```
(** {v
verbatim
text
v} *)
```
renders to:
```
    verbatim
text
```

Also, this should be rended the same way (existing indentation should be detected and normalized):
```
(** {v
      verbatim
      text
    v} *)
```

src/markdown/generator.ml
Julow: There is no `<a id...>` on headings with `--generate-links` enabled.

src/markdown/generator.ml
Julow: This is not fixed. Links are broken when the content is not normal text:
```
(** {{:href} {b test}} *)
```
renders to (with `generate-links`):
```
**test**
```

src/markdown/generator.ml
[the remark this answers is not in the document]
Julow: This is not fixed. The closing tag is missing: (`</sup>`)
```
<sup>test
```

src/markdown/generator.ml
Julow: The indentation of list items is still broken:
```
(** {ul
     {- p1

        p2}
    } *)
```
renders to:
```
- p1
p2
```

src/markdown/generator.ml
lubegasimon: That’s because we would end up with links in back ticks which will confuse the reader, that is, `` [text](href) `` (`text` will be a link, but it won’t be displayed as a link rather as source code.

src/markdown/generator.ml
lubegasimon: Hmm! only `}` behaves not right but yes, that’s a good catch. Thanks!

src/markdown/generator.ml
Julow: Do you have an example of links inside "source code" ?
Anyway, removing the backticks isn't the solution because we want them.

lubegasimon:
Closed in favour of https://github.com/ocaml/odoc/pull/791

PR #727, opened 2021-09-13
Fix resolution involving deeply nested substitutions

### LINKED ISSUE #690: Dead references after deep destructive substitutions
When substituting a signature item inside a submodule
```ocaml
module type S = sig
  module M: sig
    type t
  end
  type t = M.t
end
module type T = S with type M.t := int
```
the reference to the submodule item `M.t` are not updated outside of the submodule, which creates a dead reference inside the module type `u`.

jonludlam:
Fixes #690

There has been logic to fix this sort of issue for a long time, but it
broke at some point.

Mostly the fix is that when calculating the signatures of `With`
module expressions, we always recalculate rather than using any
previously calculated signature - this is because the previously
calculated signature makes a round-trip from Component.* to Lang.* and
Lang.ModuleType.simple_signature doesn't preserve removed items as the
Components layer does.

By itself this would not have fixed #690 though - The way the rest of
the fix works is by ensuring that at each level of the expansion we
have enough to figure out that a replacement is necessary and what it
should be. For example, in issue #690, @Octachron provides the
following example:

```ocaml
module type S = sig
  module M: sig
    type t
  end
  type t = M.t
end
module type T = S with type M.t := int
```

Before this fix, the expansion of `T` would look like this:

```ocaml
module M : sig end
type t = M.t
```

Clearly the `M.t` can't be resolved from this. Instead, this fix
works by ensuring that the expansion of `T` looks like this:

```ocaml
module M: sig type t end with type t := int
type t = M.t
```

From this we have enough information to replace the `M.t` with an
`int` at the point we do the resolution of the RHS of type `t`.

We don't really want the output to show
`sig type t end with type t := int` however, so in the document
layer these expressions are replaced with `sig ... end` as we had
before.

### DIFF
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -1430,14 +1430,22 @@ module Make (Syntax : SYNTAX) = struct
           ++ Link.from_path (m :> Paths.Path.t)
           ++ O.txt " " ++ O.keyword "end"
 
+    and is_elidable_with_u : Odoc_model.Lang.ModuleType.U.expr -> bool = function
+      | Path _ -> false
+      | Signature _ -> true
+      | With (_, expr) -> is_elidable_with_u expr
+      | TypeOf _ -> false
+
     and umty : Odoc_model.Lang.ModuleType.U.expr -> text =
      fun m ->
       match m with
       | Path p -> Link.from_path (p :> Paths.Path.t)
+      | Signature _ ->
+        Syntax.Mod.open_tag ++ O.txt " ... " ++ Syntax.Mod.close_tag
+      | With (_, expr) when is_elidable_with_u expr ->
+        Syntax.Mod.open_tag ++ O.txt " ... " ++ Syntax.Mod.close_tag
       | With (subs, expr) -> mty_with subs expr
       | TypeOf { t_desc; _ } -> mty_typeof t_desc
-      | Signature _ ->
-          Syntax.Mod.open_tag ++ O.txt " ... " ++ Syntax.Mod.close_tag
 
     and mty : Odoc_model.Lang.ModuleType.expr -> text =
      fun m ->
@@ -1470,6 +1478,8 @@ module Make (Syntax : SYNTAX) = struct
                  ++ mty arg_expr ++ O.txt ")" ++ O.txt " " ++ Syntax.Type.arrow
                  )
             ++ O.txt " " ++ mty expr
+        | With { w_expr ; _ } when is_elidable_with_u w_expr ->
+          Syntax.Mod.open_tag ++ O.txt " ... " ++ Syntax.Mod.close_tag
         | With { w_substitutions; w_expr; _ } -> mty_with w_substitutions w_expr
         | TypeOf { t_desc; _ } -> mty_typeof t_desc
         | Signature _ ->
--- a/test/xref2/deep_substitution.t/run.t
+++ b/test/xref2/deep_substitution.t/run.t
@@ -1,0 +1,46 @@
+Test to check that when doing a deep substitution inside a signature,
+that items that reference that element are still correctly resolved
+even if they're in a different path.
+
+  $ cat m.mli
+  module type S = sig
+    module M: sig
+      type t
+    end
+    type t = M.t
+  end
+  module type T = S with type M.t := int
+  
+
+In this, we want to check that the type `t` in module type `T` has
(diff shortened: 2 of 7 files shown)


test/xref2/deep_substitution.t/run.t
```diff
@@ -0,0 +1,46 @@
...
+  module type T = S with type M.t := int
+  
+
+In this, we want to check that the type `t` in module type `T` has
+its RHS correctly replaced with an `int`
+
+  $ ocamlc -c -bin-annot m.mli
+  $ odoc compile m.cmti
+  $ odoc link m.odoc
+  $ odoc html-generate m.odocl --indent -o .
```
Julow, line 21: We have a helper script for this:
```suggestion
  $ compile m.mli
```
The output is slightly different because it sets the parent. If that's annoying, perhaps we can change the script and update every other tests ?

src/document/generator.ml
```diff
@@ -1430,14 +1430,22 @@ module Make (Syntax : SYNTAX) = struct
           ++ Link.from_path (m :> Paths.Path.t)
           ++ O.txt " " ++ O.keyword "end"
 
+    and is_elidable_with_u : Odoc_model.Lang.ModuleType.U.expr -> bool = function
```
Julow, line 1433: Why not render nested substitutions ? It is rendered for `T`:

```
# Module Test
module type S = sig ... end
module type T = S with type M.t := int
```

but the documentation for `T` doesn't mention any substitution at all:
```
# Module type Test.T
module M : sig ... end
type t = int
```

Showing the substitution is not that bad:

```
# Module type Test.T
module M : M.S with type t := int
type t = int
```

It would be even more useful in case of non-destructive substitution because the definition of `t` would be in the same documentation page:
```
# Module type Test.T
module M : M.S with type t = int
type t = M.t
```

Julow (approved):

src/document/generator.ml
jonludlam: Hmm, it was certainly not my intention to remove the 'with type' always, just the construct `sig ... end with type ...`

I'll double check.

jonludlam: I think I misinterpreted what you mean - though I'm still not quite sure! To be clear, we've got:

```ocaml
module type S = sig
  module M: sig
    type t
  end
  type t = M.t
end
module type T = S with type M.t := int
```

You wrote that you thought this would be better:

```ocaml
# Module type Test.T
module M : M.S with type t := int
type t = int
```

I don't think you mean `M.S` - I think you might have meant `S.M` - but we can't do this because we can't have paths to elements within module types. Is this correct?

test/xref2/deep_substitution.t/run.t
jonludlam: We do have a helper script, true - but this way the test is a bit more self contained, and I often find this useful when I'm working on them because I find building them out of tree to be essential when developing (and debugging) these tests.

src/document/generator.ml
Julow: By `M.S`, I meant `S.M` or even `Test.S.M`, sorry for the confusion.

> Hmm, it was certainly not my intention to remove the 'with type' always, just the construct `sig ... end with type ...`

`sig ... end with type ...` is turned into just `sig ... end` but I'd expect `Test.S.M with type ...` even if `M` is not a module type.

> but we can't do this because we can't have paths to elements within module types.

On the definition of `T`, we have the path to `S` and the substitution, why can't nested substitution be represented the same way ?
Perhaps we can represent it like `(module type of Test.S.M) with type ...` internally ?

jonludlam: The problem isn't `M`, it's `S` - `S` is a module type so we can't do `S.M`.

Julow: That wouldn't be valid OCaml but it would be a valid reference (so can work internally). Do we require that every rendered paths should be valid OCaml paths ?

jonludlam: Even internally it's a Path not a Reference, and in any case I strongly prefer sticking to valid OCaml than to deviate. If the language supports it some day we can do it - for example I'm keeping an eye on the transparent ascription PR as that will help with some tricker parts of handling `module type of`.

Julow: Alright :) Sorry for the digression.

PR #729, opened 2021-09-16
Fix breadcrumbs for pages called 'index.mld'

jonludlam:


jonludlam:
That 'Format' commit also fixes the driver.md issue

dbuenzli:
I didn't look at the patch but tested it. It seems the markup is missing an enclosing `nav.odoc-nav`

src/html/tree.ml
```diff
@@ -70,37 +70,49 @@ let page_creator ?(theme_uri = Relative None) ?(support_uri = Relative None)
...
+           index.mld that would otherwise clash with their parent. In particular,
+           dune and odig both cause this situation right now. *)
+        let up_url = "../index.html" in
+        let parent_name = x.name in
         [
-          Html.a ~a:[ Html.a_href (href up_url) ] [ Html.txt "Up" ];
-          Html.txt " – ";
+          Html.a ~a:[ Html.a_href up_url ] [ Html.txt "Up" ];
+          Html.txt " - ";
+          Html.txt parent_name;
```
dbuenzli, line 88: This I guess `Html.nav ~a:[ Html.a_class [ "odoc-nav" ]]`

Julow: I was reviewing and was about to submit a patch :)

Julow (commented):
I've made a test for this feature (that will be easy to break in the future) and the rendering is inconsistent, as @dbuenzli noted: https://github.com/jonludlam/odoc/pull/37

Also, HTML is the only generator with this feature but should this be in the model ? This PR makes the page called "index" special and this affect how Odoc is used. I think the model should recognize this special kind of page so the generator code is more straightforward.

jonludlam:
Thanks for the fixes @Julow !

> This PR makes the page called "index" special and this affect how Odoc is used. I think the model should recognize this special kind of page so the generator code is more straightforward.

`index.mld` was special as soon as the parent/child code went in - this is fallout from that. In general we probably shouldn't have `index.mld` pages - instead e.g. for `astring` the mld file should be `astring.mld` and all other pages and modules should use that as parent. This is what we do in voodoo. This is why I added the warning.

dbuenzli:
Thanks ! The rendering seem ok now.

> `index.mld` was special as soon as the parent/child code went in - this is fallout from that. In general we probably shouldn't have `index.mld` pages - 

[This convention](https://erratique.ch/software/odig/doc/packaging.html#odoc_api_and_manual) was established quite a long time ago by `odig` so that package owners could customize their package page. I'm not sure it's such a good idea to add one more file that hard codes the name of the package in your source tree. 

In any case it would be nice if we could somehow agree on these on things. I don't think it's a good idea to have dozens of different conventions out there.

Julow:
I like the `index.mld` convention better because it is a standard name (easier to remember and to recognize), the web has the same convention and I think would it would make build systems impl and interface easier.

But I think it'll be better if Odoc didn't need to handle this case at all. The `index.mld` page doesn't have to be compiled to `page-index.odoc` or `../page-<pkg>.odoc`, simply `page-<pkg>.odoc` and the right `--parent` flags should work.

jonludlam:
To be clear we're not changing what is recommended for authors yet, and if and when we do I'd like to do it by updating the existing convention rather than creating a new one, and we will require backward compatibility.

This is only really affecting how the driver behaves if it has opted into the parent/child way to specify output structure (hence the warning is only when a parent has been explicitly specified).

PR #731, opened 2021-09-22
Strengthen module types

jonludlam:
I had thought I'd already done this!


jonludlam:
I think we might still be missing strengthening of class types too

src/xref2/strengthen.ml
```diff
@@ -83,11 +83,7 @@ and module_ :
 and module_type :
     Cpath.module_type -> Component.ModuleType.t -> Component.ModuleType.t =
  fun prefix m ->
-  let expr =
-    match m.expr with
-    | None -> Some (ModuleType.Path { p_path = prefix; p_expansion = None })
-    | Some _ -> m.expr
-  in
+  let expr = Some (ModuleType.Path { p_path = prefix; p_expansion = None }) in
```
Julow, line 86: Should we keep `m.expr` into `p_expansion` ? What if `m.expr` is something complicated like `With` or `TypeOf` ?

Julow (approved):
LGTM

src/xref2/strengthen.ml
jonludlam: Even if it _was_ something complicated now, it's simple now, and calculating the signature or expansion just looks at p_path and gets _it's_ expansion - [here](https://github.com/ocaml/odoc/blob/master/src/xref2/tools.ml#L1229-L1234) and [here](https://github.com/ocaml/odoc/blob/master/src/xref2/expand_tools.ml#L127-L131)

Julow: Alright, we don't need to keep the expansion around because it's getting recomputed anyway.

PR #732, opened 2021-09-22
OpaqueModule is not necessarily an error

jonludlam:
It's quite reasonable for a module to have no expansion.


Julow (approved):
LGTM.

I wonder if we should remove `OpaqueModule` from the errors types. Perhaps like `` (_, [ `OpaqueModule | `Error of tools_error ]) result ``

PR #733, opened 2021-09-23
Better docs

jonludlam:
Preview here: https://jonludlam.github.io/odoc/

PR #735, opened 2021-09-23
fix and update CONTRIBUTING.md

asavahista:
basically a draft - there is still a lot that potentially needs to be changed.
for example the roadmap was last updated 3 years ago (and the link is to the closed roadmap issue - maybe the link could be changed to one of the projects instead).
actually, most of the doc is still 3 years old so it might be worth going over the entire thing to make sure the overview is complete and correct

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -168,10 +168,10 @@ writing new tests, and want to know what they are actually touching. To use it,
 ## Project structure
 
 odoc is divided into several sub-libraries, each of which is a directory
-under `src/`. Most of these have a *main file*, whose name is the directory
-name prefixed with "`odoc__`". That main file is the interface for the entire
+under `src/`. Many of these have a *main file*, whose name is the directory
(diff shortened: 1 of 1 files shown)


lubegasimon:
Thanks for your efforts @asavahista, indeed documentation pull requests are a real deal. Though I’m certainly not sure of how much of this content will be overwritten by https://github.com/ocaml/odoc/pull/733.

asavahista:
ah, indeed it's completely obsoleted by that

jonludlam:
Thanks for the contribution anyway, @asavahista. Just bad timing that it came when we're overhauling lots of docs! Or good timing if you look at it another way: I'm looking for feedback on that docs PR - would you care to take a look over them and see if they look OK?

asavahista:
@jonludlam yeah, actually i've already been looking around a bit - just a heads up though, i'll probably only be looking for (possible) syntactic errors first as those are easier to find than semantic errors

jonludlam:
That'd be great, thank you!

Issue #730, opened 2021-09-16
.mld file errors are off by one


dbuenzli:
Sample: 

```
[doc-build odoc][0]: (brzo log --id 66)
File "/Users/dbuenzli/sync/repos/uucp/doc/unicode.mld", line 382, characters 35-52:
Failed to resolve reference unresolvedroot(Stdlib).Uchar.t Couldn't find "Stdlib"
```
This points to the [line before](https://github.com/dbuenzli/uucp/blob/4017b6dd71f29b98b32b29f08d508929ebcad240/doc/unicode.mld#L382) the actual reference.

asavahista:
i think this is the source:
https://github.com/ocaml/odoc/blob/79ee6133dd58ce2891eaa8f8a8de06cf7f9ba76b/src/loader/odoc_loader.ml#L7
this is in agreement with [`zero_pos` in the stdlib](https://github.com/ocaml/ocaml/blob/79f6c2228b4a30d472472f3160148ba600fb67d0/stdlib/lexing.ml#L140-L145).
changing `pos_lnum` from `0` to `1` works for the example given so i believe it is the issue, but i'm not sure so i won't be opening a pr (at least until someone can confirm) - it shouldn't be hard to make the pr anyway

asavahista:
pr has been merged so this *should* be fixed in master

PR #736, opened 2021-09-24
fix line number off-by-one-error

asavahista:
**warning:** not extensively tested. if others could also test that'd be great.

please let me know if i need to add regression tests (or anything else)

it's a very small and simple change so just committing to master and closing this also works

### DIFF
--- a/src/loader/odoc_loader.ml
+++ b/src/loader/odoc_loader.ml
@@ -4,7 +4,7 @@ module Error = Odoc_model.Error
 let read_string parent_definition filename text =
   let location =
     let pos =
-      Lexing.{ pos_fname = filename; pos_lnum = 0; pos_cnum = 0; pos_bol = 0 }
+      Lexing.{ pos_fname = filename; pos_lnum = 1; pos_cnum = 0; pos_bol = 0 }
     in
     Location.{ loc_start = pos; loc_end = pos; loc_ghost = true }
   in
--- a/test/integration/compile.t/line_numbers.mld
+++ b/test/integration/compile.t/line_numbers.mld
@@ -1,0 +1,32 @@
+{0 line numbers}
+{[]}
+
+
+
+
+
+{Bad Markup}
+
+
+
+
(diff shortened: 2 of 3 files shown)


asavahista:
will close #730

asavahista:
fixed the one test that was affected; the failing ci runs are the usual debian 4.02 to 4.05 - i'm assuming `with type foo := bar` in that one failing test was introduced in 4.06? if this is something that can be selectively disabled then someone might want to do that

test/xref2/references_to_pages.t/run.t
```diff
@@ -1,9 +1,9 @@
 # References to pages and items in pages
 
   $ compile p.mld good_references.mli bad_references.mli
-  File "p.mld", line 6, characters 5-11:
+  File "p.mld", line 7, characters 5-11:
```
Julow, line 4: This test is wrong, these references should be changed to: (the errors will disappear)
```
{!Good_references} {!Good_references.t}
```

Julow (approved):
LGTM !

Adding a test would be nice. The place to do that would be `test/integration/compile.t/`, you'd need to add an `.mld` file with an error (eg. not beginning with a heading) and a `odoc compile` command in `run.t`.

asavahista:
seems like not starting with a heading makes it stop completely (and not show any line numbers) - i'll just use some other error instead then

test/xref2/references_to_pages.t/run.t
Julow: This test and the other tests failing in CI will be fixed in https://github.com/ocaml/odoc/pull/738

asavahista:
(hopefully) removed the fix for the unrelated test which is now in #738

jonludlam:
Slightly surprisingly, the [examples for odoc-parser](https://ocaml-doc.github.io/odoc-parser/#example-usage) got this right. Surprising because the code was extracted from here :-)

Issue #594, opened 2021-02-18
References through `open`


Julow:
Currently there is no support for unqualified references to items brought into the scope by an `open`.
This was reported by @dbuenzli: https://github.com/ocaml/odoc/issues/585#issuecomment-777599121

For example:
```ocaml
module A : sig
  type t
end

open A (* We currently ignore this, internally we have a [Lang.Open.t] with [expansion = []]. *)

type u = t (* This works as expected and renders as [A.t] because OCaml resolves it for us. *)

(** {!t} This doesn't resolve. *)
```

### Expected behavior ?

```ocaml
open A

(** {!t} *)
```

In this example, should `t` resolve and should it render as `t` (as it's written) or `A.t` (as it would renders if it were a type expression) ?

What about this:

```ocaml
(** {!t} *)

open A
```

References can point to items defined below them, and even to `include`d items. Should `open` have the same behavior ?

Fixing this issue would probably break one of:
- References renders as they are written
- References renders to valid paths that could be typed checked again

dbuenzli:
> In this example, should `t` resolve and should it render as `t` (as it's written) or `A.t` (as it would renders if it were a type expression) ?

In my opinion it should render as `t` – in general, as written. For these reasons:

1. This allows to make some signatures less heavyweight. In particular for those definitions that draw from modules that are in a namespacing module, it allows to elide the latter. The actual link will still get to the exact place (we could also consider writing the full reference in a tooltip), so it shouldn't be confusing.
2. You still have the choice to make it render as `A.t` by using `{!A.t}`.

> Expected behavior ?

In general it's nice that `ocamldoc` language allows forwards references, but it's still unclear to me how these work exactly, it would be nice to have formal documentation about the exact notion of scope of the `ocamldoc` language and how it differs with the one from OCaml.

In any case I don't feel the need to go overboard with differences and in general it seems wise to keep the notion of scope of OCaml and `ocamldoc` as close as possible. Besides I highly suspect that:

> ```ocaml
> (** {!t} *)
> 
> open A
> ```

wouldn't work in `ocamldoc` (I didn't check) and I don't think it's such a good idea to support it.

Issue #332, opened 2019-03-06
Reconcile `odoc`'s notion of scope with `ocaml`'s notion of scope.


dbuenzli:
It seems that since the stdlib has been namespaced using module aliases in 4.07, references of the form `{!M}` were `M` is a `Stdlib.M` module are not resolved.

An instance of this problem can be seen in the `Fmt` module which as a lot of references to the `Format` module. See for example:

  https://b0-system.github.io/odig/doc/fmt/Fmt/index.html#val-pf

and the corresponding doc string:

https://github.com/dbuenzli/fmt/blob/d1c1bb26e00f22c160e6cc5d6b6376bce8426115/src/fmt.mli#L23

dbuenzli:
Related to https://github.com/ocaml/odoc/issues/42

Issue #42, opened 2016-12-31
Cross-references issues


dbuenzli:
There are still a few problems with cross references (`{{!U.id}text}` ocamldoc syntax).

1. It seems that you cannot refer to yourself in a compilation unit `U` which the `ocamldoc` language allows (i.e. references of the form `{{!U.id}text}`) in the doc strings of `U` itself). Not sure whether this was a good  idea in the first place though. (E.g. in ocamldoc the second `Ptime` in [this](http://docs.mirage.io/ptime/Ptime/index.html) module preamble links to iteself, not very useful in this case though).
2. References to other compilation units that a unit formally doesn't depend on are not resolved 
(e.g. `Ptime_clock` in [this](http://docs.mirage.io/ptime/Ptime/index.html) module preamble). This concerns intra and inter package references.
2. In ocamldoc `{!U.id}` resolution first looks for a module specification `U.id` and then for a documentation section `id` anchor. It seems to be the converse in `odoc` (see [this reference](http://docs.mirage.io/astring/Astring/index.html#val-(^)), [src](https://github.com/dbuenzli/astring/blob/842a4bc7c7a28bdc1c412d01bf7f3441c3758d31/src/astring.mli#L24-L25), [target](https://github.com/dbuenzli/astring/blob/842a4bc7c7a28bdc1c412d01bf7f3441c3758d31/src/astring.mli#L242-L249))
3. When a reference  `{{!U.id}text}` is unresolved, `U.id` is spliced in. `text` should instead, `span`ned  with a `xref-unresolved` class so that I can be made visible through CSS. Adding a `title` attribute to the span with the text `unresolved $id` allows to show the unresolved reference in a tool tip.

dbuenzli:
W.r.t. to self reference, the problem is that what `ocamldoc` does is ambiguous. Here's a concrete example. The `S.Make` here: 

http://docs.mirage.io/react/React/S/index.html#module-type-EqType
http://erratique.ch/software/react/doc/React.S.html#special

I corrected this the `react` sources https://github.com/dbuenzli/react/commit/426e44451d81a7b4fe18806dfd9238ed0071fede

trefis:
I just noticed one disturbing self reference in ocamlary. Have a look at the preamble of the `Buffer` module.
The reference should link to `Buffer.t` in the stdlib (at least, I believe this is the intent) but gets only partially resolved to the `Buffer` module being defined.

dbuenzli:
[...]

PR #739, opened 2021-09-24
Resolve definition point comments in the context of any expansion

### LINKED ISSUE #713: Unqualified preamble references are not resolved when the preamble is outside the module.
In the following module the reference to type {!t} is resolved in the second case but not in the first case.

```ocaml
(** The module M. 

    Resolve type {!t}. *)
module M : sig 

  type t
end
```

```ocaml
module M : sig 
(** The module M. 

    Resolve type {!t}. *)

  type t
end
```

This seems to be a regression since I suddenly have quite a bunch of unresolved xrefs in my module preambles (example in astring [here](https://b0-system.github.io/odig/doc/astring/Astring/String/index.html), [src](https://github.com/dbuenzli/astring/blob/ec7a266a3a680e5d246689855c639da53d713428/src/astring.mli#L176-L187)).

jonludlam:
Fixes #713 

I tested this on the deps of odoc and found changes in `Astring.String` and `Cmdliner.{Arg, Manpage, Term}`, which looked correct. However, I also found changes in tyxml that were not so obviously right - e.g. in `Html_sigs`. Compare these:

https://jonludlam.github.io/odoc/deps/tyxml/Html_sigs/module-type-T/Xml/index.html
https://ocsigen.org/tyxml/4.4.0/api/Html_sigs.T.Xml

In both of these the links from `elt` and `attrib` go to the parent page, which looks correct to me. With this change, that comment is now resolved in the context of the module `Xml` and therefore they both link to the definitions of `elt` and `attrib` in that page instead of the parent. This looks incorrect to me (neither has a type variable!)

### DIFF
--- a/src/model/paths.ml
+++ b/src/model/paths.ml
@@ -739,6 +739,12 @@ module Fragment = struct
             | Branch (base, m) ->
                 (ModuleName.to_string base, Some (`Module (m, name))))
         | `OpaqueModule m -> split m
+
+      let rec identifier : t -> Identifier.Path.Module.t = function
+        | `Subst (_, s) -> identifier s
+        | `Alias (i, _) -> Path.Resolved.Module.identifier i
+        | `Module (m, n) -> `Module (Signature.identifier m, n)
+        | `OpaqueModule m -> identifier m
     end
 
     module ModuleType = struct
--- a/src/model/paths.mli
+++ b/src/model/paths.mli
@@ -500,6 +500,8 @@ module Fragment : sig
       type t = Paths_types.Resolved_fragment.module_
 
       val split : t -> string * t option
+
+      val identifier : t -> Identifier.Path.Module.t
     end
 
     module ModuleType : sig
--- a/src/xref2/expand_tools.ml
+++ b/src/xref2/expand_tools.ml
@@ -213,7 +213,7 @@ let expansion_of_u_module_type_expr env id expr =
 let expansion_of_module_alias env id path =
   let open Paths.Identifier in
   aux_expansion_of_module_alias ~strengthen:false env path
-  >>= handle_expansion env (id : Module.t :> Signature.t)
+  >>= handle_expansion env (id : Path.Module.t :> Signature.t)
   >>= fun (env, r) -> Ok (env, false, r)
 
 let expansion_of_module_type_of_desc env id t_desc =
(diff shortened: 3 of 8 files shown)


jonludlam:
N.b. in the underlying source the references `elt` and `attrib` are both unqualified: https://github.com/ocsigen/tyxml/blob/master/lib/html_sigs.mli#L83

jonludlam:
The differences I've noted so far have been both good and bad. Given that a) the implementation is a bit unpleasant - mainly because b) it's quite different from how reference resolution works elsewhere, and c) this behaviour is a bit surprising, I want to double check that it's actually a good idea to do this. The original argument was that it broke things that previously worked -- and that _is_ a reasonable argument -- but it also broke things that were working correctly _before_ this change. Admittedly my package sample size is small, so I can't say whether this patch does more harm than good overall. 

One possibility: I note that for @dbuenzli's affected libraries, the comment is written just above the explicit signature of the module: e.g. in [astring](https://github.com/dbuenzli/astring/blob/master/src/astring.mli#L176-L187) and in [cmdliner](https://github.com/dbuenzli/cmdliner/blob/master/src/cmdliner.mli#L29-L41), whereas for tyxml the [module type is just a path](https://github.com/ocsigen/tyxml/blob/master/lib/html_sigs.mli#L83-L93).

I think the behaviour in this PR is considerably less surprising for @dbuenzli's case than for tyxml's case (though still _slightly_ odd). As a compromise we could switch to the new behaviour only on existence of an explicit signature like in Astring and Cmdliner. It's still relatively easy to explain the behaviour and I suspect the code change in odoc could be less invasive.

The other solution is to do nothing, but change the source. For Astring and Cmdliner I think that amounts to putting that comment within the signature of the modules rather than just above it. Of course, this does mean that previous versions would remain broken.

@dbuenzli, @Drup, @lpw25 - opinions?

dbuenzli:
> The other solution is to do nothing, but change the source. For Astring and Cmdliner I think that amounts to putting that comment within the signature of the modules rather than just above it

You can also keep it above and simply qualify the identifiers with the module (which I already did in some libraries I had to release). My only gripes with all that is that it somehow kills the whole behaviour we defined for module preambles which was the idea that the following where somehow fully equivalent:

 ```ocaml
(** This module bla … *)
module B : sig 
end
```
and 
```ocaml
module B : sig 
(** This module bla … *)
end
```

But I don't think explaining to people that in the first case your are outside of the module so you need to qualify is *that horrible* – it's not absurd for an OCaml programmer.

I don't have a good grasp of the problems involved. Personally I don't see it as a big deal to change my sources in the future, as long as the actual scoping rules are finally nailed down and made known. See #332, #42, #594.

jonludlam:
OK, I think on balance the scoping rules make more sense to me as they currently are rather than with this PR merged. Let's go the route of changing the source instead.

PR #742, opened 2021-09-30
Stay compatible with two versions of the parser

Julow:
In the next version of 'odoc-parser', the representation of code blocks
changes. We can make Odoc work with both versions easily.

With this change, Odoc will be compatible with the current and next versions of Mdx and OCamlformat.


jonludlam:
Neat, thanks!

PR #751, opened 2021-10-08
Promote 'module_type_subst' test

Julow:
This test is constrained to OCaml `>= 4.13.0` and was never promoted.

It seems to be working fine but the HTML generator creates *a lot* of files.


Julow:
Yes and they look fine :) I'm not sure the amount of page was intended, though.

Issue #750, opened 2021-10-07
References to outside of module types must be fully qualified


Julow:
```ocaml
module X : sig

  module type T = sig
    (** {!t} *)
  end

  type t

end

module Y : X.T
```

The reference that's inside `X.T` won't resolve in the expansion of `Y`.

Annoyingly, the error message points to the reference and not to `Y` at all:
```
File "test.mli", line 4, characters 8-12:
Failed to resolve reference unresolvedroot(t) Couldn't find "t"
```

jonludlam:
Module types are where these issues are likely to show up the most, but the same problem applies anywhere where we build expansions. The problem is that the references are all resolved after the expansion has happened, so the original context is gone. Stdlib currently has a few of these sorts of issues - e.g. https://github.com/ocaml/ocaml/blob/45612f14afc1abaa19d8585ec1f48c3cbbfbc652/stdlib/map.mli#L242

PR #761, opened 2021-10-20
Fix bogus warning about labels ambiguous with themselves

Julow:
Labels don't follow the usual shadowing rule:
- The first definition wins
- The second definition is not allowed

Because of this, when the same signature is added to the environment twice (eg. when resolving the expansion of an include), an extra check is necessary to ensure that the same label is not added twice.

This bogus warning occurs in many projects (eg. the compiler, repr, irmin) and is very visible due to ocaml-ci upgrading to Odoc 2.0.0 soon after the release.

### DIFF
--- a/src/xref2/env.ml
+++ b/src/xref2/env.ml
@@ -189,9 +189,20 @@ let add_to_elts ?shadow kind identifier component env =
 let add_label identifier heading env =
   (* Disallow shadowing for labels. Duplicate names are disallowed and reported
      during linking. *)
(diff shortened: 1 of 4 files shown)


Julow:
I first tried avoiding adding things to the environment twice by resolving expansions with an empty environment, but that would have extended https://github.com/ocaml/odoc/issues/750 to `module type of M`.

jonludlam:
Superseded by #773

PR #762, opened 2021-10-20
Add context around errors from expansion of includes

Julow:
When an error occurs when resolving an expansion, the error message doesn't mention input file being linked but where the error is written, which is not enough to fix the problem.

For example, when linking `A`, the error will look like:

    File "c.ml", line 1, character 1:
    While resolving the expansion of include at File "a.ml", line 1, character 0
    While resolving the expansion of include at File "b.ml", line 1, character 0
    While resolving the expansion of include at File "c.ml", line 1, character 1
    Some error

### DIFF
--- a/src/loader/cmi.ml
+++ b/src/loader/cmi.ml
@@ -1025,3 +1025,14 @@ let read_interface root name intf =
   let id = `Root (root, Odoc_model.Names.ModuleName.make_std name) in
   let items = read_signature Env.empty id intf in
   (id, items)
+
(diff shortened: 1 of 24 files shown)


jonludlam:
I think I'd rather have a more comprehensive solution rather than special-casing includes (as discussed offline). As such I'm going to close this PR - we can still refer to it of course :-)

Julow:
I think we should merge this anyway. This PR adds the "Error" and "Warning" prefix and some other code that will be useful.
Also, the added context on includes is still useful while we don't have a generalized solution (if it doesn't make it in 2.0.1).

PR #764, opened 2021-10-25
Better environment handling

jonludlam:
Ensures we only ever add items to the environment once.

### DIFF
--- a/src/xref2/env.ml
+++ b/src/xref2/env.ml
@@ -96,6 +96,8 @@ module Elements : sig
       elements of the same name won't be shadowed. This is used for labels,
       which doesn't allow shadowing. *)
 
+  val remove : [< Identifier.t ] -> t -> t
+
   val find_by_name :
     (Component.Element.any -> 'b option) -> string -> t -> 'b list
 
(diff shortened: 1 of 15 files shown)


src/xref2/env.ml
```diff
@@ -136,6 +166,9 @@ end = struct
 end
 
 type t = {
+  linking : bool;
+  (* True if this is a linking environment - if not,
+     we only put in modules, module types, types, classes and class types *)
```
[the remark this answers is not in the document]
jonludlam: We might be able to split it into two - the 'core' elements and the 'extras' perhaps? I think this is perhaps beyond the scope of this particular PR though.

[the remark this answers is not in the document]
Julow: Alright, let's improve that later.

src/xref2/env.ml
```diff
@@ -112,6 +114,17 @@ end = struct
   let empty = (StringMap.empty, IdMap.empty)
 
   let add ?(shadow = true) kind identifier elem (names, ids) =
+    let check_for_duplicates () =
+      try
+        ignore (IdMap.find (identifier :> Identifier.t) ids);
```
Julow, line 119: Could be `if IdMap.mem ...`.

Julow (approved):
Let's merge :)

src/xref2/env.ml
jonludlam: Erm, yes... :-)

jonludlam: Heh, fixed :-) thanks!

Issue #669, opened 2021-05-04
Feature Request: different style for external and internal links


impossiblynew:
_Please remove this if this is not the appropriate avenue for a feature request._

I am currently writing a library that contains a lot of links to external resources, and I notice that while navigating my documentation that it's irritating how easy it is to accidentally click a link that sends you to a new page when you though you were just going to be navigated to a different part of the documentation.

Would it be possible to style external links differently in such a way as to make it obvious before clicking whether the link is internal or external?

asavahista:
for anyone that wants to pick this up, this is probably the place you'd want to change:
https://github.com/ocaml/odoc/blob/40aa7587a5140a88d77c2ce400f426acf5da79a8/src/html/generator.ml#L115-L118

it may be best to implement it as a class and use an `::after` pseudo-selector to add a symbol (svg using `currentColor`?) representing "external link" https://github.com/ocaml/odoc/blob/79ee6133dd58ce2891eaa8f8a8de06cf7f9ba76b/src/odoc/etc/odoc.css#L295-L300

Julow:
> it may be best to implement it as a class and use an ::after pseudo-selector to add a symbol (svg using currentColor?) representing "external link" 

Sounds reasonable. Contribution are welcome :)

asavahista:
not sure if this should also be implemented (or an option) for the latex generator.

also not sure where to source the icons - font awesome sounds like a safe bet but not sure if the [license](https://fontawesome.com/license/free) is acceptable

Julow:
External links are already formatted specifically in latex:
```
\href{my\_url}{my\_url}\footnote{\url{my\_url}}
```

Their license is the most permissible possible, it'll work with Odoc. It's also self-contained, no license work needed at all:

> Downloaded Font Awesome Free files already contain embedded comments with sufficient attribution, so you shouldn't need to do anything additional when using these files normally. 

Odoc has a command that writes the static files (`odoc support-files`, currently only the CSS file) but I'd prefer if the icon could be inlined into the CSS to avoid a request for such a small file.

yawaramin:
I just checked, and odoc generates relative paths for all internal links. The only other thing it needs to do is, for link markup like `{{: https://example.com} Example.com}`, check if the link starts with `http://`/`https://`, and generate a `rel="external"` attribute for those. Then the stylesheet can take care of the colouring.

Note, I am not suggesting `target="_blank"` with 'opens in new tab' markers, I agree that's a bad idea.

[...]

PR #767, opened 2021-10-28
Adding a font-awesome icon for external links

panglesd:
This is a PR to address Issue #669.

It adds an icon after external links. The icon is taken from here: https://fontawesome.com/v6.0/icons/up-right-from-square (it's the only "external link" icon that is not a Pro icon):

image

I tried to follow both suggestions: making the svg inline in the css, and using `currentColor`. However, this is not that easy: one cannot manipulate the color of an svg pseudo-element that easily.
I came up with the following ~~ugly hack~~ solution: the svg is a mask, and the background color is `currentColor`.
This comes with a price: with both its prefix and unprefixed version, [caniuse](https://caniuse.com/mdn-css_properties_mask-image_svg_masks) report that only 95.41% of users have a browser that supports this functionnality. To distinguish between desktop and mobile, its respectively 96.52% and 95.16%.
Here are some details:
- 0.73% as IE users
- 0.43% as old firefox versions (version before 53)
- 1.19% as "opera Mini" users (in fact, support is unknown, but I can't check!)
- 0.28% as "Android Browser" users
- 1.07% as "UC Browser for Internet" users 

For those people, the filter won't work, so instead of the `currentColor` external link icon, they will have a `currentColor` filled square...

How acceptable do you think that is? If that's not acceptable, there are to options:
- Give up on `currentColor`. We can use the widely supported (99.95% for desktop) `content` property. The icon will always be black.
- Give up on inlining the svg in the css, adding the svg as an element (instead of a pseudo-element) and giving it the color `currentColor`.


EmileTrotignon:
Could this be rebased @panglesd ?

panglesd:
I don't think a mutual agreement was reached in https://github.com/ocaml/odoc/issues/669, so I would be in favor of letting this PR rot. The feature is of pretty low priority/interest (but it was a good exercise to enter in this codebase!).

panglesd:
Let's tidy the open PRs.

panglesd:
I removed the added icon but kept the class on external links, and rebased.

Feel free to merge (or close) this PR!

Julow (approved):
A rebase is needed and the tests should be promoted on < 4.10.

This is harmless and improves customizability. Let's merge!

PR #768, opened 2021-10-29
Small fixes

Julow:
- The `canonical_module.t` test was wrong due to a bug in the `compile` helper script
- Upgrade to OCamlformat 0.19.0
- Remove every uses of `OCAML_MAJOR` and `OCAML_MINOR`, use the more robust `OCAML_VERSION` instead.
- and a few other changes too small for a dedicated PR


jonludlam:
Really nice, thanks!

PR #771, opened 2021-11-04
Allow referencing labels in the top comment

### LINKED ISSUE #740: Can't reference labels in top-comments
Both references in this example won't resolve:

```ocaml
module M : sig
  (** {1:label Heading in top-comment} *)
end

module type MT = sig
  (** {1:label Heading in top-comment} *)
end

(** {!M.label}
    {!MT.label} *)
```

They'll resolve if something is added before both comments.

jonludlam:
This PR splits the first comment in a signature at the firstheading. This means everything before will be put into the preamble of the page, and the rest of the comment becomes part of the signature. Since the top comment has never been put into the environment, previously this meant that the labels couldn't be referenced. Now they are part of the signature this starts to work.

Fixes #740

### DIFF
--- a/test/generators/html/odoc.css
+++ b/test/generators/html/odoc.css
@@ -1,0 +1,1 @@
+../../../src/odoc/etc/odoc.css
--- a/src/loader/cmt.ml
+++ b/src/loader/cmt.ml
@@ -182,8 +182,14 @@ and read_class_signature env parent params cltyp =
             [] csig.csig_fields
         in
         let items = constraints @ List.rev items in
-        let items, doc = Doc_attr.extract_top_comment_class items in
+        let items, (doc, doc_post) = Doc_attr.extract_top_comment_class items in
+        let items =
+          match doc_post with
+          | [] -> items
+          | _ -> Comment (`Docs doc_post) :: items
+        in
         Signature {self; items; doc}
+
     | Tcty_arrow _ -> assert false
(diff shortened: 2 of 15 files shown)


test/generators/html/odoc.css
```diff
@@ -0,0 +1,782 @@
+@charset "UTF-8";
```
Julow, line 1: I think we should include these files as symlink, otherwise, they'd get out of date.

Julow (approved):
This changes the behavior of the "second part of the first comment" (defined in https://github.com/ocaml/odoc/pull/640) in the case of functors.

The page rendered like this before this PR:

```
{0 Module M.X}

Synopsis.

{TOC}

# The "content" starts here

{1 The second part of the first comment}

{1 Parameters}

module Arg : ...

{1 Signature}
```

now, it is rendered like if it was a separate floating comment:

```
{0 Module M.X}

Synopsis.

{TOC}

# The "content" starts here

{1 Parameters}

module Arg : ...

{1 Signature}

{1 The second part of the first comment}
```

I think this change is fine. If something needs to be improved here, it's the rendering of the functor parameters. (for example using stronger headings)

test/generators/html/odoc.css
jonludlam: ah good point yes! I'll fix that.

Issue #760, opened 2021-10-20
Section labels are not properly namespaced by modules


dbuenzli:
On [this source](https://github.com/dbuenzli/mtime/blob/2d2673d5e3050e85ec98ed39d302407ab4181dbe/src/mtime.mli) `odoc` complains about the `spans` and `arith` labels as being ambiguous. But the two occurences are in different modules.

---
```
[doc-build odoc][0]: (brzo log --id 61)
File "/Users/dbuenzli/sync/repos/mtime/_b0/brzo/ocaml-doc/odoc/mtime/mtime.odoc":
Couldn't find the following modules:
  Stdlib
File "/Users/dbuenzli/sync/repos/mtime/src/mtime.mli", line 33, characters 6-36:
Label 'spans' is ambiguous. The other occurences are:
  File "/Users/dbuenzli/sync/repos/mtime/src/mtime.mli", line 33, character 6
  File "/Users/dbuenzli/sync/repos/mtime/src/mtime.mli", line 20, character 4
File "/Users/dbuenzli/sync/repos/mtime/src/mtime.mli", line 68, characters 6-26:
Label 'arith' is ambiguous. The other occurences are:
  File "/Users/dbuenzli/sync/repos/mtime/src/mtime.mli", line 68, character 6
  File "/Users/dbuenzli/sync/repos/mtime/src/mtime.mli", line 165, character 4
```

dbuenzli:
I have no idea but it's definitively a regression. I don't think this ever worked that way either in previous versions of `odoc` or in `ocamldoc`.

Issue #259, opened 2018-11-27
Automatically generating section anchors


aantron:
Many section headings are/will be written like this:

```
{1 Foo bar}
```

i.e., without an explicit anchor

```
{1:foo-bar Foo bar}
```

I think it's desirable for odoc to generate good, legible anchors automatically from the heading text "`Foo bar`."

<br/>

This opens several questions:

- How should the anchor be generated? For example, what should be done with whitespace?
- How should anchors be disambiguated? In particular, nested sections are probably likely to have the same heading text. See #186.
- Should anything be done to encourage anchor stability, so that links remain valid?
- How do we ensure heading anchors (auto-generated or not) don't conflict with signature item anchors?

<br/>

Here are some links with precedent:

- ["Auto-generated-ids" in the Asciidoctor manual](https://asciidoctor.org/docs/user-manual/#auto-generated-ids)
- ["Extension: `auto_identifiers`" in Pandoc](http://pandoc.org/MANUAL.html#extension-auto_identifiers)
- [Stringex's .to_url method](https://github.com/rsl/stringex)
- [An interesting discussion on CommonMark](https://talk.commonmark.org/t//115/25)
- GitHub does something for this as well, e.g. ["Using odoc with OCaml" in odoc's own README](https://github.com/ocaml/odoc#using-odoc-with-ocaml)

<br/>

There seems to be a preference for replacing whitespace with `-`. Even the Asciidoctor manual uses this, though the algorithm it describes uses `_` by default.

If we choose `-`, we need to change the new-style reference kind syntax (#90) to not also use dashes, or otherwise come up with a scheme for referring to a section in a reference.

<br/>

Given that auto-generated anchors are likely to be unstable, should we encourage explicit anchoring?

If so, should we emit a warning if an anchor was generated automatically?

And, if so, should we allow suppressing the warning, e.g. `{1:_ Foo}` for "no anchor", `{1:* Foo}` for "generate and don't warn?"

<br/>

Finally, should we simply not generate anchors?

trefis:
My 2 cents:

> Should anything be done to encourage anchor stability, so that links remain valid?

No. People should just manually pick their anchors if they want stability.

> How do we ensure heading anchors (auto-generated or not) don't conflict with signature item anchors?

The same way we disambiguate signature items with the same name?
That is: we could use `label:foo` (or `anchor:foo` or whatever) to distinguish from `value:foo`.

[...]

PR #773, opened 2021-11-05
Fix wrong "ambiguous label" warning

Julow:
Fix https://github.com/ocaml/odoc/issues/760

- Fix the "ambigous label declaration" warning. The previous version was incorrectly comparing label names instead of the full identifiers.
  Now the backends must also handle the full identifiers (third point).

- Remove the warning about references to ambiguous labels.
  If the label is explicitly declared, a warning will be generated there. However, there is no warnings if the conflicting label is automatically generated. This extra warning was meant for covering the second case.
  A solution to this is to disallow references to generated labels.

- The code in the issue still pose problems in the latex backend.
  This is due to a "disambiguation" pass that suffixes the generated labels in case of ambiguity, which was written to fix the TOC but with the same wrong assumption as the other checks.
  I'm still investigating this but the solution might be different whether we allow references to generated labels or not.

- Some improvements to `Env` (remove an edge case that was used to support the point 2).

### Disallow references to auto-generated labels ?

Added in https://github.com/ocaml/odoc/issues/259

This has the advantage of simply removing the problem added by this PR (reference to ambiguous auto-generated label is broken or point to the wrong thing) and ensuring the disambiguation pass doesn't interfere with references.

Few documents might rely on them:
- Ocamldoc doesn't allow that.
- It was not possible to reference labels generated from headings with more than one word until Odoc 2.0.


jonludlam:
This superseded #761 yes?

test/xref2/refs/refs.md
```diff
@@ -998,7 +998,7 @@ Reference to 'Y' is ambiguous. Please specify its kind: constructor-Y, module-Y.
    Y)
 # resolve_ref "Everything_ambiguous_in_sig.t" (* Some kinds are missing: label, type subst (would be "type-") *) ;;
 File "<test>":
-Reference to 't' is ambiguous. Please specify its kind: field-t, module-type-t, type-t, val-t.
+Reference to 't' is ambiguous. Please specify its kind: field-t, module-type-t, type-t, val-t, val-t.
```
Julow, line 1001: The two `val-` actually correspond to a `external` and a `val`. This is a bug in the `Find` module but instead of fixing it, I'm investigating removing that module entirely and using `Env` instead.

src/xref2/env.ml
```diff
@@ -80,97 +80,97 @@ type kind =
...
 type t = {
   linking : bool;
   (* True if this is a linking environment - if not,
      we only put in modules, module types, types, classes and class types *)
   id : int;
-  elts : Elements.t;
+  elts : ElementsByName.t;
+      (** Elements mapped by their name. Queried with {!find_by_name}. *)
+  ids : ElementsById.t;
+      (** Elements mapped by their identifier. Queried with {!find_by_id}. *)
```
Julow, line 173: This is just a code improvement and not necessary. It was used by an other failed attempt at fixing the issue and I kept the commit.

test/xref2/refs/refs.md
jonludlam: Might be better to leave that change to 2.1.0?

Julow:
> This supersedes #761 yes?

It does ! #761 was fixing an edge case that this PR removed.

src/xref2/env.ml
```diff
@@ -80,97 +80,99 @@ type kind =
...
+  module IdMap = Identifier.Maps.Any
+
+  type t = Component.Element.any IdMap.t
+
+  let empty = IdMap.empty
+
+  let add identifier element t =
+    let check_for_duplicates () =
+      if IdMap.mem (identifier :> Identifier.t) t then
+        Format.eprintf "Duplicate found: %a\n%!" Component.Fmt.model_identifier
```
jonludlam, line 151: We've lost the `assert` here

jonludlam:
Once the assert is back in I think this is good to merge.

Julow:
Before merging, what about the auto-generated labels ? Ambiguous references to these are no longer warned about.

jonludlam:
Well we're certainly not removing it in this bugfix release :-) It'd be better if we did warn, but since the sitatution with this PR merged is enormously better than before I'm quite happy to go with what we've got here now.

Issue #588, opened 2021-02-12
Latex: add a template preamble file as a support file


Octachron:
The latex renderer currently creates one latex file by module, and expects the user to provide the latex preamble.

Since the minimal preamble is not that small (see https://github.com/ocaml/odoc/blob/master/test/latex/expect/visualizer.tex for a not completely minimal example), it probably makes sense to provide a template preamble file as an `odoc support-files` template.

somya104:
Hey @jonludlam ! I am Somya  and I am an outreachy applicant. 
I am interested in contributing to the odoc project though I'm still learning OCaml.
Apart from OCaml, what else do I require to know in order to work on this particular issue?
Also, I have been working on a few issues in the website project too.

jonludlam:
Oops, missed this - though I'm going to suggest the same thing as in issue #575  - that is, it may make more sense for an initial contribution at ocaml.org first and _then_ have a crack at this one, as there will be more 'getting started' help over there. Having said that though, the bits to look at for this issue are the [Support_files module](https://github.com/ocaml/odoc/blob/master/src/odoc/support_files.ml), the [dune rule](https://github.com/ocaml/odoc/blob/master/src/odoc/dune#L12-L22) that creates the `Css_file` module, and the [main.ml section](https://github.com/ocaml/odoc/blob/master/src/odoc/bin/main.ml#L192-L208) that defines the `support-files` CLI sub-command.

somya104:
Okay. Thank you so much, that was really helpful! :)

somya104:
Hi @jonludlam 
I've been making a good amount of progress in learning OCaml and so far, I am able to write and understand the code now. I have also made 2 tiny projects so that I can at least start working on this issue as soon as possible (either today or tomorrow).
I am filling the Final application for Outreachy and I've been willing to seek help from you for the last question in the form which I have provided below:
> **Outreachy internship project timeline:** Please work with your mentor to provide a timeline of the work you plan to accomplish on the project and what tasks you will finish at each step. Make sure take into account any time commitments you have during the Outreachy internship round. If you are still working on your contributions and need more time, you can leave this blank and edit your application later.

Could you please help me in building a project timeline? I am really looking forward to work on this project.

jonludlam:
[...]

PR #774, opened 2021-11-06
Adding file for larex preamble file

NiraliJoshi17:
Adding a preamble file as a support-files.tex and define a rule for this in dune file.

### DIFF
--- a/src/odoc/etc/support-files.tex
+++ b/src/odoc/etc/support-files.tex
@@ -1,0 +1,88 @@
+% Helper files for transforming the test files into a pdf
+% compile with
+% TEXINPUTS=${TEXINPUTS}:test_package+ml/ xelatex visualizer.tex
+%
+\documentclass{book}
+
+\usepackage{fontspec}
+\usepackage{xunicode}
+
+\usepackage{changepage}
+\usepackage{longtable}
+\usepackage{listings}
+\usepackage[strings]{underscore}
+
+\usepackage[colorlinks=true]{hyperref}
+\usepackage{color}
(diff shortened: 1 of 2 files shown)


jonludlam:
Hi @NiraliJoshi17, thanks for this! It's looking good so far, you've added the latex file as an OCaml module. The next steps are to install it as part of the `support-files` CLI sub command, as suggested in issue #588 

Also you'll need to reformat the dune file as it's causing the CI to fail. To do this, install `ocamlformat.0.19.0` and run

```
dune build @fmt 
dune promote
```

and commit the results.

EmileTrotignon:
Is this still relevant ?
@NiraliJoshi17

src/odoc/etc/support-files.tex
```diff
@@ -0,0 +1,88 @@
...
+    keywords={[2]{val,initializer,nonrec}}, keywordstyle={[2]\ocamlkeyword},
+    belowskip=0\baselineskip,
+    upquote=true,
+    literate={'"'}{\textquotesingle "\textquotesingle}3
+    {'\\"'}{\textquotesingle \textbackslash"\textquotesingle}4,
+  }
+  }{}
+  \newcommand{\inputchapter}[1]{\chapter{#1}
+    \input{#1}}
+ \begin{document}
```
Julow, line 70: That part seems dubious. The preamble shouldn't be a document.

jonludlam:
I'm going to close this for now. When we do a support-files like command for latex we can always come back to take a look at this PR.

PR #775, opened 2021-11-08
Briefly document how to create cross page section link odoc markup

NathanReb:
I recently started using odoc and `.mld` files to write a manual for MDX.

I'm contributing to odoc's documentation in the hope it will help others such as myself getting famliar with its markup.

I wasn't too sure where to put this exactly. I believe it's on the right page but the organisation there got me a bit lost.

I'm happy to move it to another section if you think that would be best!

### DIFF
--- a/doc/odoc_for_authors.mld
+++ b/doc/odoc_for_authors.mld
@@ -663,6 +663,34 @@ v}
 In this case the reference text would be "Foo Bar" so the paragraph would read
 "See Foo Bar for details".
 
+Sections with a label cannot be referenced via the section title, only
+using the label.
+
+Note that label names can contain [-] but if they do, you have to quote the label
+name in the reference link as follows:
+
+{v
+{2:foo-bar Foo Bar}
+...
+See {!label-"foo-bar"} for details
+v}
+
+You can use [_] instead of [-] to avoid this syntax.
+
+Sections can also be cross referenced from another page.
+If you define the following section in a [a.mld] file:
+
+{v
+{1:foo Foo}
+v}
+
+And then want to reference it from another [b.mld] page,
+you can use the following syntax:
+
+{v
+See {!page-a.foo} for details
+v}
+
 {1:doc-pages Writing Documentation Pages}
 
 Files with the [.mld] extension are called {e documentation pages} and should


NathanReb:
I realized that cross page section link don't expand the label to the section title as it does for same page section links.

I naively documented it in the last commit but if that's considered a bug I'm happy to turn this into an issue!

doc/odoc_for_authors.mld
```diff
@@ -663,6 +663,29 @@ v}
 In this case the reference text would be "Foo Bar" so the paragraph would read
 "See Foo Bar for details".
 
+Note that label names cannot contain [-], you should use [_] instead.
```
Julow, line 666: It's possible !
However, the reference syntax is particular:
```
  {!"that-section"}
  {!that_other_page."that-section"}
  {!section-"that-section"}
```

NathanReb: Do you want me to document that?

That looks a bit like some sort of abuse :stuck_out_tongue:.

If you want this documented I think it'd be nice to have some terminology to distinguish the two types of labels. Something like "plain" vs "quoted". I think I'd also need to know what characters can be used in either forms. I'm assuming that inside quoted labels you can put whatever you want except unescaped `"`, although I'm not sure why one would want to use such characters for a label, hence me questioning documenting this in the first place.

Let me know what you prefer!

Julow: It's not the place to document that in details but I think a little hint for reminding the reference syntax could be useful. 

The quotes are part of the reference syntax so it's more general than just labels but it's also quircky: it exists because the `-` is used for specifying the "kind" of earch segment of the reference, and there are "kinds" that contains it too (`module-type-Foo`). Also, there's no syntax for escaping quotes, so it's really just about the `-`.

NathanReb: Ok I'll update the corresponding sections and this one as well!

doc/odoc_for_authors.mld
```diff
@@ -663,6 +663,38 @@ v}
...
+{v
+{1:some_label Some Grand Title}
+v}
+
+And then later reference it from {b.mld} as follows:
+{v
+See {!page-a.some_label}
+v}
+
+The paragraph will read "See some_label".
```
Julow, line 696: We expect "See Some Grand Title" right ?

Julow (approved):
> I realized that cross page section link don't expand the label to the section title as it does for same page section links.

This is indeed expected to be like that but it seems to work only for sections. An issue would be great :)

The section on `-` looks good to me. I think it's better to include it has many people might be bothered by that (especially that automatically-generated labels use dashes too, but let's not say that too loud, I'd hope to disallow references to them)

doc/odoc_for_authors.mld
NathanReb: If the behaviour was consistent with section links within the page yes, that's what we would expect.

NathanReb:
> This is indeed expected to be like that but it seems to work only for sections. An issue would be great :)

So do you mean it's a bug and the current behaviour is unexpected or the opposite?

Julow:
The current behavior is unexpected.

NathanReb:
I reworked the PR a bit for more clarity. I removed the part about the buggy behaviour which I have now reported in #779.

I also added a small paragraph mentioning that section with a label cannot be referenced using the section title.

doc/odoc_for_authors.mld
```diff
@@ -663,6 +663,34 @@ v}
 In this case the reference text would be "Foo Bar" so the paragraph would read
 "See Foo Bar for details".
 
+Sections with a label cannot be referenced via the section title, only
```
Julow, line 666: Currently, they can but that's a new behavior from 2.0 and might be removed or might be kept. I think we can remove this sentence until we made a decision on this.

Julow (approved):

doc/odoc_for_authors.mld
NathanReb: Do you mean it is possible on `master` right now?

I've tested this odoc 2.0.0 and they cannot be referenced or at least not with the usual syntax. My impression was that the manual should be up to date with the version of the source it lives in, i.e. `master` here.

From a user perspective I think this should be documented since I assumed otherwise by default and was a bit surprised when I tried it and saw it fail.

I can remove it if you prefer but I'm not sure the fact that it might change in the future is a good reason not to document the current behaviour and syntax.

jonludlam: Just to be clear: are you expecting to be able to reference the following section:
```
{1:label Heading}
```
by `{!label}` _and_ by `{!heading}`?

NathanReb: I was initially. Now that I know I can't it makes total sense. I'm not implying that it should behave any differently but just that the existing documentation or lack of, made assume it would accept both, hence, I think, the need for clarification.

NathanReb:
I guess this can be merged then! Thanks for the thorough review!

jonludlam:
Thanks @NathanReb !

Issue #776, opened 2021-11-10
Odoc does not support odoc code blocks within code blocks


NathanReb:
When writing an ocaml code block in odoc markup that itself contains an odoc code block, the first end of block delimiter is interpreted as closing the encompassing block.

For example:
```ocaml
(** Here is how to write odoc code blocks:
  {[
  (** Here is an ocaml code block:
    {[
    let x = 0
    ]}
  *)
  ]}
*)
```
Will not be rendered as:
![Screenshot from 2021-11-10 10-46-07](https://user-images.githubusercontent.com/7419360/141089966-740b9ece-f902-4acc-9f12-d801ddc14570.png)

Problem is, if you try to escape the characters, it is also not properly rendered:
![Screenshot from 2021-11-10 10-48-37](https://user-images.githubusercontent.com/7419360/141090294-662a5032-a94a-4c75-aba1-77f72f07f7cc.png)

It would be nice to have a way to get this to work both in `{[ ... ]}` and `{@ocaml[ ... ]}` code blocks or to have a way to escape those characters properly.

NathanReb:
CC @Julow who helped me find out about this bug!

panglesd:
This is fixed by the `{delim[ ... ]delim}` syntax!

dbuenzli:
And nicely mentioned in the cheatsheet I should add…

PR #777, opened 2021-11-10
Add extensive documentation for code fragments markup

NathanReb:
There was some previous documentation but it was hidden in the List markup section, so well hidden I didn't even know it existed before working on this PR.

This adds a specific section for code blocks, split into sub section for each type of blocks with a bit more detailed documentation on how they are formatted and actual examples for the exact syntax.

It also adds documentation on the recently added language headers.

Some of the examples in this section had to be written in verbatim blocks because of the bug reported in #776.

### DIFF
--- a/doc/odoc_for_authors.mld
+++ b/doc/odoc_for_authors.mld
@@ -263,11 +263,50 @@ and
 + item 2
 + item 3
 v}
-For inline source code style, use square brackets: [ [ ... ] ]. For longer,
-preformatted sections of code, use the enclosing tags [ {[ .. ]} ].
-For verbatim (non-source) formatted sections, use the enclosing tags [{v ... v}].
 
+{3:code_blocks Code Blocks}
 
+There are various ways of inserting code elements in your documentation.
+
+{4:inline_code_blocks Inline Code Blocks}
+
+For inline, language agnostic source code style, use square brackets: [ [ ... ] ].
+
+{[
+(** Here, [f 0] is [None] *)
+]}
+
+All the other existing ways to insert code are meant to be used for code blocks,
+not inline code fragments. Those will be formatted as a code block and will end
+the paragraph. Attempting to use them inline will trigger a warning.
+
+{4:ocaml_code_blocks OCaml Code Blocks}
+
+OCaml code blocks can be written using the enclosing tags [ {[ ... ]} ].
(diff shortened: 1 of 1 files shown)


doc/odoc_for_authors.mld
```diff
@@ -263,12 +263,79 @@ and
...
+(** This how one binds a variable in OCaml:
+    {[
+    let x = 0
+    ]}
+*)
+v}
+
+{4:language_headers Language Headers}
+
+As of [odoc.2.0] it is possible to write blocks with explicit language headers,
```
Julow, line 299: This doesn't really come with `odoc 2.0` but with odoc-parser and it's not released yet.

This new feature is independent to Odoc so we might not even set versions constraints in 2.1 to avoid making other packages not co-installable with odoc (mdx, ocamlformat).

doc/odoc_for_authors.mld
```diff
@@ -263,12 +263,79 @@ and
...
+{@markdown[
+Here is some python code:
+```python
+def f():
+  return 0
+```
+]}
+
+These blocks can be written using the enclosing tags [ {@<language>[ ... ]} ].
+The content of those block should be properly styled if odoc supports the
```
Julow, line 311: That's not implemented yet. As soon as we implement this, we'll have to release odoc-parser and add a version constraint on Odoc, fixing my previous comment.

Julow (commented):
Nice improvement !

NathanReb:
I'm not sure end users have much interest in the odoc-parser vs odoc thing. I think what they're interested in is which version of odoc will support it. In the end I assume there will be a minimal version of odoc that supports those right? What I understand from your comments is that this is not even the case on `main` atm, is that correct?

If that's the case I'm happy to leave out this part of the PR. I can move that part into a separate PR or branch somewhere so it can be merged or re-used when the feature lands. How does that sound?

Julow:
> I think what they're interested in is which version of odoc will support it.

Currently not implemented on the main branch but might be for Odoc 2.1. If you could branch out this part, we'll merge it once we have implemented that.

NathanReb:
I created #778 and stripped out the part that mentioned language headers from this PR!

Julow (approved):
Thanks for that much needed documentation !

jonludlam:
Thanks @NathanReb !

PR #780, opened 2021-11-19
Add context around errors from expansion of includes

Julow:
Re-open of https://github.com/ocaml/odoc/pull/762, which can't be re-opened after a force-push, it seems.


jonludlam:
Merging this as although it's not the complete generic solution it's definitely better than what we've currently got, and useful in diagnosing warnings.

PR #781, opened 2021-11-19
Warn on reference to ambiguous label

jonludlam:
We already warn when defining an explicit label that conflicts with another,
but this PR adds a warning when trying to reference a label that is ambiguous.

### DIFF
--- a/src/xref2/env.ml
+++ b/src/xref2/env.ml
@@ -157,6 +157,8 @@ end = struct
     with Not_found -> None
 end
 
+type 'a amb_err = [ `Ambiguous of 'a * 'a list ]
+
 type t = {
   linking : bool;
   (* True if this is a linking environment - if not,
@@ -166,6 +168,7 @@ type t = {
       (** Elements mapped by their name. Queried with {!find_by_name}. *)
   ids : ElementsById.t;
       (** Elements mapped by their identifier. Queried with {!find_by_id}. *)
+  ambiguous_labels : Component.Element.label amb_err Identifier.Maps.Label.t;
   resolver : resolver option;
   recorder : recorder option;
   fragmentroot : (int * Component.Signature.t) option;
@@ -201,6 +204,7 @@ let empty =
     ids = ElementsById.empty;
     resolver = None;
     recorder = None;
+    ambiguous_labels = Identifier.Maps.Label.empty;
     fragmentroot = None;
   }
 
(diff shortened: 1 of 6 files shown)


src/xref2/env.ml
```diff
@@ -430,14 +446,21 @@ let lookup_by_name scope name env =
...
   in
-  match ElementsByName.find_by_name scope.filter name env.elts with
-  | [ x ] as results ->
+  match
+    (ElementsByName.find_by_name scope.filter name env.elts, scope.check)
+  with
+  | ([ x ] as results), Some c -> (
+      record_lookup_results env results;
+      match c env x with
+      | Some (`Ambiguous (x, y)) -> Result.Error (`Ambiguous (x, y))
```
Julow, line 455:
```suggestion
      | Some (`Ambiguous _ as e) -> Result.Error e
```

src/xref2/ref_tools.ml
```diff
@@ -96,19 +96,40 @@ let ambiguous_ref_warning name results =
...
+    "@[<2>Multiple sections named '%s' found. Please alter one to ensure \
+     reference is unambiguous. Locations:@ %a@]"
+    name
+    (Format.pp_print_list ~pp_sep:Format.pp_force_newline pp_kind)
+    labels
+
+let ambiguous_warning name (results : [< Component.Element.any ] list) =
+  let results = (results :> Component.Element.any list) in
+  if List.for_all (function `Label _ -> true | _ -> false) results then
+    ambiguous_label_warning name (results :> [> Component.Element.any ] list)
```
Julow, line 117: This cast is unnecessary.

src/xref2/env.ml
```diff
@@ -235,18 +239,30 @@ let add_label identifier heading env =
...
+          try
+            match
+              Identifier.Maps.Label.find identifier env.ambiguous_labels
+            with
+            | `Ambiguous (x, others) -> `Ambiguous (x, comp :: others)
+          with Not_found -> `Ambiguous (l, [ comp ])
+        in
+
+        Identifier.Maps.Label.add identifier err env.ambiguous_labels
+    | Some _ -> assert false
```
Julow, line 255: To avoid this assert, we could use `lookup_by_id s_label`, which also return useful data in the ambiguous case.

Julow (approved):
Looks good :)

src/xref2/env.ml
jonludlam: I think this assert is in place of what I actually want, which is a type error. You're right, we could avoid this assert with `s_label`, but `s_label` just does:
```
(function #Component.Element.label as r -> Some r | _ -> None)
```
so I could get the same effect by having just `Some _ -> env.ambiguous_labels`. I don't like we've got a runtime instead of a compile time error here, but I like ignoring the error entirely even less :-)

src/xref2/ref_tools.ml
jonludlam: Ah yes! thanks, this was left in from when I was trying to return a different `` `AmbiguousLabel `` error (which ultimately failed, predictably in hindsight).

src/xref2/env.ml
Julow: Agree.

PR #783, opened 2021-11-23
Ignore unquoted spaces in identifiers.

### LINKED ISSUE #749: Reference syntax for operators is too restricted
```ocaml
val ( ^ ) : string -> string -> string

(** {!(^)} *)
(** {!( ^ )} *)
```
In this example, the first reference resolves but the second doesn't.

Spaces inside the parentheses is common, OCamlformat puts them.

The only way to write a reference to `( * )` is:
```ocaml
(** {!"(*)"} *)
```

panglesd:
This PR fixes #749.

It does so by ignoring whitespaces in identifiers outside of quotation marks.
So both `{!( ^ )}` and `{!( * )}` will be treated internally as respectively `{!(^)}` and `{!(*)}`. However, still neither `{!"( ^ )"}` nor `{!"( * )"}` will resolve.

It leaves the possibility to have spaces in filenames, for instance as in `{!page-"will i am"}`.

However, this PR breaks any project that have refer to a filename with spaces: `{!page-will i am}` will now be equivalent to `{!page-william}` and thus refer to `william.mld` instead of `will i am.mld`. I hope this amounts to zero projects! (Odoc's documentation is also updated.)


test/integration/simple_dune_project.t/run.t
```diff
@@ -6,7 +6,5 @@ Build the documentation of a simple Dune library.
   _build/default/_doc/_html/dune_odoc_test/Dune_odoc_test/Bar/index.html
   _build/default/_doc/_html/dune_odoc_test/Dune_odoc_test/Foo/index.html
   _build/default/_doc/_html/dune_odoc_test/Dune_odoc_test/index.html
-  _build/default/_doc/_html/dune_odoc_test/Dune_odoc_test__Bar/index.html
-  _build/default/_doc/_html/dune_odoc_test/Dune_odoc_test__Foo/index.html
```
Julow, line 10: This test is not uptodate. This diff is gone after a runtest.

Julow (approved):
Looks good :) Nice that we can still write references with spaces in them with the quote syntax (`{!"with spaces"}`).

jonludlam:
I think the failure is because we need to rebase this PR on master to get the new 'Warning/Error' prefixes on the messages. We could also do with squashing the commits at that point and perhaps adding a CHANGES entry? I keep forgetting to do this myself :-/

panglesd:
Do I understand well that I should squash the commits now? (I though it's done automatically when merging the PR)

jonludlam:
It depends on the PR - mostly I merge PRs by rebasing them rather than squashing them, as in many cases the individual commits are intended to be logically separate, but if you're happy for this to be just one commit I can use the 'squash and merge' button instead! In either case it will need rebasing first.

jonludlam:
Ah, I see you've already rebased. Thanks!

jonludlam:
Thanks!

PR #785, opened 2021-11-25
Fix the warning/error prefix

Julow:
The prefix was incorrectly computed and was always "Error" for a class of warnings and "Warning" for others.
Now, correctly take into account the warn-error option.


jonludlam:
Thanks!

PR #787, opened 2021-11-26
Add compatibility with 4.14.0

patricoferris:
For some projects I might need to build using 4.14.X and I **hastily** ported odoc to `4.14.0+trunk`. Not sure I'll have time to properly test/finish the PR but this might be useful for anyone who does pick up that work.

### DIFF
--- a/src/loader/cmi.ml
+++ b/src/loader/cmi.ml
@@ -335,7 +335,9 @@ let prepare_type_parameters params manifest =
     | Some ty ->
         let vars = Ctype.free_variables ty in
           List.iter
-            (fun v -> if get_desc v = Tvar (Some "_") then (if List.memq ty vars then tvar_none ty))
+            (fun ty -> match get_desc ty with
+              | Tvar (Some "_") -> if List.memq ty vars then tvar_none ty
+              | _ -> ())
             params
     | None -> ()
   end;
--- a/src/loader/cmi.ml
+++ b/src/loader/cmi.ml
@@ -86,12 +86,19 @@ let fresh_name base =
   done;
   !current_name
 
+let get_desc x =
+#if OCAML_VERSION >= (4, 14, 0)
+  Types.get_desc x
+#else
+  x.Types.desc
+#endif
+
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Thanks a lot! I made a PR on your fork to this branch to restore compatibility with OCaml < 4.14: https://github.com/patricoferris/odoc/pull/1

src/loader/cmi.ml
```diff
@@ -504,8 +614,8 @@ and read_object env fi nm =
         let present_fields =
           List.fold_right
             (fun (n, k, t) l ->
-               match Btype.field_kind_repr k with
-               | Fpresent -> (n, t) :: l
+               match field_kind_repr k with
+               | f when f = field_public -> (n, t) :: l
```
kit-ty-kate, line 618: I think in this particular case it’s a bit better to avoid equality
```suggestion
#if OCAML_VERSION >= (4, 14, 0)
               | Types.Fpublic ->
#else
               | Types. Fpresent ->
#endif
                 (n, t) :: l
```

src/loader/cmi.ml
```diff
@@ -258,9 +335,7 @@ let prepare_type_parameters params manifest =
     | Some ty ->
         let vars = Ctype.free_variables ty in
           List.iter
-            (function {desc = Tvar (Some "_"); _} as ty ->
-              if List.memq ty vars then tvar_none ty
-                    | _ -> ())
+            (fun v -> if get_desc v = Tvar (Some "_") then (if List.memq ty vars then tvar_none ty))
```
kit-ty-kate, line 338: This is not equavalent. The `ty` before and after is not the same one (see `as ty` above)
```suggestion
            (fun ty -> match get_desc ty with
              | Tvar (Some "_") -> if List.memq ty vars then tvar_none ty
              | _ -> ())
```

kit-ty-kate (changes requested):

patricoferris:
Thanks @kit-ty-kate for the tidy, fixed PR! Closing in favour of that :))

PR #788, opened 2021-11-27
Add compatibility with OCaml 4.14

kit-ty-kate:
This is a refactor on top of #787 from @patricoferris to make it easier to review

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+2.0.2
+-----
+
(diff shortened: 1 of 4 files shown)


jonludlam:
Thanks both! Looking good. I'll merge this ASAP so that if we can make a 2.0.2 for 4.14 compatibility release that only has this in it. Could you pop a CHANGES entry in please?

PR #795, opened 2021-12-08
Fix error in xref

jonludlam:
This was causing the module_type_alias test to produce a functor expansion with two arguments labelled '1'.

### DIFF
--- a/test/xref2/resolve/test.md
+++ b/test/xref2/resolve/test.md
@@ -1529,8 +1529,9 @@ Resolve a functor:
                       (Odoc_model.Lang.Signature.Ordinary,
                       {Odoc_model.Lang.TypeDecl.id =
                         `Type
-                          (`ModuleType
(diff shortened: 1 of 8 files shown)


test/xref2/resolve/test.md
```diff
@@ -1561,10 +1562,7 @@ Resolve a functor:
                     (Odoc_model.Lang.ModuleType.Signature
                       {Odoc_model.Lang.Signature.items =
                         [Odoc_model.Lang.Signature.Type
-                          (Odoc_model.Lang.Signature.Ordinary,
-                          {Odoc_model.Lang.TypeDecl.id = ...; doc = ...;
-                           canonical = ...; equation = ...;
-                           representation = ...});
+                          (Odoc_model.Lang.Signature.Ordinary, ...);
```
Julow, line 1565: We should perhaps increase the threshold at which it does that.

Julow (approved):
Looks good :)

PR #796, opened 2021-12-14
Better handling of substitutions

jonludlam:
The scope of the substitutions shouldn't escape the signature. So
the substitutions are only added to the environment explicitly during
the compile/link processing of signatures, not in the more general
'Env.open_signature'. Included in this commit is a test for GH issue
793 which was a manifestation of the problem this caused.

### DIFF
--- a/test/xref2/refs/refs.md
+++ b/test/xref2/refs/refs.md
@@ -730,9 +730,9 @@ let resolve_ref = resolve_ref_of_mli {|
 <!-- $MDX version>=4.08 -->
 ```ocaml
 # resolve_ref "s1" ;;
-- : ref = `Identifier (`Type (`Root (Some (`Page (None, None)), Root), s1))
+Exception: Failure "resolve_reference: Couldn't find \"s1\"".
 # resolve_ref "s1.rf1" ;;
-Exception: Failure "resolve_reference: Couldn't find \"rf1\"".
+Exception: Failure "resolve_reference: Couldn't find \"s1\"".
 # resolve_ref "M.s2" ;;
 Exception: Failure "resolve_reference: Couldn't find \"s2\"".
 # resolve_ref "M.s2.rf2" ;;
--- a/src/xref2/compile.ml
+++ b/src/xref2/compile.ml
@@ -217,16 +217,21 @@ and signature_items : Env.t -> Id.Signature.t -> Signature.item list -> _ =
               in
               (Module (r, m') :: items, env')
         | ModuleSubstitution m ->
-            std @@ ModuleSubstitution (module_substitution env m)
+            let env' = Env.open_module_substitution m env in
+            (ModuleSubstitution (module_substitution env m) :: items, env')
         | Type (r, t) -> std @@ Type (r, type_decl env t)
-        | TypeSubstitution t -> std @@ TypeSubstitution (type_decl env t)
+        | TypeSubstitution t ->
+            let env' = Env.open_type_substitution t env in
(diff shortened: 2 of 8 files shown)


odoc.opam
```diff
@@ -25,7 +25,7 @@ delimited with `(** ... *)`, and outputs HTML.
 """
 
 depends: [
-  "odoc-parser" {>= "0.9.0"}
+  "odoc-parser" {>= "0.9.0" & < "1.0.0"}
```
Julow, line 28: I don't think we should keep this. The CI will be fixed and this will break pinning Odoc.

test/xref2/refs/refs.md
```diff
@@ -730,9 +730,9 @@ let resolve_ref = resolve_ref_of_mli {|
 <!-- $MDX version>=4.08 -->
 ```ocaml
 # resolve_ref "s1" ;;
-- : ref = `Identifier (`Type (`Root (Some (`Page (None, None)), Root), s1))
+Exception: Failure "resolve_reference: Couldn't find \"s1\"".
```
Julow, line 733: It's sad not to be able to reference these. I think we want that feature back.

Julow (commented):
The module case isn't handled, I've sent a PR to your branch: https://github.com/jonludlam/odoc/pull/39
I also shortened the test case (which I needed to do to understand).

I'm not sure this is the best way to fix the problem. The crash happens during the remove operation, making me think that the problem is in the way `close_signature` is used rather than due to substitutions in particular.
Could it be that the `close_signature` function is called on a signature that is slightly different than the one passed to `open_signature` ? (eg. substitutions removed or changed ID)

jonludlam:
So, the problem is that we end up with duplicate identifiers. The way the shadowing logic works is that if the computed signature of an include contains a type/module/etc that is also declared in the including module then the one in the include is shadowed - we make a note of these and ensure that these have internal identifiers that are then hidden in the final output.

For these inline substitutions, the computed signature does not contain them - as by that point they have been substituted - the identifier only lives inside the signature. This means the shadowing logic doesn't kick in and we end up with two identical identifiers for different types. The problem manifests as a duplicate identifier when opening the signature (for which the warning is currently commented out) and in some cases trying to remove an identifier that's already been removed, which triggers the actual failure.

References to the substituted types will work within the signature - ie, where the identifier is still 'live' - but won't work from outside of the signature. In this respect it's now the same as the usual destructive substitution on a module type.

odoc.opam
jonludlam: I'd gladly revert this if there's some other way we can make ocaml-ci pass - but it's way to easy for breakage to slip in if we can't. The master branch is currently broken because we ignored the failure from ocaml-ci thinking it was just the formatting failure, when in fact there were other tests failing.

jonludlam:
Thanks for the fix and improved test @Julow ! We might also want to test module type substitution as well?

odoc.opam
Julow: I'm not sure. It's easy to check for the CI and I don't expect it to stay broken for long. I don't think it's a big deal if the build is broken on master, because we'll notice quickly. On the other hand, it's easy to forget to revert this.

Julow:
Thanks for the explanation.
I agree that the substitution shouldn't escape from signatures. I think we should still put them in the env and be careful to remove them.

jonludlam:
FTR, the substitutions _are_ put into the environment while processing the signature - just not by the function `Env.open_signature`. The helper functions `Env.open_module_substitution` etc., are called during compilation/linking of signatures.

PR #797, opened 2021-12-14
Do not drop comments attached to open

### LINKED ISSUE #745: Comment attached to an `open` is dropped
```ocaml
module M : sig end

open M
(** Comment. *)
```
The comment doesn't appear in the documentation. I think it should be treated as a floating comment.

panglesd:
This PR fixes #745 by keeping comments attached to open.

They are handled the same way as floating comments.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,10 @@ Bugs fixed
 - Fix a missing Result constructor during compile. This will cause some
   functor arguments to have different filenames (@jonludlam, #795)
 
+Additions
+- Handle comments attached to open statement as floating comment, instead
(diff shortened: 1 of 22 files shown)


jonludlam:
Very nice! Could you please add a test and a changes entry?

Julow (approved):
Looks good to me :) It renders like a floating comment, which is what we want I think.

Can you add a test ? The best place would be `test/generator/toplevel_comments.mli`.

panglesd:
Thanks @Julow and @jonludlam.

I added a test and a change entry!

Julow (approved):

PR #798, opened 2021-12-14
Add a class and a css rule for empty includes

### LINKED ISSUE #725: Empty includes should probably be dropped
If you look at [the base docs on v3](https://v3.ocaml.org/p/base/v0.14.1/doc/Base/Int64/index.html) the very first include is empty because all its contents have been shadowed by later definitions. In such cases the include should probably just be removed.

panglesd:
This PR propose a fix of #725.

It adds a class `"shadowed-include"` to the html element containing an empty include.
This way, it is easy to change how these elements are displayed without changing odoc itself, just adding css.
By default, in this PR, it is not displayed.
However, the first suggestion of @asavahista on #725, of having a special formatting for these, could be implemented as well.

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -299,6 +299,10 @@ and items ~resolve l : item Html.elt list =
       :: rest ->
         let doc = spec_doc_div ~resolve doc in
         let included_html = (items content :> any Html.elt list) in
+        let a_class =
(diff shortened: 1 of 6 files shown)


Drup (approved):
Sounds like a reasonable solution :+1:

PR #799, opened 2021-12-14
Memoise module paths

jonludlam:
The aim here is to reduce the amount of memory used when compiling long chains of aliases, which currently causes problems.


src/xref2/compile.ml
```diff
@@ -16,9 +16,9 @@ let type_path : Env.t -> Paths.Path.Type.t -> Paths.Path.Type.t =
   match p with
   | `Resolved _ -> p
   | _ -> (
-      let cp = Component.Of_Lang.(type_path empty p) in
+      let cp = Component.Of_Lang.(type_path (empty ()) p) in
       match Tools.resolve_type_path env cp with
-      | Ok p' -> `Resolved (Cpath.resolved_type_path_of_cpath p')
+      | Ok p' -> `Resolved Lang_of.(Path.resolved_type (empty ()) p')
```
Julow, line 21: Why this change ?

Julow (commented):
Could it be a global memo ? There's a lot of `empty ()` so potentially a lot of memos that aren't shared.

jonludlam:
I considered a global memo, but it's all quite contextual - so converting from the same `Lang.* -> Component.*` will often give different results - whereas the reverse, `Component.* -> Lang.*` will likely cause a lot of leakage from memoised results that are subsequently irrelevant .

src/xref2/compile.ml
jonludlam: We had two ways to convert from `Cpath.*` to `Path.*` - I deleted the ones in `Cpath` and added the memoisation to the other (in `Lang_of`).

jonludlam:
In case it wasn't very obvious from the diff, this can make a startling difference (in the pathological case the test case is testing) - the size of the `odocl` file reduced from 36 megs to 24k :-) (https://github.com/ocaml/odoc/pull/799/commits/c7b327b014caaad2fc2df90e6ea4a25532f92e3e#diff-ad34e4af41233ad07cbd0172dc46cc763e6e50a0b0d403bf371ca7b7768a4e8dL5)

Julow (approved):
Awesome !

PR #800, opened 2021-12-15
Fix tests for old versions of ocaml (<4.08)

### LINKED ISSUE #745: Comment attached to an `open` is dropped

panglesd:
This PR fixes #745 for versions of ocaml older than 4.08. It solves the failure of tests introduced in #797 for those versions.


Julow (approved):
Nice !

PR #803, opened 2022-01-11
update docs

lubegasimon:


doc/driver.md
```diff
@@ -90,7 +90,7 @@ so we can see we will need to run `odoc compile` against several `Stdlib` module
 
 This takes the `odoc` files produced during the compilation step and performs the final steps of expansion and resolution. It is during this phase that all the references in the documentation comments are resolved. In order for these to be resolved, everything that is referenced must have been compiled already, and their `odoc` files must be on the
 include path as specified by the `-I` arguments to `odoc link`. In this example, we achieve that by compiling all modules and `.mld` files before linking anything. The output of the
-link step is an `odocl` file, which is in the same path as the original `odoc` file by default.
+link step is an `odocl` file, which is in `_build/default/_doc/_odocls` by default when invoked by `dune`.
```
Julow, line 93: I think the previous sentence was correct. It's talking about where the `.odocl` files will be placed by the code in the Mdx blocks below. Nothing to do with Dune.
It says that it'll put the `.odocl` files in the same place as their corresponding `.odoc` files.

lubegasimon: >  … placed by the code in the Mdx blocks below


I am not sure of what you’re relating too. If I remember quite well, the former sentence was inspired by the Odoc rules we were adding to Dune. In this https://github.com/ocaml/dune/pull/3927 that was closed in favour of https://github.com/ocaml/dune/pull/5045, we declared that `.odocl` files are to be added in the same path as the original `.odoc` file by default, whereas in the former, the path now is `.odocls` folder, and to clarify that to the reader, I added "when invoked by `dune`”.

Julow: This is a `.md` file with Mdx blocks that execute a simple driver. The text talks about that simple driver and not about Dune and not about how users would use Odoc.

lubegasimon: The ‘Linking’ section seems to talk about a general `linking` concept, I agree no where does it mention about Dune though.

jonludlam: I agree with @Julow, this is specific to dune, so I don't think we need this change.

Issue #804, opened 2022-01-13
Bug: Resolving hidden path inside method and for `typ_class`


hhugo:
b.ml 
```ocaml
type t
class type u = object end
class type uuuuuu = object end
```

c.ml
```ocaml
class type d = object
  method e : B.t
  method f : #B.u -> unit
end
let f : B.t = assert false
let g : #B.u -> unit = fun _ -> ()
let h : B.u -> unit = fun _ -> ()
let i : < m : B.u > -> unit = fun _ -> ()
```

generated html doc
```ocaml
class type d = object
  method e : A__B.t
  method f : A__.B.#u -> unit
end
val f : B.t
val g : A__.B.u -> unit
val h : B.u -> unit
val i : < m : B.u; > -> unit
```

I've a small reproduction case in https://github.com/hhugo/odoc/commit/f369520cbae84ca977ca55a44468177bae844fae

```
$ dune build @odoc-bug/doc --cache=disabled  --force
Warning, resolved hidden path: A__B.t
```

hhugo:
The issue doesn't seem to be present on 1.5.* . It seems to be new in 2.0.*

lubegasimon:
Indeed, that looks to be a bug!

PR #809, opened 2022-01-22
Issue 804

jonludlam:
Fixes for handling class type paths.

Closes: #804


src/loader/ident_env.cppo.ml
```diff
@@ -571,18 +574,18 @@ module Path = struct
   let read_class_type env = function
     | Path.Pident id -> read_class_type_ident env id
 #if OCAML_VERSION >= (4,8,0)
-    | Path.Pdot(p, s) -> `Dot(read_module env p, s)
+    | Path.Pdot(p, s) -> `Dot(read_module env p, strip_hash s)
```
Julow, line 577: We could have a function for that (`read_type_path p s`) and a bit of documentation. Also should it be called in `Fragment.read_type` ?

Julow (approved):
Looks good :)

src/loader/ident_env.cppo.ml
jonludlam: Good point, I'll check for fragments. Do you mean a function to return `` `Dot(...) ``? Might have to have some annoying coercions, but I can see how it looks.

Julow: > Do you mean a function to return `Dot(...)?

Yes but if the type is annoying, that's not a big deal. Perhaps abstracting `strip_hash` behind a `read_type_name s` would be enough to justify it.

jonludlam: I don't think you can specify '#' in fragments, so that's probably fine.

jonludlam: and I'm not keen on the function for reading `` `Dot ``, mostly because we wouldn't want to call it from `read_module` and `read_module_type`. I'll add some documentation though.

Issue #334, opened 2019-03-07
Command for mapping identifiers to urls anchors


dbuenzli:
For allowing editors to let their user browse the documentation of a given identifier in a generated documentation set it would be nice to have a command like `odoc [-I DIR]...  uri ID [KIND]`

`odoc` would resolve the identifier in the set of `odoc` files found in the `-I DIR`  options and write one or more relative url on stdout and/or exit with a well defined exit code if the identifier cannot be found.

/cc @trefis

trefis:
Note: `ID` would actually be a path (in the sense of odoc and the ocaml typechecker).

trefis:
And, since it's not obvious, the reason why you'd want this command instead of just guessing the url based on the path is that if your path is `Foo.Bar.t` but `Foo.Bar` is an alias for `Baz` then `Foo/Bar/index.html` won't exist, and the page you want to link to will be `Baz/index.html`.

jonludlam:
I think `id` will be a reference rather than a path, but yes, I'm all for this!

PR #812, opened 2022-01-25
Map reference to url

lubegasimon:
Fixes https://github.com/ocaml/odoc/issues/334


test/xref2/map_ref_to_url.t/run.t
```diff
@@ -0,0 +1,6 @@
+g=Generates documentation using odoc for a library: 
+  $ dune build @doc
```
Julow, line 2: It's not necessary to use Dune here, the simpler the better:
```
  $ compile foo.mli
  $ odoc uri -I . Foo.t
```

Julow: It could be nice to test the error cases too, parsing and resolving errors.

src/odoc/bin/main.ml
```diff
@@ -712,6 +712,64 @@ module Odoc_error = struct
...
+  let resolve directories reference =
+    let resolver =
+      Resolver.create ~important_digests:false ~directories ~open_modules:[]
+    in
+    let reference =
+      let open Odoc_model in
+      let warnings_options =
+        { Error.warn_error = true; print_warnings = true }
+      in
+      Semantics.parse_reference reference
```
Julow, line 729: This gives bad locations for errors:
```
  ERROR: File "", line 0, characters 1-1:
  Identifier in reference should not be empty.
```

I think `Reference.parse location reference` should be used instead.

`Semantics.parse_reference` can be removed entirely, given that it's job is to actively harm. Let the wrong locations in the other call site for now.

Julow (commented):
Very nice !

I think it would be worth adding an option for specifying the base directory, for example `--prefix`, because otherwise, the user will have to do it in 99% of the cases.

src/odoc/bin/main.ml
lubegasimon: Well, that is a  good point, just that a reference passed on the command line has no location unlike compilation units and pages. As a consequence, we parsed the reference with a dummy location. In the same sense, I might be wrong and I am looking forward to a way forward :).

Julow: That's right, the location type doesn't even allow location without a filename.
There's already a dummy location with `_none_` as filename, generated by `Location_.span []`. Perhaps `Location_.pp` can be improved to detect this dummy filename and adapt the error message ?

lubegasimon: That’s right, thank you!

panglesd: So we went for the following scenario:
- Filename is empty, in this case we do not display the error location at all: indeed, I don't think cmdliner allows to get the location of its arguments in terms of character position.
- Filename is "dummy", in this case we display the location: sometimes we do not have a file but we do have a character position/length.

That's certainly not a perfect approach, it does not uses ocaml types well, but we can still update it either here or on its own PR!

src/odoc/uri.ml
```diff
@@ -0,0 +1,54 @@
...
+          | Error e -> Error (`Msg (Odoc_document.Url.Error.to_string e))
+          | Ok url ->
+              let href = url_to_string prefix url in
+              print_endline href;
+              Ok ()))
+
+let reference_to_uri_html =
+  let url_to_string prefix url =
+    let prefix = match prefix with None -> "" | Some prefix -> prefix in
+    Odoc_html.Link.(href ~resolve:(Base prefix) url)
```
Julow, line 45: If `prefix` don't already have a final `/`, it won't be added and result in wrong paths.

src/odoc/uri.ml
```diff
@@ -0,0 +1,54 @@
...
+  in
+  resolve url_to_string
+
+let reference_to_uri_latex =
+  let url_to_string prefix url =
+    let prefix = match prefix with None -> "" | Some prefix -> prefix in
+    prefix ^ Odoc_latex.Generator.Link.label url
```
Julow, line 52: I don't think `prefix` makes sense for latex. The `label` here is not a path but the equivalent of an anchor.
It's expected that the latex files are processed again, as a whole, and that the only things that survive this step are these labels and not the paths.

src/odoc/uri.ml
lubegasimon: I agree

src/odoc/bin/main.ml
Julow: Works for me :) The empty filename check is enough for now. The character position inside the string is the right value to show, certainly not the character position within the original command line. But if we want better we'd need to change the location type, let's do that later.

src/odoc/uri.ml
lubegasimon: Do you want us to make it optional or get rid of it entirely? The former makes sense to me?

Julow: I think the prefix option shouldn't be allowed at all in "latex mode". I'm not sure whether that's possible with cmdliner though.
Perhaps it can be split into two separate commands (`html-uri`, `latex-uri`) to make sure there can be different arguments and documentation.

panglesd: I don't think there is a way to tell cmdliner that a flag is allowed only when another flag is used, but I would be happy to be wrong :)

As you said in the other comment, there is also the problem of appending the right way: what happens if there is no trailing `/`, should we add one? Should we parse the prefix to look for error?

I see two possibilities:
- Either we rename `prefix` to `root_dir`, use cmdliner [dir](https://erratique.ch/software/cmdliner/doc/Cmdliner/Arg/index.html#val-dir) to check that such a directory exists, and do the "appending logic".
We also make it do nothing on latex uri (except warning the user that `root_dir` cannot be used in conjunction with latex). 
- Or, we get rid of it entirely.

panglesd: Sorry I somehow missed your comment...
> Perhaps it can be split into two separate commands (html-uri, latex-uri) to make sure there can be different arguments and documentation.
We can also check and fail if `html_root_dir` is present in Latex mode, ourselves, not relying on Cmdliner

Julow: I don't prefer this solution because the error message will be different and that's new edge cases to document.

lubegasimon: In that case, I agree to have different commands

panglesd: I was not a fan of having several command, but I have come to agree that if it's the only way to have different set of options, that sounds the better option.

And I agree to add a `/` whenever there is no trailing slash to the prefix.

src/odoc/bin/main.ml
```diff
@@ -403,45 +403,46 @@ end = struct
   let targets = Targets.(cmd, info)
 end
 
-module Odoc_uri : sig
+let reference =
```
panglesd, line 406: Even if its duplicated, I think this definition belongs more to the inside of `Odoc_html_uri` and `Odoc_latex_uri`

src/odoc/bin/main.ml
```diff
@@ -403,45 +403,46 @@ end = struct
...
+  let doc = "The reference to be resolved and whose uri to be generated." in
+  Arg.(required & pos 0 (some string) None & info ~doc ~docv:"REF" [])
+
+module Odoc_html_uri : sig
   val cmd : unit Term.t
 
   val info : Term.info
 end = struct
-  let prefix =
+  let root_dir =
```
panglesd, line 415: Just thinking, the prefix to append can also be something else that a directory (for instance a url to the online doc in v3).
So maybe a better name would be `root_uri`? :/

lubegasimon: How does `base` sound?

Julow: Indeed. In this case adding a `/` at the end of it is wrong.

src/odoc/bin/main.ml
```diff
@@ -403,6 +403,55 @@ end = struct
...
+  val cmd : unit Term.t
+
+  val info : Term.info
+end = struct
+  let root_url =
+    let doc =
+      "A string to prepend to the generated relative url. A separating / is \
+       added if needed."
+    in
+    Arg.(value & opt (some string) None & info [ "r"; "root_url" ] ~doc)
```
Julow, line 416: Other options use this style:
```suggestion
    Arg.(value & opt (some string) None & info [ "r"; "root-url" ] ~doc)
```

Julow (approved):
Looks good :)

jonludlam:
Thanks!

PR #819, opened 2022-02-04
Ignore non doc attributes at top level genrated by ppx

jorisgio:
Currently, odoc fails to extract synopsis from a signature pre-processed by a ppx. This is because ppx add a "ocaml.ppx.context" attribute at top level before the top comment.
I'm not sure the current fix is the best way to fix, since it uses string matching "ocaml.text" to continue unrolling the list of items until it finds an attribute that is a document.

### DIFF
--- a/src/loader/doc_attr.ml
+++ b/src/loader/doc_attr.ml
@@ -147,7 +147,9 @@ let extract_top_comment internal_tags ~classify parent items =
                   |> Error.raise_parser_warnings
                 in
                 (tl, ast_docs)
-            | None -> (items, []))
+            | None ->
+                let items, ast_docs = extract ~classify tl in
+                (hd :: items, ast_docs))
         | Some `Open ->
             let items, ast_docs = extract ~classify tl in
             (hd :: items, ast_docs)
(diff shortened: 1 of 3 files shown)


src/loader/doc_attr.ml
```diff
@@ -139,7 +139,7 @@ let extract_top_comment internal_tags ~classify parent items =
   let rec extract ~classify = function
     | hd :: tl as items -> (
         match classify hd with
-        | Some (`Attribute attr) -> (
+        | Some (`Attribute ({ Parsetree.attr_name = { txt = "ocaml.text" ; _ }; _ } as attr)) -> (
```
Julow, line 142: Matching this is already the responsibility of `parse_attribute`. I think what you mean is to change the `None` case below to continue iterating rather than stopping.

jonludlam: Agreed - nice find though!

jorisgio: Indeed, i missed that. Thanks !

Julow (approved):
There's some documentation about this in `doc/odoc_for_authors.mld` in section `top_comment` that should be updated (where it talks about the exception for `open` statements).

Otherwise the code looks good.

jorisgio:
> where it talks about the exception for open statements
I'm not sure i understand what you mean

Julow:
I mean that there's documentation about what is "ignored" at the beginning of a file in `doc/odoc_for_authors.mld` at line 579. It should be updated to mention that it also ignores attributes.

jonludlam:
We could also do with a CHANGES entry.

jonludlam:
Thanks @jorisgio !

PR #820, opened 2022-02-04
Fix canonical

### LINKED ISSUE #724: Issues with @canonical
There are a couple of problems with @canonical tags right now.

Firstly simplification is not working as expected. For example, `Odoc_model.Paths.Path.Resolved.Module.t` currently renders as:

Screenshot 2021-09-09 at 12 21 09

there is an associated issue where self-canonical module aliases are not expanded unless the aliased module is hidden.

jonludlam:
Fixes #724.

I'd still like to verify that paths containing canonical modules are correct, but we weren't doing that before, either. See the new test for an example of this not working.


src/xref2/tools.ml
```diff
@@ -6,11 +6,87 @@ open ResultMonad
...
+
+(* Small helper function for resolving canonical paths.
+   [canonical_helper env resolve lang_of possibilities p2] takes the
+   fully-qualified path [p2] and returns the shortest resolved path
+   whose identifier is the same as the resolved fully qualified path.
+   [resolve] is a function that resolves an arbitrary unresolved path,
+   [lang_of] turns a resolved path into a generic resolved Lang path
+   and [possibilities] is a function that, given the fully qualified
+   unresolved path, returns a list of all possible unresolved paths
+   (including the longest one) *)
```
Julow, line 66: `possibilities` should be sorted, shortest first ?

src/xref2/tools.ml
```diff
@@ -6,11 +6,87 @@ open ResultMonad
...
+  let get_identifier cpath =
+    Odoc_model.Paths.Path.Resolved.identifier (lang_of cpath)
+  in
+  match resolve p2 with
+  | None -> None
+  | Some (rp2, _) -> (
+      let fallback_id = get_identifier rp2 in
+      let resolved = filter_map resolve (possibilities env p2) in
+      let find_fn (r, _) = get_identifier r = fallback_id in
+      try Some (List.find find_fn resolved) with _ -> None)
```
Julow, line 88: `filter_map` can be avoided by doing `resolve` and `find_fn` directly during the `List.find`.

src/xref2/tools.ml
```diff
@@ -6,11 +6,87 @@ open ResultMonad
...
+   a list of shorter possibilities to test *)
+let c_mod_poss env p =
+  (* canonical module paths *)
+  let rec inner = function
+    | `Dot (p, n) -> (
+        let rest = List.map (fun p -> `Dot (p, n)) (inner p) in
+        match Env.lookup_by_name Env.s_module n env with
+        | Ok (`Module (id, m)) ->
+            let m = Component.Delayed.get m in
+            `Identifier (id, m.hidden) :: rest
```
Julow, line 30: What is this lookup doing ? Shouldn't `rest` contain all the possiblities ?

src/xref2/tools.ml
```diff
@@ -981,88 +1057,101 @@ and reresolve_module : Env.t -> Cpath.Resolved.module_ -> Cpath.Resolved.module_
...
+         with the same _unresolved_ canonical path that we're currently trying to resolve.
+
+         This is particularly important because some modules don't know they're canonical!
+         For example the module Caml in base, which is marked as the canonical path for
+         all references to the standard library in the file [import0.ml], but is itself just
+         defined by including [Stdlib].
+
+         If a module doesn't know it's canonical, it will fail the self-canonical check, and
+         therefore not necessarily be expanded. If this happens, we call [process_module_path]
+         to stick the [`Alias] constructor back on so we'll link to the correct place. *)
```
Julow, line 1092: Shouldn't this be done by the caller ? Because like this, it'll still be wrapped in a `` `Canonical `` and that makes this function part of the complicated recursive functions.

src/xref2/tools.ml
Julow: No this doesn't make sense, ignore this.

test/xref2/canonical_alias.t/run.t
```diff
@@ -0,0 +1,150 @@
...
+              "Wrapper2"
+            ]
+          }
+        },
+        "X"
+      ]
+    }
+  }
+
+This should probably not resolve at all, but that's a problem for another day. currently it resolves as Test.Wrapper3.X
```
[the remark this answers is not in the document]
jonludlam: A warning should be emitted, yes. I think we should address this in a separate PR though.

src/xref2/tools.ml
jonludlam: Yes, I'll mention that.

src/xref2/tools.ml
jonludlam: I'd prefer to leave it here so we don't spread the logic between two functions. It'd be better if we have one function to test for self-canonical modules that we can call from here and also from `link.ml`, but there's a bit of refactoring required for that.

src/xref2/tools.ml
jonludlam: But then I'd have to call resolve again. I could write a bit more of a specialised function but I opted for legibility here.

src/xref2/tools.ml
Julow: I mean the resolving could fail and the function that handles the `` `Canonical `` case can start again by calling `reresolve_module` and at the same time removing the `` `Canonical `` constructor.

jonludlam: Ah, no, the canonical is still valid, it's just that it's going to point to the aliased module rather than the module itself, as the module itself won't be expanded because it doesn't know it's canonical :-)

Julow: I see.

PR #822, opened 2022-02-05
Adjust spacing between spec items

jonludlam:
A small style fix: Make sure that the content is more obviously associated with the correct signature item.

Before:
Screenshot 2022-02-04 at 19 10 54

After:
Screenshot 2022-02-04 at 19 12 04

### DIFF
--- a/src/odoc/etc/odoc.css
+++ b/src/odoc/etc/odoc.css
@@ -432,7 +432,11 @@ pre code {
 }
 
 div.spec, .def-doc {
-  margin-bottom: 20px;
+  margin-bottom: 10px;
+}
+
+div.odoc-spec {
+  margin-bottom: 25px;
 }
 
 .spec.type .variant {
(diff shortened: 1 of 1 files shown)


src/odoc/etc/odoc.css
```diff
@@ -432,7 +432,11 @@ pre code {
 }
 
 div.spec, .def-doc {
-  margin-bottom: 20px;
+  margin-bottom: 10px;
```
Julow, line 435: The margin for `.spec` is useless. It is slightly bigger than the `margin-top` of the documentation, but this should be changed instead. It is lower than the margin for `.odoc-spec` so won't have any effect there either.
I've adjusted a bit: https://github.com/jonludlam/odoc/pull/42

Julow (approved):

jonludlam:
Thanks @Julow ! In it goes.

PR #823, opened 2022-02-09
Cmdliner constraint

jonludlam:
Cmdliner 1.1.0 has improved its API in an eventually incompatible way, deprecating the API that odoc currently uses.

Unfortunately cmdliner 1.1.0 is 4.08 and upwards only, and since odoc supports OCaml >= 4.02 we have to stay on the old API. Therefore I am constraining cmdliner to be < 1.1.0.

### DIFF
--- a/odoc.opam
+++ b/odoc.opam
@@ -27,7 +27,7 @@ delimited with `(** ... *)`, and outputs HTML.
 depends: [
   "odoc-parser" {>= "0.9.0"}
   "astring"
-  "cmdliner" {>= "1.0.0"}
+  "cmdliner" {>= "1.0.0" & < "1.1.0"}
   "cppo" {build & >= "1.1.0"}
   "dune" {>= "2.9.1"}
   "fpath"
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "@opam/alcotest": "^0.8.5",
     "@opam/astring": "^0.8.3",
     "@opam/bisect_ppx": "^2.4.1",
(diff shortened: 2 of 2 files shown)


dbuenzli:
Why don't you silence the warnings instead ? The old API is supported.

jonludlam:
Is there an easy way to disable just the deprecation warnings on Cmdliner? I don't want to do it globally.

dbuenzli:
You can disable warnings around cmdliner code with (this will of course also disable other warning, but that should be quite local): 

```
[@@@alert "-deprecated"]

[@@@alert "+deprecated"]
```

I think constraining `cmdliner` is going to make it painful for people to work on `odoc`.

I'm sorry about that. I checked with the `opam` folks and they were ok with `4.08`, I didn't think about `odoc`.

dbuenzli:
Also I never tried that I don't know if it will work but what about: 

```
module Our_cmdliner 
[@@@alert "-deprecated"]
  module Term = Cmdliner.Term
  module Arg = Cmdliner.Arg
[@@@alert "+deprecated"]
end

open Our_cmdliner
```

dbuenzli:
> We'll be supporting <4.08 for a while yet - do you have a time in mind to remove the deprecated modules?

At least not while `odoc` needs it. 

I'm really not in a hurry to remove it. The only thing I may do is hide the deprecated docs at some points. 

The cost of maintaining the old interface is [negligible](https://github.com/dbuenzli/cmdliner/blob/master/src/cmdliner_term_deprecated.ml) it mainly forwards to internal function with different defaults.

PR #824, opened 2022-02-09
Cmdliner 1.1.0 compatibility take 2

jonludlam:
Unfortunately cmdliner 1.1.0 is 4.08 and upwards only, and since odoc
supports OCaml 4.02.3 we have to stay on the old API. This commit introduces
a stub library that re-exposes the deprecated interface, supressing the
deprecation notices.

### DIFF
--- a/src/compat/compatcmdliner.ml
+++ b/src/compat/compatcmdliner.ml
@@ -6,23 +6,61 @@
    the OCaml compiler without causing deprecation alerts *)
 
 [@@@ocaml.warning "-3"]
-module Term = struct
-  include Cmdliner.Term
 
-  type info2 = Cmdliner.Term.info
+module Term = struct
+  open Cmdliner.Term
+
+  type 'a t = 'a Cmdliner.Term.t
+
+  type info = Cmdliner.Term.info
+
   let info = info
+
   let name = name
+
   let eval_choice = eval_choice
+
   let eval = eval
+
   let exit = Cmdliner.Term.exit
+
+  let ( $ ) = ( $ )
+
+  let const = const
 end
 
 module Arg = struct
-  include Cmdliner.Arg
+  open Cmdliner.Arg
 
-  type 'a converter2 = 'a Cmdliner.Arg.converter
+  type 'a conv = 'a Cmdliner.Arg.conv
+
   let env_var = env_var
-  let pconv = pconv
-end
-[@@@ocaml.warning "+3"]
 
+  let pconv = pconv
+
+  let string = string
+
+  let value = value
+
+  let opt = opt
+
+  let opt_all = opt_all
+
+  let info = info
+
+  let some = some
+
+  let flag = flag
+
+  let required = required
+
+  let pos = pos
+
+  let file = file
+
+  let bool = bool
+
+  let ( & ) = ( & )
+end
+
+[@@@ocaml.warning "+3"]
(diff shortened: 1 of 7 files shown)


src/compat/compatcmdliner.ml
```diff
@@ -0,0 +1,28 @@
...
+(* Cmdliner 1.1.0 has deprecated the 'traditional' Term and Arg modules,
+   but is only available for OCaml 4.08 and above. This compatibility
+   module will work on 1.0.4 and 1.1.0 for all supported versions of
+   the OCaml compiler without causing deprecation alerts *)
+
+[@@@ocaml.warning "-3"]
+module Term = struct
+  include Cmdliner.Term
+
+  type info2 = Cmdliner.Term.info
```
dbuenzli, line 12: Do you really have to rename the type ?

src/compat/compatcmdliner.ml
```diff
@@ -0,0 +1,28 @@
...
+  let name = name
+  let eval_choice = eval_choice
+  let eval = eval
+  let exit = Cmdliner.Term.exit
+end
+
+module Arg = struct
+  include Cmdliner.Arg
+
+  type 'a converter2 = 'a Cmdliner.Arg.converter
```
dbuenzli, line 23: You can simply use [`Arg.conv`](https://erratique.ch/software/cmdliner/doc/Cmdliner/Arg/index.html#type-conv) in your code, it's in since v1.0.0 and not deprecated.

src/compat/compatcmdliner.ml
jonludlam: Hrm, maybe not actually. Originally I just 'included' the whole of Cmdliner.Term/Arg, but that didn't remove the warnings, so then I just redeclared them, which did. Unfortunately 4.07 complained about having 2 types the same name, so I renamed. Instead I'll try getting rid of the include and just export a more minimal module.

dbuenzli: Also the `module Term = Cmdliner.Term [@ocaml.warning "-3"]` trick you linked to didn't work ?

dbuenzli: I see. That being said it seems the use of cmdliner is quite local. I wonder if it wouldn't be easier to disable the warnings after the `open Cmdliner` and call it a day. 

```
> git grep Cmdliner
doc/deps.mld:{2 Cmdliner}
doc/deps.mld:{!modules:Cmdliner}
src/odoc/bin/main.ml:open Cmdliner
src/odoc/bin/main.ml:  val extra_args : args Cmdliner.Term.t
test/odoc_print/odoc_print.ml:open Cmdliner
```

dbuenzli:
Taking a step back I could have defined my own `cmdliner-deprecated` alert. 

This would have allowed you to simply disable that alert. But the idea of using [@@ocaml.deprecated] (which is synonym for `alert deprecated`) is that these things would eventually get automatically lifted in the doc generation. 

I think the story on how we can deprecate stuff in our libraries is completely missing in OCaml at the moment.

src/compat/dune
```diff
@@ -0,0 +1,3 @@
+(library
```
Julow, line 1: This library could be better defined in `src/odoc` to not suggest the library is public. Perhaps in a sub directory: `src/odoc/compat`

Julow (approved):
I think this is the best approach of the two. Of course it's more burden if we need more cmdliner features in the future but it's not the first time we have complicated compatibility code.

PR #830, opened 2022-02-18
Improve and simplify the layout for datatypes, with css

panglesd:
This is the continuation of #673 by @Drup.

I added css to his modification of the layout. I also dropped the commit making the tag a `div` or a `span` depending on the number of paragraph: There might be single long paragraph, or two very small paragraph.
Instead, this is treated using css only (with `flex-wrap`).

You can find [here](https://choum.net/panglesd/parsetree_test2/ppxlib/Astlib/Ast_413/Parsetree/index.html), as an example, the output on a `Parsetree` doc where comments have been (partly) odocified, as this module contains lots of big variant and product with documentation.

Hopefully, this implements #614.

### DIFF
--- a/src/odoc/etc/odoc.css
+++ b/src/odoc/etc/odoc.css
@@ -419,8 +419,8 @@ pre code {
   padding: 0.35em 0.5em;
 }
 
-.def-doc {
-  margin-bottom: 10px;
+li:not(:last-child) > .def-doc {
+  margin-bottom: 15px;
 }
 
 /* Spacing between items */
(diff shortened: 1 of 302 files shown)


src/odoc/etc/odoc.css
```diff
@@ -419,28 +419,44 @@ pre code {
...
-  margin: 0;
-  font-style: italic;
+
+.spec.type .record > .def-doc, .spec.type .variant > .def-doc {
+  min-width:50%;
+  padding-left: 22px;
+  margin-left: 10%;
+  margin-right: 0;
+  border-radius: 3px;
+  flex-grow:1;
```
Julow, line 458: I'm not sure this has any effect. The `min-width` idea already do the "column" layout and this can only grow the empty part of the div into empty space.

src/odoc/etc/odoc.css
```diff
@@ -419,28 +419,44 @@ pre code {
...
 }
-.spec.type .record p {
-  margin: 0;
-  font-style: italic;
+
+.spec.type .record > .def-doc, .spec.type .variant > .def-doc {
+  min-width:50%;
+  padding-left: 22px;
+  margin-left: 10%;
+  margin-right: 0;
```
Julow, line 456: I think it has no effect.

panglesd: Yes, it has one, just try to disable it in the demo.

Without it it would look like this:

image

(which is also acceptable)

src/odoc/etc/odoc.css
panglesd: I would prefer to keep it: If some css theme want to add a background color (I tried that first), it will look better with `flex-grow: 1`. Of course, the css theme can also add this line, but that requires the author to dig into our css file deeper...

Julow: I see, thanks. I agree it's better if the element grows to the right.

jonludlam:
Thanks!

PR #831, opened 2022-02-18
Add support for OCaml 5.00

kit-ty-kate:
Requires https://github.com/ocaml-doc/odoc-parser/pull/6


kit-ty-kate:
> Can we continue with the old API with the deprecated alerts disabled ?

It is not possible. The old API has been entirely removed in 5.0

jonludlam:
Closing in favour of #873

Issue #829, opened 2022-02-18
Spurious space around parentheses


dbuenzli:
It seems that the merge of https://github.com/ocaml/odoc/pull/782 (available in the latest odoc) was done with little attention to typographic details. 

New spurious spaces are being inserted around parentheses in some type definitions here are a few examples: 

```
val pf : Format.formatter -> ( 'a, Format.formatter, unit ) format -> 'a
val map : ( 'a -> 'b ) -> 'a list -> 'b list
```

In other contexts the parentheses are fine e.g.: 

```
val exn_backtrace : (exn * Printexc.raw_backtrace) t
val assoc : 'a -> ('a * 'b) list -> 'b
```

dbuenzli:
Fixed by #832, thanks @panglesd !

PR #832, opened 2022-02-21
Fix spurious space around parentheses

panglesd:
This fixes the spurious space around parentheses reported in #829:
``` ocaml
val pf : Format.formatter -> ( 'a, Format.formatter, unit ) format -> 'a
val map : ( 'a -> 'b ) -> 'a list -> 'b list
```
are now rendered as `ocamlformat` would, that is:
``` ocaml
val pf : Format.formatter -> ('a, Format.formatter, unit) format -> 'a
val map : ('a -> 'b) -> 'a list -> 'b list
```


Julow:
And you need to promote the tests.

jonludlam:
Thanks!

Issue #835, opened 2022-02-23
`Not_found` error in `Lang_of`


jonludlam:
See e.g. https://docs.ci.ocaml.org/job/2022-02-21/204225-voodoo-do-5fbed0

odoc: internal error, uncaught exception:
      Failure("Not_found: Hash/31422")
      Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
      Called from Odoc_xref2__Lang_of.Path.resolved_module.f in file "src/xref2/lang_of.ml", line 149, characters 15-70
      Called from Odoc_xref2__Lang_of.Path.resolved_module.f in file "src/xref2/lang_of.ml", line 158, characters 42-64
      Called from Odoc_xref2__Ref_tools.M.of_component in file "src/xref2/ref_tools.ml", line 179, characters 27-61
      Called from Odoc_xref2__Ref_tools.LP.of_element in file "src/xref2/ref_tools.ml", line 498, characters 8-26
      Called from Odoc_xref2__Ref_tools.resolve_reference_dot in file "src/xref2/ref_tools.ml", line 711, characters 2-43
      Called from Odoc_model__Error.catch_warnings.(fun) in file "src/model/error.ml", line 87, characters 18-22
      Called from Odoc_model__Error.with_ref in file "src/model/error.ml", line 65, characters 12-16
      Re-raised at Odoc_model__Error.with_ref in file "src/model/error.ml", line 70, characters 4-11
      Called from Odoc_xref2__Link.comment_inline_element in file "src/xref2/link.ml", line 178, characters 12-45
      Called from Odoc_xref2__Lookup_failures.with_ref in file "src/xref2/lookup_failures.ml", line 13, characters 10-14
      Called from Odoc_xref2__Lookup_failures.with_location in file "src/xref2/lookup_failures.ml", line 79, characters 6-69
      Called from Odoc_xref2__Link.with_location in file "src/xref2/link.ml", line 255, characters 14-73
      Called from Stdlib__List.map in file "list.ml", line 92, characters 20-23
      Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
      Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
      Called from Odoc_xref2__Link.comment_nestable_block_element in file "src/xref2/link.ml", line 207, characters 34-54
      Called from Odoc_xref2__Lookup_failures.with_ref in file "src/xref2/lookup_failures.ml", line 13, characters 10-14
[...]

PR #836, opened 2022-02-23
Don't reuse ident_map in Env.open_signature

jonludlam:
Fixes: #835


Julow (approved):
Makes sense! We are creating new empty maps everywhere else.
Do you have a small reproduction?

jonludlam:
Sadly not! I only got the stack trace when I installed the _exact_ same set of packages as was being run in voodoo and then running voodoo over it - running using `dune build @doc` in that same switch failed to cause the error. Honestly I'm not 100% sure this is the fix for the actual problem, but it at least removes a possible source of it.

PR #837, opened 2022-02-25
Promote alert tests

jonludlam:


Julow (approved):
Yes! Not sure how the previous output was possible?

jonludlam:
I think it's due to a (fixed) bug in the html renderer - the way to specify the class attribute is via the `Html.a_class` function, which takes a list of classes - the intent being that you specify each of your classes in the one attribute returned by this. However, it's possible to use this more than once on an individual element - ie:

```
Html.span ~a:[ Html.a_class [ "foo"; "bar" ]; Html.a_class ["baz"]] ...
```
which we were previously doing. @Drup fixed this as part of the changes for datatype rendering. I only know this because `Tyxml_js` behaves slightly differently so when I was doing some js_of_ocaml odoc rendering I had to modify it, but didn't get around to making the PR :-)

PR #839, opened 2022-02-28
Fix body margin on mobile

### LINKED ISSUE #838: CSS regression: large margin in mobile mode makes content unreadable
"recently" (odoc 2.0.2, odoc 2.1.0)
html pages viewed in mobile mode keep the very large margin of the desktop mode and hence become unreadable, see screenshot:
Screenshot_20220225_235043

panglesd:
This small PR fixes #838.

Here is a screenshot of the new layout on mobile:

image

### DIFF
--- a/src/odoc/etc/odoc.css
+++ b/src/odoc/etc/odoc.css
@@ -676,9 +676,8 @@ td.def-doc *:first-child {
 /* Mobile adjustements. */
 
 @media only screen and (max-width: 95ex) {
-  .odoc-content {
-    margin: auto;
-    padding: 2em;
+  body.odoc {
+    margin: 2em;
   }
   .odoc-toc {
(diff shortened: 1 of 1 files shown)


src/odoc/etc/odoc.css
```diff
@@ -676,9 +676,12 @@ td.def-doc *:first-child {
...
 
 @media only screen and (max-width: 95ex) {
+  body.odoc {
+    margin: 2em;
+  }
   .odoc-content {
-    margin: auto;
-    padding: 2em;
+    margin: 0;
+    padding: 0;
```
Julow, line 684: These are divs and it's not overriding anything, so you can drop the entire rule.

Julow (approved):
This works !

PR #841, opened 2022-03-01
Fix stop-comment when it follows the top-comment

Julow:
The commit ae62263a914d fixed the handling of stop-comments when it is at the beginning of a signature. For example, it's no longer needed to add an empty comment in https://github.com/dbuenzli/affect/blob/815b99ebf9d4f0c2921886fab76d275f982c0340/src/fiber.mli#L121

But it also added a new bug when the stop-comment follows a top-comment. Both `foo` and `bar` were hidden in this example:

```ocaml
(** Top-comment *)

(**/**)

val foo : int

(**/**)

val bar : int

(**/**)

val baz : int
```

This PR fixes that and adds tests.


jonludlam:
Thanks @Julow !

PR #843, opened 2022-03-03
Random internal idents

jonludlam:
A hacky fix for an issue where we had clashing hidden identifiers.

### DIFF
--- a/src/model/names.ml
+++ b/src/model/names.ml
@@ -43,6 +43,10 @@ end
 
 let internal_counter = ref 0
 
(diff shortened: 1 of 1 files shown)


Julow (commented):
Perhaps internal idents could be tied to a `Root` ? So we could have an assert to detect internal idents that are out of context.

jonludlam:
Let's not do this. This was initially submitted because this hacky fix worked, but it actually doesn't work, it turns out.

Issue #574, opened 2021-01-28
Comments describing class inheritance are not rendered 


jonludlam:

Julow:
There's a PR open, https://github.com/ocaml/odoc/pull/844

PR #844, opened 2022-03-04
Handle comments attached to class constraint and inherit

Julow:
Fix https://github.com/ocaml/odoc/issues/574

The changes in the `cmti` loader is straightforward however it's not fully implemented for other inputs:

- for `cmt`s, the constraints are not read from the typedtree and I haven't made the effort to associate the comments correctly. Comments attached to constraints will appear as standalone comments anywhere in the class.
- for `cmi`s, it seems that these comments are lost ?

Most of the diffs is because of boilerplate due to changing some types in the model.


jonludlam (approved):
This all looks good, and surprisingly doesn't require any rebasing!

jonludlam:
Actually in order to get the tests run it did require rebasing :-) I've split the classes test in two as well as it was failing on OCaml < 4.08

PR #847, opened 2022-03-10
Fix css of layout for datatypes

panglesd:
This PR fixes the css introduced in #830:

It find it more logical that odoc comments always have a `--main-background` color, even if they are inside a `spec` (with background `--spec-summary-background`).

I think it makes it easier to read large record fields with many comments attached to fields.

Compare the two images:
image
with
image

You can find complete examples of the [old style](https://choum.net/panglesd/parsetree_with_comments_old/html/compilerlibref/Parsetree/index.html) and [new style](https://choum.net/panglesd/parsetree_with_comments/html/compilerlibref/Parsetree/index.html) for this file.

(Maybe the thin blue bar could be removed now that the background is white... what do you think?)

### DIFF
--- a/src/odoc/etc/odoc.css
+++ b/src/odoc/etc/odoc.css
@@ -451,11 +451,12 @@ div.odoc-spec,.odoc-include {
 
 .spec.type .record > .def-doc, .spec.type .variant > .def-doc {
   min-width:50%;
-  padding-left: 22px;
+  padding: 0.25em 0.5em;
   margin-left: 10%;
   border-radius: 3px;
   flex-grow:1;
-  border-left: 0.1em solid var(--spec-summary-border-color);
+  background: var(--main-background);
+  box-shadow: 2px 2px 4px lightgrey;
 }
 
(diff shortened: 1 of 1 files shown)


src/odoc/etc/odoc.css
```diff
@@ -451,11 +451,13 @@ div.odoc-spec,.odoc-include {
 
 .spec.type .record > .def-doc, .spec.type .variant > .def-doc {
   min-width:50%;
-  padding-left: 22px;
+  padding-left: 18px;
```
Julow, line 454: If you remove the blue border, a smaller padding would be better: (the same as `.spec`)
```suggestion
  padding: 0.35em 0.5em;
```

Julow (approved):
That's better !

But I find it ugly to have both the blue border and the shadow, let's remove the border.

leviroth:
Could the doc comments be consistently placed below the constructor? I find it awkward to read when some of the comments are below and some of them are off to the right.

Issue #759, opened 2021-10-19
Spurious reasonml tagging in `.mld` files


dbuenzli:
For some obscure reason some of the code fences in `.mld` files get tagged as reasonml. For example the last code fence of this section: 

https://erratique.ch/software/brr/doc/web_page_howto.html#running

Here's the source:

https://github.com/dbuenzli/brr/blob/fc66944ba44bdd2fca6c35e8c82926e4890c7577/doc/web_page_howto.mld#L150

dbuenzli:
In fact there's no tagging in the output. 

This looks like an issue with `highlight.pack.js`'s automatic detection. 

`odoc` should likely classify`{[…]}` odoc output as `ocaml`.

PR #848, opened 2022-03-11
Handle language tag on code blocks

Julow:
Pass the language tag through the model and to highlightjs in the html generator. The value is ignored in other backends.
Code blocks with no language specified are rendered as before (highlighted).

Highlightjs should recognize `plain`, `plaintext` and `text` to disable highlighting but perhaps we should have a string specific to Odoc that we can document and handle specifically ?

The version of highlightjs vendored only support one language (OCaml) and is not updated in this PR.

### DIFF
--- a/odoc.opam
+++ b/odoc.opam
@@ -25,7 +25,7 @@ delimited with `(** ... *)`, and outputs HTML.
 """
 
 depends: [
-  "odoc-parser" {>= "0.9.0"}
+  "odoc-parser" {>= "1.0.0"}
   "astring"
   "cmdliner" {>= "1.0.0"}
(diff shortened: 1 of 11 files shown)


dbuenzli:
> Code blocks with no language specified are rendered as before (highlighted).

I think they should rather default to `ocaml`,  that's a reasonable default for the `ocamldoc` language.

Also this fixes #759.

jonludlam:
I think we should address changing of currently existing defaults in a separate issue/PR.

dbuenzli:
That's not really changing a default, that's actually getting on par with `ocamldoc` generated docsets most of which always used the `--colorize-code` option including the stdlib one.

PR #853, opened 2022-05-24
Bug fixes

jonludlam:
A set of fairly self-contained bug fixes found over the past few weeks.


test/generators/html/Recent-X.html
```diff
@@ -28,8 +28,7 @@ <h1>Module <code><span>Recent.X</span></code></h1>
     <div class="spec type anchored" id="type-t">
      <a href="#type-t" class="anchor"></a>
      <code><span><span class="keyword">type</span> t</span>
-      <span> = 
-       <span>int <a href="Recent-Z-Y-X.html#type-t">Z.Y.X.t</a></span>
+      <span> = <span>int <a href="Recent-Z-Y-X.html#type-t">L.X.t</a></span>
```
Julow, line 31: Is this https://github.com/ocaml/odoc/issues/594 ?

src/html/generator.ml
```diff
@@ -295,29 +295,34 @@ and items ~resolve l : item Html.elt list =
...
           in
           match status with
           | `Inline -> included_html
           | `Closed -> details ~open':false
           | `Open -> details ~open':true
           | `Default -> details ~open':!Tree.open_details
         in
-        let inc = [ Html.div ~a:[ Html.a_class a_class ] (doc @ content) ] in
-        (continue_with [@tailcall]) rest inc
+        (continue_with [@tailcall]) rest content
```
Julow, line 325: We loose the comment attached to the `include`. (the code `doc @ _` is not in every branches, relevant test is `test/generators/html/Toplevel_comments-Include_inline'.html`)

test/xref2/shadow.t/run.t
```diff
@@ -0,0 +1,8 @@
+This test contains a carefully crafted couple of modules that cause a clash of names in the current shadowing code.
+If the shadowing code is working correctly this should not crash
+
+  $ ocamlc -c -bin-annot -I . c.mli
+  $ ocamlc -c -bin-annot -I . d.mli
+  $ odoc compile -I . c.cmti
+  $ odoc compile -I . d.cmti
```
Julow, line 7: Should also look at the result with `odoc_print`.

Julow (commented):
Nice!

test/generators/html/Recent-X.html
jonludlam: No, this is a consequence of https://github.com/ocaml/odoc/pull/853/commits/277d8d05500ef124e274f6057d43b7a106e89ee8

src/html/generator.ml
jonludlam: Ah! Good point thanks!

test/xref2/shadow.t/run.t
jonludlam: Could do! I didn't put much thought into this - it was an example provided by Luke M that was failing in a nasty way, so I was satisfied that it didn't die horribly after this PR.

test/generators/html/Recent-X.html
Julow: Thanks.

test/xref2/shadow.t/run.t
jonludlam: On reflection I don't think I'll do this at this point, that it doesn't crash horribly is sufficient for now.

jonludlam:
In it goes :-)

PR #861, opened 2022-06-08
Set correct default for code blocks (closes #759).

### LINKED ISSUE #759: Spurious reasonml tagging in `.mld` files

dbuenzli:
Everything is in the title.


jonludlam:
My initial worry was for the world of reason/rescript -- that there may well be code for which the autodetection currently works fine, and that forcing it to be `ocaml` be default may well be the wrong choice. I also don't know how important rescript compatibility is given I understand their desire is to create a [different tool](https://rescript-association.org/projects/2020/rescript-programming-language). I'm not really plugged into that side of things though, so perhaps @ryyppy or @ostera might be able to comment?

dbuenzli:
> My initial worry was for the world of reason/rescript -- that there may well be code for which the autodetection 
currently works fine, and that forcing it to be `ocaml` be default may well be the wrong choice.

Right. I guess if that proves problematic to them they can certainly invest developer time to make a PR that allows to set the default with an environment variable (I'm not following but I heard the whole reasonml mod had basically forked and self destructed at some point).

ryyppy:
I think setting it to ocaml by default is fine. As stated correctly, ReScript is exploring other formats for their docstrings / doc generation, and Reason users are now pretty familiar with the OCaml infrastructure anyways.

jonludlam:
That's great, thanks @ryyppy !

jonludlam:
And thanks @dbuenzli :-)

PR #862, opened 2022-06-08
Remove impossible cases in the HTML generator (links within links)

favonia:
Nested links are impossible and the relevant code should be removed.

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -101,12 +101,6 @@ let rec internallink ~config ~emph_level ~resolve ?(a = []) (t : InternalLink.t)
       let elt = (elt :> phrasing Html.elt) in
       [ elt ]
 
-and internallink_nolink ~emph_level
-    ~(a : Html_types.span_attrib Html.attrib list) (t : InternalLink.t) =
-  match t with
-  | Resolved (_, content) | Unresolved content ->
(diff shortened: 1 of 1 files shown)


dbuenzli:
Could you please provide an example of what this changes (`ocamldoc` and corresponding `html` markup). Changing the markup is never innocuous for stylesheets.

dbuenzli:
Could you please provide an example of what this changes (`ocamldoc` and corresponding `html` markup). Changing the markup is never innocuous for stylesheets.

favonia:
@dbuenzli I missed the following code that would replace nested links with `` `Emphasis``. So, it seems there's no way to trigger this. So, this PR is at most a tiny style clean-up. What should I do in this case?

https://github.com/ocaml/odoc/blob/b7a98282f055198a96ee1976ff939f058d6aa881/src/model/semantics.ml#L176-L185

dbuenzli:
> So, it seems there's no way to trigger the function; this PR is at most a tiny style clean-up. What should I do in this case?

`assert false`

favonia:
Done.

favonia:
Any plan to merge this PR?

jonludlam:
Sorry for the delay in this, I've rebased and will merge there.

Issue #756, opened 2021-10-15
Support references in code blocks


MisterDA:
I noticed that the Lwt documentation has doc comments like this:

    There are straightforward ways to make promises canceled. One could create a
    promise that {e starts out} canceled, with {!Lwt.fail}[ Lwt.Canceled]. It's
    also possible to {e make} a promise canceled through its resolver, by
    calling {!Lwt.wakeup_later_exn}[ r Lwt.Canceled].

The intent is to have a little code example with a reference to the
function used. The output however isn't very nice:

    <p>There are straightforward ways to make promises canceled. One
    could create a promise that <em>starts out</em> canceled, with
    <a href="index.html#val-fail"><code>Lwt.fail</code></a><code>Lwt.Canceled</code>.
    It's also possible to <em>make</em> a promise canceled through its resolver, by calling
    <a href="index.html#val-wakeup_later_exn"><code>Lwt.wakeup_later_exn</code></a>
    <code>r Lwt.Canceled</code>.</p>

It would be much nicer to merge the consecutive `<code>` tags, and to
be able to embed references in these short code blocks. Depending on
the document format or the CSS, it can get very ugly. Is there a way
to achieve that? If not, could odoc add support for it?

Thanks.

panglesd:
This could be achieved (in code spans, not code blocks) by having a `{c ...}` (`c` for codespan) which renders as a codespan but contains odoc inline elements instead of plain text.

favonia:
@panglesd This is a great idea! Personally I prefer `{code ...}` over `{c ...}` for several reasons: (1) for me it is easier to remember and (2) it matches the HTML tag `<code>`, just like `{ul/ol/i/b... ...}` and (3) we can still add `{c ...}` later on.

PR #863, opened 2022-06-08
Annotate code that is following or followed by other code

### LINKED ISSUE #696: Collapse padding between consecutive code fragments
I have been looking for ways to add links to identities in code fragments, for example `f` in `f arg`. The best workaround I found is `{!val:f}[ arg]`, but the default theme unfortunately adds padding between `{!val:f}` and `[ arg]`.

~~Proposed workaround: annotate code using CSS classes (#863)~~

Based on the discussion in #863, we should come up with a cleaner solution.

favonia:
Closes #696. The goal of this PR is to make
```ocaml
(** [f][ ][x] *)
```
look like
```ocaml
(** [f x] *)
```
with the default theme. The motivation is to write `{!val:f}[ x]` so that `f` is linked but the text still looks like `[f x]`.

## Technical details

This PR annotates adjacent `<code>` phrases with CSS classes `following-code` and `followed-by-code` so that one can turn off the padding between `<code>` via CSS.  The comment `[f][ ][x]` will generate something like this:
```html
<code class="followed-by-code">f</code><code class="following-code followed-by-code"> </code><code class="following-code">x</code>
```
and then the padding can be turned off using the CSS selectors `code.following-code` and `code.followed-by-code`. See the changes to `src/odoc/etc/odoc.css`. The annotation algorithm does not change the asymptotic complexity and I did not notice obvious slowdown. However, a serious benchmarking might still be useful.

## Caveats

1. Due to https://github.com/ocsigen/tyxml/issues/287, `tyxml` would insert spaces when `indent` is true, but it's no longer `odoc`'s fault.
2. I did not check how other backends can benefit from the new annotations, but at very least it would not be worse. The LaTeX backend, for example, could benefit from them.

### DIFF
--- a/src/document/types.ml
+++ b/src/document/types.ml
@@ -166,6 +166,9 @@ let rec last = function
   | [ x ] -> x
   | _ :: xs -> last xs
 
+(* Checking whether an Inline.desc starts with or ends with preformatted text.
+   This is only an approximation as we did not check whether the text is empty
+   [Text ""] or the styled inline is empty [Style (_, [])]. *)
 let rec is_inline_preformatted =
   let open Inline in
   function
--- a/src/document/types.ml
+++ b/src/document/types.ml
@@ -35,9 +35,11 @@ and Inline : sig
 
   type href = string
 
+  type preformatted = { begin_ : bool; end_ : bool }
+
   type t = one list
 
-  and one = { attr : Class.t; desc : desc }
+  and one = { attr : Class.t; preformatted : preformatted; desc : desc }
 
   and desc =
     | Text of string
@@ -159,6 +161,34 @@ and Page : sig
 end =
   Page
 
-let inline ?(attr = []) desc = Inline.{ attr; desc }
+let rec last = function
+  | [] -> invalid_arg "last"
+  | [ x ] -> x
+  | _ :: xs -> last xs
+
+let rec is_inline_preformatted =
+  let open Inline in
+  function
+  | Text _ | Linebreak -> { begin_ = false; end_ = false }
+  | Entity _ | Source _ -> { begin_ = true; end_ = true }
+  | Styled (_, is) | Link (_, is) -> is_inline_list_preformatted is
+  | InternalLink il -> is_internallink_preformatted il
+  (* Ideally, the markup should be parsed *)
+  | Raw_markup _ -> { begin_ = false; end_ = false }
+
+and is_inline_list_preformatted = function
+  | [] -> { begin_ = false; end_ = false }
+  | l ->
+      {
+        begin_ = (List.hd l).preformatted.begin_;
+        end_ = (last l).preformatted.end_;
+      }
+
+and is_internallink_preformatted = function
+  | Resolved (_, is) | Unresolved is -> is_inline_list_preformatted is
+
+let inline ?(attr = []) desc =
+  let preformatted = is_inline_preformatted desc in
+  Inline.{ attr; preformatted; desc }
 
 let block ?(attr = []) desc = Block.{ attr; desc }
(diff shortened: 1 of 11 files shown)


dbuenzli:
> The motivation is to write `{!val:f}[ x]` so that `f` is linked but the text still looks like `[f x]`.

I don't understand. This is already the case.

favonia:
> > The motivation is to write `{!val:f}[ x]` so that `f` is linked but the text still looks like `[f x]`.
> 
> I don't understand. This is already the case.

The current default CSS theme contains this fragment:
```css
p code,
li code {
  background-color: var(--li-code-background);
  color: var(--li-code-color);
  border-radius: 3px;
  padding: 0 0.3ex;
}
```
which introduces a `0.3ex` padding on both sides of the code. Here's a real example from one of the libraries I wrote (except that I darkened the background for clarity):

image

dbuenzli:
> The current default CSS theme contains this fragment:

Well by all means fix the css (for example `odig`'s default stylesheet doesn't suffer from these defects) but I don't think there's the need to complicate the HTML generator for that.

favonia:
I see. I'm not sure how people want to change the default theme, then. I don't think it's technically possible to select `<code>` that is adjacent to other `<code>` using only CSS. Therefore, either we should avoid the padding altogether (e.g., the themes used by `odig` and the OCaml website v3 did that) or the CSS theme would need some help. On the other hand, I don't feel the code is that complicated---it's mostly just passing context along. In case it helps, I could further minimize the PR by making `context` an optional argument.

dbuenzli:
To be clear your classes seem to be reinventing the [`+` CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator).

favonia:
> To be clear your classes seem to be reinventing the [`+` CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator).

`+` is close but I don't think this would work because, for example, the `<code>` in `{!val:f}` is wrapped inside `<a>` and thus the inner code is not a sibling of the `<code>` generated by `[ x]`. In general there could be arbitrarily many layers of styles (bold, italic, etc.). That said, maybe some CSS guru can help us out.

Octachron:
That sounds like a quite complicated temporary workaround to avoid implementing references inside code block (e.g `[{!val:f} x]`). Is that temporary workaround worth the implementation and future-compatibility cost?

dbuenzli:
I may not be a CSS "guru" but I have lost enough useless hours of my life into it to tell you that this: 

> ```
> <code class="followed-by-code">f</code><code class="following-code followed-by-code"> </code><code class="following-code">x</code>
> ```

is the beginning of a nightmare. 

If possible wouldn't adding a class on the `a` indicating that this is an id reference and/or swap the `<a>` and `<code>` be a simpler option ?

Also, @octachron's comment.

favonia:
Fair enough. I think we can use `+` to do half of it, but I don't know how to control the padding before a series of code phrases. The issue is that there's no adjacent sibling combinator _in the opposite direction_ as far as I know.

And yes, supporting references within code (but still rendering records as they are) would be wonderful!

favonia:
Alright. I am closing this while searching for other solutions. Just to clarify, do the developers prefer links within code (which might mess up syntax highlighting...?) or magical swapping of `<a>` and `<code>`?

dbuenzli:
> That sounds like a quite complicated temporary workaround to avoid implementing references inside code block (e.g `[{!val:f} x]`). Is that temporary workaround worth the implementation and future-compatibility cost?

Just FTR the issue that tracks that is #756

PR #867, opened 2022-06-10
Fix bogus doc (closes #852).

### LINKED ISSUE #852: Docs: bad example of non-special comment between element and special comment
["`odoc` for authors"](https://ocaml.github.io/odoc/odoc_for_authors.html), in the large block of example syntax copied from the OCaml manual, has this section:

```ocaml
(* Hello, I'm a simple comment :-) *)
exception My_exception of (int -> int) * int
(** Comment for exception My_exception, even with a simple comment
  between the special comment and the exception.*)
```

This obviously doesn't make sense. It looks like the special comment has been moved from its position in the OCaml manual.

I'd just fix it, but I wanted to check: is this getting run through ocamlformat at some point? If so, that may interfere with the stability of the comment placement, and so it might not be desirable to use ocamlformat in this specific case.

dbuenzli:
Everything is in the title.

### DIFF
--- a/doc/odoc_for_authors.mld
+++ b/doc/odoc_for_authors.mld
@@ -125,8 +125,8 @@ comment.
 val f : int -> int -> int
 (** The continuation of the comment for function f. *)
 
-(* Hello, I'm a simple comment :-) *)
 exception My_exception of (int -> int) * int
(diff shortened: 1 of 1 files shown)


Julow (approved):
Looks good.

PR #870, opened 2022-06-13
Targetted optimisations

jonludlam:
Building on the other PRs, this one contains a set of optimisations.


src/model/paths_types.ml
```diff
@@ -1,128 +1,187 @@
 open Names
 (** {1 Paths} *)
 
+type 'a id = { iv : 'a; ihash : int; ikey : string }
```
Julow, line 4: Why store the hash instead of letting it be recomputed from the key ?

src/model/paths.mli
```diff
@@ -16,6 +16,8 @@
```
Julow, line 16: All these diffs make the `Paths` module even more annoying. What do you think of getting rid of it ?
It is not abstracting anything from `Paths_types`, it's just adding wrapper modules, which are not used much and could be written inline every time we want a `Map` or a `Set`.

src/model_desc/paths_desc.ml
```diff
@@ -60,95 +60,96 @@ module General_paths = struct
 
   let rec identifier : Paths.Identifier.t t =
     Variant
```
Julow, line 62: There's an opportunity to have a shorter output, this could return the key.

src/xref2/component.ml
```diff
@@ -3,37 +3,69 @@ module Maps = Odoc_model.Paths.Identifier.Maps
 module ModuleMap = Map.Make (struct
   type t = Ident.module_
 
-  let compare a b = Ident.compare (a :> Ident.any) (b :> Ident.any)
+  let compare a b =
+    let i1 = match a with `LRoot (_, i) | `LModule (_, i) -> i in
+    let i2 = match b with `LRoot (_, i) | `LModule (_, i) -> i in
```
Julow, line 8: Does this really have an effect ? I'm fine with specializing cases where the type has only one constructor (`TypeMap`) but I would be surprised if this leads to a speed up compared to the overhead of the `Map`.

src/xref2/cpath.ml
```diff
@@ -186,16 +186,17 @@ and is_resolved_module_hidden :
  fun ~weak_canonical_test ->
```
Julow, line 186: Unrelated to this PR but `weak_canonical_test` is very rarely passed through, almost always set to a constant. Would it make sense to remove this argument but do extra check at the call site ? eg. change `is_resolved_module_hidden ~weak_canonical_test:true p` into `is_weak_canonical p ||  is_resolved_module_hidden p`,

src/model/paths.ml
```diff
@@ -41,40 +47,40 @@ module Identifier = struct
...
   let label_parent n = label_parent_aux (n :> t)
 
-  let equal = ( = )
+  let equal x y = x.ihash = y.ihash && String.equal x.ikey y.ikey
 
-  let hash = Hashtbl.hash
+  let hash x = x.ihash
 
-  let compare = compare
+  let compare x y = compare x.ihash y.ihash
```
Julow, line 83: This doesn't correspond to `equal`, should compare the key when the hashes matches.

Julow (commented):
This is huge!

src/model/paths.ml
jonludlam: Oh, good spot thanks!

src/xref2/component.ml
Julow: In my testing, `Ident.int_of_any` compiles to just an argument access (doing just two loads, ignoring the constructor) and is inlined inside `Ident.compare`, which then inline the integer comparison (which is not optimized, it does two `cmp`).
The new versions seems to compiles to the same thing but then calls the polymorphic compare. 

Do you have an input to benchmark this function ?

jonludlam: What I was doing was turning off all the caches and doing `dune build @docgen` with `ODOC_BENCHMARK=true` (and with core_kernel changed to core if you've got v0.15)

jonludlam: (https://github.com/jonludlam/odoc/commit/074aa6a50faf99efc5a8b15cd96288a3eadaf801, https://github.com/jonludlam/odoc/commit/0aa4240252eec24351156d05298cbdb3dde65250 and https://github.com/jonludlam/odoc/commit/ec5b94d8ffc118be2f5061322fac385e80375961)

src/model/paths.ml
jonludlam: Fixed now

src/xref2/cpath.ml
jonludlam: I think I prefer it as one function - though we could make the `weak_canonical_test` an optional parameter.

src/xref2/component.ml
jonludlam: I have amended the commit so all that is left is the removal of a couple of constructors that weren't used.

src/model_desc/paths_desc.ml
jonludlam: True, though I wasn't thinking about printing the keys when I made them so they might not be particularly legible. In particular they're backward, so `Foo.bar` is `v_bar.m_Foo` IIRC (I thought many identifiers would have the same prefix, so reversing them might make it quicker to reject when comparing them)

jonludlam:
For completeness, the timing stats for most of the commits:

master: 620s
Get rid of map merges: 300s
Change type of Cpath canonical constructor: 290s
Hashtbls rather than maps: 300s
Handle canonical modules whilst handling canonical types: 280s
Only the destination of aliases in now resolved: 250s
Add smart constructors to identifiers and avoid polymorphic compare: 203.102 ±1.15
Simplify paths module: 206.368 ±1.016
Turn off caching in Of_Lang: 167.676 ±0.653
4.07/8 compatibility: 171.707 ±0.532 
Correct Identifier compare function: 175.155 ±0.685

first few were run run only so no stats.

src/model/paths_types.ml
jonludlam: just so we only compute it once.

src/model_desc/paths_desc.ml
Julow: This would have no effect for `hash` and `equal`, only `compare`, which is used less.

Julow (approved):
Pretty huge improvements ! Let's merge

PR #872, opened 2022-06-14
Coalesce expand tools

jonludlam:
Remove duplicated code from expand_tools.ml - take 2.


src/xref2/ref_tools.ml
```diff
@@ -139,14 +139,25 @@ let find find sg name =
...
   |> map_error (fun e -> `Parent (`Parent_sig e))
-  >>= fun sg -> Ok ((ref :> Resolved.Signature.t), `Module cp, sg)
 
 let module_type_lookup_to_signature_lookup env (ref, cp, m) =
-  Tools.signature_of_module_type env m
+  Tools.expansion_of_module_type env m
   |> map_error (fun e -> `Parent (`Parent_sig e))
-  >>= fun sg -> Ok ((ref :> Resolved.Signature.t), `ModuleType cp, sg)
+  >>= function
+  | Functor _ -> assert false
```
Julow, line 159: This is `Tools.assert_not_functor`.

src/xref2/ref_tools.ml
```diff
@@ -139,14 +139,25 @@ let find find sg name =
   | None -> Error (`Find_by_name (`Any, name))
 
 let module_lookup_to_signature_lookup env (ref, cp, m) =
-  Tools.signature_of_module env m
+  let rec handle_expansion : Tools.expansion -> _ = function
```
Julow, line 142: We could have a function in `Expand_tools` to do that. eg. `handle_expansion_sig`.

src/xref2/link.ml
```diff
@@ -15,13 +15,24 @@ let synopsis_from_comment (docs : Component.CComment.docs) =
...
+let synopsis_of_module env id (m : Component.Module.t) =
+  let open Utils.ResultMonad in
   match synopsis_from_comment m.doc with
   | Some _ as s -> s
   | None -> (
       (* If there is no doc, look at the expansion. *)
-      match Tools.signature_of_module env m with
-      | Ok sg -> synopsis_from_comment (Component.extract_signature_doc sg)
+      match
+        Tools.expansion_of_module env m >>= Expand_tools.handle_expansion env id
```
Julow, line 25: This could use the "handle_expansion_sig" to avoid computing the arguments.

Julow (approved):
Looks good :)

src/xref2/ref_tools.ml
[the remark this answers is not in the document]
jonludlam: But we can't because we don't have an id. This means we can't resolve paths like this:

```ocaml
module X : functor ( Y : sig type t end ) -> sig
        module Z = Y
end

(** {!X.Z.t} *)
```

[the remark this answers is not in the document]
jonludlam: This didn't work before either though, so let's leave this as is and we've got https://github.com/ocaml/odoc/issues/882 to remind us.

src/xref2/link.ml
jonludlam: Ironically this bit that is actually using `Expand_tools.handle_expansion` doesn't need it, I think.

src/xref2/ref_tools.ml
jonludlam: Thanks, fixed.

Issue #868, opened 2022-06-10
Drop OCaml < 4.08


dbuenzli:
I think it's time to move away from 4.02 which was released in 2015, seven years ago. 4.08 is a reasonable version, one of which brought many stdlib improvement.

Here are a few arguments: 

1. It would also to use cmdliner without that [monkey dance](https://github.com/ocaml/odoc/blob/6ac97f3148f7791ec7451785ef4dbd9ca0daf2d1/src/compat/compatcmdliner.ml). 
2. I was looking at the use of `Astring` in the repo which I would like to remove as a *direct dependency* (`Fpath` unfortunately also depends on it but I'd like to drop that dep there at some point too and/or remove it here too). Most of the uses of `Astring` are to work around the `String.$(op)` vs `String.ascii_$op` warnings (introduced `4.03`). Other than that we have `Astring.String.cut` all of which use a single character and can be replaced by `String.split_on_char` (introduced `4.04`). After that it only `Astring.String.is_{prefix,affix}` remains which we can easily define in the repo itself.
3. I'm pretty sure the 4.02 `cmt[i]` files were broken in some way.

jonludlam:
As I mentioned in #873:

"I'd quite like to see the features we know we want for the ocaml website to land before we drop support for the older versions - so for example the 'uses' work that will be used to drive search. That way, once we've got those bits and pieces in we can just have a maintenance branch that supports 4.02-4.07 that will hopefully be easy to manage, and we won't have to backport that sort of relatively large feature.

Of course, if it actually does become more of a burden then we can revisit this, but so far keeping our lower bound at 4.02 has only been a case of a few minor shims here and there."

Issue #826, opened 2022-02-11
Support for theming


dbuenzli:
For historical reasons `odoc` theming has been implemented in `odig`. 

The way it works there is described at the `cli` level in `odig`'s manual [here](https://erratique.ch/software/odig/doc/manual.html#odoc_themes); see also `odig odoc-theme --help`. The theme install structure is [here](https://erratique.ch/software/odig/doc/packaging.html#theme_install) (you can forget about the OCaml manual bit).

This functionality should be moved to `odoc` so that it becomes easier for people to generate documentation with a theme that is accessible to them. For example it's currently quite difficult to change it in a build system like `dune`.

Besides it would be nice if any `js` file in the theme directory could be added to the pages generated by `odoc` like `highlight.pack.js` is. This will enable nice experimentations, nowadays that `.js` files are also excuted when your browse them over the `file://` protocol.

PR #873, opened 2022-06-15
OCaml 4.02-5.0 support

dra27:
Alternate implementation to #831, retaining the current version matrix.

### DIFF
--- a/.ocamlformat-ignore
+++ b/.ocamlformat-ignore
@@ -1,3 +1,4 @@
+src/document/*.cppo.ml
 src/loader/cmi.ml
 src/loader/cmi.mli
 src/loader/cmt.ml
--- a/src/document/codefmt.ml
+++ b/src/document/codefmt.ml
@@ -1,3 +1,4 @@
+open! Compat
 open Types
 
 type out = Source.t
--- a/src/document/compat.cppo.ml
+++ b/src/document/compat.cppo.ml
@@ -1,0 +1,24 @@
+#if OCAML_VERSION >= (5, 0, 0)
+module Format = struct
+  include Format
+
+  type formatter_tag_functions = {
+    mark_open_tag : tag -> string;
+    mark_close_tag : tag -> string;
+    print_open_tag : tag -> unit;
+    print_close_tag : tag -> unit;
+  }
+
+  let pp_set_formatter_tag_functions formatter fns =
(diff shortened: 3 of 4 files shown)


dbuenzli:
I think we should move on the matrix. Why not make a last release for pre-4.08, the system is entirely usable for these (likely rare) people  and then move to 4.08. See https://github.com/ocaml/odoc/issues/868

jonludlam:
I've been thinking about this a bit. I'd quite like to see the features we _know_ we want for the ocaml website to land before we drop support for the older versions - so for example the 'uses' work that will be used to drive search. That way, once we've got those bits and pieces in we can just have a maintenance branch that supports 4.02-4.07 that will hopefully be easy to manage, and we won't have to backport that sort of relatively large feature.

Of course, if it actually does become more of a burden then we can revisit this, but so far keeping our lower bound at 4.02 has only been a case of a few minor shims here and there.

dbuenzli:
> That way, once we've got those bits and pieces in we can just have a maintenance branch that supports 4.02-4.07 that will hopefully be easy to manage, and we won't have to backport that sort of relatively large feature.

I guess could be a reasonable reason. Though you could also question whether you really want to provide documentation that far in the past (and not, e.g. ensure that past documentation is available through `archive.org`) or not simply limit new features like search to post 4.08 or rely on web search engines prior to that.

> Of course, if it actually does become more of a burden then we can revisit this, but so far keeping our lower bound at 4.02 has only been a case of a few minor shims here and there.

Let's not pretend those have not a cost (and if you are fine keeping minor shims here and there you should also be ok with adding [minor functions](https://github.com/ocaml/odoc/pull/842#issuecomment-1057555628) to avoid introducing new dependencies ;-) But what made me actually open https://github.com/ocaml/odoc/issues/868 was that the other day I was going to procrastinate by implementing https://github.com/ocaml/odoc/issues/826 but got put-off by the fact that I couldn't use cmdliner's arbitrary command nesting.

> so for example the 'uses' work that will be used to drive search.

Where can I get more information about that ? I've been wanting to add local search to `odig` for quite some time (which is the other reason why I want https://github.com/ocaml/odoc/issues/826, to be able add more JavaScript to pages without disturbing upstream) so I'm quite interested on what the plans are there and if I can piggy back on them.

PR #874, opened 2022-06-15
Test 4.02-4.06 in ocaml-ci

dra27:


odoc.opam
```diff
@@ -36,15 +36,16 @@ depends: [
...
-
-  "bisect_ppx" {dev & > "2.5.0"}
+  ("ocaml" {>= "4.05.0" & < "4.07.0" & dev} |
+  ("ocaml" {< "4.05.0" | >= "4.07.0" | !dev}
+   "ocamlfind" {with-test}
+   "yojson" {< "2.0.0" & with-test}
+   ("ocaml" {< "4.04.1" & with-test} | "sexplib0" {with-test})
+   "conf-jq" {with-test}
+   "ppx_expect" {with-test}
+   "bos" {with-test}))
```
dra27, line 46: The test dependencies have to be skipped entirely for the "dev" build on 4.05 and 4.06 which is what this is somewhat convoluted formula does!

Julow: What does the `dev` constraint mean to ocaml-ci ? Why not skip test dependencies in every cases for "dev" builds (not just for a range of ocaml versions) ?

dra27: `dev` is true for a pinned - which is what ocaml-ci does!

dra27: This was the wrong approach, anyway - I think the subsequent commit preventing the installation of `bisect_ppx` as a dev dependency on anything before 4.07 is the "right" way to go!

jonludlam:
The CI failure looks unrelated to this PR, so I'll merge this now.

Issue #825, opened 2022-02-10
Warning: Cannot find Stdlib


hbr:
When creating my documentation via odoc I usually get the warning that odoc cannot find module `Stdlib`. The reason seams to be that e.g. `String.t` is within the type of a function which is resolved to `Stdlib.String.t` (which is correct), but the documentation for `Stdlib` cannot be referenced.

Is there a way to avoid this warning or link to the documentation of `Stdlib`?

Julow:
The documentation for the stdlib isn't built, Odoc can't reference it. Dune doesn't build the documentation of dependencies and so the documentation is incomplete.

For now, Odig and the new [documentation website](https://v3.ocaml.org/packages) build the doc correctly.

On the other hand, the incomplete documentation is a lot smaller (doesn't contain all the transitive dependencies) and still readable.

Julow:
Agree. It is planned to make these error a lot less verbose. Also, they are not turn into errors with `--warn-error`.

PR #876, opened 2022-06-17
No dev deps

dbuenzli:
This moves `dev` deps to `with-test` deps. 

This makes release deps the same as `dev` deps and makes it easier to ask users to try `master` with `opam pin --dev odoc` without disturbing their `opam` switch too much.

### DIFF
--- a/odoc.opam
+++ b/odoc.opam
@@ -44,8 +44,8 @@ depends: [
   "ppx_expect" {with-test}
   "bos" {with-test}
 
-  ("ocaml" {< "4.07.0" & dev} | "bisect_ppx" {dev & > "2.5.0"})
-  ("ocaml" {< "4.03.0" & dev} | "mdx" {dev})
+  ("ocaml" {< "4.07.0" & with-test} | "bisect_ppx" {with-test & > "2.5.0"})
+  ("ocaml" {< "4.03.0" & with-test} | "mdx" {with-test})
 ]
 
 build: [
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ the docs locally yourself from the `odoc` directory:
 git clone https://github.com/ocaml/odoc.git
 cd odoc
 opam pin add . -n
-opam install mdx bos
 opam install --deps-only -t odoc
 dune build @docgen
 ```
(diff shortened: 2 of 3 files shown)


odoc.opam
```diff
@@ -44,8 +44,8 @@ depends: [
   "ppx_expect" {with-test}
   "bos" {with-test}
 
-  ("ocaml" {< "4.07.0" & dev} | "bisect_ppx" {dev & > "2.5.0"})
-  ("ocaml" {< "4.03.0" & dev} | "mdx" {dev})
+  ("ocaml" {< "4.07.0" & with-test} | "bisect_ppx" {with-test & > "2.5.0"})
+  ("ocaml" {< "4.03.0" & with-test} | "mdx" {with-test})
```
Julow, line 48: `mdx` won't be co-installable when Odoc uses the new version of `odoc-parser` and we never use `bisect_ppx`.

None of `mdx` and `bisect_ppx` are required by the main test command. I suggest we drop the dependency on both.

dbuenzli: I don't mind. Just waiting for @jonludlam's input on your comment (we agreed on that via slack).

jonludlam: The reason we don't test the mdx scripts with the default test command is because historically we couldn't guarantee it would work because of the circular dependency between mdx, odoc and odoc parser. This is no longer the case so I think we should fix this. The bisect_ppx one I'm someone less convinced about. Aspirationally I'd like to use it more :-)

dbuenzli: So what do you want ? 
 
> Aspirationally I'd like to use it more :-)

I would prefer if odoc devs would use `odoc` more ;-) That's what is going to make it a better tool, not manager metrics like test code coverage. Currently given all the warnings it outputs by default on any a reasonable code base, it's almost unusable. I struggle with this every day.

jonludlam: Is it the warnings about not being able to find dependency modules?

jonludlam: Also, the reason I want to use coverage is simply because our test suite right now is still completely insufficient to properly test the model - for a more reliable test I have to build docs using the driver, commit them to a git repo and diff them in order to make sure things haven't changed when I don't expect them to. The current process is to use this to extract a test case and drop it in, but a more systematic way would be to _at least_ try to make sure our test suite covers the majority of `xref2`. This won't be sufficient, but it'd be better than we've got now.

dbuenzli: > Is it the warnings about not being able to find dependency modules?

Indeed https://github.com/ocaml/odoc/issues/825

PR #877, opened 2022-06-17
Remove esy dev support and CI

dbuenzli:
This is an RFC. If you care deeply about that, scream. 

`esy` knows how to build from `opam` file so this is rather a maintenance burden. 

Also the remaining CI setup should be more than enough for testing a pure OCaml project. Let's avoid heating the planet more.


Julow (approved):
This was breaking regularly and I never had a clue at whether we had an actually working setup.

If esy don't need this, let's remove it!

jonludlam:
Nobody has complained. Bye for now, esy!

PR #881, opened 2022-06-21
Disable the missing root warning by default

jonludlam:
It can be re-enabled via the command-line option

    --enable-missing-root-warning

But since this is usually a build system issue rather than something
end-users will need to worry about, let's not complain about it.

Note that this will _not_ suppress warnings generated if references
aren't resolved - so if you explicitly have a reference to something
in a dependency library and it's not found when linking, you'll still
get a warning.

Fixes: #825

### DIFF
--- a/src/model/error.ml
+++ b/src/model/error.ml
@@ -1,5 +1,7 @@
 open Result
 
+let enable_missing_root_warning = ref false
+
 type full_location_payload = Odoc_parser.Warning.t = {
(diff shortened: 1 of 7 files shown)


Julow (approved):
Looks good :) It's true that Dune isn't going to fix this anytime soon, the warning was unfixable and noisy.

dbuenzli:
This is godsent. Thank you.

PR #883, opened 2022-06-23
Misc more optimisations

jonludlam:
A few more optimisations and fixes. Again, based on the previous PRs.


Julow (approved):
What is supposed to be in the "Get rid of unresolved paths in docs" commit ?

Otherwise, I think it's ready to merge :)

Julow:
I see, thanks.

PR #884, opened 2022-07-01
The right command to generate doc

cnmade:
is `dune build @doc`

see reference: https://dune.readthedocs.io/en/stable/documentation.html#building-documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ git clone https://github.com/ocaml/odoc.git
 cd odoc
 opam pin add . -n
 opam install --deps-only -t odoc
-dune build @docgen
+dune build @doc
 ```
 
 and find the docs in `_build/default/doc/html/index.html`


Julow (commented):
`@docgen` builds Odoc's doc with [this Mdx document](https://github.com/ocaml/odoc/blob/master/doc/driver.md?plain=1) as an example of how to drive Odoc.

I don't know what is best between `@docgen` or `@doc` but:
- The path in which to find the docs corresponds to `@docgen` (even thought it's wrong, should be `_build/default/doc/html/odoc/index.html`)
- The Mdx dependency of `@docgen` is not installed by the script

cnmade:
The default install won't use mdx i think. i tried many times, it default generate html , which @doc does.

jonludlam:
I just tested the instructions and it worked for me. `mdx` ought to get installed when you specify `-t` on the `opam install` line - does that not work for you?

cnmade:
image
I used the way .

I am not reading the readme.md first.

but , for user side, why this document not the same behaviour with dune document, very straige.

jonludlam:
OK, since you're not using the README but this PR changes it, let's close this one.

PR #886, opened 2022-07-05
Handle Math_span and Math_block constructs

gpetiot:
This PR is based on #811 (I don't have the rights to push to it)

- rebased on top of `main` branch of `odoc-parser` to handle `Math_span` and `Math_block`
- vendored the katex files similarly to what is done for highlight.js (no need to check for the integrity of the files once they are vendored)


jonludlam:
Looking good so far! I think there are a few things we still need to do on this:

1. We need some tests
2. I think we're missing the fonts in the vendored files
3. There's no distinction between display and inline maths
4. It'd be nice to only add the scripts to the html if the page actually contains some math content.

jonludlam:
I've released odoc-parser 2.0.0 with math support, so the CI should be a little happier now!

src/latex/raw.ml
```diff
@@ -206,3 +206,13 @@ let small_table pp ppf tbl =
   break ppf Line
 
 let ocamltag tag pp ppf x = create2 "ocamltag" Fmt.string pp ppf tag x
+
+let math ppf x = math x ppf
```
Julow, line 210: `math` is for rendering a single element, not a block of latex. It adds a `\` prefix that we don't want.

Julow (commented):
Nice! Some comments:

- Katex complains with
  ```
  downloadable font: download failed (font-family: "KaTeX_Size1" style:normal weight:400 stretch:100 src index:0): status=2152857618 source: file:///home/juloo/w/odoc/test/generators/html/fonts/KaTeX_Size1-Regular.woff2
  downloadable font: download failed (font-family: "KaTeX_Main" style:normal weight:400 stretch:100 src index:0): status=2152857618 source: file:///home/juloo/w/odoc/test/generators/html/fonts/KaTeX_Main-Regular.woff2
  downloadable font: download failed (font-family: "KaTeX_Size1" style:normal weight:400 stretch:100 src index:1): status=2152857618 source: file:///home/juloo/w/odoc/test/generators/html/fonts/KaTeX_Size1-Regular.woff
  downloadable font: download failed (font-family: "KaTeX_Main" style:normal weight:400 stretch:100 src index:1): status=2152857618 source: file:///home/juloo/w/odoc/test/generators/html/fonts/KaTeX_Main-Regular.woff
  downloadable font: download failed (font-family: "KaTeX_Size1" style:normal weight:400 stretch:100 src index:2): status=2152857618 source: file:///home/juloo/w/odoc/test/generators/html/fonts/KaTeX_Size1-Regular.ttf
  downloadable font: download failed (font-family: "KaTeX_Main" style:normal weight:400 stretch:100 src index:2): status=2152857618 source: file:///home/juloo/w/odoc/test/generators/html/fonts/KaTeX_Main-Regular.ttf 
  ```
  There should be a way to specify a font family to use/not use. Otherwise, we'll need to vendor it too.

- Create symlinks inside the generator tests output directory, to be able to open the browser on the test results:
  ```
  ln -s ../../../src/vendor/katex.min.css test/generators/html/katex.min.css
  ln -s ../../../src/vendor/katex.min.js test/generators/html/katex.min.js
  ```

- When a block doesn't render (for whatever reason), the result is unreadable. The example in the tests renders to this:
  ```
  % \f is defined as #1f(#2) using the macro \f\relax{x} = \int_{-\infty}^\infty \f\hat\xi\,e^{2 \pi i \xi x} \,d\xi
  ```
  I suggest wrapping math in a `<code>` element until they are rendered.

- In the test, the second math element doesn't render, perhaps due to the font issue ?

src/html/generator.ml
```diff
@@ -366,6 +372,72 @@ module Toc = struct
     List.map section toc
 end
 
+module Math = struct
```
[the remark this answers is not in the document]
gpetiot: done

src/latex/raw.ml
gpetiot: fixed

src/html/tree.ml
```diff
@@ -70,6 +68,34 @@ let page_creator ~config ~url name header toc content =
...
+        let katex_js_uri = file_uri support_uri "katex.min.js" in
+        default_meta_elements
+        @ [
+            Html.link ~rel:[ `Stylesheet ] ~href:katex_css_uri ();
+            Html.script ~a:[ Html.a_src katex_js_uri ] (Html.txt "");
+            Html.script
+              (Html.cdata_script
+                 {|
+          document.addEventListener("DOMContentLoaded", function () {
+            var elements = document.getElementsByClassName("odoc-katex-math");
```
jonludlam, line 84: This doesn't work as expected because `document.getElementsByClassName()` returns a _live_ HTMLCollection (see [the warning](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName)).

jonludlam, line 84:
```suggestion
            var elements = Array.from(document.getElementsByClassName("odoc-katex-math"));
```

src/html/tree.ml
```diff
@@ -70,6 +68,34 @@ let page_creator ~config ~url name header toc content =
...
+            for (var i = 0; i < elements.length; i++) {
+              var el = elements[i];
+              var content = el.textContent;
+              var new_el = document.createElement("span");
+              new_el.setAttribute("class", "odoc-katex-math-rendered");
+              katex.render(content, new_el, { throwOnError: false });
```
jonludlam, line 90: We need to tell katex to render display style rather than inline explicitly. We could, for example, add a class to display elements, and then:

```suggestion
              var display = el.classList.contains("display");
              katex.render(content, new_el, { throwOnError: false, displayMode: display });
```

src/latex/raw.ml
```diff
@@ -206,3 +206,13 @@ let small_table pp ppf tbl =
   break ppf Line
 
 let ocamltag tag pp ppf x = create2 "ocamltag" Fmt.string pp ppf tag x
+
+let math ppf x = Fmt.pf ppf {|$%s$|} x
+
+let equation ppf x =
+  let name = "equation" in
```
jonludlam, line 213: I wonder if this shouldn't be `equation*` - non numbered.

jonludlam (changes requested):

src/latex/raw.ml
gpetiot: done

src/html/tree.ml
```diff
@@ -87,7 +87,8 @@ let page_creator ~config ~url ~uses_katex name header toc content =
               var content = el.textContent;
               var new_el = document.createElement("span");
               new_el.setAttribute("class", "odoc-katex-math-rendered");
-              katex.render(content, new_el, { throwOnError: false });
+              var display = el.classList.contains("display");
```
jonludlam, line 90: I should have mentioned - we'll need to add this `display` class in the html renderer.

src/html/generator.ml
```diff
@@ -43,7 +43,7 @@ let mk_anchor anchor =
 let class_ (l : Class.t) = if l = [] then [] else [ Html.a_class l ]
 
 let math (s : Math.t) =
-  Html.span ~a:[ Html.a_class [ "odoc-katex-math" ] ] [ Html.txt s ]
+  Html.span ~a:[ Html.a_class [ "odoc-katex-math"; "display" ] ] [ Html.txt s ]
```
jonludlam, line 46: The `display` class should only be added in blocks - i.e. in the test, the one that says it's inline should be `class="odoc-katex-math"` and the one that says it's a block should be `class="odoc-katex-math display"`

jonludlam: I should have been more explicit in my previous comment, sorry! :-)

gpetiot: Done, but some fonts are not properly rendered even though every font file has been vendored and is correctly located in the test (don't have any log error).

Julow (commented):
- When a block fails to render, the result is still a mess. Intermediate elements should be `<code>` for inline elements and `<pre>` for blocks instead of `<span>` for both currently.

- Are all these font files necessary ? Can we get away with shipping just the `.ttf`s or fetching them from the internet ?

src/compat/compatfmt.ml
```diff
@@ -0,0 +1,18 @@
+(* Compatibility module for Fmt *)
```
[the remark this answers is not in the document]
dbuenzli: Indeed `odoc` needs less dependencies, not more. If you want the short names just define an internal `Fmt` module for the project (I often do so in my libraries).

gpetiot:
> * Are all these font files necessary ? Can we get away with shipping just the `.ttf`s or fetching them from the internet ?

Based on https://katex.org/docs/font.html I would be more tempted to only keep the `.woff` files (and they're smaller than ttf).

dbuenzli:
> Based on https://katex.org/docs/font.html I would be more tempted to only keep the `.woff` files (and they're smaller than ttf).

Just keep the `.woff2` fonts, browser support is [widespread enough](https://caniuse.com/?search=woff2).

jonludlam:
The block math example is a bit broken - in the original katex example it relies on a macro declared as an option to the katex renderer (which can be seen by clicking the cog in the example):
```json
{
    "\\f": "#1f(#2)"
}
```
Since I don't think we want a way to provide out-of-band macros like this we could just use `\newcommand` which appears to be supported - so the example becomes:
```latex
% \f is defined as #1f(#2) using the macro
\newcommand{\f}[2]{#1f(#2)}
\f\relax{x} = \int_{-\infty}^\infty
\f\hat\xi\,e^{2 \pi i \xi x}
\,d\xi
```
then it renders as it does on the katex home page.

giltho:
PR lgtm (although I'm not an odoc expert at all).
Thank you so much for taking over my half-baked work!

Curious @jonludlam why are you unhappy with out-of-band macros?
I'd actually be happy to be able to provide a macro file on the side, in whichever format, which would be injected everywhere

jonludlam:
> Curious @jonludlam why are you unhappy with out-of-band macros? I'd actually be happy to be able to provide a macro file on the side, in whichever format, which would be injected everywhere

Let's just say I'm happy not to have to support this until we've got an extremely compelling use case :-)

giltho:
I think that's fair yes :)

jonludlam:
I think this is good now, thanks!

PR #887, opened 2022-07-17
Add tag content linking

### LINKED ISSUE #827: References not resolved within `@deprecated` and other tagged paragraphs
Given an interface `two.mli`:
```ocaml
val y : unit
(** Nothing to see here. *)

val z : unit
(** Equivalent to {!One.x} and {!y}.

    @deprecated Move along to {!One.x} or {!y}.

    @return The same value as {!One.x} and {!y}.

    @raise Not_found if {!One.x} and {!y} are different. *)
```
the and a corresponding `one.mli` documenting `x`, neither of the references within the `@deprecated`, `@return` and `@raise` paragraphs are resolved by odoc 2.0.2, while the ones outside are.

The reference to the external module also shows the symptom reported in #816. The issue #818 might also be related.

sim642:
Closes #827.

I noticed that the similar `List` case in `comment_nestable_block_element` uses `List.rev_map ... |> List.rev` instead though. Should it be done here as well?

### DIFF
--- a/src/xref2/link.ml
+++ b/src/xref2/link.ml
@@ -224,13 +224,7 @@ and comment_nestable_block_element env parent ~loc:_
   | `List (x, ys) ->
       `List
         ( x,
-          List.rev_map
-            (fun x ->
-              List.rev_map
-                (with_location (comment_nestable_block_element env parent))
-                x
-              |> List.rev)
-            ys
+          List.rev_map (comment_nestable_block_element_list env parent) ys
           |> List.rev )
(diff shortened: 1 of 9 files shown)


sim642:
> It would be great to have a test, a new case in `test/generators/cases/` with resolving references would do, then run `dune runtest --auto-promote` twice.

Done!

jonludlam:
Thank you very much! We can ignore CI for now until the Math PR is merged.

PR #888, opened 2022-07-18
Treat raise tag value as reference

### LINKED ISSUE #475: Linking to exceptions in @raise
Using Odoc master, it doesn't appear to be possible to link to an exception definition from a `@raise` tag:

```ocaml
exception Foo

val f : unit -> _
(** @raise {!Foo} *)

val g : unit -> _
(** @raise Foo *)
```

image

I'd expect the former to work, but the latter would be fine too. CC @Julow.

sim642:
Closes #475.

In order to allow for maximal code reuse, the value variants are a subset of inline element variants. The `Nested` constructor was also added to enable code reuse.

### TODO
- [x] Test.
- [x] Format code.
- [ ] ~~Fix warnings during `dune build`:~~
    ```
    File "cases/markup.mli", line 168, characters 4-25:
    Warning: Failed to resolve reference unresolvedroot(Failure) Couldn't find "Failure"
    File "cases/ocamlary.mli", line 227, characters 4-39:
    Warning: Failed to resolve reference unresolvedroot(Not_found) Couldn't find "Not_found"
    File "cases/tag_link.ml", line 5, characters 4-21:
    Warning: Failed to resolve reference unresolvedroot(Foo) Couldn't find "Foo"
    ```
    Not sure what to do about it though.
- [x] Clean up git history.

### DIFF
--- a/src/ocamlary/ocamlary.mli
+++ b/src/ocamlary/ocamlary.mli
@@ -226,6 +226,9 @@ val fun_maybe : ?yes:unit -> unit -> int
 val not_found : unit -> unit
 (** @raise Not_found That's all it does *)
 
+val kaboom : unit -> unit
+(** @raise Kaboom That's all it does *)
+
 val ocaml_org : string
 (** @see <http://ocaml.org/> The OCaml Web site *)
 
--- a/test/generators/html/Ocamlary.html
+++ b/test/generators/html/Ocamlary.html
@@ -453,6 +453,24 @@
     </div>
    </div>
    <div class="odoc-spec">
+    <div class="spec value anchored" id="val-kaboom">
+     <a href="#val-kaboom" class="anchor"></a>
+     <code>
+      <span><span class="keyword">val</span> kaboom : 
+       <span>unit <span class="arrow">&#45;&gt;</span></span> unit
+      </span>
+     </code>
+    </div>
+    <div class="spec-doc">
+     <ul class="at-tags">
+      <li class="raises"><span class="at-tag">raises</span> 
+       <a href="#exception-Kaboom"><code>Kaboom</code></a> 
+       <p>That's all it does</p>
+      </li>
+     </ul>
+    </div>
+   </div>
+   <div class="odoc-spec">
     <div class="spec value anchored" id="val-ocaml_org">
      <a href="#val-ocaml_org" class="anchor"></a>
      <code><span><span class="keyword">val</span> ocaml_org : string</span>
--- a/test/generators/latex/Ocamlary.tex
+++ b/test/generators/latex/Ocamlary.tex
@@ -150,6 +150,11 @@ After exception title.
 \item[{raises \ocamlinlinecode{Not\_\allowbreak{}found}}]{That's all it does}\end{description}%
 \end{ocamlindent}%
 \medbreak
+\label{module-Ocamlary-val-kaboom}\ocamlcodefragment{\ocamltag{keyword}{val} kaboom : unit \ocamltag{arrow}{$\rightarrow$} unit}\begin{ocamlindent}\begin{description}\kern-\topsep
+\makeatletter\advance\@topsepadd-\topsep\makeatother% topsep is hardcoded
+\item[{raises \hyperref[module-Ocamlary-exception-Kaboom]{\ocamlinlinecode{\ocamlinlinecode{Kaboom}}}}]{That's all it does}\end{description}%
+\end{ocamlindent}%
+\medbreak
 \label{module-Ocamlary-val-ocaml_org}\ocamlcodefragment{\ocamltag{keyword}{val} ocaml\_\allowbreak{}org : string}\begin{ocamlindent}\begin{description}\kern-\topsep
 \makeatletter\advance\@topsepadd-\topsep\makeatother% topsep is hardcoded
 \item[{see \href{http://ocaml.org/}{http://ocaml.org/}\footnote{\url{http://ocaml.org/}}}]{The OCaml Web site}\end{description}%
--- a/test/generators/man/Ocamlary.3o
+++ b/test/generators/man/Ocamlary.3o
@@ -391,6 +391,13 @@ This is a_function with param and return type\.
 @raises Not_found: That's all it does
 .nf 
 .sp 
+\f[CB]val\fR kaboom : unit \f[CB]\->\fR unit
+.fi 
+.br 
+.ti +2
+@raises \f[CI]Kaboom\fR: That's all it does
+.nf 
+.sp 
 \f[CB]val\fR ocaml_org : string
 .fi 
 .br 


src/model/comment.ml
```diff
@@ -51,7 +51,10 @@ type tag =
   [ `Author of string
   | `Deprecated of nestable_block_element with_location list
   | `Param of string * nestable_block_element with_location list
-  | `Raise of string * nestable_block_element with_location list
+  | `Raise of
+    [ `Code_span of string | `Reference of Reference.t * link_content ]
```
Julow, line 55: Cosmetic changes: You can create a new type that is included here and in `inline_element` to avoid duplication:
```
type reference_element = [ `Reference of Reference.t * link_content ]
```

Either that or a type that doesn't have an empty content.

src/model/semantics.ml
```diff
@@ -273,8 +273,17 @@ let tag :
       ok (`Deprecated (nestable_block_elements status content))
   | `Param (name, content) ->
       ok (`Param (name, nestable_block_elements status content))
-  | `Raise (name, content) ->
-      ok (`Raise (name, nestable_block_elements status content))
+  | `Raise (name, content) -> (
+      match Error.raise_warnings (Reference.parse location name) with
+      (* TODO: location for just name *)
```
Julow, line 278: That would be a nice thing to have later.

src/model_desc/comment_desc.ml
```diff
@@ -115,12 +118,20 @@ and tag : general_tag t =
...
+      | `Reference (x1, x2) ->
+          C ("`Reference", (x1, x2), Pair (reference, link_content)))
+  in
   Variant
     (function
     | `Author x -> C ("`Author", x, string)
     | `Deprecated x -> C ("`Deprecated", x, docs)
     | `Param (x1, x2) -> C ("`Param", (x1, x2), Pair (string, docs))
-    | `Raise (x1, x2) -> C ("`Raise", (x1, x2), Pair (string, docs))
+    | `Raise (x1, x2) -> C ("`Raise", (x1, x2), Pair (raise_kind, docs))
```
Julow, line 134: You can coerce to the bigger type to avoid writing `raise_kind`:
```suggestion
    | `Raise (x1, x2) -> C ("`Raise", ((x1 :> general_inline_element), x2), Pair (inline_element, docs))
```

src/xref2/link.ml
```diff
@@ -262,8 +262,19 @@ and comment_tag env parent ~loc:_ (x : Comment.tag) =
       `Deprecated (comment_nestable_block_element_list env parent content)
   | `Param (name, content) ->
       `Param (name, comment_nestable_block_element_list env parent content)
-  | `Raise (name, content) ->
-      `Raise (name, comment_nestable_block_element_list env parent content)
+  | `Raise ((`Reference (r, reference_content) as orig), content) -> (
+      assert (reference_content = []);
```
Julow, line 266: We do not need to care. Either way is fine: use a more restricted type or accept every possible values.

src/document/comment.ml
```diff
@@ -254,8 +254,10 @@ let tag : Comment.tag -> Description.one =
   | `Param (name, content) ->
       let value = Inline.Text name in
       item ~tag:"parameter" ~value (nestable_block_element_list content)
-  | `Raise (name, content) ->
-      let value = Inline.Text name in
+  | `Raise (kind, content) ->
+      let value =
+        Inline.Nested (inline_element (kind :> Comment.inline_element))
```
Julow, line 259: I'm not a fan of this new `Nested` constructor. It can be avoided completely if `~value` is changed to be an `Inline.t`:

```patch
diff --git a/src/document/comment.ml b/src/document/comment.ml
index b0709b421..be0ff78d2 100644
--- a/src/document/comment.ml
+++ b/src/document/comment.ml
@@ -274,14 +274,11 @@ let tag : Comment.tag -> Description.one =
   let sp = inline (Text " ") in
   let item ?value ~tag definition =
     let tag_name = inline ~attr:[ "at-tag" ] (Text tag) in
-    let tag_value =
-      match value with
-      | None -> []
-      | Some t -> [ sp; inline ~attr:[ "value" ] t ]
-    in
+    let tag_value = match value with None -> [] | Some t -> sp :: t in
     let key = tag_name :: tag_value in
     { Description.attr = [ tag ]; key; definition }
   in
+  let mk_value desc = [ inline ~attr:[ "value" ] desc ] in
   let text_def s = [ block (Block.Inline [ inline @@ Text s ]) ] in
   let content_to_inline ?(prefix = []) content =
     match content with
@@ -293,25 +290,23 @@ let tag : Comment.tag -> Description.one =
   | `Deprecated content ->
       item ~tag:"deprecated" (nestable_block_element_list content)
   | `Param (name, content) ->
-      let value = Inline.Text name in
+      let value = mk_value (Inline.Text name) in
       item ~tag:"parameter" ~value (nestable_block_element_list content)
   | `Raise (kind, content) ->
-      let value =
-        Inline.Nested (inline_element (kind :> Comment.inline_element))
-      in
+      let value = inline_element (kind :> Comment.inline_element) in
       item ~tag:"raises" ~value (nestable_block_element_list content)
   | `Return content -> item ~tag:"returns" (nestable_block_element_list content)
   | `See (kind, target, content) ->
       let value =
         match kind with
-        | `Url -> Inline.Link (target, [ inline @@ Text target ])
-        | `File -> Inline.Source (source_of_code target)
-        | `Document -> Inline.Text target
+        | `Url -> mk_value (Inline.Link (target, [ inline @@ Text target ]))
+        | `File -> mk_value (Inline.Source (source_of_code target))
+        | `Document -> mk_value (Inline.Text target)
       in
       item ~tag:"see" ~value (nestable_block_element_list content)
   | `Since s -> item ~tag:"since" (text_def s)
   | `Before (version, content) ->
-      let value = Inline.Text version in
+      let value = mk_value (Inline.Text version) in
       item ~tag:"before" ~value (nestable_block_element_list content)
   | `Version s -> item ~tag:"version" (text_def s)
   | `Alert ("deprecated", content) ->
```

This also remove the `<span class="value">` around the reference, which doesn't seem useful.

Julow (commented):
Sorry for the very long delay! To catch up, I've rebased you commits on master in branch [888_rebased](https://github.com/Julow/odoc/tree/888_rebased) of my fork. You can `reset --hard` to it if you are OK. (tests are failing due to some new pattern matches not handling the new constructor)

I have some cosmetic comments that don't need to be addressed if you don't have the time. But I'd prefer not to have the `Nested` constructor.

Don't worry about the warnings when running `dune build`. When they come from generator tests, they can be ignored as long as the test output looks right.

If you can clean up the history, that would nice.

Thanks for your contribution!

src/model/comment.ml
sim642: Done!

src/model_desc/comment_desc.ml
sim642: Thanks for the hint!

src/xref2/link.ml
sim642: I've changed it to allow non-empty reference content, although that's still never created for raise tags. A more restricted type would've been slightly annoying since that wouldn't have allowed the reuse of the normal reference variant and all the code there already is to handle those.

src/document/comment.ml
sim642: I've reverted the `Nested` constructor now and applied your patch. I think in the git history I had something similar at one point but then wanted to be consistent and have the `value` attribute on everything.

sim642:
I also cleaned up the git history now to remove all the back-and-forth changes.

panglesd (approved):
Looks good to me. Thanks!

About the consistent use of the `value`: I agree with you it would have been better to be consistent for styling purpose. However, this class does not seem to be used in neither `odoc`'s css nor `odig`'s one, so let's not delay this PR more. If someone later wants to use the `value` class for styling purpose, we can make it consistent at this point.

There are some CI errors to fix, I think due to the fact that you modified `odoc.css` without running `dune runtest` and promoting (static files are turned into OCaml values, and kept in sync via a test-promote workflow).

Julow (approved):
Thanks a lot! This can be merged.

PR #889, opened 2022-08-02
Render ml and mli files as part of the documentation

gpetiot:
The idea is to create links from the generated doc to jump to the source files (.ml/.mli but rendered as HTML).
Inspired from a PR of Leo White (don't remember the #).

WIP:
- the path in the location is relative, this causes an issue with the tests (insights are welcomed)
- no anchor in the generated HTML files
- review the locations added, maybe the ones tagged with "fix this loc" should be removed?


dbuenzli:
I think this needs a little bit more thinking on the UI side of things. 

First I'm not convinced about the idea of linking to `.mli` files. What we see on an `odoc` page is basically an HTMLized `.mli` file. As such it feels redundant and quite useless to link on `.mli` sources, that's a click lost to effectively show you what you are already seeing except unrendered. 

Second you didn't provide a reference rendering to show what this effectively does but it seems that this is adding quite a bit of noise to the docsets under the form of "jump to source" links.

I didn't think this throughout but as a gut feeling I would say that it would be better to: 

1. Put the link on the identifier defined by the structure item (that is `id` in `type id = …` `val id : …`, `module Id`).
2. The destination of the link should bring you directly to the corresponding definition in the `.ml` source.

jonludlam:
I believe Leo's original branch (https://github.com/lpw25/doc-ock/tree/uses) was adding support to track both the declaration and definition of the various items, and the intent there was indeed to be able to jump to the ml file even if the page had been rendered from the mli file. I agree that we want to be heading in that direction.

gpetiot:
I've hit a few blockers on this PR:
- Couldn't find a way to generate html files corresponding to the source files. I thought it would be made in Subpages.compute but visibly not.
- Linking on the .ml files instead of .mli would be preferable indeed, I'mm not sure what is possible with the given state of odoc. odoc master branch diverged too far from the base of Leo's branch, reusing some ideas from it is difficult.
- From what I understand the links would need to be made in document/generator.ml instead, since it's the last step before the content of the doc gets merged into a string, the html generation would be too late.

In short I think I was too ambitious with this PR, if anyone wants to step in, feel free to do so :)

jonludlam:
I believe this has been superseded by #904

PR #893, opened 2022-09-19
Add support for tables

gpetiot:
Support new syntax added in https://github.com/ocaml-doc/odoc-parser/pull/11


src/document/types.ml
```diff
@@ -85,11 +85,25 @@ and Block : sig
     | Math of Math.t
     | Verbatim of string
     | Raw_markup of Raw_markup.t
+    | Table_light of Table_light.t
+    | Table_heavy of Table_heavy.t
```
Julow, line 89: Why do we need the two kinds in the document ? It seems like both syntaxes can be compiled to the same representation.

gpetiot: done

panglesd (commented):
According to [caniuse](https://caniuse.com/?search=colgroup), only 75% of viewer use a browser which support the `colgroup` tag (88% on desktop) (the main culprit is old versions of safari on iOS).

75/88% might be good enough, or not, but since the code is generated, it is as easy to give each `td`/`th` items of each line the style (or class) that align them, instead of using `colgroup`.

panglesd:
Now that https://github.com/ocaml-doc/odoc-parser/pull/11 has been merged, this PR is ready for review!

What is the best way to run the tests before the parser is released? Should we temporarily pin `odoc-parser`?

doc/odoc_for_authors.mld
```diff
@@ -541,6 +541,74 @@ in block form this becomes:
 See the {{:https://katex.org/docs/supported.html}KaTeX documentation} for the
 HTML mode LaTeX support status.
 
+{2 Tables}
+
+Odoc 2.3 introduced new markup for tables. This markup comes in two flavors: the light syntax, and the heavy syntax.
+
+The heavy syntax uses several markup: [{table ...}] to define a table, [{tr ...}] to define a line, and [{th ...}] and [{td ...}] to respectively define a header cell and a normal cell.
```
dbuenzli, line 548: to define a line -> to define a row 
a normal cell -> a data cell

(makes it easier understand what `tr` and `td` could stand for)

doc/odoc_for_authors.mld
```diff
@@ -541,6 +541,74 @@ in block form this becomes:
...
+Direct children of tables have to be lines, and direct children of lines have to be cells. Similarly, lines have to be direct children of tables, and cells direct children of lines. Cells can contain any markup.
```
dbuenzli, line 549: lines -> rows (4x)

doc/odoc_for_authors.mld
panglesd: Thanks, good point!

panglesd:
I added support for alignment in the `latex` backend, using the `array` latex package.
So, `\usepackage{array}` need to be added to the latex preamble for array alignment to work (array without alignment and no array do not need modification in the preamble).

I tested the `array` addition it on the ocaml manual and it seems to not change anything (array is supposed to be retrocompatible with original `tabular` environment).

The `man` backend does not have support for alignment.

panglesd:
I finished support for the `man` backend, including adding support for alignment.

The support is restricted to what the target syntax allows, though. Nested tables in man pages won't render, without warning.

Depends on https://github.com/ocaml-doc/odoc-parser/pull/14 to pass the tests.

src/latex/types.ml
```diff
@@ -23,14 +23,19 @@ type elt =
...
   | Ligaturable of string
 
 and section = { level : int; label : string option; content : t }
 
 and list_info = { typ : Odoc_document.Types.Block.list_type; items : t list }
 
-and table = { row_size : row_size; tbl : t list list }
+and layout_table = { row_size : row_size; tbl : t list list }
+
+and alignment = Default | Left | Right | Center
```
gpetiot, line 36: Why not using the same option type as in `document/types.ml` to stay consistent with the html generator?

src/manpage/generator.ml
```diff
@@ -304,6 +313,55 @@ and inline (l : Inline.t) =
...
+                 {[
+                   | Some `Left -> "lmlj"
+                   | Some `Center -> "cmce"
+                   | Some `Right -> "rmrj"
+                   | None -> "l"
+                  ]} *)
+              | Some `Left -> "l"
+              | Some `Center -> "c"
+              | Some `Right -> "r"
+              | None -> "l")
```
gpetiot, line 343: I couldn't find a ref for the manpage syntax, is there no other choice for the default case (like keeping it empty) ?

panglesd: The ref for tables in the gtroff format is https://manpages.debian.org/testing/groff-base/tbl.1.en.html and https://doc.cat-v.org/unix/v10/10thEdMan/tbl.pdf

I could not find a "not specified" alignment. Moreover, the number of alignment _has to match_ the number of columns, otherwise the extra columns do not get rendered...

gpetiot: Thanks! The default alignment seems to be left in the doc indeed.

gpetiot:
For me it's good to be merged as soon as https://github.com/ocaml-doc/odoc-parser/pull/14 is merged

panglesd:
Need to be resolved on #950 for CI to pass!

src/document/types.ml
```diff
@@ -89,11 +89,22 @@ and Block : sig
...
   and list_type = Ordered | Unordered
 end =
   Block
 
+and Table : sig
+  type alignment = [ `Left | `Center | `Right ]
+
+  type 'a t = {
+    data : ('a * [ `Header | `Data ]) list list;
+    align : alignment option list option;
```
trefis, line 103: Meh.
This leads to ugly code in the html backend, and some "normalisation" in the latex one.
Perhaps we could normalize when we create the elements, instead of when we render them; so `align : alignment option list`?

Furthermore: I assume the polymorphic variants are here because the parser emits them, but:
1. I'm not a fan
2. If we're already doing some datatype mapping from the parser to odoc types, let's go a bit further:  
   ```ocaml
   type alignment  = Default | Left | Center | Right
   type 'a t = {
     date : ...;
     align: alignment list;
   }
   ```

src/xref2/link.ml
```diff
@@ -7,6 +7,10 @@ module Opt = struct
   let map f = function Some x -> Some (f x) | None -> None
 end
 
+module Rec_list = struct
+  let map f x = List.rev_map f x |> List.rev
+end
```
trefis, line 12: ?

trefis (commented):
I left a few comments here and there.

I haven't looked in details at the latex and man backends, I personally trust that you've tested that they work.
However, I guess @Octachron might want to have a look? (And perhaps @dbuenzli as well, for the manpage one?)

odoc.opam
```diff
@@ -24,6 +24,10 @@ Odoc is a documentation generator for OCaml. It reads doc comments,
 delimited with `(** ... *)`, and outputs HTML. 
 """
 
+pin-depends: [
+  ["odoc-parser.dev" "git+https://github.com/ocaml-doc/odoc-parser.git#f98cfe3"]
```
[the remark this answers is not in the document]
panglesd: Yes, a release of `odoc-parser` is needed. But we also shouldn't release `odoc-parser` without releasing `odoc`...

Also, since https://github.com/ocaml-doc/odoc-parser/pull/16 has also been merged but not its odoc counterpart #950, this `pin-depends` don't work anyway...

src/document/types.ml
panglesd: I agree with removing the polymorphic variant.
I don't mind using `Default` instead of `None`.
I think I delayed the "compute the number of columns when alignment is unspecified" since it's not always needed, but I agree it's not ideal, so I implemented your proposition, which seems better.

src/xref2/link.ml
panglesd: Cleaned

panglesd:
- About the latex backend:

For alignment to work, you have to add `\usepackage{array}` to the preamble of the latex file.
The [array package](https://ctan.org/pkg/array?lang=en) is a compatible extension of the original `array` environment, it was needed to specify both alignment and cell width.
I tried to add this package to the ocaml manual preamble and build the docs, and I could not notice any difference. So I think it is a safe to require it for aligned arrays without breaking existing things.

Adding this package is not required if there is no tables with specified alignment in the documentation.

- About the manpage:

I used `tbl`, with the following documentation: https://manpages.debian.org/testing/groff-base/tbl.1.en.html
The manpage backend suffers from some limitations (alignment does not work well, nested tables do not work) but I think the support is already sufficient.

odoc.opam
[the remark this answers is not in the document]
trefis: The pin-depends has been updated, and "support" added to this PR. So this is now resolved.

trefis (approved):
Will merge tomorrow morning unless someone disagrees.

Octachron:
I have a quick glance at the latex backend and it looks fine to me (and we can always amend it if we discover issues later).

PR #896, opened 2022-10-05
title formatting is now a renderer decision

sabine:
This patch puts the responsibility of rendering the title (part of the "header" section) on the renderer, as suggested in the comment at https://github.com/ocaml/odoc/blob/a52dd8e960f99b38162d40311a9c4512d5741d43/src/document/generator.ml#L24

This refactoring is a prerequisite for adding more varied renderers, e.g., as requested in https://github.com/ocaml/odoc/issues/374 and as we need for the ocaml.org website's package docs browser.

`Odoc_document.Types.Page.t` now contains `preamble` and `kind` instead of `header` and the decision on whether (and how) to render a title is now made in the respective `Generator.Page.page` function of the corresponding renderer.

As I didn't change anything about the title rendering itself, I put the title rendering code shared between the different renderers in a new module `Doctree.PageTitle`. I am very open to suggestions where else to put it. Also, if there's anything I need to change about this PR, let me know.

### DIFF
--- a/src/document/doctree.ml
+++ b/src/document/doctree.ml
@@ -296,20 +296,19 @@ end = struct
     in
     match kind with
     | `Module -> prefix "Module"
-    | `Argument -> prefix "Parameter"
+    | `Parameter -> prefix "Parameter"
     | `ModuleType -> prefix "Module type"
     | `ClassType -> prefix "Class type"
     | `Class -> prefix "Class"
-    | `Page -> []
+    | `Page | `LeafPage | `File -> []
 
-  let make_name_from_path { Url.Path.name; parent; _ } =
+  let make_name_from_path { Url.Path.path_fragment; parent; _ } =
     match parent with
-    | None | Some { kind = `Page; _ } -> name
-    | Some p -> Printf.sprintf "%s.%s" p.name name
+    | None | Some { kind = `Page; _ } -> path_fragment
+    | Some p -> Printf.sprintf "%s.%s" p.path_fragment path_fragment
 
   let render_title (p : Page.t) =
-    format_title p.kind
-      (make_name_from_path p.url)
+    format_title p.url.kind (make_name_from_path p.url)
 end
 
 module Math : sig
--- a/src/document/doctree.ml
+++ b/src/document/doctree.ml
@@ -91,7 +91,7 @@ end = struct
       | Declaration { content; _ } -> walk_documentedsrc content
       | Include i -> walk_items i.content.content)
 
-  let compute (p : Page.t) = walk_items (p.header @ p.items)
+  let compute (p : Page.t) = walk_items (p.preamble @ p.items)
 end
 
 module Shift = struct
@@ -133,7 +133,7 @@ module Shift = struct
         let content =
           {
             page with
-            header = walk_item ~on_sub shift_state page.header;
+            preamble = walk_item ~on_sub shift_state page.preamble;
             items = walk_item ~on_sub shift_state page.items;
           }
         in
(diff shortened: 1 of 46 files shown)


Julow (approved):
LGTM :)

I don't know if `Doctree` is the best place but we don't have a better place at the moment. We can ask this question at if we have more shared rendering functions in the future.

Julow:
Needs some ocamlformatting.

src/document/doctree.ml
```diff
@@ -283,6 +283,35 @@ end = struct
...
+    let prefix s =
+      mk (Types.inline (Text (s ^ " ")) :: Codefmt.code (Codefmt.txt name)) in
+    match kind with
+    | `BaseModule -> prefix "Module"
+    | `Module -> prefix "Module"
+    | `Argument -> prefix "Parameter"
+    | `ModuleType -> prefix "Module type"
+    | `ClassType -> prefix "Class type"
+    | `Class -> prefix "Class"
+    | `Page -> []
```
dbuenzli, line 303: Aren't you dropping the page name here ?

Julow: The page title is embedded in the page it seems.

sabine: I had the tests fail with a duplicate title when I kept the code as it was. The reason is that it was dead code as the title was never rendered into the header anyways, in the case of `Page` (see line 1697 here): 
https://github.com/ocaml/odoc/blob/a52dd8e960f99b38162d40311a9c4512d5741d43/src/document/generator.ml#L1691-L1698

sabine:
Force pushed after `dune fmt`. No other changes.

jonludlam:
Sorry it's taken a while to get to this. I'm unsure about the distinction between `` `BaseModule `` and `` `Module `` - it looks like it's to prevent putting the parent pages's title into a 'base module's title. What do you think about handling this in `make_name_from_path`? e.g.:

```ocaml
let make_name_from_path { Url.Path.name; parent; _ } =
  match parent with
  | None | Some { kind = `Page; _ } -> name
  | Some p -> Printf.sprintf "%s.%s" p.name name
```

I'm not strongly opinionated on this, so happy either way.

sabine:
> What do you think about handling this in make_name_from_path?

Thanks for pointing that out. I took a closer look at why I added `` `BaseModule `` and what is going on in the surroundings of that.

As it stands, `Url.Path.name` is not the name, but a path fragment (it's differs from the name in that every functor parameter has their position prepended to its name), so I renamed that to `Url.Path.path_fragment`.

It turns out that `Page.title` is only used in one single place (HTML document title, all other titles were already being generated based on `Page.url`), so, for the sake of having a single source of truth, and less potential for confusion, this PR now also proposes to remove `title` from `Page`.

(I am not entirely happy with this part of the code where I have to strip the parameter number from `Url.Path.path_fragment` to get the old HTML page title: https://github.com/ocaml/odoc/pull/896/files#diff-6c455c2657c68cf89b442c7b3c19fef8503cf643a31ad0cb81d926d34e47ac10R407-R418.
An alternative to stripping the prefix from the path fragment is to keep `Url.path.name` and add the prefix for `` `Parameter `` to render the Url.Path.t is rendered to a path fragment. That would require adding an integer to `Url.path.kind`'s `` `Parameter ``.)

Opinions?

tmattio:
Gentle ping @Julow @jonludlam, is there anything blocking the merge of the PR? This would be quite useful to make progress on the updated package documentation on OCaml.org

jonludlam:
Looks great now, thanks! Quite happy with the addition of the integer into the `` `Parameter `` constructor.

PR #899, opened 2022-10-18
Math support files

jonludlam:
Math support currently doesn't work because `odoc support-files` doesn't actually write the vendored katex files. This PR fixes that.

Note that dune doesn't quite work yet with this as it's only expecting the original 2 files to be output, and while it will work with the first build, rebuilds don't work as expected. This is related to https://github.com/ocaml/dune/issues/1556.


src/vendor/dune
```diff
@@ -0,0 +1,26 @@
...
+  (:x odoc_vendor.ml)
+  (glob_files *.js)
+  (glob_files *.css)
+  (glob_files_rec *.woff2))
+ (action
+  (progn
+   (setenv
+    SOURCE_DATE_EPOCH
+    0
+    (run ocaml-crunch . -e js -e css -e woff2 -o %{x}.corrected -m plain))
```
Julow, line 20: This prints something, which is not useful to know given how fast it runs.
```suggestion
    (run ocaml-crunch . -e js -e css -e woff2 -o %{x}.corrected -m plain -s))
```

Julow (approved):
Looks good to me :)

src/vendor/dune
jonludlam: Just tried this, but `-s` doesn't work on older crunches, so I'll revert.

jonludlam: Well older crunches produce different output anyway, so that's not much help.

Issue #567, opened 2021-01-21
Support generating ranked search indices


lpw25:
It would be great to generate ranked search indices of the docs.

James Somers at Jane Street wrote a patch that we used to generate a search index for an elastic search instance. The code is [here](https://github.com/lpw25/odoc/tree/elasticsearch) but it was written against the old version of odoc so would need rebasing.

The reason we did not pursue the idea further at the time was that we found that the results were pretty useless without a good mechanism for ranking the results. The most natural approach would be to base the ranking on how often values and types are used in the code base. However, odoc does not currently ever examine OCaml implementations -- it only looks at interfaces -- and it is the uses in the implementations that we would wish to count.

If the patch discussed in #566 is finished then we can use the lists of used paths mentioned in that PR description as the basis for the ranking.

c-cube:
Discussing this over IRC, we mentioned possible client-side solutions (which would work on github hosted doc, unlike elastic search):

- https://stork-search.net/
- https://endler.dev/2019/tinysearch/

(both are in rust compiled to wasm)

PR #909, opened 2022-11-28
Render source code along side documentation

Julow:
Authors: @panglesd @gpetiot @Julow

This PR adds links from items in the documentation to their implementation. Every items that are described by shapes can have such a link.

The source code is rendered in its own HTML page with compile-time syntax highlighting and the algorithms to easily add more anchors or even links from code to documentation later.
Other backends don't render source code.

This uses shapes to lookup the locations in the implementations, which has been added in OCaml 4.14. The feature is disabled for older versions of OCaml.
Shapes are loaded from `.cmt` files, for which the path is guessed from the path of the loaded `.cmti` files. The feature is disabled when loading a `.cmi` or if the `.cmt` failed to load.

Source code is passed to `odoc compile` on the CLI and stored in intermediate `.odoc` files. This avoids adding dependencies to the other commands in the pipeline (the `link` and `html-generate` commands could be run later, on an other machine, etc..) and having to deal with hidden modules and expansion (during link).

During linking, some work is needed to avoid loosing source code for hidden modules and to correctly substitute "source parents" identifiers when expanding.

This is still a work in progress.


lpw25:
> Shapes are loaded from .cmt files, for which the path is guessed from the path of the loaded .cmti files. The feature is disabled when loading a .cmi or if the .cmt failed to load.

I think it would be better to allow the `.cmt` file to be provided explicitly for this purpose with a specific option. That way the build will give errors if it is supposed to support links into the source code, rather than silently failing.

> Source code is passed to odoc compile on the CLI and stored in intermediate .odoc files. This avoids adding dependencies to the other commands in the pipeline

Why not just take the source code as an additional argument to the `html-generate` command? Having two copies of every source file on disk just in case someone copies `odoc` files between machines without also copying the source files seems like a waste. You can also line this up with the provision of a `.cmt` file -- if the `.cmt` file is provided to `compile` then the source needs to be provided to `html-generate` -- to give an explicit error if something goes wrong.

Julow:
> I think it would be better to allow the .cmt file to be provided explicitly for this purpose with a specific option.

What's controlling whether the feature is enabled or not is passing the `.ml`. An option for the `.cmt` could be useful to be explicit about dependencies but it would have no effect if the source code is not also passed.

> Why not just take the source code as an additional argument to the html-generate command?

`html-generate` is not called on hidden compilation units. In this case, an other module would contain the expansion of the hidden module and now also its source code.

Julow:
I cleaned the history a bit: removed useless commits, squashed WIP commits and made sure every commits run the tests.
Then I rebased on top of master.

lpw25:
> During the link phase, odoc turns identifiers into shapes and tries to reduce them

Is this just for local identifiers? For global identifiers, I would hope that they linked to the associated documentation page, which you would get by using odoc's path resolution rather than shapes.

lpw25:
Coming back to my previous question. Would it make more sense to take the ml file at the link step? It seems a waste of disk space to keep a copy of the ml file verbatim in the odoc file when it isn't needed until after link.

Also, if you are reading the shapes from other odoc files, should all odoc files include shapes, not just the ones for which a source file has been provided?

Julow:
> Is this just for local identifiers

We turn unresolved identifiers into a projection into the shape of the current module. If it turns out that the identifiers points to something in an other compilation unit, we allow the shape engine to call back into Odoc's path resolution. We store the compilation unit's (implementation) shape into the `.odoc` files.

> It seems a waste of disk space to keep a copy of the ml file

The link step might be done in a context where the compile's step inputs are not available (eg. a "re-link" step to resolve cross-package references). I believe this makes the driver simpler to write.
The new size of `.odoc` file for the `Link` module is 205KB, the source file is 37KB, so it's indeed a noticable increase.

> should all odoc files include shapes, not just the ones for which a source file has been provided?

Links to other compilation units are rare and we don't expect the feature to be used on a part of the modules. This make that if the feature is not used, no shape is loaded.

We are thinking of always loading the shapes for https://github.com/ocaml/odoc/issues/567

panglesd:
> During the link phase, odoc turns identifiers into shapes and tries to reduce them

Just to clarify this, I meant *odoc* identifiers are turned into shapes, in order to find the location of their implementation. So, it is for links from documentation to source, not the other way around.

For links from implementation to documentation, there can be multiple declaration having the same implementation, so this is not obvious.

For links from source occurrences to documentation, it would indeed be nice to have jump-to-documentation, when possible, by finding the associated odoc identifier.

About local jump to implementation: Currently, local identifiers are directly used as anchor points/links in the rendered source (without involving shapes).

panglesd:
I made a small benchmark for the time it takes, with and without the feature.
I used the odoc driver as test-piece. Since the sources are only added for the odoc sources and not its dependencies, I removed them from the driver:
```diff
diff --git a/doc/driver.md b/doc/driver.md
index 58ed05d5f..9bc186852 100644
--- a/doc/driver.md
+++ b/doc/driver.md
@@ -190,29 +190,9 @@ Additionally we'll also construct a list containing the extra documentation page
 the hierarchy declared above.
 
 ```ocaml env=e1
-let dep_libraries_core = [
-    "odoc-parser";
-    "astring";
-    "cmdliner";
-    "fpath";
-    "result";
-    "tyxml";
-    "fmt";
-    "stdlib";
-    "yojson";
-    "biniou";
-];;
-
-let extra_deps = [
-    "base";
-    "core_kernel";
-    "bin_prot";
-    "sexplib";
-    "sexplib0";
-    "base_quickcheck";
-    "ppx_sexp_conv";
-    "ppx_hash";
-]
+let dep_libraries_core = [];;
+
+let extra_deps = []
 
 let dep_libraries =
     match Sys.getenv_opt "ODOC_BENCHMARK" with
```

I do not count the `dune build` time. Here are the results:

With sources:
```
$ hyperfine --prepare 'dune clean ; dune build' 'dune build @docgen'
Benchmark 1: dune build @docgen
  Time (mean ± σ):     542.5 ms ±   5.4 ms    [User: 494.7 ms, System: 47.8 ms]
  Range (min … max):   536.5 ms … 553.7 ms    10 runs
```

Without sources (on master):
```
$ hyperfine --prepare 'dune clean ; dune build' 'dune build @docgen'
Benchmark 1: dune build @docgen
  Time (mean ± σ):     491.4 ms ±   6.0 ms    [User: 436.4 ms, System: 55.4 ms]
  Range (min … max):   479.7 ms … 501.6 ms    10 runs
```

According to my computations, there is an overhead of 10% for using the source. I think that's acceptable (?), given that there is significantly more work to do (e.g. traversing all typedtrees once, loading other units for their shapes, syntax highlighting)!

lpw25:
> Just to clarify this, I meant odoc identifiers are turned into shapes, in order to find the location of their implementation. So, it is for links from documentation to source, not the other way around.

Good, that makes sense.

> For links from source occurrences to documentation, it would indeed be nice to have jump-to-documentation, when possible, by finding the associated odoc identifier.

Ok, so does that mean that non-local source occurrences are currently not links? That seams fine for version 1, but should definitely be looked at in future.

> About local jump to implementation: Currently, local identifiers are directly used as anchor points/links in the rendered source (without involving shapes).

I take it that means that only identifiers (e.g. `x` or the `M` in `M.foo`) are links rather than full paths (e.g. `M.foo` as a whole linking to a `foo` definition). This seems fine to me. It's possible that when the merlin occurrences work is completed that it will be easier to make this work for whole local paths too.

lpw25:
> The link step might be done in a context where the compile's step inputs are not available

I can't see why that would be the case for any of the use cases I'm aware of. And even if it were we could just get people to store the ml file next to the odoc file in such cases.

panglesd:
Another design decision worth mentioning, is about preprocessing.

As it is, odoc expects the preprocessed (textual, not PPX) source. So for instance, for projects using `cppo`, the generated file has to be given to the `--source` argument.

This has the following advantages:

- The odoc compilation units are similar to the ocaml one.
- Source browsing can be much easier: for instance, reading yojson generated implementation is much easier than the actual source, with many includes and ifdef
- Every file has valid OCaml syntax. This helps the syntax highlighter.
- The driver can generate the hierarchy it prefers, using the `--source-parent` argument.

It has the following drawbacks:

- Some information is losed, such as version-specific code for another version
- The code hierarchy might not be the same as the github one.

Even with the significant drawbacks, I consider the current solution to be better.

That would not be completely straightforward to change: source could not be attached to compilation unit, so they would be generated independently, as "assets" unit I guess. We would need to use the actual `Location.t` instead of the `Shape.Uid.t`.

panglesd:
About giving the source file at odoc-compile-time or link-time (or even html-generate-time), I don't have a super strong opinion (even though, I'd prefer leave it like it is).

I think the trade-off is the following:
- More space used
- Potentially irrelevant information stored in the `.odoc` file, which can be both a waste of time and space.

Indeed, if the sources are only given during the html-generate or linking phase, when compiling, odoc does not know whether some information (such as local jump, maybe type information if we want to add this to the UI at some point) are worth collecting. Of course, this could be specified by a flag, that the driver would need to respect after.

For me, the pipeline makes more sense as it is now, where each unit is "self-contained" in some sense. But I agree that from a space perspective, it is less ideal. If there is an agreement that it's significantly better with the source being given later in the pipeline, we should indeed change this before drivers are written.

jonludlam:
> 



> About giving the source file at odoc-compile-time or link-time (or even html-generate-time), I don't have a super strong opinion (even though, I'd prefer leave it like it is).
> 
> I think the trade-off is the following:
> 
>     * More space used
> 
>     * Potentially irrelevant information stored in the `.odoc` file, which can be both a waste of time and space.

Thinking about this some more, I think this part is quite important. We've spend a _lot_ of time trying to make odoc fast and efficient, both in time and space, and if we're not using the sources at compile/link time when we're doing a lot of loading of odoc files, then we should push this back to the `html-generate` phase.

lpw25:
> Actually, "local path" are handled (see e.g. [line 478](https://choum.net/panglesd/odoc_render_source_files/odoc/source/src/html/generator.ml.html#L478)), as it was sufficiently easy to do (see [implementation](https://github.com/ocaml/odoc/pull/909/files#diff-7b5a84398c5a82e98a9108d0bd54c77d7136877091c9a406be90b4ef3f85bc8cR43-R51)).

Note that this is doing "jump-to-declaration" rather than "jump-to-definition". So given:
```ocaml
module M : sig
  val foo : int
end = struct
  let foo = 5
end
let z = M.foo
```
the `M.foo` will link to the `val foo` rather than the `let foo`.

That might just be fine, and it is certainly easier to implement. Definitely reasonable for a first version.

src/model/paths.ml
```diff
@@ -14,6 +14,10 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
+module Ocaml_path = Path
```
jonludlam, line 17: Do we still need these?

Julow: We don't. I'll remove it.

Julow:
With @panglesd, we changed the pipeline to read the source file at `html-generate`. We need an extra argument on the `compile` command to specify the name of the source file, which is used in identifiers.

The `compile` command now has two new arguments: `--source-parent page-src_parent --source-name foo.ml`.
The `html-generate` command has a new argument: `--source src/foo.ml`.

panglesd:
Update what we have been working on for this PR:

The generation of the "source hierarchy" ([this](https://choum.net/panglesd/odoc_render_source_files/odoc/source/src/index.html)) used to be done by the driver, using a parent-child hierarchy of mld pages.

However, this proved very unpractical: It introduced a lot of conflicts in page names. The page name in the hierarchy were fixed (they had to match the directory name). Often, there are several directories with the same name, or with the same name as another page (such as an `odoc` page for a library, and another `odoc` page for a directory with the same name). 
Solving these name clashes with a carefully crafted search path would be difficult, introduce complex constraints on the search path, and limit the ability to reference a source directory in the documentation.

So, we introduced a new kind in `.odoc` files: "source trees" 
- Compiling a source tree is done via the `odoc source-tree` command, which takes as input a list of relative file paths.
- Compiling units with "links to source" enabled requires passing a source-tree as source-parent, and a relative path.
- `odoc html-generate` a source tree will generate the hierarchy.

The reference driver has been updated, and there is a [test](https://github.com/ocaml/odoc/pull/909/files#diff-e7911079ef963662cfb57e36b723072084e44d5b137a8773781d2770cca6403b) showing the mechanism. Here is a smaller example of the new pipeline:
```shell
# Compiling a parent for everything. NEW: Need to include source tree children.
$ odoc compile -c module-a -c src-source root.mld

# NEW: Compiling a source tree
$ printf "lib/a/a.ml\n" > source_tree.map
$ odoc source-tree -I . --parent page-root -o src-source.odoc source_tree.map

# Compiling with source links enabled. NEW: need to provide full relative path, and the source-tree file.
$ odoc compile -I . --source-name lib/a/a.ml --source-parent-file src-source.odoc a.cmt

# Linking everything. Linking src-source.odoc "does nothing"
$ odoc link -I . page-root.odoc && odoc link -I . a.odoc && odoc link -I . src-source.odoc

# html-generate on source enabled modules require passing the actual source
$ odoc html-generate --source a.ml a.odocl

# NEW: html-generate the source tree generates the source tree
$ odoc html-generate --indent -o html src-source.odocl
```

I believe there is nothing more on our side that we need to finish, so the PR is again ready for review, comment and merge!

jonludlam:
In the interests of consistency, I think we ought to change the semantics of the `source-tree` command. All of the other compile-like commands take the name from the input filename and derive the output filename from it, whereas as it stands, it's the _output_ filename that's semantically important for the `source-tree` command.

Julow:
Other commands take the input file name into account only if the output file name is not specified (`-o`). The output filename is semantically important in every commands. 

Drivers want to specify the output path. We decided to remove unused code and I suggest to deprecate the automatic output file name guessing instead.

doc/driver.md
```diff
@@ -44,31 +40,51 @@ let relativize p = Fpath.(v ".." // p)
 - `odoc/deps/stdlib/Stdlib/index.html` : Module page for the module `Stdlib`
 - `odoc/deps/astring/index.html` : astring main page
 - `odoc/deps/...` : other dependencies
+- `odoc/source/...` : rendered source files
 
-The `odoc` model for achieving this is that we have *pages* (`.mld` files) that have *children* which are either *further pages* (`.mld` files) or *modules* (from `.cmti` files). This {{!page-parent_child_spec} parent/child relationship} is specified on the command line. Parent pages must be *compiled* by `odoc` before their children. Then compiling a page `mypage.mld` will produce the file `page-mypage.odoc`.
+The `odoc` model for achieving this is that we have *pages* (`.mld` files) that have *children* which are either *further pages* (`.mld` files), *modules* (from `.cmti` files), or
+a *source parent*. This {{!page-parent_child_spec} parent/child relationship} is specified on the command line. Parent pages must be *compiled* by `odoc` before their children. Then compiling a page `mypage.mld` will produce the file `page-mypage.odoc`.
```
panglesd, line 46:
```suggestion
*source trees*. This {{!page-parent_child_spec} parent/child relationship} is specified on the command line. Parent pages must be *compiled* by `odoc` before their children. Then compiling a page `mypage.mld` will produce the file `page-mypage.odoc`.
```

panglesd: Actually, I'm not sure which is best: the file is named `source-tree` but is passed as `source-parent`. Maybe we could uniformize this?

In any case, a page can have multiple source-trees as children.

panglesd:
Rebased on `master`.

jonludlam:
OK, I'm happy with this now! Let's get this merged :-)

PR #911, opened 2022-11-29
Update version constraint on Dune

Julow:
Changed in 35cf29da1.


jonludlam:
Ah you're right! good spot thanks :-)

PR #912, opened 2022-11-30
Improvements to the CLI documentation

Julow:
Sort commands in sections and improve the documentation. It's https://github.com/ocaml/odoc/pull/903 without the new documentation page.


jonludlam:
Let's have this one for now, and we can figure out the extra docs and their generation separately.

Issue #694, opened 2021-06-27
Inclusion of Google Fonts by default might create a legal hazard due to GDPR


gadmm:
The default `odoc.css` links to https://fonts.googleapis.com. At https://complianz.io/google-fonts-and-gdpr-does-it-work/ and elsewhere, there are arguments that this is against EU's GDPR. Google Fonts should either be blocked by the webpage until consent is given by the user, or (simpler and more realistic) the fonts should be self-hosted, they say. (Whether this affects only European users of odoc, or all users of odoc serving to clients in the EU, frankly flies over my head.)

Users can already change `odoc.css` to point to self-hosted fonts, but they are probably unaware of the legal hazards of the default settings. In order to err safely on the side of “legal by default”,  it would be nice if:
- the inclusion of remote google fonts was opt-in,
- `odoc` assisted in self-hosting the fonts.

gadmm:
Here is an example where the remote css is replaced with self-hosted fonts, this is fairly simple: https://guillaume.munch.name/software/ocaml/odoc.css. Tools like https://google-webfonts-helper.herokuapp.com/fonts have been helpful to generate the css and to download the font files.

Julow:
I completely agree !

There's a few more reasons to avoid this:
- Google might be using requests made to the fonts to invade privacy (I mean, they certainly do but are only getting a small amount of data).
- The generated documentation is not 100% self contained because it changes its appearance whether an internet connection is available or not.
- Some people don't like when webpages that could be simple are not (contain fonts or CSS that can't be removed without breaking the page)

My opinion is that we should not load any font and use basic fonts (eg. no url in the CSS file and `font-family: Arial, sans-serif;`).
Modifying the CSS file is already encouraged by recent changes (not yet documented though) so having a simpler graphic design by default seems better than the opposite.

gadmm:
You might be interested in the experience of using system fonts from the following website: https://meta.stackexchange.com/questions/364048/we-are-switching-to-system-fonts-on-may-10-2021 (it seems to bring unexpected complications).

As a user I would prefer something that just works and is nice out of the box and does not require me looking into customizing CSS.

Julow:
My comment above is a bit out of scope. Odoc comes with a graphic design by default, so using specific fonts is reasonable.

Perhaps the fonts should be distributed with Odoc and with the generated HTML (like the CSS file) ? The 3 fonts are libre (Noticia Text, Fira Mono and Fira Sans).

[...]

PR #913, opened 2022-11-30
Embed font files in Odoc 

Julow:
Closes https://github.com/ocaml/odoc/issues/694

Remove the remote fonts from the default theme and instead embed them in the theme.

This removes requests to Google's server that users might not consent to.

The font files are part of the theme, they are affected by `--without-theme` and are installed in `share/odoc/odoc-theme/default/`.
The lack of weight variants don't seem to be a problem.

Fira Mono and Fira Sans are from the latest release on https://github.com/mozilla/Fira.
Noticia Text is from the latest release on Google Fonts.

Notes:

- This should not be merge now as it would break the doc built with Dune.
  We should merge this after the release to leave time to update Dune to use `support-files-targets`.

- I removed `highlightjs` from the files installed in `prefix/share`.
  A "theme" should be just the files that are allowed to be tweaked: the CSS and text fonts.
  More consistent with `odoc support-files --without-theme`. If we wish to make highlightjs optional, we should add a flag for that.

- The CSS file is not in the same place within Odoc's source due to some refactoring. I hope no one was relying on that.


dbuenzli:
> The lack of weight variants don't seem to be a problem.

Are you sure you are not falling back on browser font synthesis for bolds and italics (you can check by adding `:root { font-synthesis: none }` in the CSS) ? If that is the case, there's a high probability that type extremists will show up scandalized on the issue tracker (which is, as far as I'm concerned, not an entirely unfair behaviour).

Julow:
You are right, font synthesis is used. There's a few `font-weight: 500` on headings, the TOC and keywords in the code. This rule is shadowed on headings but the difference on keywords and in the TOC is clearly visible. The font synthesis don't render the weight `500` differently from `400` on my monitor at 300% zoom on Firefox.
The previous font had a difference here. It seems that we need 3 more font files ?

- Fira Sans weight 500 (H1 in TOC and H0 and I guess it was intended on H1 in general)
- Fira Mono weight 500 (keywords)
- Noticia Text weight 700 (bold markup)

Julow:
I moved the commits not related to the fonts into https://github.com/ocaml/odoc/pull/915 in case they can be merged sooner.

jonludlam:
We were missing some fonts - notably the italic ones - so I've added a few more. The original css had:

```css
@import url('https://fonts.googleapis.com/css?family=Fira+Mono:400,500');
@import url('https://fonts.googleapis.com/css?family=Noticia+Text:400,400i,700');
@import url('https://fonts.googleapis.com/css?family=Fira+Sans:400,400i,500,500i,600,600i,700,700i');
```

so I've got all of these minus the 600-weight fonts. Looking back, these fonts were first introduced in https://github.com/ocaml/odoc/pull/208 - and that same commit removed all mention of 600 in the font weights, so I don't think we ever used them.

src/html_support_files/odoc.css
```diff
@@ -4,9 +4,105 @@
    %%NAME%% %%VERSION%% */
 
 /* Fonts */
-@import url('https://fonts.googleapis.com/css?family=Fira+Mono:400,500');
-@import url('https://fonts.googleapis.com/css?family=Noticia+Text:400,400i,700');
-@import url('https://fonts.googleapis.com/css?family=Fira+Sans:400,400i,500,500i,600,600i,700,700i');
+/* noticia-text-regular - latin */
+@font-face {
+  font-display: swap; /* Check https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display for other options. */
```
Julow, line 9: I don't think we should configure this. The default is probably the best as we generate a static website with nothing dynamic.

Julow (commented):
Happy with all the font files. Is this ready for merge ?

src/html_support_files/odoc.css
jonludlam: Fair enough - I don't have a strong opinion and was just using the default from google-webfonts-helper. Using the default seems reasonable to me, so I'll push a fix. Thanks!

jonludlam:
I think this is ready. In it goes - thanks, @Julow !

PR #914, opened 2022-12-01
Docs update

jonludlam:


CHANGES.md
```diff
@@ -1,7 +1,17 @@
 Additions
 - New (experimental!) option `--as-json` for the HTML renderer that emits HTML
```
dbuenzli, line 2: Perhaps `unstable` is a better word to indicate that this might change in the future. (That's also what was decided to use upstream for OCaml 5 features)

doc/odoc_for_authors.mld
```diff
@@ -294,6 +294,42 @@ doc.
 *)
 v}
 
+{4:language_headers Language Headers}
+
+As of [odoc.2.1] it is possible to write blocks with explicit language headers,
+similar to markdown code blocks:
```
Julow, line 300: I don't think it's useful to compare with markdown. I think 3 examples is too much for a relatively common feature.

src/vendor/README.md
```diff
@@ -1,168 +1,28 @@
-# Highlight.js
+# Highlight.js CDN Assets
```
Julow, line 1: These files are not useful. I removed them and cleaned up the `vendor` directory in https://github.com/ocaml/odoc/pull/915

Julow (approved):

doc/odoc_for_authors.mld
jonludlam: I think that's reasonable. I'll modify it.

src/vendor/README.md
jonludlam: OK thanks - does that PR also contain the updated highlight.js?

Julow: No, it's the old one but I can cherry pick your commit.

Julow: I've cherry-picked your commit about updating highlightjs into https://github.com/ocaml/odoc/pull/915 and rebased it on top of the new support files directory.

doc/driver.md
```diff
@@ -6,7 +6,7 @@ The document built here includes not only the documentation of `odoc` itself, bu
 docs for a subset of `odoc`'s dependent libraries to show how this may be done. For a much more
 complete and comprehensive use of `odoc`, see the [Voodoo project](https://github.com/ocaml-doc/voodoo), the tool that is being used to build
 the package docs for
-[v3.ocaml.org](https://v3.ocaml.org/).
+[www.ocaml.org/packages](https://www.ocaml.org/packages).
```
Julow, line 9: `www.` is something of the past. ocaml.org don't do anything particular about it, the website is usable both with and without the `www.` and there's no redirect between them.

Julow (approved):

PR #917, opened 2022-12-05
Add prefix "page-" to target name when adding page content to cache

skcho:
This PR adds prefix "page-" to the target name when adding a page content to the cache.

* In `lookup_page` function, it adds prefix "page-" to the target name when lookup.

* However, when adding a page content to the cache (i.e. `build_env_for_page` -> `build` -> `add_unit_to_cache`), it did not add the prefix.

This discrepancy introduced an incorrect reference resolution when there is the same name of page and module. For example, suppose we have a module `Foo` and a page `Foo.mld` and `Foo.mld` includes a reference to `Foo`, i.e. `{!module-Foo}`.

1. It invokes `build_env_for_page` for the page `Foo.mld`. Now, the cache has a mapping for `Foo`, but its value includes the page content only.

2. When resolving `{!module-Foo}`, it tries to find it from the cache. It found, but the result is not for the unit contents.

3. It fails to make a correct link in `Foo.mld`.

### DIFF
--- a/src/odoc/resolver.ml
+++ b/src/odoc/resolver.ml
@@ -190,7 +190,12 @@ let lookup_page ap target_name =
 
 (** Add the current unit to the cache. No need to load other units with the same
     name. *)
-let add_unit_to_cache u = Hashtbl.add unit_cache (unit_name u) [ u ]
+let add_unit_to_cache u =
+  let target_name =
+    (match u with Odoc_file.Page_content _ -> "page-" | Unit_content _ -> "")
+    ^ unit_name u
+  in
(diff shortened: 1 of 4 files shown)


skcho:
Hello @jonludlam. Thank you for the suggestion. I added a test.

BTW, when I run `make test` there are many other changes due to different formatting. Do you have any idea what I am doing wrong? (My `fmt` version is 0.9.0 just in case.)
https://gist.github.com/skcho/7cdeed0478f15681d9b7d7fdb875d040

jonludlam:
Thanks! Will merge when the tests are green.

PR #920, opened 2023-01-06
latex backend: tune escaping rule for labels

Octachron:
When working on ocaml/ocaml#11875 , the escaping rule for `_` in latex label ended up to be more far more bothersome than useful (in other words, I would rather avoid text manipulation in LaTeX whenever possible).

Since escaping `_` is not strictly required in latex, this PR removes the escaping rule in odoc latex backend. 

The second commit goes one step further and make sure that all characters not allowed in latex label are escaped (in particular `%`, `^` and `&`) .


Julow (approved):
Looks good to me :)

jonludlam:
Thanks!

PR #921, opened 2023-01-06
latex math: remove line at the end of equation* environment

Octachron:
It is generally a good idea to avoid spurious blank line before the end of the `equation*` environment. 
For example:
```latex
\begin{equation*}
    \f\hat\xi\,e^{2 \pi i \xi x}
    \,d\xi
    
\end{equation*}
```
In my tests, this blank line breaks at least the amsmath version of `equation*`.
This commit fix this issue that trimming the last part of the raw equation text before cutting a newline to end the `equation*` environment.


src/latex/raw.ml
```diff
@@ -209,10 +209,20 @@ let ocamltag tag pp ppf x = create2 "ocamltag" Fmt.string pp ppf tag x
 
 let math ppf x = Fmt.pf ppf {|$%s$|} x
 
+let remove_empty_last_line x =
```
Julow, line 212: We depend on `astring` in other places, so `String.trim` would work.

Julow (commented):
This string comes directly from the `Math` node in the AST. I think it should be cleaned either in document or in the parser.

Octachron:
It might make sense to trim the string in the AST node, but it is not that straigthforward in simple text backend where the white space might be significant:
```
    let rec collatz n =
      if n = 1 then true
     else if n mod 2 = 0 then collatz (n/2) else collatz (3 * n + 1)

```
ought to be trimmed to
```
let rec collatz n =
  if n = 1 then true
 else if n mod 2 = 0 then collatz (n/2) else collatz (3 * n + 1)
```
rather than
```
let rec collatz n =
     if n = 1 then true
     else if n mod 2 = 0 then collatz (n/2) else collatz (3 * n + 1)
```

Here, I would tend to argue that this a minimal fix specific to the latex backend due to the fact that having an empty new line in the `equation*` brings havoc to some of the most used mathematics packages.

Julow (approved):
I agree, taking care of that in the latex backend makes the most sense to prevent generating invalid syntax.

src/latex/raw.ml
Julow: To trim only on the right, `astring` has `Astring.(String.drop ~rev:true ~sat:Char.Ascii.is_white x)`

Julow:
Merged, thanks!

Issue #479, opened 2020-09-01
Multipage manual in odoc


voodoos:
Since #61, one can write manuals in odoc by creatings `.mld` files and cross-referencing them.

However the left sidebar that contains the table of content will always show only the opened file structure, without links to other pages of the manual.

This makes the discovery of pages of the manual quite difficult as one has to look for links in the text. I think it would be more useful to display in the sidebar, in addition to the detailed outline of the current file, links to the other pages of the manual.

It is not obvious how the list of page-links should be made. It could be established in a number of ways:
- List only the pages which are referenced from the current page.
- List the pages referenced in the current page and the ones that reference the current page.
- List all the `.mld` files in the current folder 
- A tree-display of the `.mld` files in the folder and its subfolders

Another issue would be the name to display in the side bar. It could be:
- The name of the file
- The first header `{1 ...}` in the file

### Example
It could look like that in a three-pages manual :
![image](https://user-images.githubusercontent.com/5031221/91866010-4f5ad080-ec72-11ea-96c3-9bf95863bf18.png)
Where "Emacs reference" and "Vim reference" are not headers of the main page but links to two other pages of the manual named `Emacs reference.mld` and `Vim reference.mld`.

trefis:
I'm not fond of the suggestions you make regarding how to populate the sidebar:
> - List only the pages which are referenced from the current page.

This seems both fairly adhoc, and potentially annoying: what if I'm linking to some page that I don't want to appear there?

> - List the pages referenced in the current page and the ones that reference the current page.

Definitely not fond of such a non-local analysis (and it also seems fairly fragile), though I guess it could be done during the "link" step.

> - List all the .mld files in the current folder
> - A tree-display of the .mld files in the folder and its subfolders

There is no semantic notion of folders. Apart from that, notice that it's something we could already do with compilation units, but do not.

My personal take on the topic is that, for now, you should structure your "manual" in the same way you structure your API docs: the "landing page" should link you to all the relevant places, and it should always be easy (i.e. there should only be one step) to jump back to that page.
As for the future: perhaps in .mld (though I'm not sure why they'd be different in that respect) you want to *manually* write your side bar, or rather: you want to manual write additions to the generated sidebar. It doesn't seem crazy to me to add a new construct to the markup-language just for that (and it would also be usable in API pages, which seems nice)

dbuenzli:
I also have @voodoos needs, but I concur with @trefis. 

[...]

Issue #127, opened 2018-04-17
Improve navigation (TOC, index, etc.)


aantron:
We need:

- [ ] A good table of contents in each module, manual page, etc. I don't know if this should be a global index, or local to that page. It might depend on the project being documented.
- [ ] If local, we may or may not need a good global index.

odig already generates many different kinds of index. For an example, see [docs.mirage.io](http://docs.mirage.io).

cc @yawaramin @bluddy @pmetzger

bluddy:
I think a sidebar showing the modules of the package in a hierarchy  as [docre](https://jaredforsyth.com/reason-react/api/ReasonReact.html#type-componentSpec) does it (though preferably foldable) would go a long way. As it is, it's too easy to get lost.

dbuenzli:
> odig already generates many different kinds of index. For an example, see docs.mirage.io.

Note that `odig` generates only *packages* indexes (from the `opam` metadata). What is really missing for `odoc` to finally match the features of `ocamldoc` is an index page for all the identifiers of a module. 

> I think a sidebar showing the modules of the package in a hierarchy as docre does it (though preferably foldable) would go a long way. As it is, it's too easy to get lost.

Regarding module indexes and `docre` I personally don't find the side bar very usable and rather messy (as is their typography in general, the amount of information density shown on a screenful is extremely poor, the usual sin of supposedly good looking typography). 

For modules indexes if you take the current `odoc` navigation structure it's quite easy to get to the module list just hit the `up` link one or two times (module hierarchies are and should be shallow) and you are there which seems good enough.

Personally I think that a side bar (that collapses if the page is not wide enough) with only the `{1` and `{2` header level should be enough (and if browsers provided ui to the html outline algorithm we wouldn't even need that...).

dbuenzli:
> For modules indexes if you take the current odoc navigation structure it's quite easy to get to the module list just hit the up link one or two times (module hierarchies are and should be shallow) and you are there which seems good enough.

(Or you can simply hit the package link which is more predictable)

rizo:
I'm currently working on UX features for the documentation generated by odoc in some of my libraries and here are some of my thoughts about indices and search:

- Global package-wise indices are nice for navigation (in the same way sidebars with the directory structure in text-editors are). But they become less useful and messy when the project structure is too module-heavy and nested. (I personally hate horizontal scrolling in sidebars!)
[...]

PR #923, opened 2023-01-19
WIP: improve the navigation menu

gpetiot:
This is WIP, inspired by #127 and #479

So far, the data is extracted from the .odoc file (without links), but it can surely be improved.


dbuenzli:
I doubt it's possible to see what you are trying to do through html diffs – and thus assess whether this is an *improvement*. Would you mind coming up with a description or a design mockup of what you do and show us how it fares on a non-trivial selection of packages ?

gpetiot:
Not definitive but the design I'm going for is something akin to https://hexdocs.pm/phoenix/Phoenix.html, with a tab to navigate through the contents of the page, and a tab to navigate through the definitions.

dbuenzli:
I see thanks.

Note that `ocaml.org` also seems to be working on similar functionality, could we maybe establish a common ground ?

In particular I think it's nice if standalone `odoc` docsets are not necessarily one-to-one but at least give a good idea on how your docs are going to show up on `ocaml.org` since the navigation in presence does influences the way you design your docs.

tmattio:
> Note that ocaml.org also seems to be working on similar functionality, could we maybe establish a common ground ?
> 
> In particular I think it's nice if standalone odoc docsets are not necessarily one-to-one but at least give a good idea on how your docs are going to show up on ocaml.org since the navigation in presence does influences the way you design your docs.

Absolutely!

At the moment ocaml.org uses a custom design for the odoc-generated documentation, but a medium-term goal is to re-use what odoc generates (or at least be as close as possible). When odoc generates a sidebar with a global navigation, I think we can start exploring how to converge.

Inversely, it would be interesting to use the learnings we've had on ocaml.org to discuss the design of the navigation in odoc.
In particular, @sabine spent significant time iterating on the design, interviewing users, etc. and will certainly be able to share insights on the UX that are relevant to the current PR.

sabine:
Adding to @tmattio's explanation: We do currently have a UX/UI designer working on improving the navigation on ocaml.org and, in particular improving the package documentation navigation, so yes, I totally agree, converging on something that can be upstreamed would be a major goal.

gpetiot:
Closing this PR because it needs (if ever) to be restarted with a clear plan in mind.

PR #928, opened 2023-02-14
Add emoji before alert message

yawaramin:
And update path of CSS file in repo.


yawaramin:
Looks like this:

Screenshot of new alert emoji

yawaramin:
Bunch of snapshot test failures. Not sure how to correct. The [contributing doc](https://github.com/ocaml/odoc/blob/3d2e55582cccbb0314850356139bd0d6171ea2c8/doc/contributing.mld#L9) only says:

> You can edit CSS using your browser's {{:https://developer.mozilla.org/en-US/docs/Tools}developer}
{{:https://developer.chrome.com/docs/devtools/}tools}. Then send us a PR for the same changes made to this file.

Any advice appreciated.

panglesd:
I think that `dune runtest --auto-promote` will solve the test failures.

I don't know exactly what is `odoc`'s stance wrt. css and icons. I would say that `odoc`'s css aims at being the most basic, so that extending it is as easy as possible.
But, currently there are no icons in the whole output, and I'm not sure how well they would integrate in themes that extend odoc's one, such as odig and ocaml.org. More subtle changes (such as putting `Alert` in bold) would be better suited for `odoc`'s stylesheet I think.

That's why I suggested opening this PR on ocaml.org.

dbuenzli:
> But, currently there are no icons in the whole output, and I'm not sure how well they would integrate in themes that extend odoc's one, such as odig and ocaml.org.

There's no extension in `odig` either you use `odoc`'s theme (via `odig odoc-theme set odoc.default`) or one of the built-in `odig`'s one which are devised from scratch. In the former case this change will show up in the latter cases it won't.

Regarding `ocaml.org` don't know how it's built. But of course if they also devised their own stylesheet it won't show up there either.

panglesd:
Thanks for the correction! I always thought they were based on odoc's stylesheet.

yawaramin:
Looks like snapshot diff failures specifically because of formatting: https://ci.ocamllabs.io/github/ocaml/odoc/commit/8b5498b9054b8daa9412f090b38d7c596ae1b8ca/variant/ubuntu-22.10-5.0_opam-2.1#L1052-1052

```diff
@@ -6836,7 +6898,8 @@ let%expect_test _ =
                       "`Type": [
                         {
                           "`ModuleType": [
-                            { "`Root": [ "Foo", "`TModule" ] }, "Bar"
+                            { "`Root": [ "Foo", "`TModule" ] },
+                            "Bar"
                           ]
                         },
                         "baz"
```

yawaramin:
Downgraded to `yojson.1.7.0` and checked in the output from `dune runtest --auto-promote`.

yawaramin:
Rebased on latest master. Test is now failing with exact opposite diff of the one I got after running `dune test --auto-promote`: https://ci.ocamllabs.io/github/ocaml/odoc/commit/c271f3c6ef56af4c71188299606c226c2517637c/variant/debian-11-4.04_opam-2.1

yawaramin:
Bump, any interest in merging?

jonludlam:
Hey @yawaramin - I like this thanks, and thanks for the rebase too.

Julow (approved):
"deprecated" alerts are internally turned into the corresponding odoc tag (class is `.deprecated`). In this example, only the second alert renders with an emoji:
```ocaml
type a
[@@alert deprecated]

type b
[@@alert unsafe]
```

It makes sense for Odoc to interpret the "deprecated" alert, attribute and tag in the same way. Perhaps it can have a different emoji ?

yawaramin:
Oh, I had thought `deprecated` was just rendered as a regular `alert`, and intended to use the same emoji for both. Just added the class to the stylesheet.

Julow:
This affects deprecated tags and attributes as well:
```ocaml
type a
(** @deprecated *)

type b
[@@deprecated]
```

This is fine for me, I'm in favor of merging.

Issue #857, opened 2022-05-31
Section labelling gets confused by submodule


dbuenzli:
This is a continuation of https://github.com/ocaml/odoc/pull/773

This code: 

```ocaml
(** Outer

    See the {{!first}first outer section}. *)

(** {1:first First outer section} *)

(** Inner *)
module A : sig

  (** {1:first First inner section} *)

end
```

Now works correctly, in the HTML output the reference links to the section of the toplevel module and no warning is reported. However moving the header after the inner module:

```ocaml
(** Outer

    See the {{!first}first outer section}. *)

(** Inner *)
module A : sig

  (** {1:first First inner section} *)

end

(** {1:first First outer section} *)
```

Breaks the reference in the HTML. The reference correctly references a `#first` id but somehow `odoc` decided to label the anchor with `first_2`: 

```
<h2 id="first_2"><a href="#first_2" class="anchor"></a>First outer section</h2>
```

dbuenzli:
> I believe

I confirm.

PR #931, opened 2023-03-01
Fix 857

EmileTrotignon:
Fix issue #857

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -406,12 +406,10 @@ module Page = struct
     Utils.list_concat_map ~f:(include_ ~config) subpages
 
   and page ~config p : Odoc_document.Renderer.page list =
-    let { Page.preamble; items = i; url } = Doctree.Labels.disambiguate_page p
-    and subpages =
-      (* Don't use the output of [disambiguate_page] to avoid unecessarily
-         mangled labels. *)
-      subpages ~config @@ Doctree.Subpages.compute p
+    let ({ Page.preamble; items = i; url } as p) =
+      Doctree.Labels.disambiguate_page ~enter_subpages:false p
     in
+    let subpages = subpages ~config @@ Doctree.Subpages.compute p in
     let resolve = Link.Current url in
     let i = Doctree.Shift.compute ~on_sub i in
     let uses_katex = Doctree.Math.has_math_elements p in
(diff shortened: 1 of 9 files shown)


test/xref2/github_issue_857.t/a.mli
```diff
@@ -0,0 +1,7 @@
+module type A = sig
+
+  (** {1:first First inner section} *)
+
+end
+
+(** {1:first First outer section} *)
```
Julow, line 7: To best reproduce the issue, you need a reference to the outer section and to check that the generated link is correct.
Also, `A` doesn't need to be a module type.

Julow (approved):
Looks good :)

test/xref2/github_issue_857.t/a.mli
panglesd: In fact, the PR does not rewrite references: it only fixes the label disambiguation, which was too aggressive in the HTML case.

Label disambiguation cannot be left to the user: two titles with the same name (which is valid) would generate the same label, which would need disambiguation. On the contrary, references should not be rewritten (I think): for instance, in:
```ocaml
(** {1 A title} *)
(** {1 A title} *)

(** reference to {!a-title} *)
(** reference to {!a-title_2} *)
```
the first reference points to the first title, the second to the second, and none of them should be rewritten.

src/html/generator.ml
```diff
@@ -406,12 +406,10 @@ module Page = struct
     Utils.list_concat_map ~f:(include_ ~config) subpages
 
   and page ~config p : Odoc_document.Renderer.page list =
-    let { Page.preamble; items = i; url } = Doctree.Labels.disambiguate_page p
-    and subpages =
-      (* Don't use the output of [disambiguate_page] to avoid unecessarily
-         mangled labels. *)
-      subpages ~config @@ Doctree.Subpages.compute p
```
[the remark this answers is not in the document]
panglesd: We use the output of `disambiguate_pages`, but a flag is passed to this function to avoid unnecessary mangled labels.
We could add a comment along the lines of:
> Do not enter subpages when disambiguating to avoid unecessarily mangled labels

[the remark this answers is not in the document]
Julow: You are right, I missed that. No need to add a comment now that the interesting part is gone.

test/xref2/github_issue_857.t/a.mli
Julow: You are right. No need to add references as we know what they'll do.

Disambiguated labels are used in the TOC however, which would make sense to grep.

Issue #932, opened 2023-03-02
Missing reference syntax for extension declarations


dbuenzli:
```ocaml
(* Consider this extensible type *)

type more = ..

type more +=
| At_the_door
| Behind_the_floor


(** It seems I can refer to {!more}, {!extension-At_the_door}
    {!extension-Behind_the_floor}. But not the point of extension
    (despite the fact that it has an anchor in the rendering (namely
    #extension-decl-At_the_door)). *)


```

PR #933, opened 2023-03-02
Allow to omit parent type in constructor reference.

### LINKED ISSUE #447: Reference to constructor in module
With the new model, the reference `M.Foo` doesn't resolve but `M.t.Foo` does.

```ocaml
module M : sig
  type t = Foo
end

(** {!constructor:M.Foo} {!constructor:M.t.Foo} *)
```

In `Paths_types`, constructor parent should be `parent` or a new `constructor_parent` instead of `datatype`.

Also, the type for references allows extension and exception constructors in `{!constructor:..}` references but won't resolve them.

panglesd:
Fix #447.

Currently, it is not possible to refer to constructors without specifying the parent type, except if they are in scope.

This PR fixes this by extending parents of constructor references to signatures.

The first commit adds the test. The second fixes the issue similarly to what it is done with `field` (which already allows omitting the type parent). The last two commits are refactoring that improve the guarantee given by the subtyping.

(Made with @EmileTrotignon)

### DIFF
--- a/src/xref2/ref_tools.ml
+++ b/src/xref2/ref_tools.ml
@@ -292,8 +292,7 @@ module T = struct
     | `ClassType _ as e -> `CT (CT.of_element env e)
 
   let in_env env name =
-    env_lookup_by_name Env.s_datatype name env >>= fun e ->
-    Ok (of_element env e)
+    env_lookup_by_name Env.s_type name env >>= fun e -> Ok (of_element env e)
 
   (* Don't handle name collisions between class, class types and type decls *)
   let in_signature _env ((parent', parent_cp, sg) : signature_lookup_result)
--- a/src/xref2/ref_tools.ml
+++ b/src/xref2/ref_tools.ml
@@ -279,15 +279,6 @@ module DT = struct
   let of_component _env t ~parent_ref name = Ok (`Type (parent_ref, name), t)
 
   let of_element _env (`Type (id, t)) : t = (`Identifier id, t)
-
-  let in_env env name =
-    env_lookup_by_name Env.s_type name env >>= fun e -> Ok (of_element env e)
-
-  let in_signature _env ((parent', parent_cp, sg) : signature_lookup_result)
-      name =
-    let sg = Tools.prefix_signature (parent_cp, sg) in
-    find Find.datatype_in_sig sg name >>= fun (`FType (name, t)) ->
-    Ok (`Type (parent', name), t)
 end
 
 module T = struct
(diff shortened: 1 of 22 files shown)


src/xref2/ref_tools.ml
```diff
@@ -415,14 +406,30 @@ module CS = struct
     env_lookup_by_name Env.s_constructor name env
     >>= fun (`Constructor (id, _)) -> Ok (`Identifier id :> t)
 
-  let in_datatype _env ((parent', t) : datatype_lookup_result) name =
+  let got_a_field name =
+    (* Let's pretend we didn't see the field and say we didn't find anything. *)
+    Error (`Find_by_name (`Cons, name))
+
+  let in_parent _env (parent : label_parent_lookup_result) name =
```
Julow, line 413: Perhaps we need a `_lookup_result` that correspond to `Reference.Parent` to remove the impossible cases ?

test/xref2/github_issue_447.t/a.mli
```diff
@@ -0,0 +1,11 @@
+type u = Foo
+
+(** {!constructor-Foo} {!u.constructor-Foo} *)
+
+module M : sig
+  type t = Foo
+end
+
+(** {!M.constructor-Foo}
```
Julow, line 9: It could be worth checking for `M.Foo` and `M.t.Foo`, which works fine in my testing.

Julow (approved):
Looks good :)

test/xref2/github_issue_447.t/a.mli
panglesd: Done!

dbuenzli:
@panglesd since you are in the area do you maybe see a fix for https://github.com/ocaml/odoc/issues/932 at the same time ?

panglesd:
The new commits consist of:

- renaming types, since `parent` wasn't explicit enough
- implementing @Julow's comment https://github.com/ocaml/odoc/pull/933#discussion_r1124211379

About https://github.com/ocaml/odoc/pull/933#discussion_r1124211379, in the end I'm not sure yet that it's a good idea. The code is nicer, more specific and less case to handle. But, the error message in case of a `class` found as a parent of a constructor is less explicit: see the error message [before](https://github.com/panglesd/odoc/blob/d3aedc9cac618b03d59f62adda6979a2c16dba27/test/xref2/github_issue_447.t/run.t#L13) and [after](https://github.com/panglesd/odoc/blob/b1f8292cfd56c14eeaba45c090687051a895b79a/test/xref2/github_issue_447.t/run.t#L13) in [this case](https://github.com/panglesd/odoc/blob/b1f8292cfd56c14eeaba45c090687051a895b79a/test/xref2/github_issue_447.t/a.mli#L13..L15).

> do you maybe see a fix for https://github.com/ocaml/odoc/issues/932 at the same time ?

Not directly, but since this PR made me much more comfortable with this part of the code, I'll have a look!

Julow:
> But, the error message in case of a class found as a parent of a constructor is less explicit

This error message could perhaps be implemented in general ? The previous message is not satisfactory because it seems to be about `t.A` where it's actually about `t`:
```
Warning: Failed to resolve reference unresolvedroot(t).A is of kind class but expected signature or type
```
A new kind of error message is needed: "\<parent> is of unexpected kind".

EmileTrotignon:
I think this should be rebased, a better error message added, and then merged.

panglesd:
This is rebased, but I did not implement a better error message. Do you think it could be implemented in another PR? From what I remember from March 30, it is some non-negligible amount of work, but I don't remember why.

Issue #895, opened 2022-09-29
Incorrect method types in generated documentation


zoggy:
Hello,

method types generated in documentation are incorrect, they include the implicit object parameter.

For example, with the following definition:

```ocaml
class foo =
  object
    method bar = 1
  end
```
the documentation contains:

```ocaml
method bar : < bar : int.. > -> int
```

panglesd:
Fixed by #935

PR #934, opened 2023-03-02
fix PR#895: incorrect output of method types

zoggy:
Fix https://github.com/ocaml/odoc/issues/895 by removing the first arrow in method type.


Julow (changes requested):
This is not the right fix. The unwanted argument is not added when compiling an interface:

```ocaml
class foo :
  object
    method x : int -> int
  end
```

The bug occurs only when compiling an implementation (`.cmt`), the right place might be in `src/loader/cmt.ml`, perhaps in the `read_class_field` function.

zoggy:
Indeed. Thanks. I close this one and will make a new PR.

PR #935, opened 2023-03-03
fix issue #895: incorrect output of method types

zoggy:
fixing https://github.com/ocaml/odoc/issues/895
after some hint in previous PR https://github.com/ocaml/odoc/pull/934

### DIFF
--- a/src/loader/cmt.ml
+++ b/src/loader/cmt.ml
@@ -239,7 +239,15 @@ let rec read_class_field env parent cf =
         | Tcfk_virtual typ ->
             true, read_core_type env typ
         | Tcfk_concrete(_, expr) ->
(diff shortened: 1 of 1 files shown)


Julow (commented):
This should be done in the `Tcfk_concrete` case or it would incorrectly remove an argument from virtual methods:

```ocaml
class virtual foo =
  object
    method x x = x + 1
    method virtual y : int -> int
  end
```

Renders to:

```
method x : int -> int
method virtual y : int
```

Also, I can't be sure that this is correct in all future cases so I'd appreciate a comment explaining why.

zoggy:
I fixed to do this hack only in concrete case and added a comment.

Julow (approved):
Looks good to me :) Thanks!

PR #936, opened 2023-03-03
html: Remove extra space in class declarations

Julow:
Class declarations used to render like this in HTML:

```
class  foo : object ... end
```


panglesd (approved):
Looks good to me!

PR #938, opened 2023-03-07
Clarify "odoc for library authors" documentation

### LINKED ISSUE #937: Slightly confused by the section/label documentation for odoc
I find the current content of [Odoc for authors > Sections and Headings](https://ocaml.github.io/odoc/odoc_for_authors.html#sections) confusing.

### Mysterious heading level 6

https://github.com/ocaml/odoc/blob/3d2e55582cccbb0314850356139bd0d6171ea2c8/doc/odoc_for_authors.mld#L717-L731

Confusing: the grammar mentions heading levels from 0 to 6, but the text only mentions levels from 1 to 5. Right below there is a mention of level 0, but no mention of level 6. Is the grammar wrong and 6 is not supported? If it is right, please mention it in the textual description somewhere.

### Confusion around variables: placeholders or fixed syntax?

https://github.com/ocaml/odoc/blob/3d2e55582cccbb0314850356139bd0d6171ea2c8/doc/odoc_for_authors.mld#L727-L744

In this example, `label` is used sometimes as a placeholder for arbitrary user-chosen content, `foobar` in the example, and sometimes as fixed keyword syntax I believe. This is very confusing! 

### Inconsistent syntax for placeholders

More generally, this documentation page is inconsistent in its handling of placeholders, which are sometimes written as plain variables (`{b text}`), sometimes using an implicit placeholder syntax using angle brackets; `{@<language>[`, `{m <latex>}`. I wish you would be consistent about the syntax of placeholders.

### Is it `label`, or `section`, or both?

The syntax summary *might* suggest that `label` or `section` are synonyms of each other. Or maybe it suggests that `section` can be used in a subset of cases, with `label` being the general construct. I don't know.

https://github.com/ocaml/odoc/blob/3d2e55582cccbb0314850356139bd0d6171ea2c8/doc/odoc_for_authors.mld#L393

But then the documentation I quoted above only mentions `label:foo` and never `section:foo`. This is confusing. If both are supported, could the documentation mention both?

panglesd:
Fix #937:

- Be more consistent with placeholder syntax
- Clarify deprecated reference prefixes
- Clarify heading levels section
- Use non-deprecated `section` instead of `label` in examples.

### DIFF
--- a/doc/odoc_for_authors.mld
+++ b/doc/odoc_for_authors.mld
@@ -232,36 +232,36 @@ from OCamldoc.
 Text within the comments can be formatted using the following markup. Firstly,
 the simple typesetting markup:
 
-- [{b text}] bold
-- [{i text}] italic
-- [{e text}] emphasis
-- [{^ text}] superscript
-- [{_ text}] subscript
+- [{b <text>}] bold
+- [{i <text>}] italic
+- [{e <text>}] emphasis
+- [{^ <text>}] superscript
+- [{_ <text>}] subscript
  
 {3:lists Lists}
 Unordered lists:
 {v
-{ul {- item}
-    {- item}
-    {- item}}
+{ul {- <item>}
+    {- <item>}
+    {- <item>}}
 v}
 and ordered lists:
 {v
-{ol {- item 1}
-    {- item 2}
-    {- item 3}}
+{ol {- <item 1>}
+    {- <item 2>}
+    {- <item 3>}}
 v}
 There is also an abbreviated syntax for lists. The above could be written:
 {v
-- item
-- item
-- item
+- <item>
+- <item>
+- <item>
 v}
 and
 {v
-+ item 1
-+ item 2
-+ item 3
++ <item 1>
++ <item 2>
++ <item 3>
 v}
 
 {3:code_blocks Code Blocks}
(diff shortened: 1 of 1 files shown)


dbuenzli (approved):
Very quick scan but looks good.

jonludlam:
Thanks @panglesd! @gasche, does this look good to you?

doc/odoc_for_authors.mld
```diff
@@ -297,7 +297,7 @@ v}
 {4:language_headers Language Headers}
 
 As of [odoc.2.2] it is possible to write blocks with explicit language headers.
-These blocks can be written using the enclosing tags [{@<language>[ ... ]}].
+These blocks can be written using the enclosing tags [{@<language>[ <code block> ]}].
```
gasche, line 300: In general I like the consistent use of placeholder syntax, but specifically when you replace `...` I'm not fond of the result. The problem here is that the name of the placeholder is tautological, and at the same time slightly wrong at places -- here there is a confusion between "a code block" and "the content of a code block", I guess it should be `<code>` instead. The nice thing with named placeholders is that they carry meaning and can be referred to in the text; but in all examples that were using `...` before we don't really need either of these two aspects.

If it was my PR, I would keep the `...` unchanged.

gasche (commented):
Looks nice, thanks for the quick reply. I added a minor comment about `...`.

doc/odoc_for_authors.mld
panglesd: I was hesitant for this part: I went for changing the `...` first, but I just changed it back since you have a clearer opinion than me!

PR #939, opened 2023-03-22
Treat *.ml{i} in Linguist as OCaml, mark *.t as undetectable

kevinji:
The Linguist stats on the bottom-right of the main page are incorrectly recognizing some files as SML; this fixes them to be OCaml instead.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.ml linguist-language=OCaml
+*.mli linguist-language=OCaml
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 *.ml linguist-language=OCaml
 *.mli linguist-language=OCaml
+*.t -linguist-detectable


Julow (commented):
Thanks!

According to the search bar, many of the `.t` files deep in the `test/` directory are recognized as Perl or Raku. It seems that Linguist doesn't support cram tests, perhaps we can mark them as plain text ?

kevinji:
@Julow Just pushed a change to make `*.t` as undetectable.

Julow (approved):
Looks good to me :)

In the long term, it would be nice to have support for cram tests and odoc's `.mld` in Linguist.

panglesd:
From my tests, only some almost empty `.ml` seems to be recognized as Standard ML: https://github.com/search?q=repo%3Aocaml%2Fodoc+language%3A%22Standard+ML%22+&type=codelegacy

image

Julow:
@panglesd This looks like a bug in Linguist (OCaml should be selected over Standard ML in an OCaml project with `dune` and `opam` files) but that doesn't mean we shouldn't merge this.

panglesd (approved):
Oh sorry, I read too fast the thread, I thought it was `*.t` which was recognized as SML...

Yes, sure, let's merge!

gpetiot (approved):

PR #940, opened 2023-03-24
Fix hidden flag

panglesd:
Previously, the `--hidden` flag was used to set only the `Root.Odoc_file.compilation_unit` `hidden` field value.

Now, we also use it to set the `Lang.compilation_unit` `hidden` field value, so that they are in sync.

This allows to use the `--hidden` flag to not generate documentation, for instance when compiling a module solely to render its source (which happens for instance in the `single_mli` test, or in real life, `Astring` mli).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@ UNRELEASED
 Additions
 - Source code rendering (@Julow, @panglesd, #909)
 
+Bugfixes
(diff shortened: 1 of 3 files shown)


gpetiot (approved):

Julow (approved):
Looks good to me as well :)

Issue #779, opened 2021-11-16
Cross page section link using labels don't expand to the section title


NathanReb:
When cross referencing a section from a different page without a specified text for the link, the label will not be expanded into the corresponding section's title.
For example, if you define the following section in `a.mld`:
```
{1:some_label Some Grand Title}
```

And then later reference it from `b.mld` as follows:
```
See {!page-a.some_label}
```

The paragraph will read "See some_label".

NathanReb:
This behaviour was originally described in the documentation added in #775. @Julow confirmed it was a bug and not intentional so I created this issue instead.

dbuenzli:
Should be fixed by #1116

PR #943, opened 2023-03-29
Fix rendering of references to label

### LINKED ISSUE #941: Title splicing problems on master
It seems that when we fully qualify a section reference (or make one from another module) the title does not get spliced.
```
> cat bla.mli 
(** {2:splice_me Splice me}

    This is correct {!splice_me}.

    This is incorrect {!Bla.splice_me}. *)
```
Screenshot 2023-03-29 at 09 17 43

Not sure whether it's regression or not.

gpetiot:
First draft to fix #941

Apparently the reference is parsed and built correctly so I only patched the rendering.

This is the first time I'm touching this part of the code so I'm not sure if this is the right fix, let me know what you think.

### DIFF
--- a/test/xref2/github_issue_941.t/run.t
+++ b/test/xref2/github_issue_941.t/run.t
@@ -1,0 +1,16 @@
+A quick test to repro the issue found in #941
+
+  $ ocamlc -bin-annot -c foo.mli
+
+  $ odoc compile foo.cmti
+  $ odoc link foo.odoc
+
+  $ odoc html-generate --indent -o html/ foo.odocl
+
+The rendered html
+
+  $ cat html/Foo/index.html | grep "<h3" -A 3
+     <h3 id="splice_me"><a href="#splice_me" class="anchor"></a>Splice me</h3>
+     <p>This is correct <a href="#splice_me">Splice me</a>.</p>
+     <p>This is incorrect <a href="#splice_me"><code>splice_me</code></a>.</p>
+    </div>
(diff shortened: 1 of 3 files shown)


test/xref2/github_issue_941.t/run.t
```diff
@@ -0,0 +1,16 @@
...
+  $ odoc link foo.odoc
+
+  $ odoc html-generate --indent -o html/ foo.odocl
+
+The rendered html
+
+  $ cat html/Foo/index.html | grep "<h3" -A 3
+     <h3 id="splice_me"><a href="#splice_me" class="anchor"></a>Splice me</h3>
+     <p>This is correct <a href="#splice_me">Splice me</a>.</p>
+     <p>This is incorrect <a href="#splice_me">splice_me</a>.</p>
```
Julow, line 15: This is not the expected output. The `<code>` element was fine here, the problem is that the `~text` argument was `None`.
The right value should be generated in `Link.comment_inline_element`.

gpetiot: I'v added an ad-hoc case in the pattern matching (up to this point the resolved reference is identical to the "correct case" of the example), let me know if there is a prettier fix

Julow:
The fixes the specific issue but not the underlying problem (https://github.com/ocaml/odoc/issues/779). I added a failing test to your branch.

`Env.lookup_by_id` doesn't lookup things that are nested into modules that are not opened or in pages. It's a weird function that should perhaps be renamed/documented better.
Function that deconstruct identifiers and really lookup things up are in `Tools`. There's no such function for labels and we need one.

Issue #613, opened 2021-02-28
HTML renderer, broken references no longer show what is broken


dbuenzli:
There used to be a tooltip (i.e. a `title` attribute) indicating what failed to resolve. It would be nice to have that back.

Issue #372, opened 2019-07-28
References to variant cases: do not splice the type name in. 


dbuenzli:
If I have : 
```ocaml
module Bla : sig
  type ha = Alpha | Beta
end

Use {!Bla.Alpha} to say alpha.
```
The latter line gets rendered as:
```
Use Bla.ha.Alpha to say alpha.
```
This should simply be `Bla.Alpha`, I don't think `odoc` should invent OCaml syntax that doesn't exist.

MavenRain:
I'd like to take this one.

Plan: in `Odoc_document.Comment.Reference.render_resolved`, stop rendering the
type component when building the text of a reference to a constructor or a
record field, so `{!Bla.Alpha}` renders as `Bla.Alpha` rather than
`Bla.ha.Alpha`.  The enclosing module path is kept, and the anchor is untouched
(it is built from `Reference.Resolved.identifier`, a separate traversal, so
links keep pointing at `#type-ha.Alpha`).

Two existing cram expectations move as a result (`github_issue_447.t` and
`reference_to_polymorphic.t`), plus a new `github_issue_372.t`.

One question I'd like your call on before you review: for a polymorphic
constructor reached through a module, e.g. ``{!Bla.sw.`On}``, I currently
render ``Bla.`On``.  A polymorphic variant is never module-qualified in OCaml
either, so the alternative is to render just `` `On ``.  I went with the
smaller change (drop only the type component, uniformly), but I am happy to
switch.

PR #945, opened 2023-03-30
Add a tooltip to references with text

Julow:
Fix https://github.com/ocaml/odoc/issues/613

The tooltip shows what would be rendered instead if no text was written. This is currently the original reference.
This is most useful for references that do not resolve, because the original reference is not visible at all.

The tooltip is generally easier to read than the target URL shown by the browser on hover and might benefit from smarter rendering of references in the future (eg. https://github.com/ocaml/odoc/issues/372)


panglesd (approved):
Nice! Looks good!

PR #946, opened 2023-03-31
Remove 'Unresolved_apply' errors

### LINKED ISSUE #944: Hide errors directly caused by lookup failures
Example taken from https://github.com/ocaml/odoc/issues/715:

```
File "/Users/dbuenzli/sync/repos/astring/_b0/brzo/ocaml-doc/odoc/astring/astring.odoc":
Couldn't find the following modules:
  Stdlib
File "/Users/dbuenzli/sync/repos/astring/_b0/brzo/ocaml-doc/odoc/astring/astring.odoc":
Failed to lookup type unresolvedroot(Stdlib).Map.Make(unresolvedroot(Stdlib).String).t Parent_module: Unresolved apply
File "/Users/dbuenzli/sync/repos/astring/_b0/brzo/ocaml-doc/odoc/astring/astring.odoc":
Failed to lookup type unresolvedroot(Stdlib).Map.Make(unresolvedroot(Stdlib).String).t Parent_module: Unresolved apply
File "/Users/dbuenzli/sync/repos/astring/_b0/brzo/ocaml-doc/odoc/astring/astring.odoc":
Failed to lookup type unresolvedroot(Stdlib).Set.Make(unresolvedroot(Stdlib).String).t Parent_module: Unresolved apply
File "/Users/dbuenzli/sync/repos/astring/_b0/brzo/ocaml-doc/odoc/astring/astring.odoc":
Failed to lookup type unresolvedroot(Stdlib).Set.Make(unresolvedroot(Stdlib).String).t Parent_module: Unresolved apply
```

The interface contains:

```ocaml
val f : Set.Make(String).t -> unit
```

The `Stdlib` is not available in the search path, internally causing a large amount of lookup failures errors that are fortunately not reported.
The `Unresolved apply` however is directly caused by a lookup failure and is reported everytime it occurs.

All the errors that cascade from only lookup failures should be removed.

gpetiot:
Fix #944

According to #944, I removed the `Unresolved_apply` errors and propagated the underlying errors.
Are there more errors to remove? Not obvious to me just by looking at their names.

### DIFF
--- a/test/xref2/dune
+++ b/test/xref2/dune
@@ -19,6 +19,13 @@
  (enabled_if
   (>= %{ocaml_version} 4.06.0)))
 
+; 4.07.0 and above
+
+(cram
+ (applies_to github_issue_944)
+ (enabled_if
+  (>= %{ocaml_version} 4.07.0)))
+
 ; 4.08.0 and above
 
 (cram


test/xref2/github_issue_944.t/run.t
```diff
@@ -0,0 +1,8 @@
+A quick test to repro the issue found in #944
+
+  $ ocamlc -bin-annot -c foo.mli
+
+  $ odoc compile foo.cmti
+  $ odoc link foo.odoc
```
Julow, line 6: You can add `--enable-missing-root-warning` to ensures that there's still an error. Perhaps use module names that are not in the stdlib in the `.mli` file.

Julow (approved):
Looks good to me. 

It's unclear to me also which errors are useful. It might be worth investigating removing the `` `Parent_* `` errors.

test/xref2/github_issue_944.t/run.t
gpetiot: Done

Julow (commented):
There are some errors in CI due to the stdlib changes in 4.07. You need to restrict the test to `>= 4.07` like it's done for other tests in `test/xref2/dune`.

Julow (approved):
Looks good, thanks! I'll merge when the CI returns.

Issue #814, opened 2022-01-25
odoc .mld file, @see tags, can't separate from the rest of the document


tomjridge:
I have a .mld file which contains:

```
{0 Sparse-file package}

{1 Previous documents}

@see <https://github.com/tomjridge/public_notes/blob/master/2022-01-11_tjr_irmin_gc.pdf> "Implementing GC for Irmin", describes the basic structure of a pack file, what a sparse file is, etc

@see <https://github.com/ngoguey42/irmin/blob/rfc-obls/offset_based_layered_store.md> "Offset-based layered store", proposal by [@ngoguey]

@see <https://docs.google.com/presentation/d/e/2PACX-1vR6C-haCfJcosL7dz-QvjStSazjiKyz0aRCpcqImU10yNgVm0Fbu4pvfx2HaP4pJ2ENbgLuBquEm3Nq/pub?start=false&loop=false&delayms=3000> Slides to present the proposal

@see <https://github.com/tomjridge/sparse-file> An implementation of a sparse file, suffix file, worker process, irmin simulation


{1 Overview}
```

The problem is that the last `@see` gives a warning: 
```
File "../../../../src/index.mld", line 14, characters 0-2: '{1 ...}' (section heading) is not allowed in '@see'.`
```
Because parsing the tag presumably runs on to include the `{1 Overview}` heading. I guess in normal OCaml comments, the tags are last, and so the processing ends with the end of comment `*)`. In a .mld there is no end of comment, so parsing is maybe more complicated?

Julow:
Tags in general have to come at the end of a comment. But a .mld is not really a comment, the rules should be different.

We could:
- Allow some tags in mlds
- Allow some tags to be anywhere in a comment
- Entirely disallow tags in mlds

Also, there's a problem with the parsing of `@see` tags in your example. It says that the heading at the end of the example is not allowed in the rhs of the `@see` tag. This does not look like a good parsing rule to me.

dbuenzli:
> * Entirely disallow tags in mlds

I would rather be in favour of that, and deprecate tags in general, everywhere. They don't really bring any benefit, doc strings should be for humans not for machines.

The only ones that are remotely useful are `@raise`, `@since` and `@before` but in fact it would be much more beneficial to have these as attributes (like `[@@ocaml.deprecated]`), so that they can easily be acted upon by tooling.

[...]

PR #948, opened 2023-03-31
Warn against tags in pages

Julow:
Related issue: https://github.com/ocaml/odoc/issues/814

Most tags don't make sense in a page (deprecated, param, raise, ...) and the syntax prevents tags from being added in the middle of a page.

This adds a warning but do not disallow tags. It is important to not remove any text from pages and turning tags into paragraphs would not add anything interesting.


jonludlam:
Looks good to me, thanks!

Issue #578, opened 2021-01-28
@hidden would be a useful tag to have on items


jonludlam:
Rather than putting the item between (**/**) comments - it's easier for ppxes.

Julow:
@ccatherinee In your code, you can use `List.exists` instead of `List.fold_left` with an `if init` inside.

Your first example is a syntax error, so there's nothing to worry about. Your third example attaches the `[@@hidden]` attribute to the module type `S` (it is attached to the `Tsig_modtype`) and has nothing to do with what comes after.

OCaml has a syntax for "standalone" attributes: `[@@@hidden]`. These are `Tsig_attribute` in the typed tree and would mean "hide everything that comes after the attribute".

ccatherinee:
Hi @Julow - thank you so much for the detailed response. Just to clarify, does this mean my understanding of 
```
module type S = sig
  type t [@@hidden]
  val f : t
end
```
is correct (ie only the type t part should be hidden, but not the entire module) or do these sentences 

> OCaml has a syntax for "standalone" attributes: [@@@hidden]. These are Tsig_attribute in the typed tree and would mean "hide everything that comes after the attribute". 

mean that 
```
module type S = sig
  type t [@@hidden]
  val f : t
end
```
is a syntax error and we can only have 
```
module type S = sig
  type t [@@@hidden]
  val f : t
end
```
Furthermore, does 

> hide everything that comes after the [standalone] attribute

mean all remaining code in the file or just code in that module which comes after the standalone attribute? Thank you so much for your time!

lpw25:
I'm slightly surprised that we're going for:
```ocaml
type t [@@hidden]
```
rather than:
```ocaml
(** @hidden *)
type t
```
which would be more consistent with things like:
```ocaml
(** @inline *)
include Foo
```

Sorry to ask about this late in the day. Although, if we did decide to do this instead, I suspect adjusting the implementation to do so wouldn't need to change much.

jonludlam:
I'm flummoxed. I can't see why I turned this from a tag (originally here: https://github.com/ocaml/odoc/pull/571#discussion_r566101709, and in the title of this issue!) into an attribute. The only annoying thing about an attribute is that we'll have to change odoc-parser to do this... :-/

jonludlam:
Turns out using an attribute is rather a lot more annoying than an attribute as we haven't parsed any comments at the point we're hiding items.

PR #950, opened 2023-04-03
WIP: Add hidden attribute

3Rafal:
Draft of #578

### DIFF
--- a/src/loader/doc_attr.ml
+++ b/src/loader/doc_attr.ml
@@ -79,6 +79,7 @@ type parsed_attribute =
   | `Doc of payload  (* Attached comment. *)
   | `Stop of Location.t  (* [(**/**)]. *)
   | `Alert of  string * payload option * Location.t
+  | `Hidden of Location.t
     (* [`Alert (name, payload, loc)] is for [\[@@alert name "payload"\]] attributes. *) ]
 
 (** Recognize an attribute. *)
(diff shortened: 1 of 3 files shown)


Julow (commented):
The implementation looks good and the tests seem sufficient.

The `[@@@hidden]` seems unnecessary as we already have `(**/**)`, which can always be used instead.
If we keep this feature, I suggest making "on" and "off" markers more explicit as there's room on the syntax:

```ocaml
[@@@hidden]

type a (* this is hidden *)

[@@@hidden]

type b (* this is visible *)

[@@@hidden on]

type a

[@@@hidden off]

type b
```

As @lpw25 said in https://github.com/ocaml/odoc/issues/578#issuecomment-1406278762, why use an attribute (`[@@hidden]`) instead of a tag (`(** @hidden *)`) ?
Odoc has other similar tags (the current bad name is "internal") like `@canonical`, `@inline`, etc...
Pinging @jonludlam as you decided on the syntax.

Julow:
Thanks a lot for contributing to Odoc :)

3Rafal:
@Julow , if we want to use `(**@hidden*)` tag, then the change will involve odoc-parser. I can do a PR to parser which handles it. Do you think I can proceed, or should we wait for @jonludlam ?

3Rafal:
I have created a PR for a new tag: https://github.com/ocaml-doc/odoc-parser/pull/16

src/loader/doc_attr.ml
```diff
@@ -96,6 +97,7 @@ let parse_attribute : Parsetree.attribute -> parsed_attribute option =
       match load_payload attr_payload with
       | Some p -> Some (`Doc p)
       | None -> None)
+  | "hidden" | "ocaml.hidden" -> Some (`Hidden attr_loc)
```
Julow, line 100: We discussed with @jonludlam and found that we don't need the attribute but prefer the tag instead. Do you mind opening a PR for that ?

PR #953, opened 2023-04-07
Add common languages to highlightjs

Julow:
Change the languages highlighted in code blocks:

- Drop Python to save space. It's not a language commonly used with OCaml compared to for example Javascript, which is not included.
- Add C, which is commonly used with OCaml.
- Add Json, which is used in several major libraries.
- Add SQL and GraphQL, for which popular libraries exist.
- Add Shell Session which seems most useful with Mdx.

OCaml, ReasonML, HTML and Markdown are still supported. The size of highlightjs grows from 33KB to 44KB.

Ping @jonludlam as you chose the current languages.


panglesd (approved):
I haven't reviewed how you changed the highlight.pack.js file.

One small comment: if python is removed from highlightjs, then the example in https://ocaml.github.io/odoc/odoc_for_authors.html#language_headers should be updated.

There are several libraries which could use the python highlighting (pyml, py, pyast, pyre-ast, ocaml-in-python, ...) but from a [github search](https://github.com/search?q=%40python+%28path%3A*.mli+OR+path%3A*.mld+OR+path%3A*.ml%29&type=code) no ocaml code is using python syntax highlighting. So I think it can safely removed without breaking any existing doc.

In conclusion, I think your set of languages is already better than what is currently included. I would naively think that we could include python and javascript as well.
If the size is a problem, the languages support could be split into several files (I think that's what is done when fetching highlightjs from CDNs) but that might cause other problems.

PR #954, opened 2023-04-13
Generate endline for html files

### LINKED ISSUE #952: `No newline at end of file` in html files
When generating test cases for #950 , I've found out that generated html files don't end with a new line.
This results in git warning: `No newline at end of file` and is shown as red marker on GitHub:  
Screenshot 2023-04-05 at 10 22 05

The issue doesn't occur for other outputs (tex, man).

3Rafal:
Resolves #952


src/html/html_page.ml
```diff
@@ -206,7 +210,11 @@ let src_page_creator ~breadcrumbs ~config ~url ~header name content =
   let html =
     Html.html head (Html.body ~a:[ Html.a_class [ "odoc-src" ] ] body)
   in
-  let content ppf = htmlpp ppf html in
+  let content ppf =
+    htmlpp ppf html;
+    (* Tyxml's pp doesn't output a newline a the end, so we force one *)
+    Format.pp_force_newline ppf ()
```
Julow, line 216: Note: This is not adding code duplication but rather applying the same changes to already duplicated code.

If you have ideas for de-duplicating these two functions, this would be welcome.

Julow (commented):
The code looks good, thanks!

There are still some test outputs to promote, for example `Bugs_pre_410` is not run on `>= 4.10`. To avoid installing all the versions on your machine, you can edit the files and let the CI check for you. (`for f in test/generators/html/*.html; do echo >> $f; done`)

src/html/html_page.ml
3Rafal: Yes, there is a bit of duplication, but I'm not sure if we should do it right now. 

The diff of these two functions is pretty complicated and I don't see an easy way to abstract over it:
```diff
@@ -1,13 +1,9 @@
-let page_creator ~config ~url ~uses_katex header breadcrumbs toc content =
+let src_page_creator ~breadcrumbs ~config ~url ~header name content =
   let theme_uri = Config.theme_uri config in
-  let support_uri = Config.support_uri config in
-  let path = Link.Path.for_printing url in
-
   let head : Html_types.head Html.elt =
     let title_string =
-      Printf.sprintf "%s (%s)" url.name (String.concat "." path)
+      Format.asprintf "Source: %s%a" name path_of_module_of_source url
     in
-
     let file_uri base file =
       match base with
       | Types.Absolute uri -> uri ^ "/" ^ file
@@ -15,10 +11,8 @@
           let page = Url.Path.{ kind = `File; parent = uri; name = file } in
           Link.href ~config ~resolve:(Current url) (Url.from_path page)
     in
-
     let odoc_css_uri = file_uri theme_uri "odoc.css" in
-    let highlight_js_uri = file_uri support_uri "highlight.pack.js" in
-    let default_meta_elements =
+    let meta_elements =
       [
         Html.link ~rel:[ `Stylesheet ] ~href:odoc_css_uri ();
         Html.meta ~a:[ Html.a_charset "utf-8" ] ();
@@ -32,48 +26,21 @@
               Html.a_content "width=device-width,initial-scale=1.0";
             ]
           ();
-        Html.script ~a:[ Html.a_src highlight_js_uri ] (Html.txt "");
-        Html.script (Html.txt "hljs.initHighlightingOnLoad();");
       ]
     in
-    let meta_elements =
-      if uses_katex then
-        let katex_css_uri = file_uri theme_uri "katex.min.css" in
-        let katex_js_uri = file_uri support_uri "katex.min.js" in
-        default_meta_elements
-        @ [
-            Html.link ~rel:[ `Stylesheet ] ~href:katex_css_uri ();
-            Html.script ~a:[ Html.a_src katex_js_uri ] (Html.txt "");
-            Html.script
-              (Html.cdata_script
-                 {|
-          document.addEventListener("DOMContentLoaded", function () {
-            var elements = Array.from(document.getElementsByClassName("odoc-katex-math"));
-            for (var i = 0; i < elements.length; i++) {
-              var el = elements[i];
-              var content = el.textContent;
-              var new_el = document.createElement("span");
-              new_el.setAttribute("class", "odoc-katex-math-rendered");
-              var display = el.classList.contains("display");
-              katex.render(content, new_el, { throwOnError: false, displayMode: display });
-              el.replaceWith(new_el);
-            }
-          });
-        |});
-          ]
-      else default_meta_elements
-    in
     Html.head (Html.title (Html.txt title_string)) meta_elements
   in
-
   let body =
     html_of_breadcrumbs breadcrumbs
     @ [ Html.header ~a:[ Html.a_class [ "odoc-preamble" ] ] header ]
-    @ html_of_toc toc
-    @ [ Html.div ~a:[ Html.a_class [ "odoc-content" ] ] content ]
+    @ content
+  in
+  (* We never indent as there is a bug in tyxml and it would break lines inside
+     a [pre] *)
+  let htmlpp = Html.pp ~indent:false () in
+  let html =
+    Html.html head (Html.body ~a:[ Html.a_class [ "odoc-src" ] ] body)
   in
-  let htmlpp = Html.pp ~indent:(Config.indent config) () in
-  let html = Html.html head (Html.body ~a:[ Html.a_class [ "odoc" ] ] body) in
   let content ppf =
     htmlpp ppf html;
     (* Tyxml's pp doesn't output a newline a the end, so we force one *)
```

Maybe we should follow the [Rule of three™](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) and leave it as is for now? :wink:

PR #956, opened 2023-04-19
Basic support for OCaml 5.1

Octachron:
This PR updates the `cmt` and ident loader to be compatible with three OCaml 5.1 changes:

- First, generative application `F()` is now represented by its own node
- Second, there is no more hash type stored in signatures
- Third, type paths for `inline record`s are no longer encoded as invalid type name but use a distinct constructor.

Note that this PR is still a draft since the cone of dependencies of tests have not been updated yet.


anmonteiro:
Nice, kicked off a test run here https://github.com/nix-ocaml/nix-overlays/pull/763

jonludlam:
I've just run this on alpha2 and it works for me - I wasn't able to run the semantics test due to a missing ppx_expect dependency, but that only tests reference semantics which aren't affected by this change. Looks good to go to me - @Octachron do you concur?

Octachron:
I have fixed some spurious white space, but the PR looks good to go indeed.

jonludlam:
I've added a very minor backward compat commit

Issue #816, opened 2022-02-02
Broken reference sometimes get incorrectly rendered


MisterDA:
Using odoc.2.0.2, I have a doc reference such as `{!Stdlib.Invalid_argument}`. When I build my project and odoc can't find the cross-ref, it sometimes renders the reference as `<code>Stdlib</code>.Invalid_argument` instead of `<code>Stdlib.Invalid_argument</code>`. I think it doesn't do that mistake consistently, but I'm not sure.

PR #892, opened 2022-09-04
Render unresolved reference fields in `Source`

sim642:
Closes #816.

### TODO
- [ ] Fix warnings on `dune build`:
    ```
    File "cases/ocamlary.mli", line 1076, characters 6-42:
    Warning: Failed to resolve reference unresolvedroot(CollectionModule).InnerModuleA.foo Couldn't find "foo"
    File "cases/ocamlary.mli", line 1075, characters 6-20:
    Warning: Failed to resolve reference unresolvedroot(Set).S.empty Couldn't find "Set"
    File "cases/ocamlary.mli", line 1074, characters 6-18:
    Warning: Failed to resolve reference unresolvedroot(Hashtbl).t Couldn't find "Hashtbl"
    File "cases/ocamlary.mli", line 1073, characters 6-32:
    Warning: Failed to resolve reference unresolvedroot(Stdlib).Invalid_argument Couldn't find "Stdlib"
    ```
    Not sure what to do about it though.
- [ ] #613 as suggested in #816?

PR #957, opened 2023-04-20
Fix rendering of unresolved references

Julow:
Closes https://github.com/ocaml/odoc/issues/816 https://github.com/ocaml/odoc/pull/892

This is an alternative to https://github.com/ocaml/odoc/pull/892, which uses the recent `render_unresolved` function introduced in https://github.com/ocaml/odoc/pull/945.


jonludlam:
Looks good thanks!

PR #958, opened 2023-04-21
Remove expansion from "unexpanded" `module type of`

lukemaurer:
We've been keeping each `module type of` expression's expansion around, even in the nominally unexpanded version of the type. This was needed in case the expression was invalidated so that we had the expansion on hand as a fallback. However, it led to exponential blowup in the case of deeply nested `module type of`s.

Happily, we can do without the expansion if we do a little more work when substituting: We already track in the `subst` which module paths have been invalidated. When we need to invalidate a module path, we can stash the expression for the module's type in that `subst`. This is effectively the same data that we were keeping in the "unexpanded" expansion. And then when we need to expand an invalidated `module type of`, we can use the stashed expression.


jonludlam:
Thanks Luke! Looks like it needs some rebasing (and maybe squashing?)

Julow:
With @jonludlam we figured that this makes the `Type_of` module unnecessary. I pushed a commit to remove it directly to your branch.
I also added some `jq` queries to reduce the size of the test, to avoid false-negative in the future. Did I remove too much ?

lukemaurer:
> With @jonludlam we figured that this makes the `Type_of` module unnecessary. I pushed a commit to remove it directly to your branch. I also added some `jq` queries to reduce the size of the test, to avoid false-negative in the future. Did I remove too much ?

I like the idea of cutting down the test using `jq`, but yeah, this cuts out too much. The specific criteria (which I should have written in `run.t` to begin with!) are:

* In most cases, the printed module type should have a type `t` but _not_ a value `plus`. (In a few of them, it's a nested module that should look like that.)
* In `In_functor_parameter.P.G`, in the type of the functor argument, there _should_ be both a type `t` and a value `plus`.

Can that be expressed neatly using `jq`?

src/xref2/subst.ml
```diff
@@ -142,6 +142,15 @@ let rename_class_type : Ident.path_class_type -> Ident.path_class_type -> t -> t
         t.type_;
   }
 
+let u_module_type_expr_of_module_type_expr (e : Component.ModuleType.expr) :
```
jonludlam, line 145: There's a similar function here: https://github.com/ocaml/odoc/blob/master/src/model/lang.ml#L533-L538

lukemaurer:
Found a bug. Converting back to draft for the moment.

jonludlam:
This is superceded by #1081

PR #964, opened 2023-06-01
Make the tests more robust to changes to yojson

jonludlam:
A minor change between yojson 1.7.0 and 2.0.0 caused our tests to break as the output was formatted slightly differently. This attempts to fix that by doing compact output in some places and piping through jq in others.

### DIFF
--- a/odoc.opam
+++ b/odoc.opam
@@ -54,7 +54,7 @@ depends: [
   "fmt"
 
   "ocamlfind" {with-test}
-  "yojson" {< "2.0.0" & with-test}
+  "yojson" {with-test}
   ("ocaml" {< "4.04.1" & with-test} | "sexplib0" {with-test})
   "conf-jq" {with-test}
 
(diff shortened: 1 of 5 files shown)


Julow (commented):
Piping into `jq` everywhere is too much boilerplate. What do you think of doing the piping inside `odoc_print` ? We could also add an option for specifying a jq query.

jonludlam:
We're already piping through jq in almost every instance. It doesn't seem a huge burden, and adding it as an option to odoc_print will be a bit of a pain in passing through options and whatnot.

Julow (approved):
I think enforcing the formatting is better than a convention and `odoc_print` already takes options. Though, let's not complicate this PR, which is fixing an immediate problem.

Leonidas-from-XIV:
Another option could be to do the JSON comparison on a structural level instead of a textual, though of course that is not foolproof either as equivalence of two JSON files is a rather tricky question. But it would most likely deal with formatting changes just fine.

PR #966, opened 2023-06-02
Include missing preformatted text elements

toastal:


jonludlam:
Hi @toastal, thanks for your contribution! I'm not sure why we would need these though - we don't generate these tags at all. What's the use-case you're thinking of?

toastal:
Use case is that almost everything is reset in the CSS and these seemed missing. `<kbd>` in particular is the correct element for talking like: press `/` to open the TUI search (with `<code>` being the wrong element). I opened these merge requests independently late last night so any that aren’t useful can be thrown out.

jonludlam:
I'm closing this as we don't use those elements.

PR #968, opened 2023-06-02
Humanist font fallbacks

toastal:
If Fira Sans is the preferred font, a humanist sans serif, similar humanist sans serifs on OSs should be chosen over a Grotesque like Helvetica and Arial. Seravek is available on macOS/iOS and Source Sans Pro is on Android. Other available options are (subjectively) far enough away from Fira Sans that it would be better to defer to the user agent and user settings than trying to shoehorn in Gill Sans or a common Linux font.

### DIFF
--- a/src/html_support_files/odoc.css
+++ b/src/html_support_files/odoc.css
@@ -260,7 +260,7 @@ header {
 }
 
 nav {
-  font-family: "Fira Sans", Helvetica, Arial, sans-serif;
+  font-family: "Fira Sans", Seravek, "Source Sans Pro", source-sans-pro, sans-serif;
 }
 
 /* Basic markup elements */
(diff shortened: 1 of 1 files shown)


Julow (commented):
Are these commonly installed fonts ? If the fallbacks are grotesque, what do you think of shortening the property to this ?
```css
font-family: "Fira Sans", sans-serif;
```

Note that we now ship the font files (https://github.com/ocaml/odoc/pull/913) so they should load fine offline or behind adblockers.

Julow:
I've opened a PR implementing my suggestion: https://github.com/ocaml/odoc/pull/1028
What do you think ?

toastal:
Personally a bit partial to this one as falling back to other Humanist fonts is a bit more interesting than just falling back to the default, but I don’t have the strongest feeling other than removing Grotesk fonts.

Julow:
I don't see why Humanist fonts are better fallbacks than user's chosen font (or chosen by the distro). I think the problem is the presence of a fallback.

jonludlam:
I'm happy with falling back to the system defaults, so let's go with @Julow's approach.

Julow:
Thanks for the explanation!

PR #969, opened 2023-06-06
Source identifiers

jonludlam:
The `Lang.Locations` module seems to be better represented as an identifier than a separate thing. This PR fixes that and also cleans up the paths interface and removes a bunch of dead code.

### DIFF
--- a/src/model/paths.ml
+++ b/src/model/paths.ml
@@ -51,14 +51,18 @@ module Identifier = struct
     | `Method (_, name) -> MethodName.to_string name
     | `InstanceVariable (_, name) -> InstanceVariableName.to_string name
     | `Label (_, name) -> LabelName.to_string name
+    | `SourcePage (_, name) -> name
+    | `SourceDir (_, name) -> name
+    | `SourceLocation (_, anchor) -> DefName.to_string anchor
+    | `SourceLocationMod x -> name_aux (x :> t)
 
   let name : [< t_pv ] id -> string = fun n -> name_aux (n :> t)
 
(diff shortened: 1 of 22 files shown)


src/model/paths.ml
```diff
@@ -255,23 +259,27 @@ module Identifier = struct
...
 
   module SourcePage = struct
     type t = Id.source_page
     type t_pv = Id.source_page_pv
-    let equal = equal
-    let name { iv = `SourcePage (p, name); _ } = SourceDir.name p ^ name
-    
+  end
+
+  module SourceLocation = struct
```
Julow, line 280: What do you think of removing the `Paths` module ?

It slowed us down significantly while implementing source code rendering with @panglesd.
It's only made of aliases and is annoying to edit (a lot of boilerplate).
It makes `Paths_types` harder to understand and makes Merlin's printed types harder to read.

With all these downsides, I'd argue that it's not worth the slightly improved paths: `Id.SourceLocation.t` instead of `Id.source_location`.
Maps are almost only used in `Component`, which could as well define them inline.

Julow (approved):
Indeed! 

The code is not formatted (`dune build @fmt`)

jonludlam:
Argh, I _always_ forget to format! I need to configure my editor to run it on save...

src/model/paths.ml
jonludlam: Yeah, I'm leaning in that direction too, especially now that most of the additional functions added in the namespace modules hve gone.

PR #971, opened 2023-06-08
HTML: Match rendering of polymorphic variant with normal variants

### LINKED ISSUE #970: Variant <ol> renders with numbers
Screenshot 2023-06-07 at 11 11 01

panglesd:
Fix #970 

Mostly CSS, I just needed to add one class.
Quite cumbersome due to the differences in document on how polymorphic variants and normal variants are turned into a document value. But modifying the CSS seemed still less work (and with less unexpected consequences in other backend) than modifying the document generator to "unify" normal and polymorphic variants.

Previous:

image

Now:

image


trefis (commented):
Also, silly question sorry but related to Jon's issue: why do we `<ol>`?

src/document/generator.ml
```diff
@@ -703,7 +703,9 @@ module Make (Syntax : SYNTAX) = struct
         let markers = Syntax.Comment.markers in
         try
           let url = Url.Anchor.polymorphic_variant ~type_ident item in
-          let attrs = [ "def"; Url.Anchor.string_of_kind url.kind ] in
+          let attrs =
+            [ "def"; "poly-variant"; Url.Anchor.string_of_kind url.kind ]
+          in
```
dbuenzli, line 708: I'm not sure exactly why you had to change the generator to fix this issue (`odig`'s stylesheet have no problem with this without that) but, why not…

dbuenzli: Thinking more about this I'm not sure it's really a good idea to add more to the markup to distinguish polymorphic variants from regular ones. I don't really see a good use case for it.

trefis:
> I think you should add the example with a short comment and a long comment for the case to the ocamlary

👍 

> > Also, silly question sorry but related to Jon's issue: why do we `<ol>`?
> 
> Let's add my own silly question: what else would you suggest to use ?

I would have thougth `<ul>` ... but I don't know what I'm talking about, so really, don't mind me.

src/document/generator.ml
panglesd: I think it is related with odoc doing something more complex, when the size of the comment is small, the comment is in the same line and takes 50% of the space, otherwise it takes more space and is wrapped below (see the screenshot or [some file where it happens](https://ocaml-ppx.github.io/ppxlib/ppxlib/Astlib/Ast_500/Parsetree/index.html)).
I don't know if that's worth the trouble, but I think that's the difference with `odig`, which displays comments in a simpler way (from a quick look at the stylesheet, I might be wrong).

There might be a way to do it without, but since variant and fields had their classes, it felt natural to add one for polymorphic variants

panglesd:
> I think you should add the example with a short comment and a long comment for the case to the ocamlary

I'll do that

> I would have thougth `<ul>` ... but I don't know what I'm talking about, so really, don't mind me.

The html generator has to use `ol` for ordered lists: otherwise, user-input `{ol {li bli}}` would not give numbers.
So, we would need to change the document generator, from ordered lists to unordered lists, when rendering variant lists.
But that means that all backends would be modified, and I guess that latex uses the fact that its ordered.
(Otherwise, it would be very strange to use ordered lists and then remove the numbers!)

src/html_support_files/odoc.css
```diff
@@ -550,12 +550,21 @@ div.odoc-spec,.odoc-include {
...
 
+.spec.type .poly-variant > code {
+  min-width: 2ch;
+}
+.spec.type .poly-variant > code + code {
+  min-width: calc(40% - 2ch - 0.6ex);
+  /* In polymorphic variants, the | is not included in the first code block, so we
+   need to accound for its size */
+}
+
```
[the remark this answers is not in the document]
panglesd: Yes, this is fragile (and was already).

The reason the html generated in case of polymorphic variant is structured differently from the one for variants, is that it is already the case at the intermediate document representation.

I was afraid of changing the intermediate representation, as that would require me to check that it does not break anything in other backends. But maybe that is a better solution.

dbuenzli:
> I would have thougth <ul> ... but I don't know what I'm talking about, so really, don't mind me.

If one believes in semantic markup (no one does :-) then I'd say at least for *variants* it's more "correct". There is a notion of order between case (but for polymorphic ones it won't match, the `li` order).

> I think it is related with odoc doing something more complex, when the size of the comment is small, the comment is in the same line and takes 50% of the space, otherwise it takes more space and is wrapped below (see the screenshot or [some file where it happens](https://ocaml-ppx.github.io/ppxlib/ppxlib/Astlib/Ast_500/Parsetree/index.html)).

As far as I see `odig` works the same way[^1].

[^1]: Except not flush right, but I would advise against doing that. If you have many small comments, the large blank make it difficult for the eye to catch the right comment on the right. The same kind of problem you get when you typeset a table of contents with page number flushed right and which people often solve with horrible dot leaders, rather than simply mention the page right after the entry. Obsessing with alignment may makes things *look* better from a distance, but often makes it less usable for the reader.

Screenshot 2023-06-08 at 15 16 50

> There might be a way to do it without, but since variant and fields had their classes, it felt natural to add one for polymorphic variants

I think it would be more natural not to distinguish variants and polymorphic variants. These are just list of cases, it's unlikely you want to style them differently. Adding one more class is just adding more selectors to handle and makes your stylesheets more complex, which, trust me, you definitively do not want.

Basically I think `odoc` just just render polymorphic variants exactly the way variants are. Unless I'm mistaken that just means, in the current state, to add `variant` to the class of the  `li` generated for them. If your stylesheet already handles variants correctly then nothing else must be changed.

panglesd:
Thanks for the advice.
I tried to simplify the css. The fact that, in odoc, the background for text and for code is not the same is making things more complex than in `odig`. For instance, using `odig`'s way of having the "long comment/short comment" behaviour ends up with white blocks of every dimension; which in my opinion is disturbing:
image

So, I went with the even simpler approach of having only "long comments":
image

This was also requested in https://github.com/ocaml/odoc/pull/847#issuecomment-1133663289, so at least it will please one person.

Let me know what you think!

I also added a link to the ocamlary module from the index page.

dbuenzli:
> Let me know what you think!

If you need to keep the blocks it's certainly better the way you suggest.

> I also added a link to the ocamlary module from the index page.

Thanks!

wikku:
The same problem remains with extensions of extensible variants, so you would need to add more selectors for `.extension` to the CSS. But by then I think it becomes clear that the CSS rules are wrong, and should be selecting actual lists elements if you want to disable the numbering (#987).

panglesd:
I cherry picked @wikku's commit that makes the generated html for polymorphic variants match the one for normal variants.

I also added the `variant` class to extension constructor so that the same css is applied to extension constructors and normal constructors.

So, the rendering of normal constructors, extension constructors and polymorphic variants should be exactly the same, and stay the same no matter how we modify the css.

The rest of this PR is:
- adding the ocamlary module to the pages of odoc's doc
- simplifying the css so that docstrings attached to constructors are always displayed on the next line

PR #973, opened 2023-06-20
Parser remerge

jonludlam:
Back in 2021 we split odoc-parser out of odoc. The rationale behind this was that we were attempting to break an annoying cyclic dependency where both ocamlformat and mdx used the parser, and this meant whenever the library was out of sync with what those tools were expecting we couldn't use them. 

However, while simply splitting the repositories and opam packages fixed the problem from a packaging perspective, the fundamental issue of incompatibility still existed and really _couldn't_ be fixed unless we found a way to have two different versions of the library coexisting. That method ended up being vendoring the library into ocamlformat so that it no longer depends (on a package level) on odoc-parser. With that fixed, we are left with no real advantage to having two repositories other than ensuring the library remains easy to vendor, and with plenty of disadvantages. So this PR is to recombine the repositories back into one.

I've made this PR by reverting the commit that originally removed the parser, and replaying the commits from `odoc-parser.git` on top to recreate the history. I chose not to simply merge the repositories as both repos contain the parser-relevant commits from before the split, but not exactly the same hashes, so we would have ended up with a more messy history with some not-quite-duplicate commits. The end result should be that the source of parser should be exactly the same as the source files in [ocaml-doc/odoc-parser#ccb39b9c7a061ad0b86ae9427bad092bfe747966
](https://github.com/ocaml-doc/odoc-parser/commit/ccb39b9c7a061ad0b86ae9427bad092bfe747966), modulo whitespace changes from the ocamlformat upgrade.

We're in the happy position that the current releases of odoc-parser and odoc are such that we can coalesce the version numbers so we'll have 2.3.0 as the next release of both, which is one less thing to worry about.


odoc.opam
```diff
@@ -37,12 +37,9 @@ description: """
 odoc is part of the [OCaml Platform](https://ocaml.org/docs/platform), the recommended set of tools for OCaml.
 """
 
-pin-depends: [
-  ["odoc-parser.dev" "git+https://github.com/ocaml-doc/odoc-parser.git#f98cfe3"]
-]
 
 depends: [
-  "odoc-parser" {>= "2.0.0"}
+  "odoc-parser" {= version}
```
Julow, line 42: Thinking for later: Now that the sources are in the same repository, nothing (except the complicated dune rules required) prevent us from distributing the parser in two different packages `odoc.parser` and `odoc-parser`. `odoc` would no longer depend on `odoc-parser`, which would only be defined by a complicated `(copy ...)` rule.

Julow (commented):
LGTM!

odoc.opam
jonludlam: Yep there are definitely some interesting options open in the future.

PR #974, opened 2023-06-20
Add labels to basic text block (such as paragraphs and code blocks)

panglesd:
This allows to generate anchors, and link to specific places in the rendered html.
This is useful for search results in doc comments or mld page. Clickable anchors are not generated.

Somehow part of #972.


Julow (commented):
Some `< 4.10` test files are not uptodate.

src/model/semantics.ml
```diff
@@ -216,23 +216,45 @@ let rec inline_element :
 
 and inline_elements status elements = List.map (inline_element status) elements
 
+let generate_label =
+  let current_label = ref 0 in
+  fun kind status ->
+    let lbl =
+      match kind with
+      | `Paragraph -> "p"
```
jonludlam, line 224: Are these likely to clash with any other hand-written labels?

src/xref2/component.ml
```diff
@@ -474,11 +486,14 @@ end =
   CComment
 
 and Label : sig
+  (** In order to generate content for links without content *)
+  type content = Heading of Odoc_model.Comment.paragraph | NestableBlock
+
   type t = {
```
jonludlam, line 492: The attrs have been moved onto the `` `Heading `` constructor right? Any idea why they weren't there to begin with? With your fix it's much more similar to `Comment.block_element` now, so I'm confused why we didn't do that from day one!

jonludlam:
I'm a little nervous about allowing these new labels as valid references, it seems a bit brittle. I'd much prefer it that if you wanted to link to a paragraph or code block or whatever that we could add in an explicit label. I can see the use of them in search, where we'll want to be able to link to anything automatically, of course.

jonludlam:
I still think it's quite useful to be able to link directly from the search to the relevant part, especially in a large file. Though if that's the primary use-case then it's a little questionable why we're adding the links in the model rather than only in the html output, since it does mean a bit of pain going back and forth betweens `Odoc_model.Comment` and `Component.CComment` types.

panglesd:
One possibility would be to use the last heading as target when clicking on a search result corresponding to a paragraph. It would be a bit less precise, but still better than just linking to the page.
Auto-generated label for headings are also a bit brittle, although less than the one I'm introducing. But they are much easier to copy-paste, since they have clickable anchors to get the link.

Disambiguation of headings also need to be done earlier than after the `document` is generated, as is currently the case.
I guess for headings, since they can be referenced, it makes sense to disambiguate them at the model level, at least for the auto-generated one.

dbuenzli:
> One possibility would be to use the last heading as target when clicking on a search result corresponding to a paragraph.

That could be an idea!

> Auto-generated label for headings are also a bit brittle, although less than the one I'm introducing.

Indeed. I don't think they are a good idea either. But somehow it's either that or having less usable pages because people don't care. There's no real good solution that I'm aware of.

panglesd:
> why we're adding the links in the model rather than only in the html output

The compilation of the search index is done from the `.odocl`, so the links should be knowable without going through the whole html/document generation.

One possibility would be to add a common first phase to `html-generate` and `compile-index`, which gives labels to the basic blocks. It's a pass that would be done twice (one for each command), but it would avoid the back and forth with `Odoc_model.Comment` and `Component.CComment`.

dbuenzli:
> I wonder what is @jonludlam and @Julow opinion on using the last heading as target. I'm hesitant between this and the current approach.

Note that this kind of stuff is hard to assess whether it will work well or not with users. There's little you can do but try and see how it fares in practice…

jonludlam:
Last heading or most recent anchor of any sort? I'm in favour of using the anchors we have right now and if that turns out to be a bad user experience then we can revisit this decision.

jonludlam:
> The compilation of the search index is done from the `.odocl`, so the links should be knowable without going through the whole html/document generation.

Aren't we outputting an html snippet in the index?

panglesd:
> Aren't we outputting an html snippet in the index?

Yes, but we are generating them separately for each entry. Disambiguation needs the document as a whole to work.

I'm putting this PR on pause to focus on the "last heading" or "last anchor" solution.

jonludlam:
> Yes, but we are generating them separately for each entry. Disambiguation needs the document as a whole to work.

Right, makes sense, thanks.

jonludlam:
I'll close this PR now then - we can do disambiguation in another PR.

PR #975, opened 2023-06-22
Rudimentary support for assets

trefis:
Context: #59 , #972

**No support for referring to assets yet.**
This only provides a way to get odoc to place them "in the right place" in the directory structure; so it's only marginally useful.

Current interface:
- children given to `odoc compile` can now be of the form `asset-filename`
- `html-generate` and `html-targets` commands now accept `--asset FILE` arguments, which are paired (based on filename) with the relevant child asset. Warnings are emitted for unknown and missing assets.
- files passed to `html-generate` through `--asset` are copied in the directory of their parent; would it be better to put them in an `_assets` subdirectory?

This is illustrated in `test/pages/assets.t/run.t`

### DIFF
--- a/src/document/url.ml
+++ b/src/document/url.ml
@@ -87,13 +87,13 @@ module Path = struct
     | Identifier.Signature.t_pv
     | Identifier.ClassSignature.t_pv ]
 
-  type source_pv =
+  type any_pv =
     [ nonsrc_pv
     | Identifier.SourcePage.t_pv
     | Identifier.SourceDir.t_pv
     | Identifier.AssetFile.t_pv ]
 
-  and source = source_pv Odoc_model.Paths.Identifier.id
+  and any = any_pv Odoc_model.Paths.Identifier.id
 
   type kind =
     [ `Module
@@ -123,13 +123,13 @@ module Path = struct
 
   let mk ?parent kind name = { kind; parent; name }
 
-  let rec from_identifier : source -> t =
+  let rec from_identifier : any -> t =
    fun x ->
     match x with
     | { iv = `Root (parent, unit_name); _ } ->
         let parent =
           match parent with
-          | Some p -> Some (from_identifier (p :> source))
+          | Some p -> Some (from_identifier (p :> any))
           | None -> None
         in
         let kind = `Module in
@@ -138,7 +138,7 @@ module Path = struct
     | { iv = `Page (parent, page_name); _ } ->
         let parent =
           match parent with
-          | Some p -> Some (from_identifier (p :> source))
+          | Some p -> Some (from_identifier (p :> any))
           | None -> None
         in
         let kind = `Page in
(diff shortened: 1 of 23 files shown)


src/document/url.ml
```diff
@@ -377,6 +384,9 @@ module Anchor = struct
     | { iv = `SourcePage (p, _name); _ } | { iv = `SourceDir (p, _name); _ } ->
         let page = Path.from_identifier (p :> Path.source) in
         Ok { page; kind = `Page; anchor = "" }
+    | { iv = `AssetFile (p, _name); _ } ->
+        let page = Path.from_identifier p in
+        Ok { page; kind = `File; anchor = "" }
```
trefis, line 389: I copied what was done for ``` `SourcePage ``` here, but honestly it looks weird.

src/document/url.mli
```diff
@@ -35,7 +35,10 @@ module Path : sig
     | Identifier.ClassSignature.t_pv ]
 
   type source_pv =
```
trefis, line 37: I guess this should be renamed now?

src/model/paths.ml
```diff
@@ -372,6 +378,9 @@ module Identifier = struct
         [> `LeafPage of ContainerPage.t option * PageName.t ] id =
       mk_parent_opt PageName.to_string "lp" (fun (p, n) -> `LeafPage (p, n))
 
+    let asset_file : Page.t * string -> AssetFile.t =
+      mk_parent Fun.id "asset" (fun (p, n) -> `AssetFile (p, n))
```
trefis, line 382: I'm not sure what this `"asset"` is used for / I don't know what I'm doing.

test/pages/assets.t/run.t
```diff
@@ -0,0 +1,91 @@
...
+Now we can odoc-compile the module odoc file passing that file as parent.
+
+  $ odoc compile test.cmti -I . --parent index
+
+Link and generate the HTML (forgetting the asset!):
+
+  $ for i in *.odoc; do odoc link -I . $i; done
+  $ for i in *.odocl; do odoc html-generate $i -o html; done
+  File "img.jpg":
+  Warning: asset is missing.
```
trefis, line 38: I wondered if this should have been a fatal error, since I didn't know how to make it be one I managed to convince myself that it shouldn't be.

src/model/paths.ml
panglesd: If I understand correctly, this string is used to generate unique keys that are used for memoization optimizations. Your choice of `"asset"` looks valid to me!

src/model_desc/paths_desc.ml
```diff
@@ -74,6 +74,8 @@ module General_paths = struct
               ( "`LeafPage",
                 ((parent :> id_t option), name),
                 Pair (Option identifier, Names.pagename) )
+        | `AssetFile (parent, name) ->
+            C ("`AssetFIle", ((parent :> id_t), name), Pair (identifier, string))
```
panglesd, line 78:
```suggestion
            C ("`AssetFile", ((parent :> id_t), name), Pair (identifier, string))
```

src/odoc/html_page.ml
```diff
@@ -54,4 +58,46 @@ let extra_documents args unit ~syntax =
...
+      children
+  in
+  let rec extract paths name =
+    match paths with
+    | [] -> raise (Missing_asset name)
+    | x :: xs when Fpath.basename x = name -> (xs, (name, x))
+    | x :: xs ->
+        let rest, elt = extract xs name in
+        (x :: rest, elt)
+  in
```
panglesd, line 76:
```suggestion
  let rec extract paths name =
    match paths with
    | [] -> (paths, (name, None))
    | x :: xs when Fpath.basename x = name -> (xs, (name, Some x))
    | x :: xs ->
        let rest, elt = extract xs name in
        (x :: rest, elt)
  in
```

src/odoc/html_page.ml
```diff
@@ -54,4 +58,46 @@ let extra_documents args unit ~syntax =
...
+               (Paths.Identifier.name parent_id)
+               (Fs.File.to_string asset)))
+        unmatched;
+      List.map
+        (fun (name, path) ->
+          let asset_id = Paths.Identifier.Mk.asset_file (parent_id, name) in
+          let url = Odoc_document.Url.Path.from_identifier asset_id in
+          Odoc_document.Types.Document.Asset { url; src = path })
+        paired
+  | exception Missing_asset name ->
```
panglesd, line 92: Handling the exception here is a bit strange. Only a warning is issued, but this stops the process of assets whenever one is not found...

test/pages/assets.t/run.t
panglesd: I think `odoc` tries to avoid failing too badly in case of errors.

So I don't think it should be a hard failure, but that should not stop processing the other assets (see above)

panglesd (approved):
Nice!
I'm not sure why assets could not be children of compilation unit. But that should not be a too restrictive restriction, so I'm happy with it!
The only thing I would like to see changed is the non-handling of other assets in case of a missing asset.

panglesd:
Also, there are compatibility issues that would need to be taken care of...

dbuenzli:
What's the story from a user and driver (say `odig` which works out of installs) perspective ?

src/document/url.ml
trefis: > Is it the `anchor = ""` which is bothering you?

No, it's that we're getting a url to the parent of the asset, not the asset itself.

trefis:
> What's the story from a user and driver (say `odig` which works out of installs) perspective ?

I guess that's still open?
In the absence of references to / lookup for assets it would indeed be best if the various drivers used the same convention for picking the parents of their asset, but this PR isn't prescriptive in any way in that regard.

I do hope we can add references to assets in the not too distant future to make this question disappear, in the meantime my intuition is that right now the easiest is probably to use the "package root" as parent of the assets.

Does that somewhat answer your question, or did I totally miss your point?

dbuenzli:
> In the absence of references to / lookup for assets it would indeed be best if the various drivers used the same convention for picking the parents of their asset, but this PR isn't prescriptive in any way in that regard.

And yet that's what we need. It was already decided to not prescribe anything for the parent/child relationship and as I suggested at that the time that's not a good idea. We could have had perfectly self-described documents but are moving away from that more and more. 

Look at what happend with upstream not prescribing anything for libraries: eco-system mess. Now `odoc` is perpetuating this.

src/document/url.ml
panglesd: Indeed, that's a bug, thanks for spotting it! It was not harmful because We only call `Path.from_identifier` on the `` `SourcePage`` variant, never `Anchor.from_identifier`.

```suggestion
    | { iv = `AssetFile _; _ } as p ->
        let page = Path.from_identifier p in
        Ok { page; kind = `File; anchor = "" }
```

Could you correct `SourcePage` and `SourceDir` as well?

trefis:
> It was already decided to not prescribe anything for the parent/child relationship and as I suggested at that the time that's not a good idea. We could have had perfectly self-described documents but are moving away from that more and more.

I get your point. However, after years of being away from odoc and having not been involved in this decision, I just came back for this "small" PR.
I do not plan to try and argue against decisions which have been made years ago, and certainly not in a PR like this one.

That being said:
- I hope this PR will still be beneficial to you, and to everyone else
- while odoc itself isn't prescriptive, other tools can and will be; in particular dune (and probably odig) and we are the ones who will write the behavior there, so if you have a *precise* proposal/request now is still a good time to make it known (or repeat it).

panglesd:
I agree that the decision to make `odoc` more prescriptive is out of the scope of this PR, which just adds support for assets in a way that conforms to the current state of odoc.

The code looks good, let's merge, thanks @trefis!

dbuenzli:
> * while odoc itself isn't prescriptive, other tools can and will be; in particular dune (and probably odig) and we are the ones who will write the behavior there, so if you have a _precise_ proposal/request now is still a good time to make it known (or repeat it).

This has been written for ages [here](https://erratique.ch/software/odig/doc/packaging.html#odoc_api_and_manual) and it's mentioned in #59. Interacting with this project remains as frustrating as ever I have to say.

PR #977, opened 2023-06-23
Remove duniverse rules from the Makefile

Julow:
This removes seemingly outdated rules and make the Makefile more readable and usable as documentation.

Is there any user of these rules ?


jonludlam:
We've got opam monorepo for this functionality.

PR #982, opened 2023-07-21
Add a source-root arg to html-generate

jonludlam:
The relative path has already been given when the .cmt was originally compiled, so to avoid having to pass it again this argument allows the root directory of the source tree to be passed, and then odoc finds the actual source file by looking up that relative file within the tree.


src/odoc/html_page.ml
```diff
@@ -16,23 +16,45 @@
...
 open Odoc_model
 
+type source = File of Fpath.t | Root of Fpath.t
+
+let pp fmt = function
+  | File f -> Format.fprintf fmt "File: %a" Fpath.pp f
+  | Root f -> Format.fprintf fmt "File: %a" Fpath.pp f
+
+let to_string f = Format.asprintf "%a" pp f
+
```
panglesd, line 26: Maybe this could be put inside a submodule? The `Html_page.to_string` and `Html_page.pp` names can be misleading, as they are not function dealing with "html pages"...

src/odoc/bin/main.ml
```diff
@@ -714,19 +730,38 @@ module Odoc_html_args = struct
     Arg.(
       value & opt_all convert_fpath [] & info [ "asset" ] ~doc ~docv:"file.ext")
 
+  let source_root =
+    let doc =
+      "Source code root for the compilation unit. It must have been compiled \
+       with --source-parent passed."
```
panglesd, line 736: I feel it might help to add some explanation what is a "source code root".
```suggestion
      "Source code root for the compilation unit. Used to find the source file \
       from the value of --source-name it was compiled with. Incompatible with \
       --source-file."
```
Also, [line 718](https://github.com/ocaml/odoc/pull/982/files#diff-47026f52e88243e29f89adcdd872b0e32fe9f2368df000cc0db39c986e05b111R718) it should be `--source-parent-file` instead of `--source-parent` (or maybe "`--source-parent-file` and `--source-name` to be more explicit)

panglesd (approved):
Looks good to me!

jonludlam:
Thanks @panglesd - applied.

PR #986, opened 2023-07-26
Add `compile-deps --has-src`

panglesd:
Dependencies are not the same when source code rendering is enabled or not. The `--has-src` flag specifies which behaviour to adopt. The logic to lookup the cmt is factorized between `compile` and `compile-deps`.


trefis (approved):
LGTM.

I personally would have kept the cli the same between `compile` and `compile-deps` (so the later would take `--source-name` and `--source-parent-file`). It isn't needed, but it makes it easier to swap one for the other.

jonludlam:
Hold on, can we please just consider having an explicit `--cmt` argument for both compile and compile-deps? @Julow  already thought we had an explicit arg and I think it's clearer to specify it on the command line than just find it next to the cmti file?

Julow:
The reasons against taking the same arguments as `compile` is that the user would have to compute values for them that are never used. It also adds error cases that seem arbitrary.

Taking `--cmt` would remove code that have to guess the path and would make build systems more explicit. I think we decided not to add this flag because there were already a lot: `--cmt foo.cmt --source-name foo.ml --source-parent-file page-src.odoc`. 
Is it reasonable to store `cmti`s in a different directory than the corresponding `cmt`s ? If yes, this flag might be required.

trefis:
> Hold on, can we please just consider having an explicit --cmt argument for both compile and compile-deps?

Can this be done/discussed in a separate PR/issue?
The implicit lookup is already present in `compile`, this PR already has the code to mirror the behavior in `compile-deps`, I'd be in favor of merging the bugfix and doing potential improvements later.

> The reasons against taking the same arguments as compile is that the user would have to compute values for them that are never used. It also adds error cases that seem arbitrary.

I was just sharing my prefences, but I don't have a strong opinion on the matter. Though I would argue that the values will need to be computed for (and be used by) `compile` anyway. So it's not really wasted work, or arbitrary if there is a failure. It's just consistent behavior and advance warning :)

panglesd:
> Can this be done/discussed in a separate PR/issue?

I agree this should be discussed in another PR/issue. Adding `--cmt` feels it can be done independently of this.

> It's just consistent behavior

I agree that it feels more consistent, but at the same time:
- All other `compile` options are refused by `compile-deps`, so anyway you cannot just turn some `odoc compile ...` into `odoc compile-deps ...` and expect it to work.
For instance, `odoc compile -I . --source-name ... foo.cmti` works while `odoc compile-deps -I . --source-name ... foo.cmti` does not since the `-I .` option will be refused.
So, it would feel like inconsistent behaviour: some options from `compile` are accepted in `compile-deps` "as is", and some options are not.
- Having a flag to represent a boolean value seems much more "semantic" than giving `source-name` and `source-parent-file`.

I would be in favour of keeping it like that.

PR #988, opened 2023-08-04
Create .git-blame-ignore-revs

kevinji:
Ignore all ocamlformat commits for `git blame`.

For reference: [Ignore commits in the blame view](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)


Julow (approved):
Looks good to me :)

PR #990, opened 2023-08-21
Span labels in function arguments (`?id:` and `id:`)

### LINKED ISSUE #879: Identifier labels (`?id:`, `id:`) should be <span>ned.
So that we can colorize them like most syntax highlighters do.

I thought about doing a PR myself but nothing evident showed up in a quick glance. Do the labels show up as structured information or you need to munge the type identifiers ?

panglesd:
Fix #879.

If someone has some taste in the css colors, please suggest something! Currently, here is what I chose:

image

image


src/document/generator.ml
```diff
@@ -21,12 +21,12 @@ open Types
...
+let tag tag t = O.span ~attr:tag t
+
 let label t =
   match t with
-  | Odoc_model.Lang.TypeExpr.Label s -> O.txt s
-  | Optional s -> O.txt "?" ++ O.txt s
-
-let tag tag t = O.span ~attr:tag t
+  | Odoc_model.Lang.TypeExpr.Label s -> tag "opt-arg" (O.txt s)
+  | Optional s -> tag "named-arg" (O.txt "?" ++ O.txt s)
```
dbuenzli, line 29: Please avoid inventing new terminology, I'd suggest `label` and `optlabel` as per [language specification](https://v2.ocaml.org/releases/5.0/htmlman/lex.html#sss:labelname).

dbuenzli: (Also you need to swap the names)

panglesd: Good points, thanks. I updated the PR!

gpetiot:
It looks like some tests are not updated, but everything else looks good to me!

gpetiot (approved):

test/generators/html/Type.html
```diff
@@ -89,11 +95,17 @@ <h1>Module <code><span>Type</span></code></h1>
       <span><span class="keyword">type</span> labeled_higher_order</span>
       <span> = 
        <span>
-        <span>(<span>l:int <span class="arrow">&#45;&gt;</span></span> int)
+        <span>(
```
Julow, line 98: Not related to this PR but this reminds me that the `--indent` option breaks the formatting: https://github.com/ocsigen/tyxml/issues/288

Julow (approved):
Looks good to me :)

PR #993, opened 2023-08-28
Stable anchors in links to implementation.

panglesd:
Instead of relying on internals of the compiler to generate anchors in source code rendering, this PR makes "semantic" anchors; thus improving stability between OCaml versions/changes in the implementation.

It is done in the following way:
- For local identifiers (unqualified identifiers):
  - For definition, we generate an id from the name and the position, since names might not be unique. Example of such id: `local_elt_7165`
  - we build a table from `Ident.t` to id,
  - that we use for occurrences (since occurrences don't have access to the location of the definition, an information needed to generate the anchor) 
- For global identifiers (that have a `Uid.t`)
  - When traversing the typedtree, we record the parent structure to generate semantic id from the names. Example of such id: `module-Reference.value-render_resolved`.
  - We build a table `loc_to_id` that we combine with the `uid_to_loc` table to have a `uid_to_id` table, and store this table in `.odoc` files.
  - After resolving shapes, we use the corresponding table to get the id and the link right for occurrences.

Note: This PR is extracted from https://github.com/jonludlam/odoc/tree/source-locations-3, leaving the "Global jump" part aside. It contains some refactoring.


test/sources/source.t/run.t
```diff
@@ -135,21 +135,18 @@ Ids generated in the source code:
...
+  id="module-A"
+  id="module-B"
+  id="module_type-T"
+  id="module_type-U"
+  id="type-ext"
+  id="def_10"
+  id="def_11"
+  id="class-cls"
+  id="class-cls'"
+  id="class_type-ct"
```
panglesd, line 152: The number of anchors has decreased, need some investigation!

panglesd: `exception Exn` seems to be missing.

test/sources/source.t/run.t
```diff
@@ -135,21 +135,18 @@ Ids generated in the source code:
...
+  id="value-y"
+  id="value-z"
+  id="local_a_47"
+  id="module-A"
+  id="module-B"
+  id="module_type-T"
+  id="module_type-U"
+  id="type-ext"
+  id="def_10"
+  id="def_11"
```
panglesd, line 149: The `def_%d` identifiers are for `Uids` whose location was not found during the Typedtree traverse.

src/loader/ident_env.cppo.ml
```diff
@@ -360,77 +387,91 @@ let rec extract_structure_tree_items hide_item items =
...
   List.map (function
     | `Type _ 
     | `Module _
     | `ModuleType _
     | `Value _
     | `Class _
+    | `Exception _
+    | `Extension _
     | `ClassType _ as x -> [x]
     | `Include xs -> xs) items |> List.flatten
```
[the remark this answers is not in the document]
panglesd: No, because `Include` has type `item list` (and not `items list`) so it cannot contain any `Include`...

panglesd:
Thanks @Julow for the review!

I've added one "nontrivial" commit: it seems that functor parameter do have a Uid in the `uid_to_loc` table, so I added that to the traverse. But I had to change a bit the environment so that ids for functor parameter are only created once.

src/model/paths_types.ml
```diff
@@ -42,7 +42,8 @@ module Identifier = struct
 
   type source_location_pv =
     [ `SourceLocationMod of source_page
-    | `SourceLocation of source_page * DefName.t ]
+    | `SourceLocation of source_page * DefName.t
+    | `SourceLocationInt of source_page * LocalName.t ]
```
Julow, line 46: The purpose of the new constructor is not obvious, a comment would be nice. Or perhaps to rename it `SourceLocationInternal` ?

src/loader/shape_.mli
```diff
@@ -0,0 +1,16 @@
+open Odoc_model
```
Julow, line 1: It's unclear that the new name is better. It's purpose did not change and the new alias is conflicting.

src/loader/shape_.mli
```diff
@@ -0,0 +1,16 @@
...
+#else
+
+type t
+
+#endif
+
+val lookup_def :
+  (string -> (Lang.Compilation_unit.t * t) option) ->
+  Identifier.NonSrc.t ->
+  Identifier.SourceLocation.t option
```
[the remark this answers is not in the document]
Julow: You are right! I went too fast to the conclusion.

src/loader/implementation.ml
```diff
@@ -0,0 +1,475 @@
...
+        let loc_opt = Shape.Uid.Tbl.find_opt uid_to_loc uid in
+        match loc_opt with
+        | Some loc ->
+            (Odoc_model.Lang.Source_info.Definition id, pos_of_loc loc) :: acc
+        | _ -> acc)
+      uid_to_id []
+  in
+  defs @ poses
+
+let string_of_full_name_ty : Odoc_model.Paths.Identifier.full_name_ty -> string
```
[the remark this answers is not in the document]
panglesd: I addressed that in https://github.com/ocaml/odoc/pull/993/commits/f5fd9aef68be77f2812fb5e0e79bf31e282f8cdd if you want to have a look!
Basically, I removed `full_name` (in any case, it did not included enough information to generate unique anchors), and reused the `Odoc_document.Url.Anchor.kind`.

panglesd:
I added support for having anchors to constructors, but they will only get one when https://github.com/ocaml/ocaml/pull/12508 is merged (more specifically, https://github.com/ocaml/ocaml/pull/12508/commits/6c1fa737cb147ee3f0395fb6402250cdfc22317c).

src/model/paths_types.ml
panglesd: I like the explicit name: it makes it easier to understand also in occurrences of the constructor.

src/loader/shape_.mli
panglesd: You are rigth! Reverted.

jonludlam:
I know I suggested the `foo_<location>` syntax for anchors for local identifiers, but I'm now wondering whether this is actually stable in the presence of line-ending normalization on windows? @dra27?

src/loader/implementation.ml
```diff
@@ -0,0 +1,517 @@
...
+
+let pos_of_loc loc = (loc.Location.loc_start.pos_cnum, loc.loc_end.pos_cnum)
+
+type annotations =
+  | LocalDefinition of Ident.t
+  | LocalValue of Ident.t
+  | DefJmp of Shape.Uid.t
+
+module Analysis = struct
+  let ( @ ) = List.rev_append
```
jonludlam, line 13: I don't like the redefinition of `@` to be semantically different from the usual `@`! It's confusing.

Also I'm not sure that we care too much about the ordering of things do we? There are a bunch of `List.rev`s below that can be removed if we don't.

panglesd: I guess we can turn the "lift" into a "fold" and carry an accumulator around, that would be the best option.

Yes, the order does not matter.

src/odoc/odoc_link.ml
```diff
@@ -41,9 +41,10 @@ let from_odoc ~resolver ~warnings_options input output =
...
       Odoc_file.save_page output ~warnings page;
       Ok (`Page page)
-  | Unit_content (m, shape) ->
-      link_unit ~resolver ~filename m shape
+  | Unit_content m ->
+      link_unit ~resolver ~filename m
       |> handle_warnings ~input_warnings ~warnings_options
       >>= fun (m, warnings) ->
-      Odoc_file.save_unit output ~warnings (m, shape);
+      let m = {m with Odoc_model.Lang.Compilation_unit.shape = None} in
```
panglesd, line 48: Doesn't it make more sense to modify the compilation unit record only in the `link_unit` function?
When a record is modified in many place, I find it difficult to have an idea of which field will go into the file...

Moreover, it would be nice to add a small comment, just in case it is not clear later (`shapes are not used after the link phase`).

src/loader/implementation.ml
panglesd: Turned the traverse into a fold with an accumulator in 9361f0a. Hopefully it's better like that, with less list concatenations!

panglesd:
> I know I suggested the foo_<location> syntax for anchors for local identifiers, but I'm now wondering whether this is actually stable in the presence of line-ending normalization on windows?

I tested and yes, the "same" file with windows and unix ending won't have the same locations. (One could argue that it's not really the same file...)

We cannot use filename/line number/pos_bol since the anchor might not be unique, due to line directives.
Maybe the best option is to use our own counter for those anchors!

Julow (approved):
A lot has changed in a small amount of time making the review difficult.
But I think this is ready to merge!

src/odoc/odoc_link.ml
```diff
@@ -41,9 +41,16 @@ let from_odoc ~resolver ~warnings_options input output =
...
       |> handle_warnings ~input_warnings ~warnings_options
       >>= fun (m, warnings) ->
-      Odoc_file.save_unit output ~warnings (m, shape);
+      (* Remove the shape here so that we only depend upon odoc types
+         rather than odoc and ocaml types. This means we should be able
+         save an odocl file with odoc x.y compiled with one version of
+         the compiler and load it in odoc x.y compiled with a different
+         version of the compiler. This is an important use case for
+         voodoo. *)
+      let m = { m with Odoc_model.Lang.Compilation_unit.shape = None } in
```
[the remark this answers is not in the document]
jonludlam: Ooh yes, I had originally thought that wouldn't be needed but you're right, we do need to sort that.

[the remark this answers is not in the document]
jonludlam: Hrm, I'm rather more nervous about putting an empty map in there than a None. Might need some more options...

src/loader/odoc_loader.ml
```diff
@@ -45,24 +42,44 @@ exception Not_an_interface
 
 exception Make_root_error of string
 
+#if OCAML_VERSION >= (4, 14, 0)
```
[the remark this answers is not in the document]
jonludlam: Done.

dra27:
> > I know I suggested the foo_ syntax for anchors for local identifiers, but I'm now wondering whether this is actually stable in the presence of line-ending normalization on windows?
> 
> I tested and yes, the "same" file with windows and unix ending won't have the same locations. (One could argue that it's not really the same file...)
> 
> We cannot use filename/line number/pos_bol since the anchor might not be unique, due to line directives. Maybe the best option is to use our own counter for those anchors!

Does the compiler keep actual locations as well? Not looked, but presumably Merlin used _actual_ locations (insensitive to `#line` directives)?

In general, the key thing for being LF/CRLF compatible is to ensure that the No measurement can ever span a new line. So `pos_bol` is unusable but `pos_cnum` is fine. Similarly, for spans you need to start lnum+cnum and end lnum+cnum rather than just end offset.

src/odoc/odoc_link.ml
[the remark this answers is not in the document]
Julow: Looks nice :)

jonludlam:
OK, I think any other issues with this are minor and can be fixed up in follow-up PRs. In it goes!

PR #994, opened 2023-09-01
Update version constraint on Yojson

Julow:
The version 1.6.0 contains a breaking change that we rely on.
This was caught by ocaml-ci's lower-bound check.


Julow:
The lower-bound check is now passing.

panglesd (approved):
Good catch!

PR #996, opened 2023-09-04
Explicitely pass `cmt` file for source code rendering of `cmti`s

panglesd:
Previously, it was "guessed" by looking up a file with the same name but a `.cmt` extension.

`odoc compile-deps` can now take several files and output the merge of all dependencies.


src/odoc/bin/main.ml
```diff
@@ -833,9 +865,9 @@ end)
 
 module Depends = struct
   module Compile = struct
-    let list_dependencies has_src input_file =
+    let list_dependencies input_file =
```
trefis, line 868: Nitpick but: `input_files`* ?

trefis (approved):
LGTM.

test/sources/compile_deps.t/run.t
```diff
@@ -11,18 +11,18 @@ compile-deps
...
   A 21e6137bd9b3aaa3c66960387b5f32c0
-  $ odoc compile-deps a.cmt
-  Stdlib 79b0e9d3b6f7fed07eb3cc2abb961b91
   CamlinternalFormatBasics 8f8f634558798ee408df3c50a5539b15
-  B 903ddd9b7c0fa4ee6d34b4af6358d1e1
+  Stdlib 6d7bf11af14ea68354925f3a37387930
+  $ odoc compile-deps a.cmt
   A 21e6137bd9b3aaa3c66960387b5f32c0
+  B 903ddd9b7c0fa4ee6d34b4af6358d1e1
+  CamlinternalFormatBasics 8f8f634558798ee408df3c50a5539b15
```
jonludlam: This looks very brittle! I guess it's OK until we merge the stable anchors though.

src/odoc/compile.ml
```diff
@@ -120,17 +120,19 @@ let resolve_and_substitute ~resolver ~make_root ~source ~hidden
   let filename = Fs.File.to_string input_file in
   (* [impl_shape] is used to lookup locations in the implementation. It is
```
jonludlam: Not sure this comment is much use any more.

jonludlam: Or perhaps we should just move it down to line 149ish.

src/odoc/bin/main.ml
```diff
@@ -187,14 +192,34 @@ end = struct
...
+        (source_parent_file, source_name, source_cmt, Fs.File.get_ext input)
+      with
+      | Some parent, Some name, None, ".cmt" -> Ok (Some (parent, name, input))
+      | Some parent, Some name, Some cmt, ".cmt" ->
+          if Fpath.equal cmt input then Ok (Some (parent, name, input))
+          else
+            Error
+              (`Cli_error
+                "--cmt has to be equal to the input file when this one has \
+                 .cmt extension.")
```
jonludlam, line 205: This seems redundant - why don't we always read the shape if we're handed a cmt file?

src/odoc/bin/main.ml
```diff
@@ -187,14 +192,34 @@ end = struct
...
+      | Some parent, Some name, Some cmt, _ -> Ok (Some (parent, name, cmt))
+      | Some _, Some _, None, _ ->
+          Error
+            (`Cli_error
+              "--cmt has to be passed when --source-parent-file and \
+               --source-name are passed and the input file is not a cmt file.")
```
jonludlam, line 211: Shouldn't we handle the case where you want to render source but we don't have the cmt?

src/odoc/bin/main.ml
```diff
@@ -187,14 +192,34 @@ end = struct
...
           Error
             (`Cli_error
               "--source-parent-file and --source-name must be passed at the \
                same time.")
-      | None, None -> Ok None
+      | None, None, Some _, _ ->
+          Error
+            (`Cli_error
+              "--cmt should only be passed when --source-parent-file and \
+               --source-name are passed.")
```
jonludlam, line 221: Also, we might want to handle this case too - e.g. for autogenerated sources (ie, `foo.ml-gen`) - in which case you'll need the cmt to handle shape lookup but maybe you don't want the source rendered.

test/sources/compile_deps.t/run.t
```diff
@@ -9,19 +9,21 @@ code rendering is enabled.
...
-  Stdlib 6d7bf11af14ea68354925f3a37387930
-  $ odoc compile-deps a.cmt
-  A 21e6137bd9b3aaa3c66960387b5f32c0
-  B 903ddd9b7c0fa4ee6d34b4af6358d1e1
-  CamlinternalFormatBasics 8f8f634558798ee408df3c50a5539b15
-  Stdlib 6d7bf11af14ea68354925f3a37387930
+  $ odoc compile-deps a.cmti | cut -d ' ' -f 1 | sort
+  A
+  CamlinternalFormatBasics
+  Stdlib
```
jonludlam, line 15: Ah yes, this looks less brittle :-D

jonludlam:
Let's fix the redundant arguments and whatnot in the stable anchors PR. This one can go in as-is and we can rebase on top of it.

src/odoc/bin/main.ml
panglesd: If we allow for a `--cmt` argument for compiling `.cmti` files, then the following command can be used: `odoc compile --cmt <file1.cmt> <file2.cmt>`.

There are several reactions we can have:
1. Forbid the `--cmt` argument when the input is a `.cmt` file, and use the input.
1. Only accept that if the input is consistent: either there is no`--cmt` argument, or `--cmt` and input are the same.
1. Make `--cmt` have precedence over the input file

I decided to go for 2. But yes, we don't need the `--cmt` argument if we are handed a cmt file.

src/odoc/bin/main.ml
panglesd: We can, in which case there will be no jump to location, and if I'm not mistaken, no way to make a link to it.
It will still be accessible by "hardcoding" the link to it.

src/odoc/bin/main.ml
panglesd: I planned to do that with occurrence count (we might want to count occurrence but do not render the source).
So yes, we should do that. (even though I would prefer all sources rendered, including the auto-generated one).

PR #999, opened 2023-09-15
Move the sidebar to be stuck to left of the content

EmileTrotignon:
I propose to change the position of the sidebar that contains the table of content to be closer to the content.
It used to be like that : 
image
Here, the sidebar is quite far from the content. With a smaller screen this may be better.

My proposal is the following :
image
I think it is better for multiple reasons. First of all, you need to move your eyes and mouse less to use the table of content. Also, the previous design did not make it clear that the sidebar contained the table of content for the current page. Because of its position, it felt like a general sidebar for the whole website, and not linked to the current content. My proposal alleviate that issue by placing putting the sidebar closer to the content and under the title of the page.

It would also allow for a more consensual design for the search bar in #972


src/html_support_files/odoc.css
```diff
@@ -740,18 +759,18 @@ td.def-doc *:first-child {
...
   background: var(--toc-background);
   overflow: auto;
   color: var(--toc-color);
   padding-left: 2ex;
   padding-right: 2ex;
+  grid-row: 2;
+  height: fit-content;
+  border: solid 1px var(--border);
+  border-radius: 5px;
+  position:sticky;
```
Julow, line 772: Sticky is interesting but it has a behavior I find confusing:
If the side bar is longer than the screen is high but is smaller than the content, it's not possible to see the bottom of the side bar until we reach the bottom of the page.

Is here a way to make the side bar sticky only if the bottom of it is visible ?
Otherwise, I think I slightly prefer having a unsticky side bar. It's much more intuitive.

Note: We might make the side bar longer in the future.

src/html_support_files/odoc.css
```diff
@@ -253,17 +253,36 @@ body {
 }
 
 body.odoc {
-  max-width: 100ex;
+  max-width: 122ex;
+  display: grid;
+  grid-template-columns: 28ex 90ex;
```
Julow, line 258: What is the goal of this? Shouldn't the other property already control the width ?

This cause overflow when the window is half the width of my screen.

Julow (approved):
I like the new style very much! The previous side bar was invisible. This feels more conventional.

There was unwanted margin when the window is half-width that was probably here to accommodate the previous side bar. I have an improved margin/padding in this branch: https://github.com/Julow/odoc/tree/999-review
Feel free to cherry-pick.

gpetiot:
I notified the ocamlorg team about this issue in case they got some insight about the css

src/html_support_files/odoc.css
EmileTrotignon: > Is here a way to make the side bar sticky only if the bottom of it is visible ?

I agree that would be nice. It is certainly possible with javascript. I will look for a css only solution.
I think the sticky is important because the bar is use for navigation inside the page itself.

src/html_support_files/odoc.css
EmileTrotignon: This is how the grid layout usually work, but there this the `fr` unit I should use here for small width cases

Julow: The grid elements definitely have a size already. Removing this property removes the overflow in my testing.

EmileTrotignon: Yes you are in fact right, the content does not have a width, but the sidebar does, and the body has a max witdh.

src/html_support_files/odoc.css
```diff
@@ -246,24 +246,42 @@ body {
...
-  margin-top: 20px;
-  margin-bottom: 50px;
+  margin-left: auto;
+  margin-right: auto;
+  padding: 0 4ex;
 }
 
 body.odoc {
-  max-width: 100ex;
+  max-width: 122ex;
```
panglesd, line 255: If I can count, this makes the space for the content smaller: `100ex` before, vs `90ex` (`122ex - 28ex - 4ex`).

I liked the large content better!
```suggestion
  max-width: 132ex;
```

src/html_support_files/odoc.css
```diff
@@ -246,24 +246,42 @@ body {
...
 
+.odoc-content {
+  grid-row: 2;
+}
+
+.odoc-content > *:first-child {
+  /* This make the first thing in the content align with the sidebar */
+  padding-top: 0;
+  margin-top: 0;
+}
```
panglesd, line 273: The header (the first child) is not always small, and that makes strange result, I think.
See for instance the "odoc for authors" page of odoc's doc:

image

Why not, but I fear a much longer header would make it look strange, if the toc is pushed too much below...

src/html_support_files/odoc.css
panglesd: Another solution is to have a maximal height for the left bar of the size of height, and add a scroll bar.

However, I do not love any of the solution!
- Sticky, bottom of long nav bar inaccessible unless at the bottom of the page is a "no" for me
- "Custom stickiness", with some js solving the above problem: why not, but not very satisfying
- Not sticky: simple and works, but not the best since we quickly lose the table of content in long pages

src/html_support_files/odoc.css
EmileTrotignon: I am not sure I see what you mean

src/html_support_files/odoc.css
EmileTrotignon: I liked it smaller, but thats fine

src/html_support_files/odoc.css
EmileTrotignon: Ah ! indeed. This is unrelated to that particular rule though. This can be changed by tweaking the grid layout (grid-row and grid-columns rules) (and if we do this rule becomes useless)
I agree it is a bit weird like that, but I think I still like it. I will ponder that a bit more.

sabine:
Ok so I'm back and I checked this out.

IMO sticky sidebar as a concept here is an okay solution since you don't have the problem OCaml.org had with a footer spanning all columns of the design (which would push the sticky sidebar out at the top when you go into the footer).

But here's what you need to do to make it work:

* need to use `max-height:100vh` to make sure we can scroll the sidebar if it's too large
* when using positioning like `top: 50px` (please make it smaller than 50px), you would need to use `max-height: calc(100vh - 50px)` instead. Also, now you can scroll only within the actual content box (the one that has the background). So what you'd probably want to do instead is to have a transparent wrapping container for the `max-height: 100vh` which is sticky and scrollable. Then, have the content box with background sit inside that container with some margin.
* You can play around with bottom margin or padding of the surrounding container to make sure that the scrolling goes as far as it feels "right". Usually, you would have to have enough bottom padding to make up for the height of the page "header", i.e. the space between the very top of the page and the location where the sticky container sits before you scroll - otherwise, people cannot scroll the bottom of the content of the sidebar into view.
Screenshot 2023-10-02 at 10-20-09 Ocamlary (Ocamlary) However, in this case, it's irregular because the preamble sits above the sidebar... maybe a third to a half of the page height would do as bottom padding. On pages with little content, the only consequence is that there's more white space at the bottom - which seems tolerable.

In general, it's simpler to work with fixed position layouts.

gpetiot:
On `_build/default/_doc/_html/odoc/odoc_for_authors.html` the menu is not scrollable and the bottom of it might indeed not be displayed, I'm not sure if that is what Paul-Elliot was also reporting.
The whole css looks a bit overcomplicated to me, could we do something simpler like: https://www.w3schools.com/css/tryit.asp?filename=trycss_navbar_vertical_fixed that would mean having the preamble on the same column as the body.

EmileTrotignon:
@gpetiot I do not think the css you linked to me would be able to achieve the result I desire, at least not without changing the html. I think Sabine's suggestion would fix the issue you mentionned.

src/html_support_files/odoc.css
Julow: There's no perfect solution and I think the scrollable side bar is the best compromise of usability and simplicity.

Julow (approved):
I love the new side bar as it is now :) I'm in favor of merging.

panglesd:
This needs rebasing. I tried a "naive" resolution of conflicts, but it makes the search bar at the bottom of the page...

panglesd:
Also, the changes introduced in the html layout in the search PR changes the way it looks compared to your last screenshot:

image

EmileTrotignon:
I have made a new search compatible-layout. It looks like the following : 
image

src/html_support_files/odoc.css
```diff
@@ -782,20 +784,18 @@ td.def-doc *:first-child {
...
 }
 
-:root {
-}
 
 .odoc-toc {
-  --toc-top: 20px;
-
+  /* TODO : make this smaller if there is no search bar */
+  --toc-top: calc(var(--search-bar-height) + var(--search-padding-top) + 20px);
```
Julow, line 790: Is the search bar sticky as well ? If not, this is not needed at all, `--toc-top: 20px` is sufficient.

test/search/html_search.t/run.t
```diff
@@ -239,7 +239,7 @@ We should now put the scripts where it was:
   $ cp fuse.js.js html/
 
 One way to visually try the search is to indent
-$ cp -r html /tmp/
+  $ cp -r html /tmp/
```
Julow, line 242: We don't want that.

Julow:
This needs an other rebase. I think the commits could be squashed as the history is currently a bit messy.

test/search/html_search.t/run.t
EmileTrotignon: Oops, for sure.

src/html_support_files/odoc.css
EmileTrotignon: >  is not properly spaced from the content is hovers
i do not understand this part
> When the search bar is not present, this calculation adds space above the TOC that doesn't look good either.
I can fix this.
> It's only surrounded by a white shadow that makes it hard to spot where the contents begins 
This is probably fixable, although I do not remember having this issue
>  doesn't fit with Odoc's theme.
Its true the input element is practically unstyled, but I plan to style it in my colors pr. As of now, it is dependent on browser defaults, which are very different and sometimes ugly.

I like the sticky sidebar, for the reason that I think that searching for something while looking at the doc is very useful. If it is not sticky, if you read the doc and the want to search for something, then you have to scroll to the top, and might forget what you wanted to do, and you also cannot type the query while looking at the doc (for instance copying a typename in a type-search query).
Also, the sidebar was not made sticky by this PR, and changing it is a little out of scope.

Julow: > is not properly spaced from the content is hovers

I mean that it's hard to distinguish where the content starts below the search bar. 

> If it is not sticky, if you read the doc and the want to search for something, then you have to scroll to the top, and might forget what you wanted to do, and you also cannot type the query while looking at the doc

I don't think this workflow is well implemented currently. The dropdown showing the search results is above the page and might interfere. On mobile, almost none of the page is visible while typing the search query.
This waste vertical space at all time, not just when I intent to search.

>  Also, the sidebar was not made sticky by this PR, and changing it is a little out of scope.

Agree! I just noticed that the search bar was sticky while reviewing this calculation. Let's not do change the search bar in this PR.

EmileTrotignon: > On mobile, almost none of the page is visible while typing the search query.
True, the search bar should not be sticky on mobile
> This waste vertical space at all time, not just when I intent to search.
That is true, there is a tradeoff here. Maybe with some js magic we can have both, but I am not sure.

Julow: I would trade having to scroll to the top for more vertical space.

EmileTrotignon:
I have squashed everything, I believe this is now ready for merging

src/html_support_files/odoc.css
```diff
@@ -759,19 +783,29 @@ td.def-doc *:first-child {
   line-height: 1.2;
 }
 
+.odoc-search + * + .odoc-toc {
```
Julow, line 786: A comment saying that this is to accommodate the sidebar would be useful.

Julow (approved):
This needs a rebase and then can be merged.

I'd like a comment in the code if you have the time.

src/html_support_files/odoc.css
EmileTrotignon: this is done

Julow:
Perfect, let's merge!

PR #1000, opened 2023-09-15
Creates a dedicated rule to generate support files

EmileTrotignon:
Currently, regenerating support files requires using the test rule, which feels weird because changing the css has very little to do with tests. It also allows to generate the support files without running the test more easily.

### DIFF
--- a/src/html_support_files/dune
+++ b/src/html_support_files/dune
@@ -2,7 +2,7 @@
 ; of the vendored projects.
 
 (rule
- (aliases runtest html-support-files)
+ (aliases runtest support-files)
  (enabled_if
   (> %{ocaml_version} 4.08))
  (deps
--- a/src/html_support_files/dune
+++ b/src/html_support_files/dune
@@ -2,7 +2,7 @@
 ; of the vendored projects.
 
 (rule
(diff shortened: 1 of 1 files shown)


src/html_support_files/dune
```diff
@@ -2,7 +2,7 @@
 ; of the vendored projects.
 
 (rule
- (alias runtest)
+ (alias html-support-files)
```
panglesd, line 5: Can it be both (we want the tests to detect an unpromoted html_support_files)?

panglesd: Nit-picking, but why so long?
```suggestion
 (alias support-files)
```

Julow (commented):
But how do we know whether the support files are uptodate ? The previous rule was tested in CI for free.

src/html_support_files/dune
trefis: Dune's [manual](https://dune.readthedocs.io/en/stable/dune-files.html#rule) suggests using `(aliases runtest support-files)`.

panglesd (approved):
Thanks. This will be useful to solve merge conflicts for html support files without having to recompile everything.

Issue #985, opened 2023-07-25
Syntax for including images


panglesd:
This issue is for discussing the syntax for including images in odoc. Since the parser has been remerged into this repo, I open it here!

### External and internal images

I would like to support "external" and "internal" images, just as for links. The "src attribute" for external images would be given as a link, and as a reference (to an asset) for internal images.

### Alternative text

I think that we should encourage the use of an alternative text, for accessibility. The [recommanded rules](https://html.spec.whatwg.org/multipage/images.html#alt) for alternative text are:
- A replacement for the image when the image is "informative"
- An empty alt text when the image is "decorative"
- No alt text when providing an alt text is not really possible (for instance, a user-uploaded image.)

In the case of odoc, I guess we can always provide an alt text (case 1 and 2), so the syntax should hopefully incentivize people to add one.

### Compatibility with metadata extension

At some point, during the dev meetings and private discussions, the need to add metadata to the odoc language was mentioned (it first came when discussing how to specify alignments in table columns, in "heavy syntax").

Some things were mentioned:
- `{tag[key1=value1 key2=value2] content}`. For instance `{table[align="LCR"]...}`.
- `{{tag key1=value1 key2=value2} content}`. For instance `{{table align="LCR"}...}`. This looks a bit similar to `{{!ref}content}` and `{{:link}content}`.

While this was only informal discussions, I think the new syntax for image should be designed not to prevent the addition of metadata!

### The syntax

With these three requirements in mind, here is one proposition (option A) `{{img!ref}alt text}`:

```
Using a reference:
Home town: {{img!carouge.png}Carouge}.

Using a link:
Home town: {{img:http://carouge.com/carouge.png}Carouge}.

With an empty alternative text:
I lived in {{img!carouge.png}} Carouge.
I lived in {{img:http://carouge.com/carouge.png}} Carouge.
```

I think it is nice for the following reason:
- The syntax for external and internal images is quite consistent with links and references
- It incentivize people to write alternative text, as `{{img!ref}}` feels wrong (`{img!ref}` could arguably be added).
- It is compatible with at least one option for adding metadata: `{{img!carouge.png title="A photo of Carouge"}Carouge}`

### Other possibilities

- Option B: `{img{!ref}alt text}`

```
Using a reference:
Home town: {img{!carouge.png}Carouge}.

Using a link:
Home town: {img{:http://carouge.com/carouge.png}Carouge}.

With an empty alternative text:
I lived in {img{!carouge.png}} Carouge.
[...]

PR #1002, opened 2023-09-20
Asset references

panglesd:
Assets were introduced in #975.

This PR adds references to assets.

Resolving of reference to asset work like this: 
- Search the asset in the current page's asset (or the explicitely given page)
- if not found, recurse in the parent page.

In case of unqualified reference (e.g. `"caml.gif"`, `page-name."caml.gif"`) the asset possibility is tested last.

References to assets is a prerequesite to:
- Search #972 
- Medias #985


test/xref2/references_to_assets.t/run.t
```diff
@@ -0,0 +1,68 @@
+In this file, we test the resolving of asset references.
+
+More precisely we test resolving an an asset reference where the asset lives:
+- in the current page (index.mld references ocaml.gif)
+- in a parent page (test.mli references ocaml.gif)
+- in a sibling page (test.mli references ocaml_not.gif, through page-other_page.ocaml_not.gif)
```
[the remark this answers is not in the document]
trefis: Also: you have references to children pages.

trefis (approved):
LGTM modulo nitpicks

dbuenzli:
Do you actually have a design doc somewhere on how all this is supposed to work because I'm afraid to say it doesn't make much sense to me. In an opam install how do you know which assets is attached to which page ?

panglesd:
No, I don't have a design doc, but I finally realized that all these parent-child relationship do not make sense in terms of opam installs, and that we will need to do something about that before going further...
Thanks for the insight!

dbuenzli:
If I understand correctly these parent-child relationships are mainly there to handle the package versioning on `ocaml.org` (or at least that's how it was sold to me when I said it wouldn't be such a good idea not to have self-describing documents). 

If we follow the [odig conventions](https://erratique.ch/software/odig/doc/packaging.html) (which as far as I know `dune` does). Then maybe to generate the docs of a package we can simply:

1. Make `index.mld` the root of a parent-child relationship for the package (note odig generates an `index.mld` if there is none provided). 
2. Any module found in the package is made a child of `index.mld`
3. Any page found in `odoc-pages` (cf convention) is made a child of it. 
4. Any asset found (using file extensions) in `odoc-pages` and/or `odoc-assets` is attached to `index.mld`.

Since IIUC this PR a child can access it's parent assets then any asset is available for use in all the doc strings and manuals of the package. 

For users it's a simple model: your package simply has a folder with `odoc` pages and assets which are all accessible flately. This should be entirely sufficient for 98% of the packages out there.

Does that make sense ?

panglesd:
Yes, this makes sense to me.

I think simply adding some documentation about this on odoc's doc, maybe on the one about the driver, would suffice.

Maybe, one day, one would want to use this parent-child relationship even inside a package, but I'm convinced that the package-flat hierarchy is enough for 99% of the usecases.

panglesd:
(which reminds me: this PR is lacking documentation update on assets)

panglesd:
The state of this PR is:
There are some refactoring suggested by @jonludlam: Populate the environment for once at the beginning, to have all assets available, instead of looking them up recursively on demand.

Also, I think a decision regarding the referencing and resolving of pages should be made before making a hard decision on this one!

doc/parent_child_spec.mld
```diff
@@ -175,17 +175,17 @@ installed and might be used by a different driver.
 In order for drivers to build consistent documentation for a package, the
 following convention should be followed.
 
-- [.mld] pages are installed in a package's [share] directory, under the
+- [.mld] pages and assets are installed in a package's [share] directory, under the
```
panglesd, line 178: Should assets be installed in a particular directory, or with a particular name?
With the proposed convention, assets with extension `.mld` are recognized as pages. But I think this is fine. If really needed, we could find some ways to work around the restriction with extension `.mld_`... ?

panglesd:
I implemented @jonludlam's suggestion to populate the environment with available assets, which I agree makes the code better.

I also added an explanation on the convention for assets in installed packages.

Currently, the resolving of assets is going "up" in the parent-child hierarchy, looking for the asset. This might not be the "final" strategy: we might want to change this to be closer to what is done with pages (see the discussion on that in #1011). However, I believe that it is safe to start with this strategy, as I don't expect a change in the strategy to break (future) existing documentation.

So in my opinion, this PR is ready to be merged.

src/xref2/env.ml
```diff
@@ -84,6 +84,7 @@ type kind =
   | Kind_Exception
   | Kind_Extension
   | Kind_Field
+  | Kind_asset
```
gpetiot, line 87: nit-picking but if we follow the convention of other constructors: `Kind_Asset`

gpetiot (approved):
A few typos, and a few suggestions, otherwise looks good to me!

src/xref2/ref_tools.ml
```diff
@@ -784,7 +824,12 @@ let resolve_reference_dot_sg env ~parent_path ~parent_ref ~parent_sg name =
       Error (`Find_by_name (`Any, name))
 
 let resolve_reference_dot_page env page name =
-  L.in_page env page name >>= resolved1
+  match (L.in_page env page name, page) with
```
Julow, line 827: That's not the usual shape of functions in this module. To make this more efficient and maintainable, it should match on the result of a single search like `resolve_reference_dot_type` below.
The correct `find` function is not in the `Find` module yet, it is currently inlined in `L.in_page` and should be extended for assets.

src/xref2/ref_tools.ml
```diff
@@ -806,10 +851,30 @@ let resolve_reference_dot env parent name =
   | (`C _ | `CT _) as p -> resolve_reference_dot_class env p name
   | `P _ as page -> resolve_reference_dot_page env page name
 
+let resolve_page_reference env (r : Reference.Page.t) =
+  match r with
+  | `Resolved _ -> failwith "unimplemented"
```
Julow, line 856: The `` `Resolved `` constructor for pages is never constructed so it could be dropped from the type.

test/xref2/references_to_assets.t/run.t
```diff
@@ -0,0 +1,71 @@
+In this file, we test the resolving of asset references.
```
Julow, line 1: Nothing in this test shows a working asset reference. Shouldn't `caml.gif_hidden` be renamed ?
Also, the only reference that seems to be intended to work do not test page parents (there's no `other_page."caml.gif"`).

Julow (commented):
Nice :)

src/model/reference.ml
```diff
@@ -298,6 +299,26 @@ let parse whole_reference_location s :
...
+        match kind with
+        | `TUnknown -> `Dot (label_parent next_token tokens, identifier)
+        | _ ->
+            let suggestion =
+              Printf.sprintf "'page-%s' should be first." identifier
+            in
+            not_allowed ~what:"Page label"
+              ~in_what:"the last component of a reference path" ~suggestion
+              location
+            |> Error.raise_exception)
```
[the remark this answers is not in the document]
panglesd: Thanks, there was something wrong here!

Pages reference cannot have a parent, and I want to preserve this behavior in this PR. Should be better now.

test/xref2/references_to_assets.t/run.t
panglesd: Yes, at the end there are working asset references. The assets are not passed in the html generation phase, which creates "dead links" just as there are dead links when you do not html-generate some of the pages), but the test checks that the links point to the right place.
Do you think it would be better to actually pass some image in the html generation, to compare the path more accurately?

panglesd: I went ahead and passed assets in the test.

panglesd:
Closing now since asset referencing will change subsequently soon. But some of the code will be used!

PR #1005, opened 2023-09-26
Medias in odoc

panglesd:
Based on #1002.
See #985 for discussion about the syntax.

## The syntax

The syntax is the following, where `<media>` can be either `image`, `video` or `audio`:
```
{<media>!reference_to_asset}

{{<media>!reference_to_asset}replacement/alternative text}

{<media>:link_to_asset}

{{<media>:link_to_asset}replacement/alternative text}
```

The media elements are block elements, and as such should be separated from the rest of the content with new lines.
They cannot be put inline (so, for now, no `{{:link-to-something}{image!ref}}`...) or even inside  other block elements (such as table, lists, ...).
The reason for this is that without attributes (e.g. to define the size) it would be quite hard to style the inserted media as an inline in a consistent way...
I think it is enough for most cases. Searching github for `{html%` and embedding of images seems to comfort this.

## The rendering

Only html display medias, using the html5 elements `audio`, `video` and `img`. Even images are not displayed in latex (it lacks "asset" support).

For images, the "replacement text" is used an alt text. The replacement text is also used in case of unresolved reference.

For latex and manpage backend, the replacement text is always used.

## Future work

I wondered whether something such as `{href!reference.to."asset.png"}`, which would render as `../../reference/to/asset.png` (for instance) would be a good addition? It would allow more html embedding, such as `{html% %}`.
What do you think?

Having a way to specify the size (but this requires some syntax discussion!), or to add inline media, would be nice.


src/document/comment.ml
```diff
@@ -326,6 +334,38 @@ let attached_block_element : Comment.attached_block_element -> Block.t =
...
+        | `Reference (`Resolved r) -> (
+            let id =
+              Odoc_model.Paths.Reference.Resolved.(identifier (r :> t))
+            in
+            match Url.from_identifier ~stop_before:false id with
+            | Ok url -> Target.Internal (Resolved url)
+            | Error exn ->
+                (* FIXME: better error message *)
+                Printf.eprintf "Id.href failed: %S\n%!"
+                  (Url.Error.to_string exn);
```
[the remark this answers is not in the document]
panglesd: That's taken from a "warning" thrown in similar situations.
However, yes, we could finally fix those `FIXME` instead of propagating them through the code...
Related: #1010 which removed similar unconditional eprintfs.

src/document/types.ml
```diff
@@ -90,6 +94,9 @@ and Block : sig
     | Verbatim of string
     | Raw_markup of Raw_markup.t
     | Table of t Table.t
+    | Image of Target.t * Inline.t
+    | Video of Target.t * Inline.t
+    | Audio of Target.t * Inline.t
```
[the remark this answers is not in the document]
panglesd: I think both implementations are very similar but I can try to see if it can make some part of the code more factorized!

dbuenzli:
> They cannot be put inline (so, for now, no `{{:link-to-something}{image!ref}}`...) or even inside other block elements (such as table, lists, ...).

That seems rather limiting. I don't really care about the inline bit but you definitively want to put these things into lists, for example:

```
This can be either:
{ul
{- A sine wave {audio!sine.wav}}
{- A triangle wave {audio!tri.wav}}}
```

I don't really care if the assets element ends up being a block level element rather than an inline but I definitively want it into list elements.

Julow:
I'm in favor of medias to be (nestable) block elements. These elements take time to show up and might never do. I think it's important to make sure they are visible, for example by nesting them in a block with a border and a large padding.
We also need space for a caption in the syntax and in the rendering.

panglesd:
I turned medias into nestable block elements: it is now possible to add them inside lists and tables.
They will still be displayed as blocks. I haven't added any fixed size, border...
About the caption, I think this should be left for _figures_. I could add a new media, `figure` whose "replacement text" is the caption.

src/html/generator.ml
```diff
@@ -222,6 +243,55 @@ let rec block ~config ~resolve (l : Block.t) : flow Html.elt list =
...
+        in
+        let block =
+          match target with
+          | External url -> image url
+          | Internal (Resolved uri) ->
+              let url = Link.href ~config ~resolve uri in
+              image url
+          | Internal Unresolved ->
+              let a = [ Html.a_class [ "xref-unresolved" ] ] in
+              [ Html.span ~a (inline ~config ~resolve alt) ]
```
gpetiot, line 292: For unresolved images, what do you think of displaying placeholder images like: 
Image_not_available (1)

src/xref2/ref_tools.ml
```diff
@@ -349,7 +349,7 @@ module L = struct
             when name = LabelName.to_string name' ->
               Ok (`Identifier label)
           | _ -> find tl)
-      | [] -> Error (`Find_by_name (`Page, name))
+      | [] -> Error (`Find_by_name (`Label, name))
```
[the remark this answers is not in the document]
panglesd: Yes, it could be made in another PR. Not sure that's worth it, given how small it is! But if anyone confirm it's helpful, I'll separate it into another PR!

src/html/generator.ml
panglesd: When there is no alt text, yes that's a good idea! (are we allowed to use this particular image?)

When there is alt text, I'm not sure what's best!

dbuenzli: Personally I think it's better to let things break. This makes it mechanically discoverable (404) if you add a stub, then the page has no problem as far as machine are concerned.

Julow: Readers also need to distinguish missing images. Perhaps the alt text could be made mandatory ?
That's what I was most worried about when I suggested to add border, padding and a caption around image blocks.

dbuenzli: > Readers also need to distinguish missing images.

Browsers already have their own way of notifying users about broken image sources (typically via a small icon representing a broken document), there's no need to invent anything here.

dbuenzli: Ah but of course this should not be an `a` this should be an `img` tag with the class `xref-unresolved` and perhaps with the `src` attribute being simply the undefined reference (so that the broken icon shows up, it does not with an empty `src`).

panglesd: Currently it's a `span` (the `a` is for attribute) which include visual feedback that something is wrong, but nothing tells the user or the browser that an image should be there. So yes, I think an `img` with broken url and the alt text as alt text (if there is one) makes sense.
I'll check if "broken url" can be made a real concept, and how the browser react!

panglesd:
Here is the state of this PR!

It is mostly finished, I think. There is one remaining question: what do we do with unresolved medias.

@dbuenzli suggested to use an "unresolved" URL, so that the browser is responsible for showing the breakage.
I think it is a good idea, I just have some trouble accepting to forge an invalid url. Putting the reference as an url is making my brain output "Error: This expression has type reference but an expression was expected of type url" even if both ultimately are strings. It also creates a web request which, maybe, could work and show something unintended?

I have tested how my browser (firefox) displays an image with an url that returns 404. When there is no alt text, it shows an icon of a broken image. When there is an empty alt-text, it shows nothing. When there is a non-empty alt-text, it shows the alt-text as an image. All of this is a very good behaviour in my opinion.
With video, it is also very clear when the file is not found.
With audio however, the normal controls are shown, but as if the file were of length 0s, nothing happens when you click on the "play" button, there is no indication that a file could not be loaded.

Also, to note, I have tried some "invalid" urls, such as `\\`, and it shows the broken image without needing to make a network request. But this is very browser implementation specific...

I won't have time to work on this more in the coming month.

dbuenzli:
> @dbuenzli suggested to use an "unresolved" URL, so that the browser is responsible for showing the breakage.
> I think it is a good idea,

In fact I forgot that `odoc` would already tell you that something is wrong, this reference doesn't exist. So there's little interest in these 404, there is already a way to discover these errors when you produce the docs. So maybe you can just output text "Missing {image,audio,video} asset" classified with `xref-unresolved` in a span or a div depending whether it's a block or inline element.

src/model_desc/paths_desc.ml
```diff
@@ -194,6 +196,7 @@ module General_paths = struct
       | `TType -> C0 "`TType"
       | `TUnknown -> C0 "`TUnknown"
       | `TValue -> C0 "`TValue"
+      | `TAsset -> C0 "`TValue"
```
gpetiot, line 199:
```suggestion
      | `TAsset -> C0 "`TAsset"
```

src/parser/ast.ml
```diff
@@ -57,7 +60,12 @@ and nestable_block_element =
     * [ `Light | `Heavy ]
     * nestable_block_element with_location list list
   | `Table of table
-  | `Math_block of string  (** @since 2.0.0 *) ]
+  | `Math_block of string  (** @since 2.0.0 *)
+  | `Media of
+    reference_kind
+    * media_href with_location
+    * inline_element with_location list
+    * media (** @since 2.3.0 *)]
```
gpetiot, line 68: probably `2.4.0`?

src/parser/token.ml
```diff
@@ -124,6 +141,21 @@ let print : [< t ] -> string = function
   | `Tag `Hidden -> "'@hidden"
   | `Raw_markup (None, _) -> "'{%...%}'"
   | `Raw_markup (Some target, _) -> "'{%" ^ target ^ ":...%}'"
+  | `Simple_media (`Reference _, `Image) -> "{image!...}"
+  | `Simple_media (`Reference _, `Audio) -> "{audio!...}"
+  | `Simple_media (`Reference _, `Video) -> "{video!...}"
+  | `Simple_media (`Link _, `Image) -> "{image!...}"
```
gpetiot, line 147:
```suggestion
  | `Simple_media (`Link _, `Image) -> "{image:...}"
```

src/parser/token.ml
```diff
@@ -145,6 +177,24 @@ let describe : [< t | `Comment ] -> string = function
   | `Simple_reference _ -> "'{!...}' (cross-reference)"
   | `Begin_reference_with_replacement_text _ ->
       "'{{!...} ...}' (cross-reference)"
+  | `Simple_media (`Reference _, `Image) -> "'{image!...}' (image-reference)"
+  | `Simple_media (`Reference _, `Audio) -> "'{audio!...}' (audio-reference)"
+  | `Simple_media (`Reference _, `Video) -> "'{video!...}' (video-reference)"
+  | `Simple_media (`Link _, `Image) -> "'{image!...}' (image-link)"
```
gpetiot, line 183:
```suggestion
  | `Simple_media (`Link _, `Image) -> "'{image:...}' (image-link)"
```

src/parser/token.ml
```diff
@@ -145,6 +177,24 @@ let describe : [< t | `Comment ] -> string = function
...
+  | `Simple_media (`Link _, `Audio) -> "'{audio:...}' (audio-reference)"
```
gpetiot, line 184:
```suggestion
  | `Simple_media (`Link _, `Audio) -> "'{audio:...}' (audio-link)"
```

src/parser/token.ml
```diff
@@ -145,6 +177,24 @@ let describe : [< t | `Comment ] -> string = function
...
+  | `Simple_media (`Link _, `Video) -> "'{video:...}' (video-reference)"
```
gpetiot, line 185:
```suggestion
  | `Simple_media (`Link _, `Video) -> "'{video:...}' (video-link)"
```

src/parser/token.ml
```diff
@@ -145,6 +177,24 @@ let describe : [< t | `Comment ] -> string = function
...
+  | `Begin_media_with_replacement_text (`Reference _, `Image) ->
+      "'{{image!...} ...}' (image-reference)"
+  | `Begin_media_with_replacement_text (`Reference _, `Audio) ->
+      "'{{audio!...} ...}' (audio-reference)"
+  | `Begin_media_with_replacement_text (`Reference _, `Video) ->
+      "'{{video!...} ...}' (video-reference)"
+  | `Begin_media_with_replacement_text (`Link _, `Image) ->
+      "'{{image:...} ...}' (image-reference)"
```
gpetiot, line 193:
```suggestion
      "'{{image:...} ...}' (image-link)"
```

src/parser/token.ml
```diff
@@ -145,6 +177,24 @@ let describe : [< t | `Comment ] -> string = function
...
+  | `Begin_media_with_replacement_text (`Link _, `Audio) ->
+      "'{{audio:...} ...}' (audio-reference)"
```
gpetiot, line 195:
```suggestion
      "'{{audio:...} ...}' (audio-link)"
```

src/parser/token.ml
```diff
@@ -145,6 +177,24 @@ let describe : [< t | `Comment ] -> string = function
...
+  | `Begin_media_with_replacement_text (`Link _, `Video) ->
+      "'{{video:...} ...}' (video-reference)"
```
gpetiot, line 197:
```suggestion
      "'{{video:...} ...}' (video-link)"
```

src/xref2/ref_tools.ml
```diff
@@ -729,26 +770,55 @@ let resolve_reference_dot env parent name =
   | (`C _ | `CT _) as p -> resolve_reference_dot_class env p name
   | `P _ as page -> resolve_reference_dot_page env page name
 
+let resolve_page_reference env (r : Reference.Page.t) =
+  match r with
+  | `Resolved _ -> failwith "unimplemented"
```
gpetiot, line 775: Let's not forget this

gpetiot (commented):
A few typos left in the code.

Otherwise, it looks good to me, is something else missing or blocking this PR?

jonludlam:
> Otherwise, it looks good to me, is something else missing or blocking this PR?

Yes, I'd just like to try the suggestion in #1002 before we merge that, then this can be rebased on top.

favonia:
I have a related request which I believe will come up once this PR is merged: I hope there's a way for odoc to download external images, audios, and maybe videos and generate offline-viewable pages. The reason is that I might want to generate a self-contained documentation site for offline reading without putting everything in the same git repo.

Maybe this can be done in a different PR, but I hope the current design is compatible with this request.

PR #1008, opened 2023-09-27
Update CHANGES

jonludlam:

PR #1009, opened 2023-09-27
Add changelog check in CI

panglesd:
Add a changelog check in CI, to avoid having to do #1008 before releasing.
To disable the check on a PR, we should add the "no changelog" label.


Julow (approved):
I'm completely in favor! Contributors and reviewers are the right persons to write good change entries.

Julow:
I've created the label.

gpetiot (approved):
LGTM, we use the same on ocamlformat, very useful!

PR #1010, opened 2023-09-27
Remove unconditional eprintfs and fix a broken reference

jonludlam:


Julow (approved):
Looks good but needs to be formatted.

panglesd:
Wouldn't it be better to keep the warning but use the "warning framework" to control it via CLI?

(If it complicates the too much and is not worth it, I'm OK with merging the changes after reformatting!)

jonludlam:
It's slightly complicated - we don't have a warnings framework for the renderers bit. We do for the compile phase, but it just didn't seem worth it right now. It made it very noisy when building docs for external dependencies in dune!

PR #1012, opened 2023-10-06
Improvements to benchmarks and profiling

Julow:
The first commit comment-out a command that should have been.

The second commit gives predictable and useful names to landmarks output files. Previously, names contained random numbers.

The third commit adds a benchmarks: The time spent running the slowest command. This metric might help recognize large performance regressions in rare cases that we might have missed otherwise.


Julow:
I'll merge this as it only touch `driver.mld` and is useful now.

PR #1014, opened 2023-10-09
Set Component.Delayed.eager to true

Julow:
On my machine, this doesn't have an impact on the benchmarks.
I open this PR so it can run on the benchmarks infrastructure and be compared to previous runs.


Julow:
The benchmarks first reported +3.3% compile time and +8.9% link time. I then replaced the implementation of Delayed with the identity function, the numbers are +2.5% for compile and +8.1% for link, confirming it's not noise.

I explain such a low number with that `Delayed` is useful on some modules only, where it would have a bigger impact. Delayed have been tuned on pathological cases, that are not part of the benchmarks.
Also, the substitution algorithm that requires `Delayed` to not blow up in some cases looks very costly in every cases.

Julow:
Since https://github.com/ocaml/odoc/pull/1012 we can see that the slowest compile invocation takes 36% more time.

Julow:
This was just a test and isn't meant to be merged.

PR #1019, opened 2023-10-13
Print 'private' keyword for private type extensions

### LINKED ISSUE #1017: Document `private` constructors
With odoc version 2.2.1, the documentation generated from e.g.

```ocaml
type _ Effect.t += private Yield : unit Effect.t
```

does not show that the constructor `Yield` is private.

This is an important part of the interface and should be documented.

gpetiot:
This should fix #1017

### DIFF
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -646,6 +646,9 @@ module Make (Syntax : SYNTAX) = struct
         O.keyword "type" ++ O.txt " "
         ++ Link.from_path (t.type_path :> Paths.Path.t)
         ++ O.txt " +=" ++ O.sp
+        ++
+        if t.private_ then O.keyword Syntax.Type.private_keyword ++ O.sp
+        else O.noop
       in
       let content =
         O.documentedSrc prefix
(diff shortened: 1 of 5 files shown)


src/document/generator.ml
```diff
@@ -646,6 +646,11 @@ module Make (Syntax : SYNTAX) = struct
         O.keyword "type" ++ O.txt " "
         ++ Link.from_path (t.type_path :> Paths.Path.t)
         ++ O.txt " +=" ++ O.sp
+        ++
+        if t.private_ then
+          O.keyword Syntax.Type.private_keyword
+          ++ match t.constructors with [] -> O.noop | _ :: _ -> O.sp
```
panglesd, line 652: I think that extensions never have an empty list of new constructor, so it's safe and more readable to write:

```suggestion
          O.keyword Syntax.Type.private_keyword ++ O.sp
```
(I guess ocamlformat would format it this way)

panglesd (approved):
Thanks for the useful improvement!

panglesd:
If you rebase your PR on top of master, we could check whether the CI complains about the absence of changelog entry, since https://github.com/ocaml/odoc/pull/1009 has been merged!

Issue #1020, opened 2023-10-15
Dark mode for code snippets


polytypic:
Code snippets in odoc HTML output are poorly readable (IMHO) in dark mode.

Here is a code snippet in dark mode:

<img width="610" alt="Screenshot 2023-10-15 at 16 50 53" src="https://github.com/ocaml/odoc/assets/5804945/9985d2b0-34b2-4dfd-81c0-e7c01028531b">

Here is the same snippet in light mode:

<img width="613" alt="Screenshot 2023-10-15 at 16 51 21" src="https://github.com/ocaml/odoc/assets/5804945/5dc33836-8860-4b45-ac1c-043342d89f60">

Aside from the background and main foreground color the highlight colors appear to be unchanged and seem to designed for light mode.

Julow:
@EmileTrotignon has improved the code blocks themes in https://github.com/ocaml/odoc/pull/1023
What do you think ?

polytypic:
I assume the PR changes to the color scheme in [this](https://github.com/ocaml/odoc/pull/1023#issuecomment-1801696489) message.  That looks much improved to me! 👍 

Hmm... I think I personally might like a little bit higher contrast in the dark mode (especially for keywords), but overall the improvement is substantial and I'd rather have the new scheme from the PR than wait for a perfect [bikeshed](https://bikeshed.com/) color. :D

EmileTrotignon:
> I'd rather have the new scheme from the PR than wait for a perfect [bikeshed](https://bikeshed.com/) color. :D
And the change also comes with a nice refactoring to the ways the color work, that should hopefully make it much easier to make more changes in the future.

PR #1023, opened 2023-10-19
change color to gruvbox theme

EmileTrotignon:
This PR changes the color theme in odoc to match the gruvbox color set ( https://github.com/morhetz/gruvbox )
The dark theme looks like that : 
image
And the light theme like that : 
image
It was inspired by issue #1020 , but does more than what was asked there.

### DIFF
--- a/src/html_support_files/odoc.css
+++ b/src/html_support_files/odoc.css
@@ -98,138 +98,175 @@
 
   scroll-padding-top: calc(var(--search-bar-height) + var(--search-padding-top) + 1em);
 
-  --main-background: #FFFFFF;
+  /* light gruvbox theme colors */
+  --bg: #f9f5d7;
+  --bg_h:   #fbf1c7;
+  --bg_s: #f2e5bc;
+  --bg1:  #ebdbb2;
+  --bg2:  #d5c4a1;
+  --bg3:  #bdae93;
+  --bg4:  #a89984;
 
-  --color: #333333;
-  --link-color: #2C94BD;
-  --source-color: grey;
-  --anchor-hover: #555;
-  --anchor-color: #d5d5d5;
-  --xref-shadow: #cc6666;
-  --header-shadow: #ddd;
-  --by-name-version-color: #aaa;
-  --by-name-nav-link-color: #222;
-  --target-background: rgba(187, 239, 253, 0.3);
-  --target-shadow: rgba(187, 239, 253, 0.8);
-  --pre-border-color: #eee;
-  --code-background: #f6f8fa;
+  --fg:  #282828;
+  --fg1: #3c3836;
+  --fg2: #504945;
+  --fg3: #665c54;
+  --fg4: #7c6f64;
 
-  --li-code-background: #f6f8fa;
-  --li-code-color: #0d2b3e;
-  --toc-color: #1F2D3D;
-  --toc-before-color: #777;
-  --toc-background: #f6f8fa;
-  --toc-background-emph: #ecf0f5;
-  --toc-list-border: #ccc;
+  --red:    #9d0006;
+  --green:  #79740e;
+  --yellow: #b57614;
+  --blue:   #076678;
+  --purple: #8f3f71;
+  --aqua:   #427b58;
+  --orange: #af3a03;
+  --gray:   #928374;
 
-  --spec-summary-border-color: #5c9cf5;
-  --spec-label-color: green;
-  --spec-summary-background: var(--code-background);
-  --spec-summary-hover-background: #ebeff2;
-  --spec-details-after-background: rgba(0, 4, 15, 0.05);
-  --spec-details-after-shadow: rgba(204, 204, 204, 0.53);
-
-  --search-results-border: #bbb;
-  --search-results-shadow: #bbb;
-
-  --search-snake: #82aaff;
-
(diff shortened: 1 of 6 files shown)


panglesd:
Thanks!

(The PR needs `dune build @support-files`.)

I do like the dark mode!

However, the light mode seems a little yellow to me. Those colors might work well for an editor, but I don't think they work well for a website (but maybe after looking at it for sufficiently long my opinion would change).

In any case, I think the "code block theme" part is a clear improvement!

So I'm hesitating between several scenarios:
1. Merging only the "code block" part, solving #1020 
2. Merging code blocks (for all mode) and dark mode
3. Merging the PR with everything.

The problem with 2. is that the light and dark version do not correspond anymore... (do they correspond currently?)

We could also do 1., followed by #826, followed by the addition of a gruvbox theme?

EmileTrotignon:
I have done a bit of cleanup, I remove unused variables, and picked a color from the palette for almost everything.
I personnaly, the scenario I favor is to merge everything, but I am not opposed to changing the light-theme color palette to something that pleases everyone. This may sounds like option 2 if we go back to the previous colors for everything but code, but the way I organized the code is better and should be merged even if the colors are the same in the end.

src/html_support_files/odoc.css
```diff
@@ -621,10 +657,10 @@ div.def-doc>*:first-child {
 
 /* FIXME: Does not work in Firefox. */
 .odoc-include summary::-webkit-details-marker {
-  color: #888;
+  color: #888; /* todo : use color from palette */
   transform: scaleX(-1);
   position: absolute;
-  top: calc(50% - 5px);
+  top: calc(50% - 5px);include
```
panglesd, line 663: Typo? Or CSS4?

```suggestion
  top: calc(50% - 5px);
```

src/html_support_files/odoc.css
```diff
@@ -386,7 +422,7 @@ a.anchor {
 }
 
 .xref-unresolved {
-  color: #2C94BD;
+  color: var(--blue-dim);
```
panglesd, line 425: Could this be a semantic variable? Such as `--xref-unresolved`?

panglesd (commented):
> [...] I am not opposed to changing the light-theme color palette to something that pleases everyone. [...], but the way I organized the code is better and should be merged even if the colors are the same in the end.

I agree that the code is much better organized your way! For sure, this part should be merged.

Let's wait a little bit to see if someone else likes or complains about the light color theme.

src/html_support_files/odoc.css
```diff
@@ -903,84 +939,80 @@ td.def-doc *:first-child {
...
 }
 
 .VAL, .TYPE, .LET, .REC, .IN, .OPEN, .NONREC, .MODULE, .METHOD, .LETOP, .INHERIT, .INCLUDE, .FUNCTOR, .EXTERNAL, .CONSTRAINT, .ASSERT, .AND, .END, .CLASS, .STRUCT, .SIG {
-  color: #859900;;
+  color: var(--hljs-keyword);;
 }
 
 .WITH, .WHILE, .WHEN, .VIRTUAL, .TRY, .TO, .THEN, .PRIVATE, .OF, .NEW, .MUTABLE, .MATCH, .LAZY, .IF, .FUNCTION, .FUN, .FOR, .EXCEPTION, .ELSE, .TO, .DOWNTO, .DO, .DONE, .BEGIN, .AS {
-  color: #cb4b16;
+  color: var(--orange-dim);
```
panglesd, line 999: There are modifications to the highlight color for render source code, which sometimes include "color variable" and sometimes various "semantic variable".

I think it would be nice to have only semantic variable, when applicable specific to render source code

src/html_support_files/odoc.css
EmileTrotignon: no its a typo, thanks

EmileTrotignon:
I have taken into accounts comments that say that changing the whole theme might be too big of a change, so i propose the following : 
image
image

This changes the color theme to gruvbox but only for code snippets, and in the case of the light theme, does not use the yellow gruvbox backgrounds, but keeps the old one. It should feel much more familiar, but still include the very nice refactoring I did. This refactoring is a bit impaired by the necessity to have two color palette, but it is still nicer than the previous situation.

Julow (approved):
The new colors are a lot more pleasing to me!

Thanks for only changing the code blocks as I think the gruvbox colors don't fit with documentation (contrasts are too low).

src/html_support_files/odoc.css
```diff
@@ -95,138 +95,163 @@
...
+    --aqua-dim:   #689d6a;
+    --gray-dim:   #a89984;
+    --orange-dim: #d65d0e;
+      
+    /* odoc colors */
+    --odoc-blue: #5c9cf5;
+    --odoc-bg: #202020;
+    --odoc-bg1: #252525;
+    --odoc-fg: #bebebe;
+    --odoc-fg1: #777;
```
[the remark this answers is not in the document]
EmileTrotignon: These color are the odoc blue and odoc background. --bg is the gruvbox background and --blue is the gruvbox blue.
I could have used` --gruvbox-blue` and `--blue` instead, but it seemed unpractical to me.

Julow:
Let's merge!

PR #1024, opened 2023-10-19
Changelog entry of #1019 belongs to the next release

gpetiot:
Follow-up of #1019


panglesd (approved):
Thanks for the fix. Checked, and it seems that was the only one.

Updated the discuss post as well.

PR #1025, opened 2023-10-20
Reference driver: Generate docs for the parser module

panglesd:
The parser's doc were not included anywhere since the remerge.

This PR simply includes the parser in the list of modules built by the driver.


Julow (approved):
Looks good to me :)

PR #1026, opened 2023-10-20
Update odoc-parser's constraint on dune

panglesd:
Since the `dune-project` file now has

```
(lang dune 3.7)
```

it generates a warning to build the odoc parser with a too old version of dune.


Julow (approved):
Looks good to me. In the future we might want to let Dune generates the opam files.

PR #1027, opened 2023-10-25
Handle differences in sig_class for OCaml 5.1. (to master branch)

panglesd:
Cherry pick changes from #1018.

Is everyone happy with me not picking the rename from `cmi.ml` to `cmi.cppo.ml`?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,9 @@ Tags:
 ### Added
 - Display 'private' keyword for private type extensions (@gpetiot, #1019)
 
+### Fixed
+- Fix 5.1 support (@tmcgilchrist, #1018)
+
 # 2.3.0
 
 ### Added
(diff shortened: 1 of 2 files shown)


Julow (approved):
Looks good.

> Is everyone happy with me not picking the rename from cmi.ml to cmi.cppo.ml?

Git generally handles that very well and I like that the file name indicates how it's compiled. I wouldn't mind a general change to either always use `.cppo.ml` or to never use it. (in an other PR)

panglesd:
> Git generally handles that very well and I like that the file name indicates how it's compiled. I wouldn't mind a general change to either always use .cppo.ml or to never use it. (in an other PR)

I agree! Currently there is no consistency, and I wanted to avoid an incomplete fix in some direction. But let's get consistency in the naming in another PR.

PR #967, opened 2023-06-02
Preformatted elements fallback to UA monospace

toastal:
Courier looks mighty ugly on a lot of systems. When in doubt, it’s better to just trust the user agent. This also allow users to set their own preferred option—which is better than a bad option.

PR #1028, opened 2023-10-26
Style: Remove font fallback to Helvetica, Arial

Julow:
These fonts are not common nowadays and do not look native.

Let the user agent set the fallback font in the rare cases the distributed font cannot be used (eg. missing on server or unsupported by client).


jonludlam:
This needs rebasing.

Julow:
I suggest that I rebase after https://github.com/ocaml/odoc/pull/967 is merged.

jonludlam:
Ah! I hadn't spotted that you'd just done that. Let's get that one in, yes.

Julow:
Rebased.

PR #1030, opened 2023-10-27
Add value and constructor path

panglesd:
The ability to resolve those path will be useful in many render source code new features. In particular, this is a prerequisite to #976.

### DIFF
--- a/src/xref2/compile.ml
+++ b/src/xref2/compile.ml
@@ -31,6 +31,17 @@ and value_path : Env.t -> Paths.Path.Value.t -> Paths.Path.Value.t =
       | Ok p' -> `Resolved Lang_of.(Path.resolved_value (empty ()) p')
       | Error _ -> p)
 
+and constructor_path :
+    Env.t -> Paths.Path.Constructor.t -> Paths.Path.Constructor.t =
+ fun env p ->
+  match p with
+  | `Resolved _ -> p
+  | _ -> (
+      let cp = Component.Of_Lang.(constructor_path (empty ()) p) in
+      match Tools.resolve_constructor_path env cp with
+      | Ok p' -> `Resolved Lang_of.(Path.resolved_constructor (empty ()) p')
+      | Error _ -> p)
+
 and module_type_path :
     Env.t -> Paths.Path.ModuleType.t -> Paths.Path.ModuleType.t =
  fun env p ->
--- a/src/xref2/compile.ml
+++ b/src/xref2/compile.ml
@@ -74,6 +74,11 @@ and class_type_path : Env.t -> Paths.Path.ClassType.t -> Paths.Path.ClassType.t
       | Ok p' -> `Resolved Lang_of.(Path.resolved_class_type (empty ()) p')
       | Error _ -> p)
 
+let () =
+  (* Until those are used *)
+  ignore value_path;
+  ignore constructor_path
+
 let rec unit env t =
   let open Compilation_unit in
   let source_info =
--- a/src/xref2/compile.ml
+++ b/src/xref2/compile.ml
@@ -21,6 +21,16 @@ let type_path : Env.t -> Paths.Path.Type.t -> Paths.Path.Type.t =
       | Ok p' -> `Resolved Lang_of.(Path.resolved_type (empty ()) p')
       | Error _ -> p)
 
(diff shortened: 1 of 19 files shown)


src/xref2/link.ml
```diff
@@ -143,6 +153,52 @@ let type_path : Env.t -> Paths.Path.Type.t -> Paths.Path.Type.t =
...
+        match Tools.resolve_constructor_path env cp with
+        | Ok p' ->
+            let result = Tools.reresolve_constructor env p' in
+            `Resolved Lang_of.(Path.resolved_constructor (empty ()) result)
+        | Error e ->
+            Errors.report ~what:(`Constructor_path cp) ~tools_error:e `Lookup;
+            let _ =
+              ignore e;
+              failwith "todo"
+            in
```
gpetiot, line 194: To clean up.

src/xref2/tools.ml
```diff
@@ -1110,6 +1252,148 @@ and resolve_type :
...
+      | `FType_removed _ -> (* failwith "error" *) Error `Find_failure
+      | `FType (_, t) ->
+          handle_constructor_lookup env (ConstructorName.to_string id) parent t
+      (* let result = *)
+      (*   match Find.value_in_sig parent_sig (ValueName.to_string id) with *)
+      (*   | `FValue (name, t) :: _ -> *)
+      (*       Some (`Value (parent, name), `FValue (name, Subst.value sub t)) *)
+      (*   | [] -> None *)
+      (* in *)
+      (* of_option ~error:`Find_failure result *))
```
gpetiot, line 1391: To clean up.

src/xref2/tools.ml
```diff
@@ -886,6 +972,62 @@ and lookup_type :
...
+  |> map_error (fun e -> (`ParentC e :> simple_constructor_lookup_error))
+  >>= fun t ->
+  match t with
+  | `FType (_, t) ->
+      handle_constructor_lookup env (ConstructorName.to_string name) parent t
+      >>= fun (_, x) -> Ok x
+  | `FType_removed _ -> Error `Find_failure (* failwith "Error" *)
+
+(* handle_value_lookup env (ValueName.to_string id) p sg *)
+(* >>= fun (_, `FValue (name, c)) -> Ok (`FValue (name, Subst.value sub c)) *)
```
gpetiot, line 1029: To clean up.

src/xref2/compile.ml
```diff
@@ -53,6 +74,11 @@ and class_type_path : Env.t -> Paths.Path.ClassType.t -> Paths.Path.ClassType.t
       | Ok p' -> `Resolved Lang_of.(Path.resolved_class_type (empty ()) p')
       | Error _ -> p)
 
+let () =
+  (* Until those are used *)
+  ignore value_path;
+  ignore constructor_path
```
gpetiot, line 80: Would it break things to modify `source_info_infos` like in #976, or would it allow to check more things in the tests?

src/xref2/link.ml
```diff
@@ -143,6 +153,52 @@ let type_path : Env.t -> Paths.Path.Type.t -> Paths.Path.Type.t =
...
+            p)
+
+let () =
+  (* Until those are used *)
+  ignore value_path;
+  ignore constructor_path
```
gpetiot, line 200: Same question here, would it break things to update the `source_info` in `unit` like in #976?

gpetiot (approved):
I left a few questions (and some things should be cleaned up) but otherwise it looks good to me!

src/xref2/link.ml
panglesd: Woops, slipped into my staging area... thanks!

src/xref2/compile.ml
panglesd: I'm not sure about the question! Initially, #976 was a single PR, but I turned it into three PRs:
- one for the path resolution (this one)
- one for the refactoring #1031
- #976 which will be rebased on top of master when the two other one have been merged.

So no, it won't break things to have `source_infos` as in #976, on the contrary that is the goal!

gpetiot: Go it, thanks!

jonludlam:
I'm not very convinced with all the datatype stuff; I think we've not got the best set of types here. Having said that, we're already in this situation and this PR won't make it more tricky to rip it out, and there are other PRs that depend upon this. Hence, let's figure this out later.

PR #1031, opened 2023-10-27
Refactor typedtree traverse to use compiler traverse

panglesd:
This is hopefully less maintenance, increased compatibility, shorter code, and the guaranty that we don't miss any occurrence of some deeply nested core type (for instance) for source features such as #976.


Julow (approved):
Looks good! The code indeed looks much more maintainable.

I couldn't find new links but I noticed that this now generates more anchors (for example in structures passed as argument to functors).

panglesd:
> I noticed that this now generates more anchors

Could you give me an example? In particular, is the new anchor a "local anchor", or a "global anchor"?
I expect it to be a local anchors. The global anchors should be the same, but it is very possible that the old manual traverse was missing local anchors.

Julow:
By diffing the output of `@docgen`, I only see changes in local anchors. For example, in `src/document/ML.ml` there's new local anchors for `close_tag_closed`, etc..
Here's the observed diff: https://gist.github.com/Julow/230e1e240453137f8cbcc5072e75c47a

panglesd:
Thanks. Indeed, two anchors are generated... let me investigate!

panglesd:
The post-processing of source info is now more careful to not add two definition at the same place. In case of "conflicts", we use the "local anchor", which is more readable, rather than the `def_%d` anchor.

src/loader/typedtree_traverse.ml
```diff
@@ -0,0 +1,66 @@
+#if OCAML_VERSION >= (4, 14, 0)
```
gpetiot, line 1: This can be done in a later PR, but why not using dune rules like this instead:
```lisp
(rule
  (enabled_if (< %{ocaml_version} 4.0.5))
  (action (copy file.404.ml file.ml)))

(rule
  (enabled_if (>= %{ocaml_version} 4.0.5))
  (action (copy file.405.ml file.ml)))
```

so we don't need the cppo syntax and we can apply ocamlformat on this file (and implementation.ml)

gpetiot (approved):
It looks good to me!

src/loader/typedtree_traverse.ml
Julow: I like the idea of a code without cppo though I think this wouldn't play well with Merlin. OCamlformat could perhaps support some form of cppo directives ? (eg. as structure items)

src/loader/implementation.ml
```diff
@@ -4,367 +4,182 @@
...
-
-  and exception_ _env _parent acc _e = acc
-
-  and module_declaration env _parent acc md =
+    | Tsig_module mb -> module_declaration env parent mb
+    | Tsig_recmodule mbs -> module_declarations env parent mbs
+    | Tsig_modtype mtd -> module_type_declaration env parent mtd
+    | Tsig_modtypesubst mtd -> module_type_declaration env parent mtd
+    | Tsig_value _ | Tsig_type _ | Tsig_typesubst _ | Tsig_typext _
+    | Tsig_exception _ | Tsig_modsubst _ | Tsig_open _ | Tsig_include _
```
Julow, line 32: Why is `Tsig_modsubst` not iterated over ?

src/loader/implementation.ml
```diff
@@ -373,7 +188,7 @@ let postprocess_poses source_id poses uid_to_id uid_to_loc =
       (fun uid id acc ->
         let loc_opt = Shape.Uid.Tbl.find_opt uid_to_loc uid in
         match loc_opt with
-        | Some loc ->
+        | Some loc when not (LocHashtbl.mem loc_to_id loc) ->
```
Julow, line 191: Is this to prevent generating overlapping anchors for the same loc ? A comment would be nice.

Julow (approved):
Looks good :)
I love the new `local_...` anchors! Though, the diff is now too large to be inspected, I believe the code to be correct.

src/loader/implementation.ml
panglesd: That's a good question. I'm uncomfortable writing all those traverse by hand, since it's easy to miss something, particularly on the language constructs I don't know so well. I'll check and add the traverse if needed!

panglesd: It is not iterated because there is nothing to iterate inside!
```
and module_substitution =
    {
     ms_id: Ident.t;
     ms_name: string loc;
     ms_manifest: Path.t;
     ms_txt: Longident.t loc;
     ms_attributes: attributes;
     ms_loc: Location.t;
    }
```
In the parsetree, it is used to represent
```
  | Psig_modsubst of module_substitution  (** [module X := M] *)
```
and I assume it represents the same in the typedtree.

Julow: I see! It must be added to the env but that's already done in `Ident_env`, so there's no problem. Sorry for the false alarm.

panglesd:
While trying to add a comment as suggested in https://github.com/ocaml/odoc/pull/1031#discussion_r1376513646, I realized that the code was not clear enough...

In 48f8f06 I tried to improve that. @Julow is it clearer like that?

jonludlam:
> The post-processing of source info is now more careful to not add two definition at the same place. In case of "conflicts", we use the "local anchor", which is more readable, rather than the `def_%d` anchor.

Just want to check that that's not going to lead to broken links? As in, if something is linking here and we've dropped the `def_...` anchor in favour of the local one, won't that break the link?

panglesd:
> Just want to check that that's not going to lead to broken links? As in, if something is linking here and we've dropped the def_... anchor in favour of the local one, won't that break the link?

No, that's not going to lead to any dead link. All source ids are stored in a `loc to id` table, from which we generate both the links and the anchors.

jonludlam:
In it goes!

PR #1032, opened 2023-10-27
Update tyxml constraint

panglesd:
Since table support, odoc is incompatible with tyxml 4.3.0.

See https://github.com/ocaml/opam-repository/pull/24686.


Julow (approved):
Let's merge!

PR #1033, opened 2023-10-31
Define resolve_parent to memoize calls to prefix_substitution that were not (Dot cases)

gpetiot:


Julow (commented):
This has a very large impact on the slowest module to compile (`Html_sigs` in tyxml), which contains a large number of types.

`src/xref2/test.md` is failing and I don't understand why at the moment (`dune build @runmdx`). `t` no longer resolves in this example:
```
module A : sig
  module M : sig type t end
  module N = M
  module O = N
end

type t = A.O.t
```

gpetiot:
I've defined `resolve_and_lookup_parent` that calls `resolve_parent` (a glorified call to resolve_module, but the return type is not the same so it looks worth it to me to memoize it separately) and `lookup_parent` (pre-existing), and apparently `dune build @runmdx` now runs.

src/xref2/tools.ml
```diff
@@ -915,6 +915,21 @@ and lookup_class_type :
...
+    Env.t ->
+    Cpath.module_ ->
+    ( Cpath.Resolved.parent * Component.Signature.t * Component.Substitution.t,
+      [ `Parent of parent_lookup_error ] )
+    result =
+ fun ~mark_substituted ~add_canonical env parent ->
+  resolve_module ~mark_substituted ~add_canonical env parent
+  |> map_error (fun e -> `Parent (`Parent_module e))
+  >>= fun (parent, _) ->
+  lookup_parent ~mark_substituted env (`Module parent)
```
Julow, line 930: The double-lookup seems weird at first but I agree it's a good idea to take advantage of the existing caches. This seems to retain the same performances as before with less code.

A small comment would be nice.

Julow (commented):
This shouldn't be a draft anymore!

Can you rebase on top of https://github.com/ocaml/odoc/pull/1034 to see the result on bigger libraries?

src/xref2/tools.ml
gpetiot: Is there a double lookup? The code I removed looks identical to the code of lookup_parent (`Module case), except it uses `expansion_of_module` instead of `expansion_of_module_cached`, but `lookup_parent` is already memoized.

Julow: `resolve_module` does a first lookup (memoized), giving a resolved path and a signature. The signature is discarded and `lookup_parent` does a second (memoized) lookup, giving the same signature and the computed substitutions.

I think this is acceptable as both lookups are memoized and we do need the resolved path from the first lookup. The unnecessary work is very small and the avoided code duplication will help us do the changes we planned to the `lookup_parent` function.

Julow (approved):
This seems to be a general improvement to the correctness and performances. The double lookup reduces the amount of code and the number of caches, which seems to be the right direction.
I'd rather merge this before more work can be done on this part of the code.

Julow (approved):
There have been concerns about the double lookup but this PR is already an improvement. More improvements should be made on top and should not block this.
Let's merge before it rots.

PR #1034, opened 2023-10-31
Improve benchmarks

Julow:
- The metrics added in https://github.com/ocaml/odoc/pull/1012 are not useful as it repeats an other existing metric. Instead, the 5 slowest commands are considered (roughly the top 1%).
- Build the documentation for `core` in the benchmarks, which contains modules on which Odoc is very slow.
- Fix driver error not reported since https://github.com/ocaml/odoc/pull/978


Makefile
```diff
@@ -15,7 +15,7 @@ test :
 
 .PHONY : bench
 bench:
-	@dune build @bench
+	@ODOC_BENCHMARK=true dune build @bench
```
panglesd, line 18: I wonder if a comment here would make sense. The meaning of this variable is not documented anywhere, you have to look at the `driver.mld` code.

doc/driver.mld
```diff
@@ -300,14 +300,7 @@ let dep_libraries_core = [
...
 let extra_deps = [
-    "base";
-    "core_kernel";
-    "bin_prot";
-    "sexplib";
-    "sexplib0";
-    "base_quickcheck";
-    "ppx_sexp_conv";
-    "ppx_hash";
+    "core";
```
panglesd, line 303: How were the previous "extra deps" chosen? If the reason was valid, why not keeping them, adding "core"?

doc/driver.mld
```diff
@@ -816,99 +813,104 @@ This last block analyze the running times so that they can be submitted to
...
+              ("min", `Int min); ("max", `Int max); ("avg", `Int avg);
+            ] );
+        ("units", `String "b");
+        ( "description",
+          `String ("Size of file produced by 'odoc " ^ cmd ^ "'") );
+        ("trend", `String "lower-is-better");
+      ];
+  ]
 
 (** Analyze the size of files produced by a command. *)
```
panglesd, line 890: This comment looks like is the result of a (previous) copy-paste!.

doc/driver.mld
Julow: I think the intent of the previous list was to have a world that is as large as core but that is not core, because it failed to build at some point. I might be wrong.

doc/driver.mld
Julow: Nice catch!

Makefile
Julow: Good idea, done.

doc/driver.mld
jonludlam: No, it was that core had some dependencies and some were vaguely interesting, so might as well build them too. Particularly anything that core/core_kernel might have linked to (so sexplib, bin_prot, etc).

Julow: Alright, I added them back. Though I did not add more as that would make the list very large without adding interesting test cases.

jonludlam: My thinking was mainly that certain modules take vastly longer than others, so including a new library either means it's very quick (and not interesting) or that some modules now take ages, which is very interesting, and when you fix it, becomes a useful test case :-)

jonludlam:
OK, green tick. In it goes.

PR #1036, opened 2023-11-10
Don't compute the unused list removed_items in Tools.fragmap

gpetiot:
A small optimization + cleanup found while working on the sig optimization.


Julow (approved):
Did you measure any performance difference ?

Otherwise, the code looks good and is cleaner.

gpetiot:
> Did you measure any performance difference ?

ocaml-benchmarks doesn't show any big difference.

Rebasing this branch on `landmarks` shows the same: running compile on html_sigs stays between 13.5-13.7

PR #1040, opened 2023-11-16
Fix and show benchmark errors

Julow:
The benchmarks are failing due to a missing dependency. I changed the benchmark script to avoid hidding errors.


jonludlam:
Thanks!

PR #1041, opened 2023-11-17
Add instrumentation with landmarks

Julow:
Also, activates bisect_ppx on some modules where it was missing.

To run the driver with instrumentation,

- Set `let instrument = true` in doc/driver.mld's prelude.
- Use `dune build --instrument-with landmarks @docgen`.

The results are stored in _build/default/doc/landmarks/ and can be viewed with landmarks' web viewer.


jonludlam:
Nice, thanks!

PR #1042, opened 2023-11-21
Add a test for the size of `module type of` expressions

jonludlam:


test/xref2/module_type_of_size.t/gen.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
Julow, line 1: I guess this file wasn't meant to be there ? Otherwise, `compile.sh` and `odoc.sh` shouldn't be included.

test/xref2/module_type_of_size.t/compile.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
```
Julow, line 1: That's not a portable `#!` line. I suggest not having one and to run this script using `bash ./compile.sh`.

Julow (commented):
I have some small comments but otherwise looks good.

test/xref2/module_type_of_size.t/gen.sh
Julow: After all, I find `gen.sh` more readable than the full test.
It could even be inlined in `run.t`, generating `compile.sh` and `odoc.sh` wouldn't be necessary as commands could be called directly in the loop.

jonludlam: It's there in case we ever wanted to increase the size of the test, or modify the contents of the mli files. I wrote it to do this test so I thought I might as well include it :-)

test/xref2/module_type_of_size.t/compile.sh
jonludlam: Huh? I don't understand - I think this script is pretty portable. The `gen.sh` needs to be changed to `#!/bin/bash` though!

jonludlam: it doesn't use /bin/bash, it's /bin/sh - does that not work on your machine?

jonludlam:
Closing in favour of #1079

PR #1044, opened 2023-11-24
Remove odoc_html_support_files.ml

EmileTrotignon:
This PR removes the need to promote odoc_html_support_files.ml from the source tree.

It is still present in the build.

I am aware that I might have missed something that make this a bad idea, but I wanted to open that discussion.


src/html_support_files/dune
```diff
@@ -5,34 +5,32 @@
  (aliases runtest support-files)
  (enabled_if
   (> %{ocaml_version} 4.08))
+ (target odoc_html_support_files.ml)
```
Julow, line 8: odoc.opam is not uptodate as `crunch` is still listed `with-test`. It would be nice to test various versions of crunch and find an eventual lower bound.

Julow (approved):
The reason for this was to avoid having `ocaml-crunch` as a build-dependency instead of a test-dependency. 

I'm in favor of this change as it adds few dependencies and remove a often conflicting file.

Julow:
This requires a change entry as well as an approval from more than one person.

Julow:
@jonludlam As you implemented the ocaml-crunch rule, what do you think of merging this ?

panglesd:
I'm agnostic in this debate (which has... a conflict with `odoc_html_support_files.ml`!)

Counting the recursive one, this PR adds the following dependencies:
- ocaml-crunch
- ptime

which I think is reasonable, although the benefit is not essential. It can easily be revert if needed.

jonludlam:
Since we're already having to do some work on release branches (ie, substituting in the version in the crunched files and then updating the generated ml file), we could just as well do the crunching and remove the dependency, if we felt it that important.

PR #1045, opened 2023-11-26
css: adjust the line height to improve readability

sorawee:
When a heading is long, it will be word-wrapped. The current value of line-height is very high, causing lines that are word-wrapped to appear as new items.

The fix is to reduce the line-height, and instead use margin as a way to distance items apart.

Before:

Screenshot 2023-11-26 at 3 10 40 PM

After:

Screenshot 2023-11-26 at 3 11 13 PM


src/html_support_files/odoc.css
```diff
@@ -1077,7 +1077,7 @@ td.def-doc *:first-child {
 }
 
 .odoc-toc ul li {
-  margin: 0;
+  margin: 8px 0;
```
Julow, line 1080: This creates gaps in the line formed by each item's left border. Changing the `padding` instead would be better.

Julow (commented):
> The contribution guide probably should be updated. It said that CSS change should be self-contained.

Good catch!

src/html_support_files/odoc.css
sorawee: Thanks! It does look better that way.

sorawee:
An updated before-after screenshots:

Screenshot 2023-11-28 at 4 18 22 AM

src/html_support_files/odoc.css
```diff
@@ -1070,14 +1070,15 @@ td.def-doc *:first-child {
...
+  padding-top: 4px;
 }
 
 .odoc-toc ul {
   list-style-type: none;
 }
 
 .odoc-toc ul li {
-  margin: 0;
+  padding: 4px 0;
```
Julow, line 1081: Shouldn't this be measured in `em` ? Perhaps `0.2em` as this is half of the removed amount.

Julow (commented):
This is a after-before screenshot, right?

sorawee:
Fix pushed!

Julow (approved):
This is a nice improvement, thanks!

Julow:
I think a changelog entry would be useful.

sorawee:
Ah, I thought the change was too minor to warrant a changelog entry. But yes, I added the entry.

Thanks again for the review!

Julow:
Thanks!

PR #1046, opened 2023-11-27
Fix crash on functor returning an alias

Julow:
Fix https://github.com/ocaml/odoc/issues/905

The following structure is not given a type and hits an 'assert false':

    module Foo (X : sig
      val foo : int
    end) =
      String

This gives it the type:

    module Foo (X : sig ... end) : module type of Stdlib.String

### DIFF
--- a/src/loader/cmi.ml
+++ b/src/loader/cmi.ml
@@ -938,7 +938,9 @@ let rec read_module_type env parent (mty : Odoc_model.Compat.module_type) =
         in
         let res = read_module_type env (Identifier.Mk.result parent) res in
         Functor( f_parameter, res)
-    | Mty_alias _ -> assert false
+    | Mty_alias p ->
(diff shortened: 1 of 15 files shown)


EmileTrotignon:
This looks good to me

EmileTrotignon (approved):
Lets fix the ci and then get this merged

panglesd (commented):
At first sight, it seems that there is also something to change in https://github.com/ocaml/odoc/blob/master/src/loader/cmti.ml#L569

Is it the case? In any case, it does not seem to be checked by the test, which only load a `cmt`...

Julow:
I haven't found a way to hit this case yet. If you do, I'd like that to be fixed in a second PR.

panglesd (approved):
Looks good! A changelog entry and let's merge!

PR #1049, opened 2023-11-29
Use a map to memoize looked-up items in signatures

gpetiot:
Opening this PR for ocaml-benchmarks. Follow-up of #1039


Julow:
We'll close this prototype due to a lack of time.

Issue #1043, opened 2023-11-24
Table with missing closing bracket hangs


art-w:
When using the new table syntax, incomplete markup appears to hang rather than signal a syntax error:

```ocaml
(** {t missing the closing bracket *)

(** also happens with {table with no closing bracket *)
```

According to `dune build --verbose @doc`, it seems to get stuck at this step: (and consummes 100% cpu with increasing memory?)

```
$ odoc compile -I . -I ../../_doc/_odoc/pkg/test --pkg test -o test.odoc test.cmti
```

In comparison, some of the other markup like `{1` or `{ul` report some weird warnings which are not hinting at the missing `}`, but they do not hang:

```ocaml
(** {ul {- foo} *)
```
```
File "test.mli", line 1, characters 16-17:
Warning: End of text is not allowed in '{ul ...}' (bulleted list).
```

(tested with odoc 2.3.1 and master)

Julow:
Thanks to @lubegasimon for solving this issue :)

PR #1050, opened 2023-11-30
Warn when a table is not closed

lubegasimon:
Fixes https://github.com/ocaml/odoc/issues/1043

### DIFF
--- a/src/parser/parse_error.ml
+++ b/src/parser/parse_error.ml
@@ -82,7 +82,7 @@ let truncated_code_block_meta : Loc.span -> Warning.t =
 let truncated_code_block : Loc.span -> Warning.t =
   Warning.make ~suggestion:"add ']}'." "Missing end of code block."
 
-let end_not_allowed :
-    ?suggestion:string -> in_what:string -> Loc.span -> Warning.t =
- fun ?suggestion ~in_what ->
-  Warning.make ?suggestion "End of text is not allowed in %s." in_what
+let end_not_allowed : in_what:string -> Loc.span -> Warning.t =
+ fun ~in_what ->
+  Warning.make ~suggestion:"add '}'." "End of text is not allowed in %s."
+    in_what
--- a/src/parser/parse_error.ml
+++ b/src/parser/parse_error.ml
@@ -81,3 +81,7 @@ let truncated_code_block_meta : Loc.span -> Warning.t =
 
 let truncated_code_block : Loc.span -> Warning.t =
   Warning.make ~suggestion:"add ']}'." "Missing end of code block."
+
+let unclosed_table : Loc.span -> Warning.t =
+  Warning.make ~suggestion:"try to add '}' at the end of table content."
+    "Unclosed table '{t ...' or '{table ...'"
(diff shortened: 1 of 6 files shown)


src/parser/syntax.ml
```diff
@@ -118,13 +118,17 @@ module Table = struct
...
-  let until_rbrace input acc =
+  let until_rbrace_or_eof input acc =
     let rec consume () =
       let next_token = peek input in
       match next_token.value with
       | `Right_brace ->
           junk input;
           `End (acc, next_token.location)
+      | `End ->
+          Parse_error.unclosed_table next_token.location |> add_warning input;
```
Julow, line 129: This function seems to only be used when parsing tables but its name doesn't say that it's specific to tables. Could the error message be changed not to be specific to tables ? Perhaps by taking one more argument here.

Julow (commented):
Looks good :) Why is this a draft ?

src/parser/syntax.ml
lubegasimon: Indeed, thanks

lubegasimon:
> Looks good :) Why is this a draft ?

There is another part referenced in the same issue.
> In comparison, some of the other markup like {1 or {ul report some weird warnings which are not hinting at the missing }, but they do not hang:
...

src/parser/syntax.ml
```diff
@@ -1277,6 +1284,7 @@ and explicit_list_items :
         | `End ->
             Parse_error.not_allowed token_after_list_item.location
               ~what:(Token.describe `End) ~in_what:(Token.describe token)
+              ~suggestion:"try to add '}' at the end of the list content."
```
Julow, line 1287: This brings some code duplication. It would look better if a new error was defined in Parse_error, with the suggestion, eg. `Parse_error.end_not_allowed`. The suggestion string could be changed not to be specific to lists.

Julow (approved):
Two in one :) I have a small comment but otherwise looks good.

src/parser/syntax.ml
```diff
@@ -1227,8 +1236,10 @@ and explicit_list_items :
     let next_token = peek input in
     match next_token.value with
     | `End ->
-        Parse_error.not_allowed next_token.location ~what:(Token.describe `End)
+        Parse_error.end_not_allowed next_token.location
+          ~what:(Token.describe `End)
           ~in_what:(Token.describe parent_markup)
+          ~suggestion:"try to add '}' at the end of the list content."
```
Julow, line 1242: I hoped that we wouldn't need to pass `~what` and `~suggestion`.

lubegasimon: I hoped to use the same error function -- `end_not_allowed` for both unclosed table and list, I thought it would be important to specify the markup that is not closed (which is what the `~what` is for). But I now see that it's not important, I will remove that.

lubegasimon: I think `~what` is important

src/parser/parse_error.ml
```diff
@@ -81,3 +81,8 @@ let truncated_code_block_meta : Loc.span -> Warning.t =
 
 let truncated_code_block : Loc.span -> Warning.t =
   Warning.make ~suggestion:"add ']}'." "Missing end of code block."
+
+let end_not_allowed : in_what:string -> Loc.span -> Warning.t =
+ fun ~in_what ->
+  Warning.make ~suggestion:"add '}'." "End of text is not allowed in %s."
```
Julow, line 87: I've found a way to avoid the redundant `~suggestion` argument. What do you think ?

Julow (approved):
Unless you disagree with my change, I'll merge this soon.

panglesd (approved):
Looks good to me. Thanks @lubegasimon !

PR #1051, opened 2023-12-05
fixes anchor appearing bellow search bar

EmileTrotignon:
It used to be that the when following a link to an anchor, the anchor would appear a the top of the view, which was under the new search bar. This adds a margin of a 100px to such jumps.

### DIFF
--- a/src/html_support_files/odoc.css
+++ b/src/html_support_files/odoc.css
@@ -95,7 +95,10 @@
 
 :root,
 .light:root {
-   --main-background: #FFFFFF;
+
+  scroll-padding-top: 100px;
+
+  --main-background: #FFFFFF;
 
   --color: #333333;
   --link-color: #2C94BD;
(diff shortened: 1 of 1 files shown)


src/html_support_files/odoc.css
```diff
@@ -95,7 +95,10 @@
 
 :root,
 .light:root {
-   --main-background: #FFFFFF;
+
+  scroll-padding-top: 100px;
```
Julow, line 99: Shouldn't this be in `em`, like the font size and the margin of items ?

Julow (commented):
Needs a rebase.

src/html_support_files/odoc.css
panglesd: More than a specific unit, shouldn't this be computed from `--search-bar-height` and `--search-padding-top` to be sure it is enough? Something like:

```suggestion
  scroll-padding-top: calc(var(--search-bar-height) + var(--search-padding-top) + 20px);
```
Where `20px` can be replaced by a size in `em` if needed.

panglesd (approved):

PR #1053, opened 2023-12-06
Benchmark: Add all core dependencies

jonludlam:
This commit adjusts the set of libraries used as a benchmark. It adds approximately the minimal set of libraries to ensure a self-consistent set of odocs without unresolved module dependencies.

I found this particularly useful when testing some recent changes - changes that showed no difference in output when run on the 'normal' set of packages, but a large number of differences when run on the extended list.


doc/library_mlds/base_bigstring.mld
```diff
@@ -0,0 +1,2 @@
+{0 Base_bigstring}
```
Julow, line 1: These pages are too hard to create and delete. The driver should generate them instead.

doc/library_mlds/ppx_expect.make_corrected_file.mld
```diff
@@ -0,0 +1 @@
+{0 ppx_expect.make_corrected_file}
```
panglesd, line 1: Is this file added on purpose?

jonludlam: Nope, thanks!

jonludlam: Fixed.

Julow (approved):
More unnecessary pages can be removed: `base`, `base_quickcheck`, `core`, `ppx_hash`, etc.. as they are not expected in a build without `extra_deps`.

Otherwise, looks good.

PR #1056, opened 2023-12-07
tells if a search entry is child of a module type

EmileTrotignon:
Adds a function that tells if an entry is the child of a module type
this is useful because search entry from module types might be given a different priority in search results


jonludlam:
From odoc dev meeting - we'll close this and put the implementation into sherlodoc directly.

PR #1059, opened 2023-12-11
html: Output meta charset before any stylesheet

Julow:
This is a dirty fix for a problem affecting the release.

The generator meta element has a variable length due to the `%%VERSION%%` watermarking and that makes the test fail after a `dune subst`.
The generator meta and the charset meta elements are sometimes formatted on the same line, depending on the length of the version string. This do not happen anymore.

The first commit is refactoring.

### DIFF
--- a/src/html/html_page.ml
+++ b/src/html/html_page.ml
@@ -93,8 +93,8 @@ let default_meta_elements ~config ~url =
   let theme_uri = Config.theme_uri config in
   let odoc_css_uri = file_uri ~config ~url theme_uri "odoc.css" in
   [
-    Html.link ~rel:[ `Stylesheet ] ~href:odoc_css_uri ();
     Html.meta ~a:[ Html.a_charset "utf-8" ] ();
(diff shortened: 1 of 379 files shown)


panglesd (approved):
The refactoring commit LGTM.

The fix hack looks ok to me. This order works "by chance" and there might be future modifications which reintroduce such version-string-unstability in the tests.
However, at least the fix does not harm anything in the current code base.

(The other fix I could think about either are overly complex for the issue, or degrade something (for instance, removing the formatting in the generated html degrades reading the diff))

Julow:
Perhaps a CLI flag that disables the `generator` meta could help. It was too much for before the release but now we can think of a cleaner solution.

Julow:
I'll merge this as well, to reduce the discrepancy between the 2.4 branch and master. The second commit should be reverted if we hit this issue again or find a better solution.

Julow:
This creates a issue on `<= 4.08`: https://github.com/ocaml/odoc/actions/runs/7180556228/job/19553083098

PR #1060, opened 2023-12-11
Compatibility with 4.02

Julow:


panglesd:
How much do we want to maintain compatibility with older versions, now that `ocaml-ci` does check only on `4.14` and `5.1`?

`odoc` is only used to build the doc, dropping compatibility shouldn't completely break build of a package.

I think that without "full CI coverage" of supported versions, it will be very hard to maintain full compatibility.

Julow:
I'm in favor of dropping compatibility with OCaml `< 4.14` but that wasn't planned for this release and is not an option for now.

You are right that we need full coverage as almost every versions occur in `enabled_if` dune fields and in cppo conditions.
To limit the waste of power, I included every pair versions. Do you think that's a good idea ?

panglesd:
> Do you think that's a good idea ?

That's fine to me! I only remember compatibility issues on ranges of versions of length at least two.

panglesd (approved):
CI is green, let's merge and (hopefully) undo most of it for next release!

PR #1061, opened 2023-12-12
Update lower bounds and fix tests on < 4.10 and macos

Julow:
This was found by Opam's CI.
For an unknown reason, it was not caught by https://github.com/ocaml/odoc/actions/runs/7180162543/job/19551836931


Julow:
The third commit removes an unstable output, again found by opam-ci.

Julow:
The fourth commit fixes the way the tests are run by Opam for `odoc-parser` in a hacky way. The best would be to associate each test with the right package.

PR #1062, opened 2023-12-12
Don't make parser tests installable

Julow:
This is broken since https://github.com/ocaml/odoc/pull/1061 as I focused on tests and forgot to try building without.
Related dune issue: https://github.com/ocaml/dune/issues/4738


Julow:
This unlocked the release and can be merged.

PR #1064, opened 2023-12-18
doc: Use Sherlodoc as search engine

Julow:
This is a WIP driver support for the Sherlodoc search engine. It uses this version: https://github.com/art-w/sherlodoc/pull/4

This changes driver.mld to use sherlodoc. The database is 11MB large.

Sherlodoc's sources must be cloned into the dune workspace in order for it to use the same version of Odoc as the driver. Installing it via Opam results in magic number clashes.

### DIFF
--- a/doc/driver.mld
+++ b/doc/driver.mld
@@ -278,7 +278,7 @@ let html_generate ?(ignore_output = false) ?(assets = []) ?(search_uris = []) fi
   in
   let search_uris =
     List.fold_left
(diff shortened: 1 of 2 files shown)


Julow:
This is live in the v2.4 branch: https://ocaml.github.io/odoc/

panglesd:
> The database is 11MB large.

Small remark: only 8.33MB are sent through the wire, due to compression.

Julow:
Superseded by https://github.com/ocaml/odoc/pull/1086

PR #1065, opened 2023-12-18
doc: Add the generate page 'API Reference'

Julow:
This page lists the internal libraries and is generated by the driver. It's live in the v2.4 branch: https://ocaml.github.io/odoc/api_reference.html

### DIFF
--- a/doc/driver.mld
+++ b/doc/driver.mld
@@ -354,10 +354,10 @@ let extra_docs = [
     "driver";
     "parent_child_spec";
     "features";
-    "interface";
     "odoc_for_authors";
     "dune";
     "ocamldoc_differences";
+    "api_reference";
 ]
 
 let parents =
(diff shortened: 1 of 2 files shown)


panglesd (approved):
Since it's already live, let's merge it! It's clearly an improvement.

Now we have a combination of a generated page without any explanations (the `api_reference` page) and 38 handwritten pages without any explanations (the `library_mlds/*/mld` pages.
It could be nice to handwrite some of them to add general explanations, and let the other be auto-generated.

Julow:
Are the pages in `library_mlds` linked to ? Getting the data to put in these pages is not easy and I'd prefer not to make the driver more complex and simply drop the pages.

Julow:
I'll merge this so it doesn't get in the way of future improvements.

PR #1067, opened 2023-12-21
Separate implementation and interface in different (odoc) compilation units

panglesd:
This PR separate the compilation pipeline for interface and implementation. This is a backward incompatible change, but only for rendering source code. It introduces a new command, `compile-src`, which compiles to a new kind of odoc file: implementation files, named `src-<name>.odoc(l)`.

A good way to see the changes is to look at the diff in the source rendering tests, but here is a list of the changes:

- The `compile` command do not take the `--source-name` and `--source-parent-file` commands.
- The new `compile-src` command take `--source-path` and `--source-parent-file` command instead. It can only take a `<name>.cmt` file as input, and creates a value of type `Odoc_model.Lang.Implementation.t` which is stored in `src-<name>.odoc`. Basically, such type is a record containing an identifier, the source information extracted from the typedtree, the implementation shape and the `uid_to_id` map.
- When looking for a shape, we look in the `src-....odoc` files in the path.
- When html-generating, the `--source` and `--source-root` are restricted to implementation `odocl` files.

A notable change is that `--source-parent-file` and `--source-path` is required for implementation files, which forces to have them even when we only want to count occurrences, not render them.

Another change is that `--source` and `--source-root` are not anymore specific to the html rendering process. The reason is that with the current organization of the code, only the renderer can depend on backend-specific flags, not the transformation model -> document. Of course it is possible to change that, but that require further modifications which I'm not sure we want.

The reason for this change is the following:
- It separates two independent pipeline that were merged into one,
- It avoids having a single `.odocl` generate files in very different output directories, which can be a problem for some build systems.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@
 - Improve jump to implementation in rendered source code, and add a
   `count-occurrences` flag and command to count occurrences of every identifiers
   (@panglesd, #976)
+- Separate compilation of interface and implementation files, using a new
+  `compile-src` command (@panglesd, #1067).
 
 ### Changed
 
--- a/doc/driver.mld
+++ b/doc/driver.mld
@@ -132,25 +132,25 @@ In this section [odoc] is used to generate the documentation of [odoc] and some 
 Let's start with some functions to execute the three phases of [odoc].
 
 Compiling a file with [odoc] requires a few arguments: the file to compile, an
-optional parent, a list of include paths, a list of children for [.mld] files,
-optional parent and name for source implementation, and an output path. Include
-paths can be just ['.'], and we can calculate the output file from the input
-because all of the files are going into the same directory. If we wish to count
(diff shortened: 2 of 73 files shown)


panglesd:
Note: the driver needs to be rewritten to use the new pipeline...

panglesd:
This is ready for review.

One notable change that can be discussed: the `--count-occurrences` (which was used for collecting occurrences information without the will to render source code) command has disappeared.

Instead, we just `compile-src` the implementation, give it to `odoc count-occurrences`, but do not call `odoc html-generate` on it.
However, this means that we still have to give a source parent, even when we do not intend to render it. I would make the source parent optional, and forbid the `html-generate`ion of implementation without source parent. What do you think?

Julow (approved):
I did a full review and I think this is ready to merge!

panglesd:
I rebased and marked the commands as experimental, as they are likely to change in the future (likely before we release)!

Issue #1013, opened 2023-10-06
odoc 2.3.0 gives "Rule failed to generate the following targets"


talex5:
After upgrading to odoc 2.3.0, I get this error building Eio:
```
$ dune build @doc
File "_doc/_html/eio/Eio__core/_unknown_", line 1, characters 0-0:
Error: Rule failed to generate the following targets:
- _doc/_html/eio/Eio__core/index.html
```

Any idea what could cause that?

The problem also appears in CI (e.g. https://ocaml.ci.dev/github/ocaml-multicore/eio/commit/4db6533fa661ef348ae3197cc6c465622ae02703/variant/%28lint-doc%29#L429-432)
```
/src: (run (shell "ODOC_WARN_ERROR=false opam exec -- dune build @doc || (echo \"dune build @doc failed\"; exit 2)"))
File "_doc/_html/eio/Eio__core/_unknown_", line 1, characters 0-0:
Error: Rule failed to generate the following targets:
- _doc/_html/eio/Eio__core/index.html
```

jonludlam:
hum interesting, why was dune trying to build `Eio__core/index.html` - that's a hidden module :-(

jonludlam:
Ah, because you've got an explicit module called `Eio__core`. Hum. Why do you need the double underscore? Odoc treats that specially to mean the module should be hidden.

jonludlam:
Oh, I see - it _should_ be hidden - the top of `eio_core.mli`:

```ocaml
(** Private internal module. Use {!Eio} instead. *)

(** @canonical Eio.Switch *)
module Switch : sig
  (** Many resources in Eio (such as fibers and file handles) require a switch to
```

talex5:
Yep. I'm happy for it to be hidden in the docs. Its contents get included at the start of `eio.ml`, so people don't refer to it directly. Previously, odoc included it in the docs (e.g. https://ocaml-multicore.github.io/eio/eio/index.html), but it would be better to hide it.

panglesd:
Small recap of this issue:

- Previously, odoc would generate an empty html file for hidden compilation units
- Since #909 (more precisely [this modification](https://github.com/ocaml/odoc/pull/909/files?diff=split&w=0#diff-0c6891cf6f87c0b10cf953a6538f37801282ae8b2077133bab2b0962f086e649R9-R12)), odoc no longer does that. The idea was that `odoc html-generate` needs to be called on hidden modules to generate the source code rendering, and we don't want all those empty html files.
[...]

PR #1069, opened 2024-01-11
Revert to outputing a content-empty file when rendering a hidden module

panglesd:
Hidden modules stopped creating an output file with "render source code", since calling `html-generate` on them was used to generate the associated source code.

However, this breaks dune rules in some cases (see issue #1013), as dune always expect an output to the execution of the command. See #1013 for a case where it happens.

With the separation of implementation and interface pipelines (see #1067), we will stop generating those content-empty files when rendering source code.

I think this fix should be included as a hotfix to the `2.3` and `2.4` branches, to maintain compatibility with dune rules (and ensure `eio` can build its doc).

@talex5 could you confirm that this patch fixes your issue?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,12 @@
   `count-occurrences` flag and command to count occurrences of every identifiers
   (@panglesd, #976)
 
+# Fixed
+
+- Revert to outputing a file (without content) when rendering a hidden
+  compilation unit. This fixes cases where the dune rules would
+  fail. (@panglesd, #1069)
+
 # 2.4.0
 
 ### Added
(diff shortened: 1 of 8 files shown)


Julow (approved):
Looks good! 

What do you think of deprecating this behavior so it can be removed again later ?

jonludlam:
I'm not sure it's necessary. We'll certainly need a fix to dune for this issue - even with this PR merged dune will produce a completely empty file for modules with double underscores, which is of no use to anyone. Once that's fixed (and we've merged the PR changing how we do source rendering), nobody will ever be calling `html-generate` on hidden modules any more, so we might not need to revert this change. It doesn't seem a bad principle that calling `html-generate` will produce _something_ - perhaps an html file saying "This module has been hidden". What do you think?

panglesd:
I agree that having `odoc html-generate` always output something seems something we want, and that telling the reader why the documentation does not appear (as opposed to outputing an empty file) seems better. I'll add that to the PR.

Julow:
I agree, it make sense to always generate something and the driver can easily skip generating these. The new paragraph is a nice touch !

PR #1072, opened 2024-01-17
Remove tyxml from odoc_html_frontend

EmileTrotignon:
Also include a bugfix : name with no prefix are not displayed with a `.` in front.

Removing Tyxml is interesting for the frontend : it results in smaller executable. 
As the typing is different, in an exposed library, it would be beneficial for this to be stabilized quickly.


panglesd:
How much difference is there in the executable size, when produced in "release" mode? I would have though `tyxml` is not too big, given that it is very often used client-side.

trefis:
> Removing Tyxml is interesting for the frontend : it results in smaller executable.

A note about the context here: this lib is compiled to js when used by sherlodoc.

EmileTrotignon:
The difference is more than 10k. I do not think it is huge, and would be fine with it, but Arthur disagrees.
Besides, the important part is the interface, and the new interface is still compatible with tyxml.

panglesd:
> Besides, the important part is the interface, and the new interface is still compatible with tyxml.

I'm not sure what you mean here!

> The difference is more than 10k. 

For a bit of precision, 10k out of how much?

EmileTrotignon:
It was actually ~108k out of 212k which is a big improvement.

EmileTrotignon:
> > Besides, the important part is the interface, and the new interface is still compatible with tyxml.
> 
> I'm not sure what you mean here!
> 

The current interface locks us into using Tyxml. With mine we can use anything, and change the implementation later, without having to change sherlodoc.

Drup:
> As for why tyxml is that big, it forces uses of format which is a big library.

Mmh, I remember it wasn't so hard to avoid pulling Format for Tyxml_js in the past, but that's not something I track carefully on the tyxml side. Maybe I should find a way to enforce it :thinking:.

(hello all :wave: )

src/search/odoc_html_frontend.ml
```diff
@@ -19,9 +104,10 @@ let of_strings ~kind ~prefix_name ~name ~rhs ~typedecl_params ~doc =
         ]
   and prefix_name =
     match prefix_name with
+    | None -> []
+    | Some "" -> []
     | Some prefix_name ->
         [ span ~a:[ a_class [ "prefix-name" ] ] [ txt (prefix_name ^ ".") ] ]
```
panglesd, line 110: There is indeed a bug here. But wouldn't it be better to remove the addition of `"."` to the prefix name:

```suggestion
  and prefix_name =
    match prefix_name with
    | None -> []
    | Some prefix_name ->
        [ span ~a:[ a_class [ "prefix-name" ] ] [ txt prefix_name ] ]
```
and modify [`names_of_id`](https://github.com/ocaml/odoc/blob/master/src/search/html.ml#L174) accordingly?

panglesd (approved):
At the same time, it's nice not to force the consumer of the library to use tyxml, at the same time, it's nice to provide the data in a structured form rather than a string.

Also, re-implementing tyxml in `odoc_html_frontend.ml` makes future modification cumbersome.

However, if tyxml adds a lot to the output (in its current form), and given that it's unlikely that search engine will really use these "data in a structured form" soon, I'm OK with this PR (though, not very enthusiastic!)

src/search/odoc_html_frontend.ml
EmileTrotignon: Yes, it is better i agree.

EmileTrotignon: On second thought, no its not better. I believe there a few place where we String.split the prefix name. This would behave very poorly

EmileTrotignon:
The ci was failing on formating, so I called dune fmt. It made a diff in an unrelated file for reasons I ignore.

src/search/odoc_html_frontend.ml
panglesd: In sherlodoc or in odoc? Could you provide links to such places?

That would maybe mean that the prefix name is turned into a string too early, and we should keep it as a list?

EmileTrotignon: We should probably keep it as a list.

EmileTrotignon: I went and checked more closely. We do indeed use String.split on the name, but we do not use the name functions provided by odoc.search, we call fullname which is a list, so it does not change anything to switch that around. We could even say that the bug was sherlodoc's fault, we could have given None instead of Some "".

CHANGES.md
```diff
@@ -6,12 +6,18 @@
   `count-occurrences` flag and command to count occurrences of every identifiers
   (@panglesd, #976)
 
+### Changed
+
+- `Odoc_html_frontend` does not use tyxml, for smaller javascript sizes.
+  (#EmileTrotignon, #1072)
```
panglesd, line 12:
```suggestion
  (@EmileTrotignon, #1072)
```

PR #1073, opened 2024-01-23
Ident_env: Assume unknown types are core types

Julow:
Types that couldn't be found in the environment are looked up in the list of predefined core types.
This lookup is not expected to fail, as indicated by the presence of an `assert false`.

This removes the core type lookup and assumes all unknown types are core types.

The second commit removes unused Predefined identifiers, paths, references and exceptions. This code required some maintenance in the past.


Julow:
The third commit removes the core type lookup in `Tools`. Instead, core types are constructed on the fly and unknown core types are accepted.
This removes more code from `Predefined`, which now has a small API.

With this change, Odoc no longer crash when encountering unknown core types (eg. using a patched compiler).

gpetiot (approved):
Looks good to me, is there a way to add a cram test showing this with a custom core type?

Julow:
The only way to add custom core types that I know of is to patch the compiler, so no test is possible.

But I believe this is already tested. For example, `int` is no longer predefined and our tests contains ints.

src/loader/ident_env.cppo.ml
```diff
@@ -659,22 +655,17 @@ let find_extension_identifier env id =
 let find_value_identifier env id =
   Ident.find_same id env.values
 
+(** Lookup a type in the environment. If it isn't found, it's assumed to be a
+    core type. *)
```
panglesd, line 659: That means we are confident there is no error in our building of the environment.
Would it make sense to include a simple check that `Ident.name id` is part of a list of predefined types? Kind of similar to what was before, but much more lightweight and easily removable.

panglesd (approved):
There are so many things in this module that were not used! (core exceptions, paths, references, ...)
Do you have an idea why they were introduced?

In any case (whether my comment in the code is addressed or not) I'm in favor of merging this PR.

src/loader/ident_env.cppo.ml
Julow: This patch is intended to be used with patched compilers and not in prevision of new core types. Not having a check is intended.

To more easily debug wrongly built environments, we'd just need to differentiate core types from types that didn't resolve in the generated HTML. We could wrap core types in a `span` with a class (they don't need to be styled differently for this purpose but could be) but I think we can already debug this case without this.

panglesd: I agree that the check is somehow redundant with OCaml having successfully compiled the file, and that it will be annoying for patched compilers.

Julow: Thanks for reviewing :) Now that we agree, I'll merge!

PR #1075, opened 2024-01-29
Find odoc files in accessible path more efficiently

panglesd:
Previously, when $n$ `-I` arguments were given, looking for a `.odoc` file was of complexity $n$, so looking for $n$ files was of complexity $n^2$, which could take 99.99% of a compile command in some extreme case (when a lot of `-I` are given to the compile command).

This commit turns that in $O(n)$ where $n$ is the number of files available in all directories, by doing an "`ls`" in each directory and caching the result.

So, it performs _slightly_ worse in the case of a single `-I` with many files that are not accessed `.odoc` in it, but drastically better in case of many `-I` with (accessed) `.odoc` files in them.


panglesd:
There is a TODO which I think could be fixed in another PR: what to do in case of an inexistent/unavailable directory given to `-I`.

There was no warning, and it is still the case, but at some point it would be good to emit a warning.

Julow (approved):
The benchmark don't seem to contain a reproduction for the mentioned bug but the numbers look ok: compile time increased slightly and link time decreased by about the same (+0.003s per compile, -0.006s per link).

The code looks good.

jonludlam:
Could we please have a test case that demonstrates the problem and the fix?

panglesd:
I added a test with 1000 `-I` and 1000 different lookups, so 1000000 call to `Unix.stat`.
It used to take more than 5s to run, now it takes less than 0.1s.

panglesd:
@jonludlam I added a test for the case that is now worse, and it seems to be OK.

jonludlam:
`timeout` doesn't exist on my mac. Also, some of the tests are failing because 0.1 seconds is too aggressive! :-D

A more portable timeout-like thing could be:

```
(./slow_odoc_invocation.sh; echo 'done')  & sleep 1; kill $! 2> /dev/null || true
```

see e.g. http://blog.mediatribe.net/en/node/72/index.html

panglesd:
Thanks for the mac trick.

The failing CI does not seem to be relevant for this PR:
> [ERROR] The compilation of jst-config.v0.16.0 failed at "dune build -p jst-config -j 3".

jonludlam:
great thanks!

PR #1076, opened 2024-01-31
Add occurrence count in the json output for search engine

panglesd:
Since #972 there is a `compile-index` command which outputs a json index containing all searchable entries. The intended purpose of this json index is to pass it to a search engine.

Since #976 there is also a `count-occurrences` command which creates a table containing the number of occurrences for each identifier.

This PR makes the `compile-index` be able to use the information from `count-occurrence`. It adds a `--occurrences` argument to get the path to the table.

The first commit isolate the occurrence code in its own library.
The second commit implements the `--occurrences` argument of ``compile-index`.


panglesd:
Note that there will be a conflict with #1067 which we will need to resolve (in whichever PR gets merged last).

src/search/json_index/json_search.ml
```diff
@@ -167,11 +167,27 @@ let of_entry ({ Entry.id; doc; kind } as entry) html =
...
   in
+  let occurrences =
+    match occurrences with
+    | Some occ ->
+        `Object
+          [
+            ("direct", `Float (float_of_int occ.Odoc_occurrences.Table.direct));
+            ("indirect", `Float (float_of_int occ.indirect));
+          ]
+    | None -> `Null
```
gpetiot, line 178: Nit-picking but it would look more consistent to me if we had 0 for both counters here (see the output in the tests).

gpetiot (approved):
Minor comment about the displaying of the result, otherwise looks good to me!

src/search/json_index/json_search.ml
panglesd: What about "no 'occurrences' field at all"?
I'm not very keen on choosing a wrong value when it was not possible to compute one because the table was not given...

gpetiot: Well, at least if we don't put this field at all it reduces the diff of this PR!

panglesd: With 6205f97, if the occurrence table is not given, the field is not included.
However, if the occurrence table is given, but does not contain the occurrence, a use of 0 is counted.

doc/driver.mld
```diff
@@ -754,7 +754,11 @@ Some more details about the json format:
        {- ["display"], which is a json object. It contains two fields:
           {ul
             {- ["url"], a string. It is the URL to the entry in the documentation, relative to the base of the documentation}
-            {- ["html"], also a string. It is the html odoc uses to display the entry in the search results.}}}}}}
+            {- ["html"], also a string. It is the html odoc uses to display the entry in the search results.}}}
+       {- Additionally, the ["occurrences"] field exists if and only if the [--occurrences] flag was given to the [odoc compile-index] command. When it exists, it contains a json object, with two fields:
+          {ul
+            {- ["direct"], an integer. It is the number of direct use of the entry. For instance, [open M] and [Make(M)] contain direct uses of [M], while [1 + M.v] contains an indirect use of [M] and a direct use of [M.v].}
```
gpetiot, line 760:
```suggestion
            {- ["direct"], an integer. It is the number of direct uses of the entry. For instance, [open M] and [Make(M)] contain direct uses of [M], while [1 + M.v] contains an indirect use of [M] and a direct use of [M.v].}
```

gpetiot (approved):

EmileTrotignon:
What's the status of this  ?

panglesd:
It was ready for review. Now there are many conflicting files.

panglesd:
It is now rebased.

src/search/json_index/json_search.ml
```diff
@@ -192,11 +206,26 @@ let output_json ppf first entries =
...
 
-let unit ppf u =
+let unit ?occurrences ppf u =
+  let get_occ id =
+    match occurrences with
+    | None -> None
+    | Some occurrences -> (
+        (* We don't want to include the [sub] field of occurrence tables. We use
+           a "polymorphic record" to avoid defining a type, but still get named
+           fields! *)
```
Julow, line 216: This is original but I find it confusing and heavy in syntax. What about defining a record with just that instead ?

This record could even replace the `item` type in `Table` as the `sub` field is never used outside of this module.
Here's an other idea: https://github.com/Julow/odoc/commit/9cafc01c8f803828f15923dca18020010da43924

Julow (approved):
Looks good! Finally a way to consume the occurrences data.

src/search/json_index/json_search.ml
panglesd: I was thinking that the `sub` field could be useful to some application of the occurrence table.

However, there are no such application. I removed `sub` from the exposed type in the last commit, and use that type instead of this weird tuple. If an application requires exposing the `sub` field, we'll do it at this moment.

PR #1077, opened 2024-02-06
Adding documentation on json search index format

panglesd:
This should help consumers of the search engine index.

I noticed that the source for the `odoc_json_index` library is not rendered...


gpetiot (approved):
A few nitpicks, but otherwise looks good to me!

PR #1078, opened 2024-02-08
fix to shape lookup of missing items

jonludlam:
Looking up non-existent things in shapes takes ages - don't do it!


jonludlam:
I don't think this needs a changelog.

panglesd (approved):
I could not find a case where we want to lookup the shape of an internal identifier, so this looks good to me, and is a nice performance improvement (for some reasons)!

PR #1079, opened 2024-02-08
Add a few more tests in anticipation of fixes

jonludlam:
This includes some changes to odoc_print. We might want to change some more tests to use this rather than jq.


jonludlam:
Oh hang on, this needs a fix or two.

test/odoc_print/odoc_print.ml
```diff
@@ -204,9 +256,37 @@ let a_inp =
...
+  let doc = "Show expansions in short output" in
+  Arg.(value & flag & info ~doc [ "show-expansions" ])
+
+let a_long_paths =
+  let doc = "Show long paths in short output" in
+  Arg.(value & flag & info ~doc [ "long-paths" ])
+
+let a_show_canonical =
+  let doc = "Show modules canonical reference in short output" in
+  Arg.(value & flag & info ~doc [ "show-canonical" ])
```
Julow, line 273: This options seems unused. Why shouldn't it be `true` by default ?

test/odoc_print/odoc_print.ml
```diff
@@ -204,9 +256,37 @@ let a_inp =
...
+
+let a_show_include_expansions =
+  let doc = "Show include expansions in short output" in
+  Arg.(value & flag & info ~doc [ "show-include-expansions" ])
+
+let a_show_removed =
+  let doc = "Show removed items in signature expansions in short output." in
+  Arg.(value & flag & info ~doc [ "show-removed" ])
```
Julow, line 281: This one also seems unused. I think it should be true by default, the output of the tests wouldn't be changed currently.

test/odoc_print/odoc_print.ml
```diff
@@ -204,9 +256,37 @@ let a_inp =
...
+  let doc = "Short output." in
+  Arg.(value & flag & info ~doc [ "short" ])
+
+let a_show_expansions =
+  let doc = "Show expansions in short output" in
+  Arg.(value & flag & info ~doc [ "show-expansions" ])
+
+let a_long_paths =
+  let doc = "Show long paths in short output" in
+  Arg.(value & flag & info ~doc [ "long-paths" ])
```
Julow, line 269: This option isn't used. What's its use case ?

test/odoc_print/odoc_print.ml
```diff
@@ -204,9 +256,37 @@ let a_inp =
...
+
+let a_show_canonical =
+  let doc = "Show modules canonical reference in short output" in
+  Arg.(value & flag & info ~doc [ "show-canonical" ])
+
+let a_show_include_expansions =
+  let doc = "Show include expansions in short output" in
+  Arg.(value & flag & info ~doc [ "show-include-expansions" ])
```
Julow, line 277: What's the inconvenient in merging this option with `--show-expansions` ? The output is quite short already.

test/xref2/module_type_of_size.t/gen.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
Julow, line 1: As I said in https://github.com/ocaml/odoc/pull/1042, this test is not portable and is annoying to work with due to many files and the scripts. Here's an alternative that should be more portable and is easier to read and change: https://github.com/jonludlam/odoc/pull/43

Julow (commented):
The new output looks really nice and eliminate the `jq` queries that could easily break with changes in Lang.

src/xref2/component.ml
```diff
@@ -578,251 +598,434 @@ module Fmt = struct
...
+      (config -> Format.formatter -> b -> unit) ->
+      Format.formatter ->
+      a ->
+      b ->
+      unit =
+   fun c txt fn1 fn2 ppf x y ->
+    if c.short_paths then Format.fprintf ppf "%a" (fn1 c) x
+    else Format.fprintf ppf "%s(%a,%a)" txt (fn1 c) x (fn2 c) y
+
+  let wrap2r :
```
Julow, line 627: I'm a bit worried about the added maintenance in this module. Component changes often and the printing functions are slowing down other work.
These functions need some documentation.

jonludlam:
Mostly the extra options were those that were either useful while debugging or I anticipate will be useful in replacing the jq test cases. The marginal cost of adding extra params is dwarfed by the cost of introducing any params at all so I figured I might as well have a few in there now to make the later conversion of tests easier. In addition the defaults are chosen so that 'short' output is shortest!

test/xref2/module_type_of_size.t/gen.sh
jonludlam: I find your version slightly less readable, and it's also more tricky to try outside of 'dune runtest'.

Julow:
I think the many options can be surprising while writing tests. What do you think of removing `--show-include-expansions` ?

src/xref2/component.ml
jonludlam: That's true, I'll add some doc comments.

test/xref2/module_type_of_size.t/gen.sh
Julow: Trying outside of `dune runest` is not a goal of cram tests. Perhaps the test should exist in two different forms.
Why is it less readable ? It contains basically the same code but with 90% of it removed.

jonludlam: I prefer my tests to be very explicit and clear, and your version, to me, obfuscates it with the source on one line with shell variable substitution going on. Please can we leave it as it is, and if you have any problems maintaining the test just punt it to me :-D

jonludlam:
I kept the 'show-include-expansions' separate from the 'show-expansions' as this is required to show the complete contents of a module, whereas the 'show-expansions' will show _everything_ within the signature, including all submodules. I found this useful particularly when comparing the output with `ocamlc -i`

test/xref2/module_type_of_size.t/gen.sh
Julow: Ok, let's leave it as it is.

panglesd (approved):
Looks good to me.

The maintenance cost of all those printing function is still heavy but not that much heavier than today, and the output in tests is much nicer.

I'm fine with the tests and the unused commands.

PR #1081, opened 2024-02-08
Overhaul of module-type-of and shadowing

jonludlam:
This PR contains a relatively big overhaul of how we do module-type-of expansions in particular, but also how we reuse previously calculated expansions for other module type expressions too. In doing so some other changes were required:

- Remove the booleans `mark_substituted` and `add_canonical` from the resolution and expansion functions
- Separate the concepts of shadowed and hidden identifiers
- Add `removed_items` into `Lang.Signature.t` so it can be persisted instead of having to be recalculated
- Simplify the representation of `sig ... end with .. ` (not strictly required but it's another potential source of exponential size problems)

Note that, despite the name of the branch, this _hasn't_ fixed the issue that the `module type of` expressions aren't quite right, though their expansions are correct. A little bit more logic is required to assess whether we need to show an ascription operation.

This PR includes the content of PRs #1079 and #1078 .


src/model/lang.ml
```diff
@@ -101,7 +102,7 @@ and ModuleType : sig
       | Path of Path.ModuleType.t
       | Signature of Signature.t
       | With of substitution list * expr
-      | TypeOf of typeof_t
+      | TypeOf of type_of_desc * Path.Module.t
 
     (* Nb. this may have an expansion! *)
```
jonludlam, line 107: This comment needs to be removed!

src/xref2/cpath.ml
```diff
@@ -395,3 +395,64 @@ and unresolve_type_path : type_ -> type_ = function
 and unresolve_class_type_path : class_type -> class_type = function
   | `Resolved m -> unresolve_resolved_class_type_path m
   | y -> y
+
```
jonludlam, line 398: Perhaps leave in but commented out? this isn't being used. Also add docs!

src/xref2/tools.ml
```diff
@@ -1690,6 +1690,40 @@ and unresolve_subs subs =
...
+  | Some cp -> (
+      expansion_of_module_path env ~strengthen p >>= fun exp ->
+      match expansion_of_module_path env ~strengthen:false cp with
+      | Ok _orig_exp ->
+          Ok exp
+          (* Ok (ascribe env ~expansion:exp ~original_expansion:orig_exp) *)
+      | Error lookup_error ->
+          Format.eprintf "Here we are!\n%!";
+          Error lookup_error)
+(*
```
jonludlam: Get rid of this, if there's no value in it.

src/xref2/tools.ml
```diff
@@ -1815,19 +1883,15 @@ and fragmap :
     | Alias (path, _) ->
         expansion_of_module_path env ~strengthen:true path
```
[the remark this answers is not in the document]
jonludlam: Also, do we need to qualify the RHS of the substitution? It might not be valid at this point?

src/xref2/tools.ml
```diff
@@ -2247,7 +2311,10 @@ and resolve_module_fragment :
         match expansion_of_module env m' with
         | Ok (_m : expansion) -> f'
         | Error `OpaqueModule -> `OpaqueModule f'
-        | Error (`UnresolvedForwardPath | `UnresolvedPath _) -> f'
+        | Error
+            ( `UnresolvedForwardPath | `UnresolvedPath _
+            | `UnresolvedOriginalPath _ ) ->
```
jonludlam, line 2215: Should we still have this extra error?

src/model/names.ml
```diff
@@ -81,7 +89,8 @@ module Name : Name = struct
 
   let is_hidden = function
```
jonludlam, line 93: Perhaps rename this?

src/loader/cmt.ml
```diff
@@ -448,12 +448,12 @@ and read_module_binding env parent mb =
   let canonical = (canonical :> Path.Module.t option) in
   let hidden =
 #if OCAML_VERSION >= (4,10,0)
-    match canonical, mb.mb_id with
-    | None, Some id -> Odoc_model.Names.contains_double_underscore (Ident.name id)
+    match canonical, mid.iv with
+    | None, (`Module (_, n) | `Parameter (_, n) | `Root (_, n)) -> Odoc_model.Names.ModuleName.is_hidden n
     | _, _ -> false
```
jonludlam, line 453: Put a Some here for exhaustiveness checking

src/loader/cmt.ml
```diff
@@ -448,12 +448,12 @@ and read_module_binding env parent mb =
...
 #else
-    match canonical with
-    | None -> Odoc_model.Names.contains_double_underscore (Ident.name mb.mb_id)
+    match canonical, mid.iv with
+    | None, (`Module (_, n) | `Parameter (_, n) | `Root (_, n)) -> Odoc_model.Names.ModuleName.is_hidden n
```
jonludlam, line 456: And here.

src/model/names.ml
```diff
@@ -54,7 +54,10 @@ end
 
 let internal_counter = ref 0
 
-module Name : Name = struct
+module type NameTy = sig
```
jonludlam, line 57: Perhaps something like:
```
module ModuleName = struct
  include ModuleName

  let is_hidden = ... 
```
?

src/xref2/tools.ml
```diff
@@ -1882,6 +1886,13 @@ and fragmap :
     map_module_decl m.type_ new_subst >>= fun type_ ->
     Ok (Left { m with type_ })
   in
+  let map_removed = function
```
jonludlam, line 1823: Why are we doing this instead of creating the correct item in the first place?

test/xref2/expansion.t/run.t
```diff
@@ -0,0 +1,900 @@
+  $ ocamlc -c -bin-annot test.mli
+  $ odoc compile test.cmti
+  $ odoc_print test.odoc
```
jonludlam, line 3: Rewrite this test using the new output of odoc_print

src/xref2/tools.ml
```diff
@@ -78,7 +78,12 @@ let canonical_helper :
       ('resolved * 'result) option =
  fun env resolve lang_of possibilities p2 ->
   let resolve p =
-    match resolve env p with Ok rp -> Some rp | Error _ -> None
+    (* Format.eprintf "Resolving...\n%!"; *)
```
jonludlam, line 81: Tidy this up

src/xref2/tools.ml
```diff
@@ -1368,16 +1320,21 @@ and handle_canonical_module_real env p2 =
```
jonludlam, line 1320: The comment immediately above here needs updating as it still mentions ~add_canonical

src/xref2/tools.ml
jonludlam: answer: because we need to do `sub_of_removed` and that needs the Ident.t rather than just the name.

src/xref2/tools.ml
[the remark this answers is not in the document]
jonludlam: Not required indeed.

src/xref2/subst.ml
```diff
@@ -691,56 +692,52 @@ and u_module_type_expr s t =
...
+              List.map (with_module_type_substitution s) w_substitutions;
+            w_expansion = option_ simple_expansion s w_expansion;
+            w_expr = u_module_type_expr s w_expr;
           }
-      with MTOInvalidated ->
-        module_type_expr s (module_type_of_simple_expansion e))
-  | TypeOf { t_desc; t_expansion = None } ->
-      TypeOf
-        { t_desc = module_type_type_of_desc_noexn s t_desc; t_expansion = None }
+    | TypeOf desc -> TypeOf (module_type_of_type_desc s desc)
```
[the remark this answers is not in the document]
jonludlam: In fact, it turns out it needs removal :-)

jonludlam:
I think I've resolved the issues we found during review, and also a couple of others:
- I've fixed the infinite loop that @lukemaurer found in async_kernel
- I've removed all of the `MTOInvalidated` stuff in subst as this is no longer needed.

test/generators/html/Ocamlary.html
```diff
@@ -2319,7 +2320,9 @@ <h4 id="advanced-type-stuff">
     <div class="spec type anchored" id="type-dep5">
      <a href="#type-dep5" class="anchor"></a>
      <code><span><span class="keyword">type</span> dep5</span>
-      <span> = <a href="Ocamlary-Dep13-class-c.html">Dep13.c</a></span>
+      <span> = 
+       <a href="Ocamlary-Dep11-module-type-S-class-c.html">Dep13.c</a>
+      </span>
```
panglesd, line 2325: It's correct but slightly odd to link from `Dep13.c` to `Dep11.S.c` when there is a `Dep13` page containing `c`...

panglesd (approved):
Although my expertise on this area is not very high, I approve this PR! It looks good to me.
It fixes a problem with module type expansions, and the other modifications in this PR are also good improvements to the codebase.
Thanks @jonludlam for the work!

Issue #1022, opened 2023-10-17
Provide a Complete Module Index


sabine:
One recurring feature request we get for OCaml.org is the wish to have a **full index** (which we would use on to the left navigation on the ocaml.org package documentation) **that includes all the definitions within the modules / classes / etc**.

I think it's a valid request that also enables a diff of the module index between package versions.

sabine:
> We have plans to improve the navigation side bar in the future (though not concrete enough yet) that would include listing submodules and siblings modules. It would hopefully be good enough for ocaml.org. Do you also plan to generate a page listing all the definitions ?

Just putting this here so it's not forgotten. No concrete plans to do this right now.

A thing we can do with a complete index is to compute diffs between two package versions to know when a given value has been added or removed between versions.

panglesd:
> @panglesd Do you think the work on occurrence counting could help list the definitions ?

The occurrence command list the occurrences as found in the implementation files. If no occurrence is found, the table won't include the definition.

Maybe the index for search entries could be used instead...

PR #1084, opened 2024-02-20
Add a marshalled output for index generation

panglesd:
This PR adds a new marshalled output for the generation of index files.

This is desirable for several reasons:
- It allows to combine several pre-built indexes, making the index generation more incremental
- It can be read by other OCaml tools using odoc as library, for instance when building a search engine index (cc @EmileTrotignon). #1022 is also relevant.

The first commit fixes a bug in the computation of the ID of the page containing a standalone comment. The second commit is the actual change.

There are some design to discuss:
- This PR adds a `--marshall` flag to the `compile-index`: by default the output is json, but it is marshalled with this flag.
- The files containing marshelled indexes have to be prefixed by `index-` and have to have the `.odoc` extension.
- Currently, the format is a hashtable. I think it would be better to have a better datastructure, similar to the one to contain occurrences.
- The complexity when merging indexes many times is not ideal. For instance the following commands would be in O(n^2)
```
$ odoc compile-index --marshall -o index-all1.odoc a1.odocl
$ odoc compile-index --marshall -o index-all2.odoc a2.odocl index-a1.odoc
$ odoc compile-index --marshall -o index-all3.odoc a3.odocl index-a2.odoc
$ odoc compile-index --marshall -o index-all4.odoc a4.odocl index-a3.odoc
$ odoc compile-index --marshall -o index-all5.odoc a5.odocl index-a4.odoc
...
$ odoc compile-index --marshall -o index-all<n>.odoc a<n>.odocl index-a<n-1>.odoc
$
```
But I think that is ok.

### DIFF
--- a/odoc-driver.opam
+++ b/odoc-driver.opam
@@ -34,7 +34,7 @@ documentation for installed packages.
 
 depends: [
   "odoc" {= version}
-  "bos" 
+  "bos"
   "fpath"
   "yojson"
   "ocamlfind"
@@ -43,6 +43,7 @@ depends: [
   "eio_main"
   "progress"
   "cmdliner"
+  "sherlodoc"
 ]
 
 build: [
--- a/odoc-driver.opam
+++ b/odoc-driver.opam
@@ -43,11 +43,6 @@ depends: [
   "eio_main"
   "progress"
   "cmdliner"
-  "sherlodoc"
-]
-
-pin-depends: [
-  ["sherlodoc.dev" "git+https://github.com/emiletrotignon/sherlodoc.git#076cc2b"]
 ]
 
 build: [
--- a/odoc-driver.opam
+++ b/odoc-driver.opam
@@ -46,6 +46,10 @@ depends: [
   "sherlodoc"
 ]
 
+pin-depends: [
+  ["sherlodoc.dev" "git+https://github.com/emiletrotignon/sherlodoc.git#076cc2b"]
+]
+
 build: [
   ["dune" "subst"] {dev}
   [
--- a/src/driver/compile.ml
+++ b/src/driver/compile.ml
@@ -16,17 +16,17 @@ type compiled = {
   include_dirs : Fpath.Set.t;
   impl : impl option;
   pkg_args : pkg_args;
-  current_package : string;
+  pkgname : string;
 }
 
 let mk_byhash (pkgs : Packages.t Util.StringMap.t) =
   Util.StringMap.fold
-    (fun _pkg_name pkg acc ->
+    (fun pkg_name pkg acc ->
       List.fold_left
         (fun acc (lib : Packages.libty) ->
           List.fold_left
             (fun acc (m : Packages.modulety) ->
-              Util.StringMap.add m.m_intf.mif_hash m acc)
+              Util.StringMap.add m.m_intf.mif_hash (pkg_name, m) acc)
             acc lib.modules)
         acc pkg.Packages.libraries)
     pkgs Util.StringMap.empty
@@ -95,7 +95,7 @@ let compile output_dir all =
     | None ->
         Logs.debug (fun m -> m "Error locating hash: %s" hash);
         Error Not_found
-    | Some modty ->
+    | Some (pkgname, modty) ->
         let deps = modty.m_intf.mif_deps in
         let output_file = Fpath.(output_dir // modty.m_intf.mif_odoc_file) in
         let fibers =
@@ -146,7 +146,7 @@ let compile output_dir all =
             include_dirs = includes;
             impl;
             pkg_args;
-            current_package = modty.m_package;
+            pkgname;
           }
   in
 
@@ -167,7 +167,7 @@ let compile output_dir all =
     List.filter_map (function Ok x -> Some x | Error _ -> None) mod_results
   in
   Util.StringMap.fold
-    (fun _ (pkg : Packages.t) acc ->
+    (fun pkgname (pkg : Packages.t) acc ->
       Logs.debug (fun m ->
           m "Package %s mlds: [%a]" pkg.name
             Fmt.(list ~sep:sp Packages.pp_mld)
@@ -191,13 +191,13 @@ let compile output_dir all =
             include_dirs;
             impl = None;
             pkg_args;
-            current_package = pkg.name;
+            pkgname : string;
           }
           :: acc)
         acc pkg.mlds)
     all mods
 
-type linked = { output_file : Fpath.t; src : Fpath.t option }
+type linked = { output_file : Fpath.t; src : Fpath.t option; pkgname : string }
 
 let link : compiled list -> _ =
  fun compiled ->
@@ -205,7 +205,7 @@ let link : compiled list -> _ =
    fun c ->
     let includes = Fpath.Set.add c.output_dir c.include_dirs in
     let link input_file =
-      let { pkg_args = { libs; docs }; current_package; _ } = c in
+      let { pkg_args = { libs; docs }; pkgname = current_package; _ } = c in
       Odoc.link ~input_file ~includes ~libs ~docs ~current_package ()
     in
     let impl =
@@ -214,7 +214,13 @@ let link : compiled list -> _ =
           Logs.debug (fun m -> m "Linking impl: %a" Fpath.pp impl);
           link impl;
           Atomic.incr Stats.stats.linked_impls;
-          [ { output_file = Fpath.(set_ext "odocl" impl); src = Some src } ]
+          [
+            {
+              output_file = Fpath.(set_ext "odocl" impl);
+              src = Some src;
+              pkgname = c.pkgname;
+            };
+          ]
       | None -> []
     in
     match c.m with
@@ -227,16 +233,66 @@ let link : compiled list -> _ =
         (match c.m with
         | Module _ -> Atomic.incr Stats.stats.linked_units
         | Mld _ -> Atomic.incr Stats.stats.linked_mlds);
-        { output_file = Fpath.(set_ext "odocl" c.output_file); src = None }
+        {
+          output_file = Fpath.(set_ext "odocl" c.output_file);
+          src = None;
+          pkgname = c.pkgname;
+        }
         :: impl
   in
   Fiber.List.map link compiled |> List.concat
+
+let odoc_index_path ~odoc_dir pkgname =
+  Fpath.(odoc_dir / pkgname / "index.odoc-index")
+let sherlodoc_js_index_path_relative_to_html pkgname =
+  Fpath.(v pkgname / "sherlodoc_db.js")
+
+let sherlodoc_js_path_relative_to_html = Fpath.v "sherlodoc.js"
+let sherlodoc_js_index_path ~html_dir pkgname =
+  Fpath.(html_dir // sherlodoc_js_index_path_relative_to_html pkgname)
+
+let sherlodoc_js_path ~html_dir =
+  Fpath.(html_dir // sherlodoc_js_path_relative_to_html)
+
+let sherlodoc_marshall_path ~html_dir =
+  Fpath.(html_dir / "sherlodoc_db.marshal")
+let index_one output_dir pkgname _pkg =
+  let dir = Fpath.(output_dir / pkgname) in
+  let dst = odoc_index_path ~odoc_dir:output_dir pkgname in
+  let include_rec = Fpath.Set.singleton dir in
+  Odoc.compile_index ~json:false ~dst ~include_rec ()
+let index odoc_dir pkgs = Util.StringMap.iter (index_one odoc_dir) pkgs
+
+let sherlodoc_index_one ~html_dir ~odoc_dir pkgname _pkg_content =
+  ignore @@ Bos.OS.Dir.create Fpath.(html_dir / pkgname);
+  let format = `js in
+  let inputs = [ odoc_index_path ~odoc_dir pkgname ] in
+  let dst = sherlodoc_js_index_path ~html_dir pkgname in
+  Sherlodoc.index ~format ~inputs ~dst ()
+
+let sherlodoc ~html_dir ~odoc_dir pkgs =
+  ignore @@ Bos.OS.Dir.create html_dir;
+  Sherlodoc.js (sherlodoc_js_path ~html_dir);
+  Util.StringMap.iter (sherlodoc_index_one ~html_dir ~odoc_dir) pkgs;
+  let format = `marshal in
+  let dst = sherlodoc_marshall_path ~html_dir in
+  let inputs =
+    pkgs |> Util.StringMap.bindings
+    |> List.map (fun (pkgname, _pkg) -> odoc_index_path ~odoc_dir pkgname)
+  in
+  Sherlodoc.index ~format ~inputs ~dst ()
 
 let html_generate : Fpath.t -> linked list -> _ =
  fun output_dir linked ->
   let html_generate : linked -> unit =
    fun l ->
-    Odoc.html_generate
+    let search_uris =
+      [
+        sherlodoc_js_index_path_relative_to_html l.pkgname;
+        sherlodoc_js_path_relative_to_html;
+      ]
+    in
+    Odoc.html_generate ~search_uris
       ~output_dir:(Fpath.to_string output_dir)
       ~input_file:l.output_file ?source:l.src ();
     Atomic.incr Stats.stats.generated_units
--- a/src/driver/odoc.ml
+++ b/src/driver/odoc.ml
@@ -5,31 +5,10 @@ type compile_deps = { digest : Digest.t; deps : (string * Digest.t) list }
 let odoc = Cmd.v "./_build/default/src/odoc/bin/main.exe"
 (* This is the just-built odoc binary *)
 
-let submit desc cmd output_file =
-  match Worker_pool.submit desc cmd output_file with
-  | Ok x -> x
-  | Error exn -> raise exn
-
-let compile_output = ref [ "" ]
-
-let compile_src_output = ref [ "" ]
-
-let link_output = ref [ "" ]
-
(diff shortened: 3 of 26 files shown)


EmileTrotignon:
I do not like the `index-<name>.odoc` syntax. It is unusual to express file type with a prefix, and it might break scripts/makefiles that use `*.odoc`. `.odoci` or `.odoc_index` would be better.

EmileTrotignon:
@Drup 
`.odoc` represent the data of a single page. `.odocl` also, but the page has been linked with others so that it is possible to resolve references.

In term of command line, you need a single module (cmti ?) to create a `.odoc`, but you need every .odoc to create a `.odocl`.
There are no others except the one in this proposal.

EmileTrotignon:
> About the index-<name>.odoc syntax: it is not that unusal, at least in odoc:

I know it is quite common in odoc, but at least for files names I really do not like it because it is a weird convention that fails to communicate its intent, so this requires knowledge of odoc to understand. It might also restrict names available for users, in a way that will be unexpected : no one will try to name a page `.index` (or at least they will understand that it might have meaning), but `index-` might happen. 

I also do not like that file with the same extension are in fact of a different "type". I think you should be able to trust that the extension tells you what kind of file a file is, and in (1), all the `.odoca` have completely different (binary) content and they cannot be used interchangeably at all.

Regarding implementation and sources, it seems to me that they are used in the same context (linking) and as such can share an extension. Their content could even be of the same ocaml type if we wanted to (maybe they are ?).

I deeply favor (2), even though I understand that it requires changes out of the scope of this PR. 

The only benefits for (1) I found in your message is the fact it is done like that in other places in odoc, but as it seems insufficient with regard to the drawbacks expressed above. Maybe there are other benefits I did not consider ?

EmileTrotignon:
The interface changes from the first commit are fine, sherlodoc only uses Fold.unit and Fold.page

panglesd:
I'm not decided on which is better, but let me still answer to some of you comments

> this requires knowledge of odoc

Using `odoc` requires knowledge of `odoc`. Only odoc drivers are expected to call the command line. So I don't think this is such a big deal!

> no one will try to name a page .index (or at least they will understand that it might have meaning), but index- might happen

This is not a problem at all. I don't think there can be any issue:
- a compilation unit named `<name>` cannot have `-` in it
- Other odoc produces artifact are prefixed: for instance, the compilation of `index-bli.mld` will generate `page-index-bli.mld` and that would not cause problems

> I also do not like that file with the same extension are in fact of a different "type". I think you should be able to trust that the extension tells you what kind of file a file is, and in (1), all the .odoca have completely different (binary) content and they cannot be used interchangeably at all.

`odoc` files have completely different binary content, and cannot be used interchangeably: for instance, only pages can be passed as parents.
Similarly, json files can have completely different content and not be used interchangeably, but it still makes sense to give them the same  extension.

---

So:
1. `index-<name>.odoca`, `occurrences-<name>.odoca`:
      - Makes it explicit which phase those files belong to: `.odoca` files belong to the "post-link" phase, which takes `.odocl` files as input.
      - Consistent with what exists in the compile (`.odoc`) and link (`.odocl`) phase
2. `<name>.odoc-index`, `<name>.odoc-occurrences`
      - @EmileTrotignon could you tell what what are the main benefits, considering what I wrote above?

panglesd:
> The interface changes from the first commit are fine, sherlodoc only uses Fold.unit and Fold.page

(sherlodoc uses `entries_of_item`, which is also changed by the bugfix.)

EmileTrotignon:
> (sherlodoc uses entries_of_item, which is also changed by the bugfix.)
I had missed that, the change still looks fine.

EmileTrotignon:
> Using odoc requires knowledge of odoc. Only odoc drivers are expected to call the command line. So I don't think this is such a big deal!

Yes, but it that case, we make up our own convention when a perfectly fine one already exists, so even though it's not unreasonable to ask people to learn things, my impression is that here, we are making it harder when its not necessary.

> Other odoc produces artifact are prefixed: for instance, the compilation of index-bli.mld will generate page-index-bli.mld and that would not cause problems

You are right that it will not cause conflicts, but it could be confusing  for people who might be trying to understand what is happening looking at `_build`, for instance if you also had `bli.mld`. Having a `bli.odoc-index` would be immediately understood for what it is.

> odoc files have completely different binary content, and cannot be used interchangeably: for instance, only pages can be passed as parents.
Similarly, json files can have completely different content and not be used interchangeably, but it still makes sense to give them the same extension.

Yes, I do not mean that they should be used completely interchangeably, this is not possible if two file have different content, but that there exists a context where they are all accepted. This is the case for json (you can always pass a json file to jq), and I hope it is for `.odoc` (a `.odoc` is always valid as something to be linked against ? It can always be used to generate and odocl ?).
There is no such context for occurrences and index files, the only thing they have in common is being produced at link time, which is not the usual meaning of a file extension.

Regarding consistency, I think that this convention is more appropriate for  `src-` and `page-` because they have shared uses, which is not the case for `index-` and `occurrences-`.

To me, the main benefit of option (2) is that it makes it clear what you can do with the file, in a way that is understandable with minimal friction. This feel like a way more important information about a file than when it was produced.

src/odoc/bin/main.ml
```diff
@@ -403,46 +403,56 @@ module Source_tree = struct
...
-         files to index."
+        "Input text file containing a line-separated list of paths to \
+         .odocl/.json files to index."
       in
       Arg.(
         value & opt_all convert_fpath []
         & info ~doc ~docv:"FILE" [ "file-list" ])
     in
+    let marshall =
+      let doc = "whether to output a json file, or an .odoc file" in
```
EmileTrotignon, line 445: Small nitpick : I think it should be "a .odoc file" (pronounced "a dot odoc file")

EmileTrotignon:
I have read the diff of the second commit and it looks good to me aside from a very small nitpick.

EmileTrotignon:
This has been overhauled significantly, and a re-review would be appropriate.

The extension for index files is now .odoc-index, and you give .odocl files with `odoc compile-index --include-rec my_dir` .

src/driver/compile.ml
```diff
@@ -222,16 +231,50 @@ let link : compiled list -> _ =
...
+  let dst = Fpath.(dir / "index.odoc-index") in
+  let include_rec = Fpath.Set.singleton dir in
+  Odoc.compile_index ~json:false ~dst ~include_rec ()
+let index odoc_dir pkgs = Util.StringMap.iter (index_one odoc_dir) pkgs
+
+let sherlodoc_index_one ~html_dir ~odoc_dir pkgname _pkg_content =
+  ignore @@ Bos.OS.Dir.create Fpath.(html_dir / pkgname);
+  let format = `js in
+  let inputs = [ Fpath.(odoc_dir / pkgname / "index.odoc-index") ] in
+  let dst = Fpath.(html_dir / pkgname / "sherlodoc_db.js") in
```
panglesd, line 254: This filename is hardcoded and reused. It would be better to return it or something like that.

src/driver/compile.ml
```diff
@@ -222,16 +231,50 @@ let link : compiled list -> _ =
...
+  Sherlodoc.js Fpath.(html_dir / "sherlodoc.js");
+  Util.StringMap.iter (sherlodoc_index_one ~html_dir ~odoc_dir) pkgs;
+  let format = `marshal in
+  let dst = Fpath.(html_dir / "sherlodoc_db.marshal") in
+  let inputs =
+    pkgs |> Util.StringMap.bindings
+    |> List.map (fun (pkgname, _pkg) ->
+           Fpath.(odoc_dir / pkgname / "index.odoc-index"))
+  in
+  Sherlodoc.index ~format ~inputs ~dst ()
```
panglesd, line 268: There are several hardcoded path there, which should be returned to avoid having them hardcoded in distinct places.

src/driver/odoc.ml
```diff
@@ -1,31 +1,11 @@
 open Bos
+open Cmd_outputs
```
panglesd, line 2: Let's not open that globally...

panglesd (commented):
I only reviewed the driver part!

It looks good to me. It would be nice to:
- add some parallelism, which is easy here (see eg https://github.com/panglesd/odoc/commit/2dadfb5ea6af86c58b1936837e9df25bb0eb4caa)
- Add a progress bar for indexes, since it can take quite a lot of time

src/odoc/bin/main.ml
```diff
@@ -478,46 +479,49 @@ module Compile_impl = struct
...
+         generated index."
       in
       Arg.(
-        value & opt_all convert_fpath []
-        & info ~doc ~docv:"FILE" [ "file-list" ])
+        value
+        & opt_all (convert_directory ()) []
+        & info ~doc ~docv:"DIR" [ "include-rec" ])
     in
-    let inputs =
```
panglesd, line 514: Why removing the old methods? Passing a list of files instead of a list of directories can be useful in some usecases, I think!

src/odoc/indexing.ml
```diff
@@ -66,3 +57,63 @@ let compile ~output ~warnings_options inputs_in_file inputs =
...
+
+let compile_to_marshall ~output ~warnings_options files =
+  let final_index = H.create 10 in
+  let unit u =
+    Odoc_model.Fold.unit
+      ~f:(fun () item ->
+        let entries =
+          Odoc_search.Entry.entries_of_item
+            (* (u.Odoc_model.Lang.Compilation_unit.id *)
+            (*   :> Odoc_model.Paths.Identifier.t) *)
```
panglesd, line 69:
```suggestion
```

src/odoc/indexing.ml
```diff
@@ -66,3 +57,63 @@ let compile ~output ~warnings_options inputs_in_file inputs =
...
+          (fun entry -> H.add final_index entry.Odoc_search.Entry.id entry)
+          entries)
+      () u
+  in
+  let page p =
+    Odoc_model.Fold.page
+      ~f:(fun () item ->
+        let entries =
+          Odoc_search.Entry.entries_of_item
+            (* (p.Odoc_model.Lang.Page.name :> Odoc_model.Paths.Identifier.t) *)
```
panglesd, line 82:
```suggestion
```

src/odoc/indexing.mli
```diff
@@ -4,13 +4,14 @@ val handle_file :
   Fpath.t ->
   unit:(Odoc_model.Lang.Compilation_unit.t -> 'a) ->
   page:(Odoc_model.Lang.Page.t -> 'a) ->
+  occ:(Odoc_search.Entry.t Odoc_model.Paths.Identifier.Hashtbl.Any.t -> 'a) ->
```
panglesd, line 7: I don't remember, is `handle_file` exposed for sherlodoc?

src/odoc/resolver.ml
```diff
@@ -92,7 +92,7 @@ end = struct
     let flat_namespace = Hashtbl.create 42 in
     let () =
       match
-        Fs.Directory.fold_files_rec_result
+        Fs.Directory.fold_files_rec_result (* bookmark *)
```
panglesd, line 95: ?

```suggestion
        Fs.Directory.fold_files_rec_result
```

test/search/html_search.t/run.t
panglesd, line 1: The tests and prose around needs to be updated!

src/odoc/bin/main.ml
panglesd: In particular, I think we need them to specify `.odoc-index` files.

src/driver/compile.ml
EmileTrotignon: I do not think it is a good idea to return them, because --search_uri wants relative paths. I have factored them in functions and values and its not repeated now.

src/driver/odoc.ml
EmileTrotignon: Ok

src/odoc/indexing.mli
EmileTrotignon: Yes it is used in sherlodoc. I would like that we do not remove it for now, sherlodoc still accepts .odocl files using this function.

src/odoc/bin/main.ml
```diff
@@ -478,46 +479,49 @@ module Compile_impl = struct
...
+            "At least one of --include-rec must be passed to odoc compile-index")
+    | _ -> Indexing.compile marshall ~output ~warnings_options include_rec
   let cmd =
     let dst =
       let doc =
         "Output file path. Non-existing intermediate directories are created. \
-         Defaults to index.json"
+         Defaults to index.odoc-index, or index.json if --json is passed (in \
+         which case, the .odoc-index file extension is mandatory)."
+        (* This is not checked for now *)
```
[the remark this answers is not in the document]
EmileTrotignon: Alright

src/odoc/bin/main.ml
EmileTrotignon: I have added them back in.

odoc-driver.opam
```diff
@@ -43,6 +43,7 @@ depends: [
   "eio_main"
   "progress"
   "cmdliner"
+  "sherlodoc"
```
panglesd, line 46: A pinned version is needed, for now, I think.

src/driver/compile.ml
```diff
@@ -222,16 +231,66 @@ let link : compiled list -> _ =
...
+          output_file = Fpath.(set_ext "odocl" c.output_file);
+          src = None;
+          pkgname = c.pkgname;
+        }
         :: impl
   in
   Fiber.List.map link compiled |> List.concat
 
+let odoc_index_path ~odoc_dir pkgname =
+  Fpath.(odoc_dir / pkgname / "index.odoc-index")
```
panglesd, line 244: I believe it would be cleaner to output index files outside of the directory made for `.odocl` files. That is:
```suggestion
  Fpath.(index_dir / pkgname / "index.odoc-index")
```
or 
```suggestion
  Fpath.(index_dir / pkgname ^ ".odoc-index")
```

Although, this is not mandatory. Just resolve the comment if you believe it is better like that.

src/odoc/bin/main.ml
```diff
@@ -511,13 +527,28 @@ module Indexing = struct
         value & opt_all convert_fpath []
         & info ~doc ~docv:"FILE" [ "file-list" ])
     in
+    let include_rec =
```
Julow, line 530: What's the point of this option ? The `inputs` term could also accept directories, which removes an exception case and is common among CLI tools.

src/odoc/bin/main.ml
```diff
@@ -476,28 +476,44 @@ module Compile_impl = struct
...
+    match (includes_rec, inputs_in_file, inputs) with
+    | [], [], [] ->
         Result.Error
           (`Msg
-            "At least one of --file-list or an .odocl file must be passed to \
-             odoc compile-index")
-    | _ -> Indexing.compile ~output ~warnings_options inputs_in_file inputs
-
+            "At least one of --file-list or --include-rec or an .odocl file \
+             must be passed to odoc compile-index")
```
Julow, line 507: This is already checked in `Indexing.compile` but in a slightly different way: If `--file-list` points to an empty file, this passes but `Indexing.compile` has an other check that do not.

Why shouldn't making an empty index be allowed ? For example, drivers might want to make an index for every packages, even empty or bugous packages. This restriction increases complexity in drivers.

src/odoc/indexing.ml
```diff
@@ -66,3 +73,58 @@ let compile ~output ~warnings_options inputs_in_file inputs =
...
+  let files =
+    List.rev_append files
+      (includes_rec
+      |> List.map (fun include_rec ->
+             Fs.Directory.fold_files_rec ~ext:"odocl"
+               (fun files file -> file :: files)
+               [] include_rec)
+      |> List.concat)
+  in
+  if files = [] then Error (`Msg "No .odocl files were included")
```
Julow, line 126: As there's several indirect ways to pass the list of inputs, to reassure driver authors, we could log the number of odocl files considered for indexing:
```
Indexing N odocl files.
```

src/odoc/indexing.mli
```diff
@@ -4,13 +4,16 @@ val handle_file :
...
   ('a, [> msg ]) result
 (** This function is exposed for custom indexers that uses [odoc] as a library
     to generate their search index *)
 
 val compile :
+  [ `JSON | `Marshall ] ->
```
Julow, line 13: What is the use case for each formats ? It would be nice to write that down for maintenance in the future.

Julow (approved):
Nice :) Made some comments but nothing that can't be changed later.

src/odoc/bin/main.ml
panglesd: In any case, this option will change to `-P` and `-L` soon: we will use the index for sidebars too, and sidebars require those options.

src/odoc/bin/main.ml
panglesd: Good point.

src/odoc/indexing.mli
panglesd: The use cases for json are for external use of indexes, which include using external search engine but is not really limited...
The usecase for the marshalled format is for odoc: currently search, soon table of contents, maybe indices of values/types/... so it is going to evolve.

I'm not sure how much it would help maintenance or hurt it to have the usecases written there, as they are likely going to evolve and the explanation will be outdated. Maybe a simple sentence (`` `Marshall `` is for uses of the index by `odoc`, `` `JSON `` is for external uses) would be sufficient?

Julow:
The CI is not passing due to the new dependency: https://ocaml.ci.dev/github/ocaml/odoc/commit/f0acee4b79947a0a4ff1643bc675e877045b98a3/variant/debian-12-5.0_opam-2.1

panglesd:
I opened an issue on ocaml-ci : https://github.com/ocurrent/ocaml-ci/issues/954

As soon as it is solved, we should reintroduce the driver's dependency on sherlodoc.

PR #1085, opened 2024-02-21
Normalize whitespaces in codespans

gpetiot:
To summarize:
- A newline followed by any whitespaces should be considered to be one space character by odoc-parser and ocamlformat.
Consecutive whitespaces not after a newline are conserved as they are.
- Trim whitespaces at the beginning and end of code spans. Not related to the issue but as we making whitespaces a bit less significant, this is worth mentioning.

### DIFF
--- a/src/parser/lexer.mll
+++ b/src/parser/lexer.mll
@@ -617,7 +617,7 @@ and code_span buffer nesting_level start_offset input = parse
           ~in_what:(Token.describe (`Code_span "")));
       Buffer.add_char buffer ' ';
       code_span buffer nesting_level start_offset input lexbuf }
-  | newline horizontal_space+
+  | newline horizontal_space*
     { Buffer.add_char buffer ' ';
       code_span buffer nesting_level start_offset input lexbuf }
 
--- a/src/parser/lexer.mll
+++ b/src/parser/lexer.mll
@@ -263,6 +263,8 @@ let horizontal_space =
   [' ' '\t']
 let newline =
   '\n' | "\r\n"
+let whitespace =
+  horizontal_space | newline
 
 let reference_start =
   "{!" | "{{!" | "{:" | "{{:"
@@ -609,13 +611,13 @@ and code_span buffer nesting_level start_offset input = parse
     { Buffer.add_char buffer c;
       code_span buffer nesting_level start_offset input lexbuf }
 
-  | newline newline
+  | newline whitespace+ as sp
     { warning
         input
         (Parse_error.not_allowed
-          ~what:(Token.describe (`Blank_line "\n\n"))
+          ~what:(Token.describe (`Blank_line sp))
           ~in_what:(Token.describe (`Code_span "")));
-      Buffer.add_char buffer '\n';
+      Buffer.add_char buffer ' ';
       code_span buffer nesting_level start_offset input lexbuf }
 
   | eof
--- a/src/parser/lexer.mll
+++ b/src/parser/lexer.mll
@@ -263,8 +263,6 @@ let horizontal_space =
   [' ' '\t']
 let newline =
   '\n' | "\r\n"
-let whitespace =
-  horizontal_space | newline
 
 let reference_start =
   "{!" | "{{!" | "{:" | "{{:"
@@ -611,13 +609,16 @@ and code_span buffer nesting_level start_offset input = parse
     { Buffer.add_char buffer c;
       code_span buffer nesting_level start_offset input lexbuf }
 
-  | newline whitespace+ as sp
+  | newline horizontal_space* (newline horizontal_space*)+
     { warning
         input
         (Parse_error.not_allowed
-          ~what:(Token.describe (`Blank_line sp))
+          ~what:(Token.describe (`Blank_line "\n\n"))
           ~in_what:(Token.describe (`Code_span "")));
       Buffer.add_char buffer ' ';
+      code_span buffer nesting_level start_offset input lexbuf }
+  | newline horizontal_space+
+    { Buffer.add_char buffer ' ';
       code_span buffer nesting_level start_offset input lexbuf }
 
   | eof
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,8 @@
 
 - Updated colors for code fragments (@EmileTrotignon, #1023)
 - Fixed complexity of looking up `.odoc` files (@panglesd, #1075)
+- Normalize whitespaces in codespans (@gpetiot, #1085)
+  A newline followed by any whitespaces is normalized as one space character.
 
 ### Changed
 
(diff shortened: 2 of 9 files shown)


dbuenzli:
> Implementing the proposition of https://github.com/tarides/tooling-tasks/issues/49

Please avoid referencing non-public propositions in a public project. 

> * A newline followed by any whitespaces should be considered to be one space character by odoc-parser and ocamlformat.

That's quite an odd convention. If you want to change the current behaviour (not sure what it is) I think you should rather ignore *all* whitespace (and make sure you can introduce one with an escape).

jonludlam:
How will this affect mdx?

dbuenzli:
> That's quite an odd convention.

Sorry I misunderstood, I thought it was for code blocks. What you propose is fine with me (and FWIW indirectly what CommonMark does).

dbuenzli:
> * Trim whitespaces at the beginning and end of code spans. Not related to the issue but as we making whitespaces a bit less significant, this is worth mentioning.

Also I don't think this should be done.

gpetiot:
> How will this affect mdx?

There won't be any impact, mdx doesn't process code spans, only the code blocks

panglesd:
For the records, here is the issue this PR is solving.

Citing @Julow (private conversation):
> [In the context of ocamlformat] Code spans can be large and might need to break. For example, we might want to format the first into the second:, and having the following two versions identical would allow for better formatting:
> ```
> (** [very_long_word b] *)
>
> (** [very_long_word
>      b] *)
> ```
> In the second case, Odoc's AST contains "very_long_word\n     b". When using the html backend, both comments render the same as the whitespaces are normalized by the browser.
>
> My suggestions:
>
> -   A newline followed by any whitespaces should be considered to be one space character by odoc-parser and ocamlformat.
>     Consecutive whitespaces not after a newline are conserved as they are.
> -   Trim whitespaces at the beginning and end of code spans. Not related to the issue but as we making whitespaces a bit less significant, this is worth mentioning.
> -   Backward compatibility is not a concern. Most already written don't have this problem. The newline is not visible when using the html backend.


See also this [PR](https://github.com/ocaml-ppx/ocamlformat/pull/1736) for the discussion on the ocamlformat side.

dbuenzli:
As I said I don't think initial and trailing whitespace should be suppressed. 1) There may be legitimate reasons to have some 2) It needlessly deviates from other lightweight markup languages like CommonMark.

panglesd:
Maybe this PR could be split in two:
- The modification solving the original issue: considering newlines followed by blanks as a single space, and
- the initial and final whitespace trimming

so one modification is not blocking the other?

I think having the syntax for inline elements play nice with the "80 char rule" is good to have, so I would be in favor of the first change. I haven't though about the second change, but as a first impression I would be in favor of having the same behavior as CommonMark.

gpetiot:
> * the initial and final whitespace trimming

I removed these changes and cleaned up the PR.

src/parser/lexer.mll
```diff
@@ -609,13 +611,13 @@ and code_span buffer nesting_level start_offset input = parse
...
+  | newline whitespace+ as sp
     { warning
         input
         (Parse_error.not_allowed
-          ~what:(Token.describe (`Blank_line "\n\n"))
+          ~what:(Token.describe (`Blank_line sp))
           ~in_what:(Token.describe (`Code_span "")));
-      Buffer.add_char buffer '\n';
+      Buffer.add_char buffer ' ';
       code_span buffer nesting_level start_offset input lexbuf }
```
panglesd, line 621: Two newlines are still not allowed. What is allowed is one newline followed by some space (all of that being counted as one space).

panglesd (changes requested):

src/parser/lexer.mll
gpetiot: This new case subsumes the old case. I don't have a good answer to how many spaces should correspond "\n\n" (or more linebreaks), 1 space for every case is the easy answer.
I'm in favor of removing the warning as well.

panglesd: After thinking about this a little bit, I think that whitespace including two newlines should emit a warning. Indeed, code spans are inline elements and we only allow newlines to break long lines.

So the following are equivalent (using `␣` as space to make it more visible):
```
[a␣␣␣b]
[a␣␣
b]
[a␣␣
␣␣␣␣␣b]
```

and the following is forbidden:
```
[a

b]
```

This is closer to what commonmark does (two newlines separate in two paragraph _before_ the inlines inside the paragraph are parsed).

src/parser/lexer.mll
```diff
@@ -609,13 +611,8 @@ and code_span buffer nesting_level start_offset input = parse
...
 
-  | newline newline
-    { warning
-        input
-        (Parse_error.not_allowed
-          ~what:(Token.describe (`Blank_line "\n\n"))
-          ~in_what:(Token.describe (`Code_span "")));
-      Buffer.add_char buffer '\n';
+  | newline whitespace+
+    { Buffer.add_char buffer ' ';
```
panglesd, line 615: Something along the lines of (not tested):
```suggestion
  | newline horizontal_space (newline horizontal_space)+
    { warning
        input
        (Parse_error.not_allowed
          ~what:(Token.describe (`Blank_line "\n\n"))
          ~in_what:(Token.describe (`Code_span "")));
      Buffer.add_char buffer ' ';
      code_span buffer nesting_level start_offset input lexbuf }
  | newline horizontal_space+
    { Buffer.add_char buffer ' ';
```

panglesd (commented):
Maybe a small mention of that in the [docs](https://ocaml.github.io/odoc/odoc_for_authors.html#inline_code_blocks) could be good! Odoc's language is underspecified but let's slowly improve that...

src/parser/lexer.mll
gpetiot: I've restored the warning, and fixed up the optional horizontal spaces, let me know what you think

src/parser/lexer.mll
```diff
@@ -609,13 +609,16 @@ and code_span buffer nesting_level start_offset input = parse
...
+  | newline horizontal_space* (newline horizontal_space*)+
     { warning
         input
         (Parse_error.not_allowed
           ~what:(Token.describe (`Blank_line "\n\n"))
           ~in_what:(Token.describe (`Code_span "")));
-      Buffer.add_char buffer '\n';
+      Buffer.add_char buffer ' ';
+      code_span buffer nesting_level start_offset input lexbuf }
+  | newline horizontal_space+
```
panglesd, line 620: A newline without horizontal space should also be converted to a single space

```suggestion
  | newline horizontal_space*
```

panglesd:
Promoted tests and updated docs, and then it's mergeable!

panglesd (approved):
This PR is ready to be merged. I'll let a few days before merging to let anyone who dislike the change speak up.

jonludlam:
Looks good to me. In it goes!

PR #1088, opened 2024-02-27
Search bar navigation

EmileTrotignon:
The odoc search bar does have navigation : you cannot use the arrow keys to select a search result.

This add javascript that allows this.

There is also a shortcut to enter search : `/` and to exit : `escape`.
Escape is very intuitive, an `/` is also used by other website like github or discuss.

There should no diff on css, as the js uses the focus mechanism and there was css for that already.

I am thinking of adding a visual cue that `/` focuses the search bar. Github has one, but I am not very happy with what I came up with. Ideas and help are welcome.


panglesd:
Thanks, that a much improved user experience!

Small feedback from using it:

- Going "up" when already at the top yields an error:
   ```
   Uncaught TypeError: results[(current_focus - 1)] is undefined
   ```
- (nitpick) It would be nice if the focus were teleported to the input when typing again: 
   - Type "abc"
   - Type "down arrow" several times... don't find what was looked for
   - type "d", refocuses on the search bar (and add "d").
- One simple way to tell about the keybinding is to use the placeholder: Turn "🔎 Search..." into "🔎 Type '/' to search" (just like in github). I find it fine.

EmileTrotignon:
@panglesd 
> Going "up" when already at the top yields an error: 

This is fixed

> (nitpick) It would be nice if the focus were teleported to the input when typing again: 

This is done as well

> One simple way to tell about the keybinding is to use the placeholder: Turn "🔎 Search..." into "🔎 Type '/' to search" (just like in github). I find it fine.

I wanted to be too much like github : my attempt was "🔎 Type [/] to search" which looked ugly. This is now done, thanks.

src/html_support_files/odoc_search.js
```diff
@@ -64,3 +64,91 @@ document.querySelector(".search-bar").addEventListener("input", (ev) => {
...
+    if (event.key === "/") {
+      event.preventDefault();
+      enter_search();
+    }
+    else if ( event.key.length === 1
+            && (ascii >= 65 && ascii <= 90) // lowercase letter
+            || (ascii >= 97 && ascii <= 122) // uppercase letter
+            || (ascii >= 48 && ascii <= 57) // numbers
+            || (ascii >= 32 && ascii <= 46) // ` ` to `.`
+            || (ascii >= 58 && ascii <= 64) // `:` to `@`
```
panglesd, line 148: I'm not sure we want that all those ascii characters trigger search...

panglesd: > > (nitpick) It would be nice if the focus were teleported to the input when typing again:
> This is done as well

Ah, maybe this was due to this! But typing on any ascii character even outside of search triggers search, maybe that is too much?

panglesd (approved):
Looks good to me! Thanks!

PR #1089, opened 2024-03-07
Add clock emoji before 'since' tag

yawaramin:
Example:

Screenshot 2024-03-07 at 00 26 03

### DIFF
--- a/src/html_support_files/odoc.css
+++ b/src/html_support_files/odoc.css
@@ -725,13 +725,13 @@ td.def-doc *:first-child {
 /* Alert emoji */
 
 .alert::before, .deprecated::before {
-  content: '⚠️ ';
+  content: '⚠️ ' / '';
 }
 
 /* Since emoji */
 
 .since::before {
-  content: '🕚 ';
+  content: '🕚 ' / '';
 }
 
 /* Lists of modules */
--- a/src/html_support_files/odoc.css
+++ b/src/html_support_files/odoc.css
@@ -728,6 +728,12 @@ td.def-doc *:first-child {
   content: '⚠️ ';
 }
 
+/* Since emoji */
+
+.since::before {
+  content: '🕚 ';
+}
+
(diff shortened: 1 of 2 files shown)


panglesd (approved):
I like that! Thanks.

I hope this does not worsen accessibility with screen readers, but we already have the warning emoji and it makes sense to fix potential problems in another work (and this does improve readability for sighted people).

yawaramin:
Hmm yeah unfortunately VoiceOver on Mac reads the emoji as 'clock face'. I had thought (hoped) it would ignore content injected via CSS. Let me investigate further.

panglesd:
While searching for info, I found that `content` seems to be okay to use for decorative purpose (see [this](https://developer.mozilla.org/en-US/docs/Web/CSS/::before#accessibility_concerns) for instance). So I think that it fits this case.

> EDIT: seems like this will work: `content: '🕚 ' / '';`

That's great! 
I could not find any information for this syntax (I could only validate it on [this website](https://jigsaw.w3.org/css-validator/validator)). Do you have any reference?

yawaramin:
Sure, it's the alt text. Explained here with examples and syntax: https://developer.mozilla.org/en-US/docs/Web/CSS/content

panglesd:
Thanks a lot, let's merge!

src/html_support_files/odoc.css
```diff
@@ -725,7 +725,13 @@ td.def-doc *:first-child {
 /* Alert emoji */
 
 .alert::before, .deprecated::before {
-  content: '⚠️ ';
+  content: '⚠️ ' / '';
```
yawaramin, line 728: Just informational, the ` / ''` here is to set the alt text to an empty string which tells the screen reader to not read out the emoji because it is strictly decorative and redundant for spoken text.

Julow: Very interesting, thanks!

PR #1090, opened 2024-03-07
Less aggressive timeout on "big search path" tests

panglesd:
CI times may vary, we don't want to have false positive too often.


jonludlam:
Thanks!

PR #1094, opened 2024-03-21
Basic support for OCaml 5.2.0

Octachron:
This PR makes the minimal changes for supporting OCaml 5.2, without taking advantages of any of the new shape features in OCaml 5.2.0 .

Another aspect that should probably be improved later on is the support for `open`s in type expressions: the present PR is only traversing the open while it could be useful to keep track of those when generating the documentation:
```ocaml
val add:  Int.( t -> t -> t)
```

### DIFF
--- a/src/model/compat.cppo.ml
+++ b/src/model/compat.cppo.ml
@@ -239,25 +239,25 @@ let empty_map = Shape.Uid.Map.empty
 let shape_info_of_cmt_infos : Cmt_format.cmt_infos -> (shape * uid_to_loc) option =
  fun x -> Option.map (fun s -> (s, x.cmt_uid_to_loc)) x.cmt_impl_shape
 #else
-let loc_of_declaration =
-  let open Typedtree in
-  function
-  | Value v -> v.val_loc
-  | Value_binding vb -> vb.vb_pat.pat_loc
-  | Type t -> t.typ_loc
-  | Constructor c -> c.cd_loc
-  | Extension_constructor e -> e.ext_loc
-  | Label l -> l.ld_loc
-  | Module m -> m.md_loc
-  | Module_substitution ms -> ms.ms_loc
-  | Module_binding mb -> mb.mb_loc
-  | Module_type mt -> mt.mtd_loc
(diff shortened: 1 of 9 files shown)


test/xref2/lib/common.cppo.ml
```diff
@@ -623,7 +627,10 @@ let mkresolver () =
   Odoc_odoc.Resolver.create
     ~important_digests:false
     ~directories:(List.map Odoc_odoc.Fs.Directory.of_string
-#if OCAML_VERSION >= (4,8,0)
+#if OCAML_VERSION >= (5,2,0)
+  (let paths = Load_path.get_paths () in
+   List.filter (fun s -> s <> "") (paths.visible @ paths.hidden))
```
Octachron, line 632: This is the conservative version, it is possible that keeping just `paths.visible` would be better for odoc.

src/model/compat.cppo.ml
```diff
@@ -235,8 +235,30 @@ type 'a shape_uid_map = 'a Shape.Uid.Map.t
 type uid_to_loc = Warnings.loc Types.Uid.Tbl.t
 let empty_map = Shape.Uid.Map.empty
 
+#if OCAML_VERSION < (5,2,0)
 let shape_info_of_cmt_infos : Cmt_format.cmt_infos -> (shape * uid_to_loc) option =
  fun x -> Option.map (fun s -> (s, x.cmt_uid_to_loc)) x.cmt_impl_shape
+#else
+let loc_of_declaration =
```
panglesd, line 242: It might be good to have this function local to `shape_info_of_cmt_infos` (to avoid having a different signature for `Compat` on 5.2 and 5.1)

test/xref2/lib/common.cppo.ml
panglesd: Apparently, those function are only used in an old `.md` test, so anything is fine.

panglesd (approved):
Thanks a lot for the PR!

src/model/compat.cppo.ml
Octachron: Good point, done.

Issue #1091, opened 2024-03-10
odoc: several variants completely missing from generated docs, but present in .odocl


edwintorok:
The documentation for Omd's `type block` is incorrect.

```ocaml
type !'attr block = 
| Blockquote of 'attr4 * 'attr5 [block](https://ocaml.org/p/omd/latest/doc/Omd/index.html#type-block) list
| Thematic_break of 'attr6
| Code_block of 'attr9 * string * string
| Html_block of 'attr10 * string
| Definition_list of 'attr11 * 'attr12 [def_elt](https://ocaml.org/p/omd/latest/doc/Omd/index.html#type-def_elt) list
```

However the source code has:
```ocaml
 type 'attr block =
    | Paragraph of 'attr * 'attr C.t
    | List of 'attr * list_type * list_spacing * 'attr block list list
    | Blockquote of 'attr * 'attr block list
    | Thematic_break of 'attr
    | Heading of 'attr * int * 'attr C.t
    | Code_block of 'attr * string * string
    | Html_block of 'attr * string
    | Definition_list of 'attr * 'attr def_elt list
    | Table of 'attr * ('attr C.t * cell_alignment) list * 'attr C.t list list
        (** A table is represented by a header row, which is a list of pairs of
            header cells and alignments, and a list of rows *)
```

Same issue with older version of API docs: https://ocaml.github.io/omd/omd/Omd/index.html#type-block

What I tried:
* adding a new variant at the beginning, e.g. `Z`: it shows up, even though `Paragraph` and `List` and `Table` do not
* removing the extra doc comment for table, doesn't help
* look at the .odocl file with `strings -a`
* generate other output formats: html-generate, latex-generate, man-generate are all affected

Steps to reproduce:
```
opam source omd
dune build @doc --cache=disabled
odoc latex-generate _build/default/_doc/_odocls/omd/omd.odocl --output-dir tmp
strings -a _build/default/_doc/_odocls/omd/omd.odocl|grep -A30 %attr1
cat tmp/omd/Omd.3o|grep -A20 'type.*attr block'
```

Output for `strings -a`, see that the variants are all present:
```
%attr1@@@
%block
W3t_block.r_Omd.p_omd@@@
$attr@A@@
*)Paragraph
        "ctor_Paragraph.t_block.r_Omd.p_omd@
%attr0
%attr1@@@
D$List
c=ctor_List.t_block.r_Omd.p_omd@
%attr2
%attr3@@@@@
v*Blockquote
[...]

PR #1105, opened 2024-03-28
Treat hidden representations differently

jonludlam:
Instead of omitting variant constructors where they've got hidden elements in their types, we now show all constructors and just emit a warning when we detect the problem.

In addition, also check the fields of records similarly.

### DIFF
--- a/src/xref2/link.ml
+++ b/src/xref2/link.ml
@@ -365,6 +365,51 @@ and open_ env parent = function
   | { Odoc_model__Lang.Open.doc; _ } as open_ ->
       { open_ with doc = comment_docs env parent doc }
 
+let warn_on_hidden_representation (id : Id.Type.t)
+    (r : Lang.TypeDecl.Representation.t) =
+  let open Paths.Identifier in
+  let rec internal_typ_exp typ_expr =
+    let open Lang.TypeExpr in
+    let open Paths.Path in
+    match typ_expr with
+    | Constr (p, ts) ->
+        is_hidden (p :> Paths.Path.t)
+        || List.exists (fun t -> internal_typ_exp t) ts
+    | Poly (_, t) | Alias (t, _) -> internal_typ_exp t
+    | Arrow (_, t, t2) -> internal_typ_exp t || internal_typ_exp t2
+    | Tuple ts | Class (_, ts) -> List.exists (fun t -> internal_typ_exp t) ts
+    | _ -> false
+  in
+
+  let internal_cstr_arg t =
+    let open Lang.TypeDecl.Constructor in
+    let open Lang.TypeDecl.Field in
+    match t.args with
+    | Tuple type_exprs ->
+        List.exists (fun type_expr -> internal_typ_exp type_expr) type_exprs
+    | Record fields ->
+        List.exists (fun field -> internal_typ_exp field.type_) fields
+  in
+
+  let internal_field t =
+    let open Lang.TypeDecl.Field in
+    internal_typ_exp t.type_
+  in
+
+  let fmt_cfg = Component.Fmt.{ default with short_paths = true } in
+  match r with
+  | Variant constructors ->
+      if List.exists internal_cstr_arg constructors then
+        Lookup_failures.report_warning "@[<2>Hidden constructors in type '%a'@]"
+          Component.Fmt.(model_identifier fmt_cfg)
+          (id :> Id.any)
+  | Record fields ->
+      if List.exists internal_field fields then
+        Lookup_failures.report_warning "@[<2>Hidden fields in type '%a'@]"
+          Component.Fmt.(model_identifier fmt_cfg)
+          (id :> Id.any)
+  | Extensible -> ()
+
 let rec unit env t =
   let open Compilation_unit in
   let content =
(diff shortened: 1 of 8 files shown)


jonludlam:
This addresses problems brought up in #1091

src/xref2/link.ml
```diff
@@ -365,6 +365,51 @@ and open_ env parent = function
...
+  let internal_field t =
+    let open Lang.TypeDecl.Field in
+    read_typ_exp t.type_
+  in
+
+  let fmt_cfg = Component.Fmt.{ default with short_paths = true } in
+  match r with
+  | Variant constructors ->
+      if List.exists internal_cstr_arg constructors then
+        Lookup_failures.report_warning "@[<2>Hidden constructors in type '%a'@]"
```
Julow, line 403: I'm not convinced a warning is useful to users. Except maybe, if there's a way to disable it.

This can happen in legit code and I don't think I would change my code based on this warning.
I suggest removing the warning.

src/xref2/link.ml
```diff
@@ -365,6 +365,51 @@ and open_ env parent = function
   | { Odoc_model__Lang.Open.doc; _ } as open_ ->
       { open_ with doc = comment_docs env parent doc }
 
+let warn_on_hidden_representation (id : Id.Type.t)
+    (r : Lang.TypeDecl.Representation.t) =
+  let open Paths.Identifier in
+  let rec read_typ_exp typ_expr =
```
Julow, line 371: `typ_contains_hidden` would be a better name ?

test/generators/html/Stop_dead_link_doc.html
```diff
@@ -112,25 +112,87 @@ <h1>Module <code><span>Stop_dead_link_doc</span></code></h1>
...
+     <code><span><span class="keyword">type</span> another_bar_</span>
+      <span> = </span>
+     </code>
+     <ol>
+      <li id="type-another_bar_.Bar__" class="def variant constructor
+       anchored"><a href="#type-another_bar_.Bar__" class="anchor"></a>
+       <code><span>| </span>
+        <span><span class="constructor">Bar__</span> 
+         <span class="keyword">of</span> 
+         <span><span class="xref-unresolved">{Another_Foo}1.t</span> option
```
Julow, line 190: I think internal idents shouldn't be outputted. The useful parts of this type declaration are: Exhaustive list of the constructors, arity of each constructors.

I suggest rendering types containing hidden paths as `_`. This is not valid OCaml but neither is the proposed output.

This example could render as `Bar of _` if we keep this specific to variants or `Bar of _ option` if we extend this to every type expressions.

Julow (commented):
I think we should recognize this as legit code and have a nicer output.

src/xref2/link.ml
jonludlam: True! I just moved the code from generator.ml and didn't think enough about the name. I'll fix this, good call.

src/xref2/link.ml
jonludlam: The problem usually comes up because someone has overridden the default wrapping that dune does, e.g. writing a manual foo.ml(i) for wrapped library foo. So I don't think it happens unless someone is _trying_ to make the interface shaped differently, and in that case we really do want people to fix it, no? Or is there another way this can happen? For an example of the problem I'm trying to fix 'in the wild', have a look at my PR to omd.

test/generators/html/Stop_dead_link_doc.html
jonludlam: Yeah I'm quite sympathetic to this point of view, but I'd rather address this in a larger scoped PR, as this is a much bigger issue than in just the variants.

src/xref2/link.ml
Julow: I don't think Odoc is the best tool to report this problem. Also, if the interface works, then Odoc should render it nicely.

It seems to me that the warning will be annoying 100% of the time in exchange for being useful rarely.

test/generators/html/Stop_dead_link_doc.html
Julow: In the meantime, we could have this for variants only ? Eg. render `Bar of _`.

src/xref2/link.ml
jonludlam: I don't understand, what other tool would report this?

lukemaurer: Odoc definitely isn't the best tool to report this problem, but unfortunately it's not clear what else can. It's very easy for a type to be used without using its name at all. And so the interface might "work" but there's _no_ nice way to render it. 

How common is it to export a constructor but not its argument type? On purpose, that is?

Julow: We don't have a tool to report this but it has been suggested that the OCaml community needs a linter.
My point is that we don't report this problem at all.

jonludlam: The problem is that in the case of omd a suitable fix involved canonical tags, so whatever tool it is that's reporting the problem, it'll have to understand them. Once we've got such a tool I'll happily delete this code.

Julow: Works for me. We can continue discussing this if it becomes an unsolvable error.

test/generators/html/Stop_dead_link_doc.html
jonludlam: That would require special-casing the rendering of paths when they're found in variants, so I don't want to go down this route at this point.

Julow (approved):
I'm not a fan of rendering the internal and unstable idents. I think the special casing for variants is worth implementing.

Other than that, this is good to go.

PR #1107, opened 2024-04-05
Enable coverage of cram tests

jonludlam:
See https://github.com/ocaml/dune/issues/3884 for some discussion about the problem / workaround

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,9 +19,11 @@ jobs:
           - 4.10.x
           - 4.12.x
           - 4.14.x
+          - 5.0.x
+          - 5.1.x
         include:
           - os: ubuntu-latest
-            ocaml-compiler: 4.14.x
+            ocaml-compiler: 5.1.x
             # We don't need to compute coverage for more than one build
             send-coverage: true
             # Mdx tests Mdx tests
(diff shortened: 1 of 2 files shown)


Julow (approved):
Works on my machine!

jonludlam:
Didn't work for the coverage reporting done to coveralls via github actions! Fixed that now though.

.github/workflows/build.yml
```diff
@@ -60,8 +62,10 @@ jobs:
         if: matrix.send-coverage == true
         run: |
           opam install bisect_ppx
+          mkdir -p $BISECT_FILE
           opam exec -- dune runtest --instrument-with bisect_ppx --force
-          opam exec -- bisect-ppx-report send-to Coveralls
+          opam exec -- bisect-ppx-report send-to Coveralls --coverage-path /tmp/coverage
```
Julow, line 67: Coveralls returns:

```
{"message":"Job #1940.1","url":"https://coveralls.io/jobs/139417004"}
```

Do you have an idea of if we could have this URL easily accessible ?

Julow (approved):
I had forgotten we ran coveralls for so long as the URL to the result is hard to find.
That works :)

.github/workflows/build.yml
jonludlam: https://coveralls.io/github/ocaml/odoc

PR #1108, opened 2024-04-12
Fix href for aliased modules in search results

### LINKED ISSUE #1106: Search entries for module aliases return dead links
Searching for `Throttle` in `Async` at http://ocaml.org or on http://doc.sherlocode.com produces an entry for `Async.Throttle` with a dead link to the url:

https://ocaml.org/p/async/v0.16.0/doc/Async/Throttle/index.html

it should instead link to:

https://ocaml.org/p/async_kernel/v0.16.0/doc/Async_kernel/Throttle/index.html

because `Throttle` in `Async` is just a module alias to `Throttle` in `Async_kernel`.

panglesd:
Fix #1106

It is not ideal that calls to `Url.from_identifier` can result in 404 pages on aliased modules. This never happened before the introduction of search. This PR fixes this in an ad-hoc way for search.

I think it would be better to fix it in the identifier themselves: the ID of an aliased module would include the ID of its alias if it exists. Otherwise, generated URL can point to the definition point: the parent page with the appropriate anchor.

Before this is implemented, I think this fix is good enough. What do you think?

Ping @EmileTrotignon for the change in API of the search library.


jonludlam:
Probably the simplest fix is to say 'stop_before=true' - this should cause links to everything that might have an expansion to go to the definition point instead of into the expansion. The current change definitely doesn't look right.

panglesd:
> Probably the simplest fix is to say 'stop_before=true'

The main problem is that standalone comments do not have an ID, and thus use the ID of their parent module/page. If we use `stop_before=true` in all cases, when clicking on a comment we will jump in the definition point of the module containing the comment, which is not ideal at all. That's why the fix is a bit more difficult than just `stop_before=true`...

> Since `Url.from_identifier` takes any identifier it cannot tell the difference between a signature identifier and a module identifier, 

I opened #1109 which also stems from this issue. Indeed fixing both #1109 and #1106 at the same time would be much better.

> so it should always assume it is a module identifier and give the url that definitely exists.

That would turn all references of modules into links to the definition of the module. Currently those references link to the expansion. That can make sens but I'm always cautious when changing the current behaviour...
I'd like to try to change the odoc IDs to be able to distinguish between a module ID and a signature of module ID.

lukemaurer:
> That would turn all references of modules into links to the definition of the module.

In particular, it would make every `[Core.List]` link to the `module List = List` line in the odoc for `core.mli`, no?

jonludlam:
Looking back on my comment I missed the vital bit of info: I meant just use `stop_before=true` for _all modules_ (except compilation units). So any search for a module ends up at the place where the module is defined, as Leo suggested. I'm not convinced changing the behaviour if there's an expansion or not is worth it.

panglesd:
I implemented your suggestion @jonludlam. Instead of using `stop_before` to true only for modules, I set it to false only for standalone docstrings. Indeed, other expansable items (such as module types) also suffer from the issue.

src/search/html.ml
```diff
@@ -3,11 +3,10 @@ type html = Html_types.div_content Tyxml.Html.elt
...
 open Lang
 
-let url id =
-  match
-    Odoc_document.Url.from_identifier ~stop_before:false
-      (id :> Odoc_model.Paths.Identifier.t)
-  with
+let url { Entry.id; kind; doc = _ } =
+  let open Entry in
+  let stop_before = match kind with Doc _ -> false | _ -> true in
```
jonludlam, line 8: Could you please put a comment in the code here explaining why you're matching `Doc`? It's not very obvious!

src/search/html.ml
```diff
@@ -3,11 +3,15 @@ type html = Html_types.div_content Tyxml.Html.elt
...
-    Odoc_document.Url.from_identifier ~stop_before:false
-      (id :> Odoc_model.Paths.Identifier.t)
-  with
+let url { Entry.id; kind; doc = _ } =
+  let open Entry in
+  let stop_before =
+    (* Docstring do not have an ID, so we use the ID from the parent signature
+       in search entries. Links to signature need [stop_before] to be false
+       (otherwise the link may point to the definition point of the module). *)
+    match kind with Doc _ -> false | _ -> true
```
jonludlam, line 12: Hmm.

I think my point was that the intent of this line of code is to prevent a link potentially going to a module/module type/class/etc that doesn't have an expansion. I was mostly expecting to see something more like `match kind with | Module _ -> true | ModuleType _ -> true ... | _ -> false` - where the wildcard is matching the things like `Exception` and so on. I don't think it's terribly obvious that `stop_before` has no impact on those things, so instead we can just match `Doc` and let `stop_before` do nothing for the other things.

jonludlam:
That's much better, thanks!

PR #1112, opened 2024-04-16
5.2.0 support and shape: use amended query API

Octachron:
This PR amends to the support for OCaml 5.2.0 to the updated version of the `Shape_reduce` queries API.


panglesd (approved):
Thanks!

PR #1100, opened 2024-03-27
Fix problem with exceptions and type extensions in extended opens

jonludlam:
This is related to issue #1066, which was partially fixed by #1082. This PR sorts the remaining two instances of the problem: exceptions and type extensions.

Fixes #1095

PR #1114, opened 2024-04-19
Refactor Ident_env.extract_signature_type_items

Julow:
This separates the item extraction from the handling of visibility and of the 'hidden' flag.
This should catch the bug fixed by https://github.com/ocaml/odoc/pull/1100


jonludlam:
This looks very sensible to me.

PR #1116, opened 2024-04-19
Use heading text for references to heading even across pages

### LINKED ISSUE #941: Title splicing problems on master

panglesd:
Fix #941 and #779.


src/xref2/ref_tools.ml
```diff
@@ -817,13 +817,19 @@ let resolve_reference =
...
         | `Value (id, _) -> identifier id
         | `Type (id, _) -> identifier id
-        | `Label (id, _) -> identifier id
+        | `Label (id, _) ->
+            let text =
+              match Env.lookup_by_id Env.s_label id env with
+              | Some (`Label (_, lbl)) -> Some lbl.Component.Label.text
+              | None -> None
+            in
+            identifier ?text id
```
jonludlam, line 832: Did this last commit fix a bug?

panglesd: The interface for `Ref_tools.resolve_reference` was changed to also return a `paragraph option`: the "paragraph"[^1] is a possible replacement text computed when resolving the reference. In the case of a title, this "paragraph" is the title.

[^1]:  `paragraph` here is just a type alias for `inline_element with_location list`. Not an actual paragraph.

However, before the last commit, this paragraph was not always output as `Some` by `resolve_reference`, so there was some [code](https://github.com/ocaml/odoc/pull/1116/files#diff-be0d75e85b8a996a6f315280c9f08df1ff70f8af28810af26cf282b28bc7f9aeL232-L238) in `Link` which looked for the title content in the missing case.

In short, the last commit does not fix a bug, it just moves this lookup of title content from `Link.comment_inline_element` to `Ref_tools.resolve_reference` so that the lookup of the title content is centralized in one place.

jonludlam:
This seems fine, as far as it goes. At some point I think it'd be good if `resolve_reference` returned the component of the thing it found rather than just some extracted text, more like `Tools.resolve_module` and the like. This would allow a tool to look up an element by reference and then show the docs for that element.

In any case, this is a step in the right direction and fixes some bugs, so in it goes. Thanks!

PR #1117, opened 2024-04-19
Import top comment from top include of a hidden module

### LINKED ISSUE #926: Modules should inherit top-comments from hidden includes too
#654 made it so that an `include` that is marked with `(** @inline *)` can import a top-comment into a module if it's the first item. I think it would be good to extend this to also work for includes of hidden modules. Such includes are already automatically expanded, and commonly occur when people use the `_intf.ml` pattern because the `_intf` module is not usually exposed by the library. I think this would save people from a lot of `@inline` annotation, or more realistically, improve a lot of documentation where people have neglected to do that.

panglesd:
Fix #926.

### DIFF
--- a/src/model/lang.ml
+++ b/src/model/lang.ml
@@ -555,24 +555,21 @@ let umty_of_mty : ModuleType.expr -> ModuleType.U.expr option = function
 (** Query the top-comment of a signature. This is [s.doc] most of the time with
     an exception for signature starting with an inline includes. *)
 let extract_signature_doc (s : Signature.t) =
-  let rec uexpr_is_hidden = function
+  let rec uexpr_considered_hidden = function
     | ModuleType.U.Path p -> Path.is_hidden (p :> Path.t)
     | Signature _ ->
         true (* Hidden in some sense, we certainly want its top comment *)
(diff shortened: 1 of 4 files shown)


src/model/lang.ml
```diff
@@ -555,10 +555,24 @@ let umty_of_mty : ModuleType.expr -> ModuleType.U.expr option = function
...
+    | TypeOf { t_desc = ModPath p | StructInclude p; _ } ->
+        Path.is_hidden (p :> Path.t)
+  in
   match (s.doc, s.items) with
   | [], Include { expansion; status = `Inline; _ } :: _ ->
       (* A signature that starts with an [@inline] include inherits the
          top-comment from the expansion. This comment is not rendered for
          [include] items. *)
       expansion.content.doc
+  | [], Include { expansion; decl = Alias p; _ } :: _
```
Julow, line 572: I think this would be better written as a `should_inline_expansion` function to avoid duplicating the match on the doc, `Include` and the RHS of `->`.

Julow (approved):
Looks good :) I have an optional comment.

PR #1118, opened 2024-05-03
Fix wrong id being given to doc comments

panglesd:
Standalone documentation comments currently do not have an id. In search results, the goal was to give them the id of the signature they are in.
However, there was a confusion in the computation of this signature, resulting in a wrong ID. This PR fixes that.

The commit was originally in #1084 but since this PR does not seem to make a lot of progress, I think it is better to extract it in a standalone PR.


jonludlam:
Always nice to see PRs that remove more lines than they add :-)

jonludlam:
I can't see clearly from the test what was broken before and now fixed? Is it worth having an explicit test? The sea of json in the existing test case is a bit meaningless right now.

panglesd:
I added a more readable test focusing on the issue.

panglesd:
Ping @jonludlam @Julow

jonludlam:
Thanks!

PR #1119, opened 2024-05-06
Add cheatsheet to the odoc manual

panglesd:
This adds a cheatsheet to the odoc manual. Have a look at the rendered page [here](https://choum.net/panglesd/odoc_cheatsheet/cheatsheet.html).

The cheatsheet is rendered by odoc, which is not really focused on rendering cheatsheets, so it does not render ideally well.
If someone wants to contribute by adding some special css on this page to make it better, feel free!

### DIFF
--- a/doc/cheatsheet.mld
+++ b/doc/cheatsheet.mld
@@ -9,7 +9,7 @@ Quick reference for the odoc language!
     {th Render as}}
   {tr
     {th Paragraphs}
-    {td {[
+    {td {@text[
       A first paragraph
 
       A second paragraph]}}
@@ -19,7 +19,7 @@ Quick reference for the odoc language!
       A second paragraph}}
   {tr
     {th {{!odoc_for_authors.sections}Headings}}
-    {td {[
+    {td {@text[
       {1 Title}
       {2 Subtitle}
       {3 Subsubtitle}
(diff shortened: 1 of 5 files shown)


doc/cheatsheet.mld
```diff
@@ -0,0 +1,225 @@
...
+Quick reference for the odoc language!
+
+{table
+  {tr
+    {th }
+    {th [odoc] syntax}
+    {th Render as}}
+  {tr
+    {th Paragraphs}
+    {td {[
```
Julow, line 12: The code blocks are highlighted in a broken way. Can you add the language annotation ? `{@text[ .. ]}` works well. `{@odoc[ .. ]}` is not recognized by the highlighter and renders like verbatim.

Julow (approved):
Awesome!

doc/cheatsheet.mld
panglesd: Good point, thanks!

panglesd:
When this is merged, I think it would make sense to backport it to 2.4 branch and update the website.

dbuenzli:
@panglesd thanks for this excellent documentation addition (note that this could not have been achieved[^1] with `ocamlmark` ;-)

[^1]: Except by dropping to HTML and thus leaving other backends in the dust.

PR #1121, opened 2024-05-14
Driver

jonludlam:
This is a new driver for testing the odoc 3.0 implementation. It's based on a combination of our current mld driver and bits of ocaml-doc/voodoo. Also included are some draft changes to the CLI. These will need a bit of polish!

Requires eio for parallelism and therefore OCaml >= 5.1, and should therefore be in a different opam package. 

I'm making this PR very early as we'll be collaborating on the driver as we make the odoc 3 implementation changes.


src/driver/library.ml
panglesd, line 1: I guess this file is meant to be removed!

src/odoc/classify.ml
```diff
@@ -0,0 +1,387 @@
...
+        modules = StringSet.add cu.cu_name lib.modules;
+        intf_deps =
+          List.fold_left
+            (fun deps (cu, _) -> StringSet.add cu deps)
+            lib.intf_deps cu.cu_imports;
+        impl_deps =
+          List.fold_left
+            (fun deps id -> StringSet.add (Ident.name id) deps)
+            lib.impl_deps cu.cu_required_globals;
+      }
```
panglesd, line 53: There is some compatibility issues with OCaml `5.2` here...

The 5.2 version would be:

```suggestion
      {
        lib with
        modules =
          StringSet.add (match cu.cu_name with Compunit s -> s) lib.modules;
        intf_deps =
          List.fold_left
            (fun deps (cu, _) -> StringSet.add cu deps)
            lib.intf_deps cu.cu_imports;
        impl_deps =
          List.fold_left
            (fun deps id -> StringSet.add id deps)
            lib.impl_deps
            (List.map (fun (Compunit s) -> s) cu.cu_required_compunits);
      }
```

panglesd (commented):
Impressive! Some basic comments below.

jonludlam:
OK, green tick, in it goes.

PR #1126, opened 2024-05-21
Cleanup implementation of --parent-id

EmileTrotignon:
This cleans up the implementation of --parent-id.

It is currently possible to declare a module as your parent id, which can lead to conflicts that are silently resolved by overwritting pages.


src/odoc/compile.ml
```diff
@@ -79,7 +85,7 @@ let resolve_parent_page resolver f =
   in
   let extract_parent = function
     | { Paths.Identifier.iv = `Page _; _ } as container -> Ok container
-    | _ -> Error (`Msg "Specified parent is not a parent of this file")
+    | { Paths.Identifier.iv = `LeafPage _; _ } -> Error (`Msg "leaf page")
```
jonludlam, line 88: Does this error message ever get exposed to the user?

panglesd: No, this is a leftover from debugging. The more exhaustive matching should be kept though.

```suggestion
    | { Paths.Identifier.iv = `LeafPage _; _ } -> Error (`Msg "Specified parent is not a parent of this file")
```

jonludlam:
I'd prefer `parents_children` over `siblings` as when I first read this I assumed that `siblings` wouldn't include self. A minor nit, but worth changing I think.

panglesd:
> when I first read this I assumed that siblings wouldn't include self

Good point!
Note that before this PR, "parents_children` variables were named `children` which caused a lot of confusion with `children`, the variables for our own children :)

jonludlam:
Heh, naming is hard :-)

jonludlam:
Option.map is 4.08+. We've talked about dropping support for older OCamls, but I'd rather do this in a separate PR than just because we happen to use a newer function.

PR #1128, opened 2024-05-22
Improvements to driver: cli

jonludlam:


panglesd (approved):
Very nice! I have some questions:

- Does "[sublibraries](https://github.com/ocaml/odoc/pull/1128/files#diff-48c7de60a33306a4dc2f449ee0e837455ac9d28c3cf25ec57947a526175d9bbeR7-R8)" take a library name as input? In which case, do the opam package name and the (main) library always have the same name?
- How often do we need to call [this `init` function](https://github.com/ocaml/odoc/pull/1128/files#diff-f19144f7f0a293f8942bfb273f6eb3d701540356e36b778f4260be7771d73acfR1-R5)? It used to be called often, I changed it in 03716d1 to be called once.
- 7c4565d needs some polish...

Once these small things are fixed, I think it is ready for merge!

jonludlam:
For sublibraries - there's some confusing terminology here. ocamlfind and opam disagree with what a package is (or at least, they don't agree on the names of some packages!)

What that code does is find all the installed libraries with a common 'dot-delimited' prefix. For example, for if called with 'base' it would return the set `[base; base.base_internalhash_types; base.caml; base.md5; base.shadow_stdlib]`. Ocamlfind refers to `base` as a `package`, but this doesn't necessarily agree with what opam believes the package name to be. For example, for findlib itself, opam believes the package name is `ocamlfind`, but findlib will talk about the `findlib` package.

I'd like to be a bit tighter with the words here, and I think we should be talking about 'libraries' when talking about findlib, and 'packages' when talking about opam.

jonludlam:
In any case, these things can be fixed up as we're working on the code, so I think this can go in. Thanks for your improvements!

PR #1129, opened 2024-05-23
Make `--source-id` optional when compiling implementations

panglesd:
`--source-id` is only needed when rendering implementations. However, compiling implementations can also be required for another feature: counting occurrences.

This PR makes `--source-id` option at compile time, but it remains mandatory to render source code.


jonludlam:
Thanks!

PR #1132, opened 2024-05-23
Odoc 3: CLI for `-L` and `-P`

panglesd:
This includes:
- The `-P` and `-L` CLI options,
- Several checks:
  - that no directory is contained another,
  - that the output file is contained in a package (this is disabled in case no `-P`, `-L` is passed)
- (Lazily) storing file paths for each package in a hashtable
- Functions for finding files them by name, or by path.
- Tests for the checks
- A test for finding by name, using a hack.

The next stage is to implement new reference syntax, and integrate those "find by name" and "find by path" functions into the resolution of references.


src/odoc/resolver.ml
```diff
@@ -194,15 +265,46 @@ let lookup_unit ~important_digests ~imports_map ap target_name =
...
+              | Some (p, _) -> Some p
+              | None ->
+                  failwith
+                    ("Page not found by name: "
+                    ^ string_of_int (List.length units)))
+          | Error _ -> failwith "Not found by name")
+      | [] -> assert false
+      | "" :: package :: path ->
+          let _res = Named_roots.find_by_path ~package ~path pages in
+          failwith "TODO"
```
panglesd, line 296: I'll fix the TODO soon...

src/odoc/bin/main.ml
```diff
@@ -574,15 +655,38 @@ end = struct
       & opt (some string) None
       & info ~docs ~docv:"PATH.odocl" ~doc [ "o" ])
 
+  let page_pkgnames =
+    let doc =
+      "Specifies a directory PATH containing pages that can be referenced by \
+       {!/pkgname} during linking. A pkgname can be specified in the -P \
```
EmileTrotignon, line 661: What does "during linking" mean here ? `{!/pkgname}` looks like odoc syntax, which is used during the writing of the doc, not linking.

panglesd: Indeed, they are referenced during the writing of the doc. The reference is resolved at link time, but this might not be relevant.
I'll remove the "during linking".

src/odoc/resolver.mli
```diff
@@ -21,10 +21,17 @@
...
+  page_pkgnames : (string * Fs.Directory.t) list;
+  lib_pkgnames : (string * Fs.Directory.t) list;
+  current_pkg : string;
+}
+
 val create :
   important_digests:bool ->
   directories:Fs.Directory.t list ->
   open_modules:string list ->
+  roots:roots option ->
```
EmileTrotignon, line 34: I think docs should be updated to explain the roots options here

src/driver/odoc.ml
```diff
@@ -81,18 +81,33 @@ let compile_impl ~output_dir ~input_file:file ~includes ~parent_id ~source_id =
...
+        let s = Format.asprintf "%s:%a" pkgname Fpath.pp path in
+        v "-L" % s %% acc)
+      Cmd.empty libs
+  in
   let cmd =
-    odoc % "link" % p file % "-o" % p output_file %% includes
+    odoc % "link" % p file % "-o" % p output_file %% includes %% docs %% libs
     % "--enable-missing-root-warning"
   in
+  Logs.debug (fun m -> m "AAAAAAAA %a" pp cmd);
```
EmileTrotignon, line 110: Is this supposed to be here ?

panglesd: Yes, definitely obviously supposed to be there.

src/odoc/resolver.mli
panglesd: Rereading this part, I noticed that some of my terminology might be wrong: for the `-P` it is the package name that is relevant, but for the `-L` it is the library name.

EmileTrotignon (commented):
I think there is no test for -L, is there a reason ?

PR #1133, opened 2024-05-23
Remove unwanted or duplicate doc in expansions

Julow:
A module's `doc` field was propagated through expansions by synthetizing a comment at the top of the expanded signature. It's now propagated via the `doc` field of the signature.

In the case where the module has a comment and the signature has a top-comment, the module's comment is not propagated.

This fixes doc appearing multiple time when the signature is expanded mutliple time (eg. due to an include-module-type-of) and doc appearing when they shouldn't (eg. in a non-inline include).
This also moves the preambles into the right section in the generated HTML.

### DIFF
--- a/src/xref2/tools.ml
+++ b/src/xref2/tools.ml
@@ -1579,7 +1579,7 @@ and expansion_of_module_path :
           let sg' =
             match m.doc with
             | [] -> sg
-            | docs -> { sg with items = Comment (`Docs docs) :: sg.items }
+            | doc -> { sg with doc }
           in
           if strengthen then
             Ok (Signature (Strengthen.signature (`Resolved p') sg'))
--- a/src/xref2/tools.ml
+++ b/src/xref2/tools.ml
@@ -1576,14 +1576,10 @@ and expansion_of_module_path :
       in
       expansion_of_module_cached env p' m >>= function
       | Signature sg ->
-          let sg' =
-            match m.doc with
-            | [] -> sg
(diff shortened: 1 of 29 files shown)


src/xref2/tools.ml
```diff
@@ -1579,7 +1579,7 @@ and expansion_of_module_path :
           let sg' =
             match m.doc with
             | [] -> sg
-            | docs -> { sg with items = Comment (`Docs docs) :: sg.items }
+            | doc -> { sg with doc }
```
panglesd, line 1582: I'm still really not familiar with this part of the code...
Could this erase some content from `sg`? (Could `{ sg with doc = doc @ sg.doc }` be safer?)
And why is this done only in `expansion_of_module_path`, and not in `expansion_of_module`, since this function already has `m`?

Julow: It's overriding the doc for this use case:
```
module P = struct
  (** Preamble for P. *)
end

(** Outside preamble for Q. *)
module Q = struct
  (** Inside preamble for Q. *)
end
```

We want to see the preamble from the first comment in `P` but we want the comment outside of `Q`. Concatenating them would result in the same bug this patch is trying to fix.

Moving that to `expansion_of_module` makes complete sense.

panglesd: Thanks for the explanation. This makes sense!

panglesd (approved):

jonludlam:
Thanks!

PR #1142, opened 2024-06-07
Parsing of path-references to pages and modules

Julow:
Add the parsing of references to pages and modules containing a slash: `{!./foo}`, `{!/foo}`, `{!//foo}`. 
These are not resolved yet.

### DIFF
--- a/src/model/paths_types.ml
+++ b/src/model/paths_types.ml
@@ -553,6 +553,11 @@ module rec Reference : sig
 
   type tag_only_child_module = [ `TChildModule ]
 
+  type tag_page_path =
+    [ `TRelativePath (* {!identifier/} *)
+    | `TAbsolutePath (* {!/identifier} *)
+    | `TCurrentPackage (* {!//identifier} *) ]
+
   type tag_any =
     [ `TModule
     | `TModuleType
@@ -592,10 +597,16 @@ module rec Reference : sig
     | `TChildPage
     | `TChildModule ]
 
+  type page_path =
+    [ `Root of string * tag_page_path
+    | `Slash of page_path * string (* {!page_path/identifier} *) ]
+  (** @canonical Odoc_model.Paths.Reference.PagePath.t *)
+
   type signature =
     [ `Resolved of Resolved_reference.signature
     | `Root of string * tag_signature
     | `Dot of label_parent * string
+    | `Page_path of page_path
     | `Module of signature * ModuleName.t
     | `ModuleType of signature * ModuleTypeName.t ]
   (** @canonical Odoc_model.Paths.Reference.Signature.t *)
@@ -629,6 +640,7 @@ module rec Reference : sig
     [ `Resolved of Resolved_reference.label_parent
     | `Root of string * tag_label_parent
     | `Dot of label_parent * string
+    | `Page_path of page_path
     | `Module of signature * ModuleName.t
     | `ModuleType of signature * ModuleTypeName.t
     | `Class of signature * ClassName.t
@@ -743,13 +755,15 @@ module rec Reference : sig
   type page =
     [ `Resolved of Resolved_reference.page
     | `Root of string * [ `TPage | `TUnknown ]
-    | `Dot of label_parent * string ]
+    | `Dot of label_parent * string
+    | `Page_path of page_path ]
   (** @canonical Odoc_model.Paths.Reference.Page.t *)
 
   type any =
     [ `Resolved of Resolved_reference.any
     | `Root of string * tag_any
     | `Dot of label_parent * string
+    | `Page_path of page_path
     | `Module of signature * ModuleName.t
     | `ModuleType of signature * ModuleTypeName.t
     | `Type of signature * TypeName.t
--- a/src/model/paths_types.ml
+++ b/src/model/paths_types.ml
@@ -553,7 +553,7 @@ module rec Reference : sig
 
   type tag_only_child_module = [ `TChildModule ]
 
-  type tag_path =
+  type tag_hierarchy =
     [ `TRelativePath (* {!identifier/} *)
     | `TAbsolutePath (* {!/identifier} *)
     | `TCurrentPackage (* {!//identifier} *) ]
@@ -597,14 +597,14 @@ module rec Reference : sig
     | `TChildPage
     | `TChildModule ]
 
-  type path = tag_path * string list
-  (** @canonical Odoc_model.Paths.Reference.Path.t *)
+  type hierarchy = tag_hierarchy * string list
+  (** @canonical Odoc_model.Paths.Reference.Hierarchy.t *)
 
   type signature =
     [ `Resolved of Resolved_reference.signature
     | `Root of string * tag_signature
     | `Dot of label_parent * string
-    | `Module_path of path
+    | `Module_path of hierarchy
     | `Module of signature * ModuleName.t
     | `ModuleType of signature * ModuleTypeName.t ]
   (** @canonical Odoc_model.Paths.Reference.Signature.t *)
@@ -629,7 +629,7 @@ module rec Reference : sig
     [ `Resolved of Resolved_reference.field_parent
     | `Root of string * tag_parent
     | `Dot of label_parent * string
-    | `Module_path of path
+    | `Module_path of hierarchy
     | `Module of signature * ModuleName.t
     | `ModuleType of signature * ModuleTypeName.t
     | `Type of signature * TypeName.t ]
@@ -639,9 +639,9 @@ module rec Reference : sig
     [ `Resolved of Resolved_reference.label_parent
     | `Root of string * tag_label_parent
     | `Dot of label_parent * string
-    | `Page_path of path
-    | `Module_path of path
-    | `Any_path of path
+    | `Page_path of hierarchy
+    | `Module_path of hierarchy
+    | `Any_path of hierarchy
     | `Module of signature * ModuleName.t
     | `ModuleType of signature * ModuleTypeName.t
     | `Class of signature * ClassName.t
@@ -653,7 +653,7 @@ module rec Reference : sig
     [ `Resolved of Resolved_reference.module_
     | `Root of string * [ `TModule | `TUnknown ]
     | `Dot of label_parent * string
-    | `Module_path of path
+    | `Module_path of hierarchy
     | `Module of signature * ModuleName.t ]
   (** @canonical Odoc_model.Paths.Reference.Module.t *)
 
@@ -758,16 +758,16 @@ module rec Reference : sig
     [ `Resolved of Resolved_reference.page
     | `Root of string * [ `TPage | `TUnknown ]
     | `Dot of label_parent * string
-    | `Page_path of path ]
+    | `Page_path of hierarchy ]
   (** @canonical Odoc_model.Paths.Reference.Page.t *)
 
   type any =
     [ `Resolved of Resolved_reference.any
     | `Root of string * tag_any
     | `Dot of label_parent * string
-    | `Page_path of path
-    | `Module_path of path
-    | `Any_path of path
+    | `Page_path of hierarchy
+    | `Module_path of hierarchy
+    | `Any_path of hierarchy
     | `Module of signature * ModuleName.t
     | `ModuleType of signature * ModuleTypeName.t
     | `Type of signature * TypeName.t
--- a/src/model/paths_types.ml
+++ b/src/model/paths_types.ml
@@ -597,9 +597,7 @@ module rec Reference : sig
     | `TChildPage
     | `TChildModule ]
 
-  type page_path =
-    [ `Root of string * tag_page_path
-    | `Slash of page_path * string (* {!page_path/identifier} *) ]
+  type page_path = tag_page_path * string list
   (** @canonical Odoc_model.Paths.Reference.PagePath.t *)
 
   type signature =
--- a/src/model/paths_types.ml
+++ b/src/model/paths_types.ml
@@ -553,7 +553,7 @@ module rec Reference : sig
 
   type tag_only_child_module = [ `TChildModule ]
 
-  type tag_page_path =
+  type tag_path =
     [ `TRelativePath (* {!identifier/} *)
     | `TAbsolutePath (* {!/identifier} *)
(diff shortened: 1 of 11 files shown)


panglesd:
Are `{!foo/bar}` also parsed? (they are not put as examples!

Julow:
Yes! As well as many variations like `{!foo/page-bar}` and `{!/foo/bar.label}`.

src/model/paths_types.ml
```diff
@@ -592,10 +597,16 @@ module rec Reference : sig
     | `TChildPage
     | `TChildModule ]
 
+  type page_path =
```
panglesd, line 600: nitpick on the name: I think it is a bit misleading, those kind of references can also be used to reference modules. From the [RFC](https://github.com/ocaml/odoc/discussions/1097):

> A module can be added to the scope by adding the library it's found in, which will make those modules accessible by the references `{!Module}` or **`{!/libname.sublib/Module}`** (not `{!sublib/Module}`). By default all modules in the package's libraries are added to the scope.

src/model/paths_types.ml
```diff
@@ -592,10 +597,16 @@ module rec Reference : sig
...
+    [ `Root of string * tag_page_path
+    | `Slash of page_path * string (* {!page_path/identifier} *) ]
```
panglesd, line 602: I don't think `identifier` should be a string: We agreed that it should be parsed as a page reference, so we can write:

```
{!page_path/page-pagename.section-sectionname}
```

src/model/paths_types.ml
panglesd: Let me be constructive and suggest something! Maybe `hierarchical_path`?

src/model/paths_types.ml
panglesd: Ah in fact maybe it should be a string, but the identifier part should be present somewhere else. (see next comment!)

src/model/paths_types.ml
Julow: True. What do you think of `path` ?

panglesd: I am slightly reluctant (we already have `Paths`, `Paths.Path`, references also technically are paths, ...).

Let me propose `hierarchy`. I think it is a bit more explicit, and avoids restating a third time that it is a path.

src/model/paths_types.ml
Julow: Indeed, there's a problem here. I think this is related to the lack of a `Module_path` constructor. The prefix can only appear on the last component of the path and the intended representation is more like this:
```
Label (Page_path (Slash (Root ("page_path", relative), "pagename")), "sectionname")
```

The result for `{!page_path/page-pagename}` is as intended but there's two problems:
- This reference with a label do not parse.
- No representation for `path/module-Bar`.

I'll look into it.

Julow:
Rebased and ready for an other review. https://github.com/ocaml/odoc/pull/1150 is rebased on top.

src/model/reference.ml
```diff
@@ -170,43 +177,95 @@ let tokenize location s : token list =
...
+    | `End_in_slash, [] ->
+        (* {!identifier'/identifier} *)
+        (`TRelativePath, next_token.identifier :: components)
+    | `End_in_slash, next_token' :: tokens' ->
+        (* {!path/identifier} *)
+        path (next_token.identifier :: components) next_token' tokens'
+    | (`None | `Prefixed _), _ ->
+        (* This is not really expected *)
+        expected ~expect_paths:true [] next_token.location
+        |> Error.raise_exception
```
panglesd, line 245: I think this is impossible, and should be an `assert false`. This also makes `expected` never called with an empty list, and we can put back the `"or "` which makes the error more readable.

panglesd: One way to avoid having this assert false would be to have the `` `End_in_slash `` token contain the whole path, and parse it later.

src/model/reference.ml
```diff
@@ -170,43 +177,95 @@ let tokenize location s : token list =
...
+  in
+
   let rec signature { kind; identifier; location } tokens : Signature.t =
     let kind = match_reference_kind location kind in
     match tokens with
     | [] -> (
         match kind with
         | (`TUnknown | `TModule | `TModuleType) as kind ->
             `Root (identifier, kind)
+        | `TPathComponent -> `Module_path (`TRelativePath, [ identifier ])
```
panglesd, line 259: Why is that a relative path, even if `identifier = ""` ?
I think module paths must always be of this form:

`{!/libname/module-M.type-t}`

src/model/paths_types.ml
```diff
@@ -620,6 +629,7 @@ module rec Reference : sig
     [ `Resolved of Resolved_reference.field_parent
     | `Root of string * tag_parent
     | `Dot of label_parent * string
+    | `Module_path of hierarchy
```
panglesd, line 632: Module path can only be of the form: `/libname/`. For example: `{!/libname/module-M.type-t}`.

I think that could be reflected in the type?

src/model/reference.ml
```diff
@@ -269,19 +337,34 @@ let parse whole_reference_location s :
...
+  let rec label_parent ({ kind; identifier; location } as token) tokens :
+      LabelParent.t =
     let kind = match_reference_kind location kind in
     match tokens with
     | [] -> (
         match kind with
         | ( `TUnknown | `TModule | `TModuleType | `TType | `TClass | `TClassType
           | `TPage ) as kind ->
             `Root (identifier, kind)
+        | `TPathComponent -> `Page_path (`TRelativePath, [ identifier ])
```
panglesd, line 360: Why is it a relative path even when `identifier` is `""`?

src/model/reference.ml
```diff
@@ -269,19 +337,34 @@ let parse whole_reference_location s :
         )
   in
 
-  let rec label_parent { kind; identifier; location } tokens : LabelParent.t =
+  let any_page { identifier; location; _ } kind next_token tokens =
```
panglesd, line 340: Shouldn't that be named `any_path`?

src/model/reference.ml
```diff
@@ -327,7 +414,12 @@ let parse whole_reference_location s :
     in
 
     match tokens with
-    | [] -> `Root (identifier, kind)
+    | [] -> (
+        match kind with
+        | #Paths.Reference.tag_any as kind -> `Root (identifier, kind)
+        | `TPathComponent -> `Page_path (`TRelativePath, [ identifier ]))
```
panglesd, line 420: Do I understand correctly that this can only happen in case of `/` ending references, such as `{!a/}` or `{!/}`? If so, I think that those references are currently not allowed.

If we allow them, shouldn't it be `` `Any_path ``? and what about `identifier = ""` ? :)

src/model/reference.ml
```diff
@@ -378,13 +470,21 @@ let parse whole_reference_location s :
...
+                    ~in_what:"on the right side of a dot" ~suggestion location
+                  |> Error.raise_exception
             in
-            not_allowed ~what:"Page label"
-              ~in_what:"the last component of a reference path" ~suggestion
-              location
-            |> Error.raise_exception)
+            (* Prefixed pages are not differentiated. *)
+            `Page_path (path [ identifier ] next_token tokens)
+        | `TPathComponent -> `Page_path (path [ identifier ] next_token tokens))
```
panglesd, line 487: Same:
> Do I understand correctly that this can only happen in case of `/` ending references, such as `{!/a/b/}`? If so, I think that those references are currently not allowed.
>
> If we allow them, shouldn't it be `` `Any_path ``? :)

test/xref2/refs/refs.md
```diff
@@ -2629,12 +2629,12 @@ val resolve_ref : string -> ref = <fun>
    m)
 # resolve_ref "type-c.method-m" ;;
 Exception:
-Failure "Expected 'class-', 'class-type-', or an unqualified reference.".
+Failure "Expected 'class-', 'class-type-', an unqualified reference.".
```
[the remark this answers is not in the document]
Julow: The printing of this error message is changed to allow adding "a path, " when appropriate, and I think I forgot to add the "or" case back. Fixed now.

src/model/paths_types.ml
```diff
@@ -553,6 +553,11 @@ module rec Reference : sig
 
   type tag_only_child_module = [ `TChildModule ]
 
+  type tag_hierarchy =
+    [ `TRelativePath (* {!identifier/} *)
+    | `TAbsolutePath (* {!/identifier} *)
+    | `TCurrentPackage (* {!//identifier} *) ]
```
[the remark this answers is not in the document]
Julow: A canonical tag would be nice indeed. I thought doc-comments were not allowed on polymorphic variants with older compilers but apparently that parses fine.

Don't look at odoc's doc please.

src/model/reference.ml
```diff
@@ -90,21 +92,22 @@ let match_extra_odoc_reference_kind (_location as loc) s :
 
    A secondary reason to delay parsing, and store strings in the token list, is
    that we need the strings for user-friendly error reporting. *)
-let match_reference_kind location s : Paths.Reference.tag_any =
+let match_reference_kind location s : reference_kind =
   match s with
-  | None -> `TUnknown
-  | Some s -> (
+  | `None -> `TUnknown
+  | `Prefixed s | `Old_prefix s -> (
```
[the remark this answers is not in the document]
Julow: Some places have `` `Prefixed `` and not `` `Old_prefix `` but it's otherwise not necessary. Old prefixes could just be passed in as new prefixes.
It's just self-documentation added to the old code.

src/model/reference.ml
```diff
@@ -170,43 +177,95 @@ let tokenize location s : token list =
...
   in
   expected_err allowed location
 
+(* Parse references that do not contain a [/]. Raises errors and warnings. *)
 let parse whole_reference_location s :
     Paths.Reference.t Error.with_errors_and_warnings =
   let open Paths.Reference in
   let open Names in
+  let rec path components next_token tokens : Hierarchy.t =
+    match (next_token.kind, tokens) with
```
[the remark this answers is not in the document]
Julow: No because that adds the `[]` case, which a lot of effort was made to avoid it. Also, unpacking the kind from that is too much syntax overhead.

src/model/reference.ml
```diff
@@ -170,43 +177,95 @@ let tokenize location s : token list =
...
+        (* {!./identifier} *)
+        (`TRelativePath, components)
+    | `End_in_slash, [ { kind = `End_in_slash; identifier = ""; _ } ]
+      when next_token.identifier = "" ->
+        (* {!//identifier} *)
+        (`TCurrentPackage, components)
+    | `End_in_slash, [] ->
+        (* {!identifier'/identifier} *)
+        (`TRelativePath, next_token.identifier :: components)
+    | `End_in_slash, next_token' :: tokens' ->
```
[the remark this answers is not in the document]
Julow: Good catch! Added the error.

src/model/reference.ml
Julow: I want to avoid adding structure in the lexer as it's manually written and quite complicated. `expected` with an empty list still gives a valid error message but I agree, let's not put specific error messages in unreachable paths.

src/model/reference.ml
Julow: Yes!

src/model/reference.ml
Julow: Indeed that's not allowed by the lexer. I'll put `assert false` here.

src/model/reference.ml
Julow: Same fix.

src/model/reference.ml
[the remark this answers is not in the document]
panglesd: I think that then `match next_token, tokens with ...` is a little bit more clear. I find those:
```
| `End_in_slash, [] when next_token.identifier = "." ->
```
hard to read, since it's hard to remember that `` `End_in_slash `` and `next_token` are linked at this point 
Those cases amounts to 3 out of 6, but we are likely to add more of then.

```
| {kind = `End_in_slash ; identifier = "" ; _ }, [] ->
```
Feels easier to read to me.

(I will obviously not die on this hill, if you are not convinced just resolve the conversation!)

panglesd (commented):
I finished reviewing, and have no more comments than the one already made!

Once they are resolved, it's ready to be merged.

src/model/paths_types.ml
panglesd: I guess that https://github.com/ocaml/odoc/pull/1142#discussion_r1672366269 answers no to the question!

src/model/reference.ml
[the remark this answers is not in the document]
Julow: Good idea, I implemented that.

src/model/reference.ml
panglesd: I do think that indeed, `{!/M.type-t}` would yield a `` `Type( ...( `Module_path (`TRelativePath, [""])...) `` !

Julow: Sorry this second sentence is answering an other comment. `identifier` cannot be `""` here and there's now an `assert false` in this case.

`{!/M.type-t}` parses to `` "`Type":[{"`Module_path":["`TAbsolutePath",["M"]]},"t"] ``.

src/model/paths_types.ml
Julow: I think this can happen with `{!/M.type-t}`. I'm not in favor of enforcing that in the parser. The resolver has to match on the path to be able to lookup `libname`, that's where the reference should be rejected.

src/model/reference.ml
Julow: Case removed, see the other comment.

src/model/reference.ml
panglesd: It took me a while but yes, I'm convinced this is indeed an `assert false`! Thanks for the explanation.

Tricky parser, good luck to the maintainers.

src/model/reference.ml
panglesd: Hmm it seems the `assert false` in this case was forgotten!

Julow: That's right! In fact none of the `` `TPathComponent `` cases are reachable, path components can only be parsed by the `path` function.

Issue #1111, opened 2024-04-16
Mind the gap


polytypic:
So, if I have this in a .mli:

```ocaml
(** Framework for interoperable effects based concurrency.

    ℹ️ Picos, i.e. this module, is not intended to be an application level
    concurrent programming library or framework.  If you are looking for a
    library or framework for programming concurrent applications, then this
    module is probably not what you are looking for.

    {1 Understanding cancelation}

    A central idea of Picos is to provide a collection of building blocks for
    parallelism safe cancelation.  Consider the following characteristic
    example:
```

The layout will have a (relatively, to my eyes) absolutely massive (some people have said I have a tendency to exaggerate) gap above the first heading (i.e. between preamble and content):
 
![image](https://github.com/ocaml/odoc/assets/5804945/d7e14bee-a5e8-4cc4-935b-016f1b3b64a9)

I mentioned this in a slack and it was suggested that this gap might be a bug introduced on some redesign of the CSS.  In the first couple of screenshots of [this PR](https://github.com/ocaml/odoc/pull/999) the screenshots do not have the gap, for example.

At any rate, I don't know whether the gap is intentional design or not, but I do personally find the gap distracting.  I'd like to structure the content of documentation with headings without having the gap between the preamble and the content.

PR #1143, opened 2024-06-07
Less gaps in css

EmileTrotignon:
Reduces gaps in css.

Answers issue #1111.

### DIFF
--- a/src/html_support_files/odoc.css
+++ b/src/html_support_files/odoc.css
@@ -300,7 +300,7 @@ body.odoc {
   display: grid;
   grid-template-columns: min-content 1fr;
   column-gap: 4ex;
-  row-gap: 2ex;
+  row-gap: ex;
 }
 
 body.odoc-src {
@@ -316,10 +316,6 @@ body.odoc-src {
   /* This make the first thing in the preamble align with the sidebar */
   padding-top: 0;
   margin-top: 0;
-}
-
-header {
-  margin-bottom: 30px;
 }
 
 header.odoc-preamble {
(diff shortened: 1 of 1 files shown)


EmileTrotignon:
Here is a screenshot of an example page : 
image

panglesd (commented):
We don't really need less gaps in CSS, we need to remove the gap introduced in #999 by the introduction of the grid system, which disabled [margin collapse](https://www.joshwcomeau.com/css/rules-of-margin-collapse/):

> the margins of adjacent grid items do not collapse

(From https://www.w3.org/TR/css-grid-1/#item-margins)

The solution you propose changes many margins/paddings (making the result slightly not spaced enough in my opinion, but that's maybe just habits of the previous style) ~~while not solving the original problem~~! Edit: it does solve it by always having the top margin equal to zero... I don't think that having all elements have a 0 value for margin-top is a valid assumption to make!

We can solve the problem either by going away from the grid system, or modifying the html layout: grouping `odoc-preamble` and `odoc-content` in a single element.
I would prefer the second option. It would not affect ocaml.org since they are using the `--as-json` option which would be unaffected by this change. However, we should be careful to check if the layout change affects `odig` stylesheet.

EmileTrotignon:
Oh ok, I did not know about margin collapse, I just git blame on the line that set the topmost margin and found it was quite old, so I thought this was not introduced recently.

I am a bit anxious about changing the order of elements, but I certainly do not want to drop grid.
Another way to do would be to have a rule like :
```css
.odoc-content:first-child {
  margin-top:0 !important;
}
```

EmileTrotignon:
I am also not i agree that assuming every element not to have a top-margin is a bad idea, it seems quite reasonable to me, its a convention that works well, and if a margin-top slips in its a very easy patch to remove it.

PR #1145, opened 2024-06-07
"Global" Sidebar

panglesd:
This is a prototype for a "global" sidebar:
- CLI
  - a new command `odoc sidebar` which takes `-L`s and `-P`s and outputs a marshalled file.
  - a new `--sidebar` argument to `odoc html-generate` to pass the marshalled file and generate a sidebar
 - Rendering: currently, renders a new right side bar with the list of pages, and units (sorted by libraries). This is a first prototype and needs to be designed, but I think that is already helpful for discoverability of pages!

The content and rendering of the marshalled sidebar can be changed with odoc versions, without altering the CLI.

Here is an example of a rendering of sidebars, from packages using mld pages: [ppxlib](https://choum.net/panglesd/odoc_with_sidebar/_html/ppxlib/doc/index.html), [brr](https://choum.net/panglesd/odoc_with_sidebar/_html/brr/doc/index.html).

I'm marking this as draft as I'd like to do a first round of autoreview, but the main functionality should be ready for review.


dbuenzli:
I'm not sure I'm really convinced by these global side bars. I think a trip to the landing page is often better (especially since you get context that is totally lost in the sidebars see e.g. brr's landing page) 

That being said the library listings in `brr` would be better sorted alphabetically rather than the seemingly random order I see here.

Also on mid narrow pages you want to collapse the layout: 

```
| nav       | toc-local  |
| preamble  | toc-global | 
| content   |            |
```

and on mobile something like: 

```
| nav        |
| preamble   | 
| toc-local  |
| content    |
| toc-global |
```
Currently on my screen size (editor and web browser side-by-side) your example makes the content unreadable: 

Screenshot 2024-06-10 at 17 03 28

panglesd:
Thanks for the suggestion!

> Also on mid narrow pages you want to collapse the layout (if not always for a less busy layout):
> ```
> | nav       | toc-local  |
> | preamble  | toc-global | 
> | content   |            |
> ```


I'm not that familiar with the grid system, but I'm not sure how to do that without introducing a gap somewhere.

```
  grid-template-areas:
    "nav      toc-local"
    "preamble toc-global"
    "content  toc-global"
```
will force `preamble` to start aligned with `toc-global,` creating a gap between `nav` and `preamble` when `toc-local` has big height.

```
  grid-template-areas:
    "nav      toc-local"
    "preamble toc-local"
    "content  toc-global"
```
Similarly, this will force `toc-global` to start aligned with `content`, creating a gap between `toc-local` and `toc-global` when preamble has big height.

I'll try to find a solution, without changing the html structure ;)

(maybe 
```
| nav       |            |
| preamble  | toc-global | 
| toc-local | toc-global |
| content   | toc-global |
```
?
)

dbuenzli:
You are right that it's problematic. Personally I'm not a very big fan of the three column layout: usually works horribly for small screens[^1].

Basically to have the two column layout you'd need:

```html
<div>
   <div class="nav"></div>
   <div class="preamble"></div>
   <div class="tocs">
     <div class="toc-page"></div>
     <div class="toc-pkg"></div>
   </div>
   <div class="content"></div>
</div>
```

but then you can't have the single column layout where you put the `toc-pkg` at the end of the page which I would find the most desirable. Perhaps we are still in a CSS blind spot here or there is still some kind of CSS++ I don't know about yet (and would likely prefer not to know about :-)

Basically you'd like to be able to assign two divs to the same area in the grid and then have them layout like say a flex-box. But when you do that they overlap. See for example the fiddle [here](https://jsfiddle.net/u950dknx/27/) that (unsucessfully) tries to accomodate the three layouts; move the width of the rendering pane to see the responses.

[^1]: I'm actually not sure it works well for large screens either: it seems to me you get a lot of mouse travel, I prefer if all navigation is located in the same area.

Regarding:

> (maybe
> 
> ```
> | nav       |            |
> | preamble  | toc-global | 
> | toc-local | toc-global |
> | content   | toc-global |
> ```

that would be extremely annoying. I don't want to have to scroll through the page toc to access page content if my screen is large enough.

dbuenzli:
> there is still some kind of CSS++ I don't know about yet (and would likely prefer not to know about :-)

Of course there is… I asked on [stackoverflow](https://stackoverflow.com/questions/78608060/how-can-i-avoid-overlap-when-assigning-two-elements-the-same-css-grid-area/78608456#78608456). So the answer is to devise the markup as follows:
```html
<div>
   <div class="nav"></div>
   <div class="preamble"></div>
   <div class="tocs">
     <div class="toc-page"></div>
     <div class="toc-pkg"></div>
   </div>
   <div class="content"></div>
</div>
```
and make judicious use of the `display:content` CSS property see [this fiddle](https://jsfiddle.net/gmuyt7jb/2). Basically using `display:content` on `tocs` allows bring its children in the toplevel `div` when you want to (namely in the three and single column layouts).

panglesd:
Thanks a lot for the solution!

(Note: Maybe we should have originally put the preamble and content in the same div, as in https://github.com/ocaml/odoc/pull/1146, and relied on this trick to allow both the grid layout and the non-grid one where margin collapse can happen. But yes, let's avoid changing the HTML layout if not absolutely required).

dbuenzli:
> Maybe we should have originally put the preamble and content in the same div

No :-) It's better to show a synopsis of the content immediately on a page rather than a wall of navigation to scroll through. In fact what we do here is what wikipedia used to do before they switched to clicky tocs and/or details on mobile.

So I find the `nav, preamble, page-toc, content` very natural and any order based on margin collapse arguments rather dubious…

src/document/generator.ml
```diff
@@ -1783,26 +1799,152 @@ module Make (Syntax : SYNTAX) = struct
...
+          v.libraries
+      in
+      let units = block (Block.List (Block.Unordered, units)) in
+      let units = [ title "Libraries"; units ] in
+      page_hierarchies @ units
+
+    let compilation_unit ?sidebar:sb (t : Odoc_model.Lang.Compilation_unit.t) =
       let url = Url.Path.from_identifier t.id in
+      let sidebar =
+        match sb with None -> None | Some sb -> Some ((sidebar t.id) sb)
```
Julow, line 1927: I think these parentheses are not doing anything. (Same below)

src/document/generator.ml
```diff
@@ -1357,9 +1363,10 @@ module Make (Syntax : SYNTAX) = struct
       Item.Declaration { attr; anchor; doc; content; source_anchor }
 
     and simple_expansion :
```
Julow, line 1365: This function should probably do something to the sidebar ? Currently, entering a sub-module do not change the sidebar. The current module isn't visible in the sidebar, which still highlight the parent module.

src/document/generator.ml
```diff
@@ -1783,26 +1799,152 @@ module Make (Syntax : SYNTAX) = struct
       in
       List.map f t
 
-    let compilation_unit (t : Odoc_model.Lang.Compilation_unit.t) =
+    module Hierarchy = struct
```
Julow, line 1802: The `dir` type is quite complex and probably leaks out of this module. What do you think of abstracting it behind a signature ?

src/document/generator.ml
```diff
@@ -1783,26 +1799,152 @@ module Make (Syntax : SYNTAX) = struct
...
+          | Error _ -> None
+          | Ok href ->
+              let target = InternalLink.Resolved href in
+              let content = [ inline @@ Text name ] in
+              let link = { InternalLink.target; content; tooltip = None } in
+              let attr = if is_highlighted then [ "current_unit" ] else [] in
+              let elem =
+                [ block (Inline [ inline ~attr @@ Inline.InternalLink link ]) ]
+              in
+              Some elem
```
Julow, line 1912: Pages contain almost the same code.

src/document/generator.ml
```diff
@@ -1783,26 +1799,152 @@ module Make (Syntax : SYNTAX) = struct
...
+      let units =
+        let item id =
+          let id = (id :> Paths.Identifier.t) in
+          let href, name, is_highlighted =
+            ( Url.from_identifier ~stop_before:false id,
+              Paths.Identifier.name id,
+              Odoc_model.Paths.Identifier.equal id root_id )
+          in
+          match href with
+          | Error _ -> None
```
Julow, line 1903: I think `Url.from_identifier` cannot return `Error` when `~stop_before:false`. We should expose this case with a new function in `Url`.

src/document/generator.ml
```diff
@@ -1783,26 +1799,152 @@ module Make (Syntax : SYNTAX) = struct
...
+                let content = Comment.link_content content in
+                (* let attr = if is_highlighted then [ "current_unit" ] else [] in *)
+                block (Inline content)
+          in
+
+          let pages = Hierarchy.to_sidebar convert hierarchy in
+          let pages = [ block (Block.List (Block.Unordered, [ pages ])) ] in
+          [ title @@ page_name ^ "'s Pages" ] @ pages
+      in
+      let rec concat_map acc f = function
```
Julow, line 1889: You've defined `concat_map` for the 6th time and `filter_map` for the 5th time in this PR.
We have 4 `Utils` modules in the projects, I think it's time for a unified "utils" library.

src/html/html_page.ml
```diff
@@ -41,10 +41,27 @@ let html_of_search () =
...
+              Html.nav
+                ~a:[ Html.a_class [ "odoc-toc"; "odoc-local-toc" ] ]
+                (html_of_toc toc);
+            ])
+  | Some c ->
+      sidebar
+        [
+          Html.nav
+            ~a:[ Html.a_class [ "odoc-toc"; "odoc-local-toc" ] ]
+            (html_of_toc toc);
```
Julow, line 62: The local toc is identical in both branch.

Julow (approved):
Looks good. I've left many minor comments but feel free to ignore them.

src/document/generator.ml
panglesd: I'm too used to ocamlformat being great at eliminating useless parentheses!

src/document/generator.ml
panglesd: Yes, but I propose to keep the current PR with the minimalest global sidebar: only toplevel entries.

It is very likely that we want to expand the sidebar, but I think that belongs to another PR (which will likely contain a lot of discussion).

src/document/generator.ml
panglesd: I fixed that in df4a803 by going through paths: `id |> Url.Path.from_identifier |> Url.from_path`.

EmileTrotignon:
Could you force an extension for sidebar artifacts ? I think `.odoc-sidebar` would be appropriate.

src/html/html_page.ml
```diff
@@ -225,10 +242,12 @@ let search_urls = %s;
   in
   content
 
-let make ~config ~url ~header ~breadcrumbs ~toc ~uses_katex content children =
+let make ~config ~url ~header ~breadcrumbs ~sidebar ~toc ~uses_katex content
+    children =
   let filename = Link.Path.as_filename ~is_flat:(Config.flat config) url in
   let content =
-    page_creator ~config ~url ~uses_katex header breadcrumbs toc content
+    page_creator ~config ~url ~uses_katex ~sb:sidebar header breadcrumbs toc
```
EmileTrotignon, line 249: Rest of the time this is sidebar, could this be made more consistent ?

EmileTrotignon (commented):
Is the driver parallel and everything ?

src/html_support_files/odoc.css
```diff
@@ -1193,21 +1210,53 @@ td.def-doc *:first-child {
 
 /* Mobile adjustements. */
 
+@media only screen and (max-width: 210ex) {
```
EmileTrotignon, line 1213: Maybe this could be 160ex because this is the size of the body in line 299. Maybe even make it into a variable ?

panglesd:
Note to self: 
- [ ] Need to fix the json output as well
- [ ] (from Sabine): put the local table of content on the right

src/document/generator.ml
jonludlam: I do find it very awkward to be threading the sidebar through all these calls in `document/generator.ml` for very little benefit. It would seem sensible to me to see if we could instead pass a sidebar argument to the renderer; or at least set it in the pages after they've been generated.

src/html_support_files/odoc.css
```diff
@@ -820,15 +821,35 @@ td.def-doc *:first-child {
   line-height: 1.2;
 }
 
+.odoc-toc.odoc-local-toc:before {
+  content: "Local content";
+
```
EmileTrotignon, line 826: Useless white space here

src/document/generator.ml
panglesd: I factored what I could in the `sidebar_toc_entry` function in [4eb39d8](https://github.com/ocaml/odoc/pull/1145/commits/4eb39d8b11d1ea259ea8d7547fba2c871965aa6b) (and the previous commit).

src/document/generator.ml
panglesd: I can do that but it will introduce unrelated diffs in this PR. I'll try and we'll see.

src/document/generator.ml
panglesd: I removed the threading of the sidebar through `document/generator.ml` in [8d2c0cf](https://github.com/ocaml/odoc/pull/1145/commits/8d2c0cf969ed61578dd0020e6e49414e86c6d243).
A sidebar value is now passed to the renderer, which turns the sidebar into a document (depending on the page generated) and then pass that to the backend renderer.

panglesd: I am still not sure that it was not better to pass the sidebar around: the value can be used for different usecases, as for instance a hypothetical `{toc}` odoc syntax that would generate a table of content.

src/document/generator.ml
Julow: It can be done in a later PR!

src/document/generator.ml
Julow: Makes sense.

src/document/generator.ml
Julow: Nice!

src/document/generator.ml
Julow: Not this one, that would be illegal in 5.2 :police_car:

src/driver/odoc.ml
```diff
@@ -116,12 +116,38 @@ let link ?(ignore_output = false) ~input_file:file ~includes ~docs ~libs () =
...
+        v "-P" % s %% acc)
+      Cmd.empty docs
+  in
+  let libs =
+    List.fold_left
+      (fun acc (libname, path) ->
+        let s = Format.asprintf "%s:%a" libname Fpath.pp path in
+        v "-L" % s %% acc)
+      Cmd.empty libs
+  in
```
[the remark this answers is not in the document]
Julow: :+1:

src/driver/compile.mli
```diff
@@ -8,4 +8,7 @@ type linked
 
 val link : compiled list -> linked list
 
-val html_generate : Fpath.t -> linked list -> unit
+val compile_sidebars :
+  Fpath.t -> Fpath.t -> Packages.set -> Fpath.t Util.StringMap.t
```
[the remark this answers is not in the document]
Julow: :+1:

panglesd:
There is a little work on the css (especially for the search bar, Emile should fix that), but otherwise, I finished what I wanted to do on this PR before merging.

jonludlam:
OK, this is good enough now! Further suggestions can be addressed in subsequent PRs.

PR #1146, opened 2024-06-10
Fix gaps between preamble and content

### LINKED ISSUE #1111: Mind the gap

panglesd:
Fix #1111 (alternative to #1143).

Since the introduction of the grid system (#999), [margin collapse](https://www.joshwcomeau.com/css/rules-of-margin-collapse/) did [not work](https://www.w3.org/TR/css-grid-1/#item-margins) between the preamble and content, creating [gaps on which people fell](https://github.com/ocaml/odoc/issues/1111):

> the margins of adjacent grid items do not collapse

This PR slightly changes the html layout, putting the `odoc-preamble` and `odoc-content` in an `odoc-main` element, which in my opinion make more sense than previously:
Previously, they were separated by the nav bar. I wonder if this was a deliberate choice, eg for accessibility reasons, or if it just happened to be like that?

This does not affects ocaml.org (which uses the json output), and from my tests it does not seem to affect odig (let's ping its author, @dbuenzli, so he's aware of the potential layout change!).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,6 +39,8 @@
 - Fixed title content not being picked up across pages when rendering references
   (#1116, @panglesd)
 - Fix wrong links to standalone comments in search results (#1118, @panglesd)
+- Fix a big gap between the preamble and the content of a page (#1146,
+  @panglesd)
 
 
 # 2.4.0
--- a/src/html/html_page.ml
+++ b/src/html/html_page.ml
@@ -210,10 +210,15 @@ let search_urls = %s;
 
   let body =
     html_of_breadcrumbs breadcrumbs
-    @ search_bar
-    @ [ Html.header ~a:[ Html.a_class [ "odoc-preamble" ] ] header ]
-    @ sidebar toc
-    @ [ Html.div ~a:[ Html.a_class [ "odoc-content" ] ] content ]
+    @ search_bar @ sidebar toc
+    @ [
+        Html.div
+          ~a:[ Html.a_class [ "odoc-main" ] ]
+          [
+            Html.header ~a:[ Html.a_class [ "odoc-preamble" ] ] header;
+            Html.div ~a:[ Html.a_class [ "odoc-content" ] ] content;
+          ];
+      ]
   in
 
   let htmlpp = Html.pp ~indent:(Config.indent config) () in
--- a/src/html_support_files/odoc.css
+++ b/src/html_support_files/odoc.css
@@ -307,8 +307,8 @@ body.odoc-src {
   margin-right: calc(10vw + 20ex);
 }
 
-.odoc-content {
-  grid-row: 4;
+.odoc-main {
+  grid-row: 3;
   grid-column: 2;
 }
 
@@ -320,11 +320,6 @@ body.odoc-src {
 
 header {
   margin-bottom: 30px;
-}
-
-header.odoc-preamble {
-  grid-column: 2;
-  grid-row: 3;
 }
 
 nav {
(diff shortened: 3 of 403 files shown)


EmileTrotignon:
Nice, the CI needs some work, but I like the PR.

dbuenzli:
I don't think this should be done. Basically all the different page elements should be at the same level, you can then easily dispatch them using CSS grid.

EmileTrotignon:
> For example Here's a design I can no longer do:
> 
> ```
> |         nav        |
> | preamble | content |
> | toc      |         | 
> ```
> 
> Please don't break odoc's markup because your CSS is broken. `odig`'s CSS has no problem with the current markup.

I think this is a good reason not to go forward with this PR, thanks for the feedback.

dbuenzli:
In general adding more `div`s to solve CSS layout problem is the wrong way to go. 

I would rather suggest someone takes a deep look at rebooting and trimming down `odoc.css`. You have 1200 lines of CSS. `odig` has 270 lines of CSS for a generally more readable result. Granted a few new things like tables are not added yet but still, I think you have a problem here. 

The thing you absolutely do not want is thousands of lines of CSS which is easy to get when people "fix" the CSS here and there (personally I don't think CSS and web design is really suitable for free software collaboration but that's another story – and the reason why I [gave up](https://github.com/ocaml/odoc/commit/1ec8f293c3e51a11dc91109a09a7da659f200456) trying to help with `odoc.css` here).

dbuenzli:
And btw. since I'm looking at odig's CSS this PR certainly does break `odig`'s CSS see [here](https://github.com/b0-system/odig/blob/8e02e740c29a37a56acbaa5d25da9ee0eb8dce75/themes/odoc.css#L83-L86). On narrow (mobile) viewports the stylesheet collapses the page structure to: 
```
| nav      |
| preamble |
| toc      |
| content  |
```

panglesd:
Thanks for your input.

I would not author odoc's CSS either. I'm just trying to fix some problems without rewriting the whole thing, which may indeed be a bad idea. Until someone (with the skills to do it) rewrite the whole thing, I'll stick with that.

On the topic of this PR: Yes, you are right, I underestimated the distinction between the preamble and the content.
So maybe the problem is more that odoc's stylesheet is not making it clear that the reason their is a gap (as in #1111) is that we are switching from the preamble to the content.

I can try to open a PR with a similar approach to what is done in odig? Or do you want to take care of that @EmileTrotignon ?

image

EmileTrotignon:
@panglesd I have a patch ready, I will open a PR.

PR #1147, opened 2024-06-10
Remove margin of first child of odoc-content

EmileTrotignon:
This fixes the gap in between the preamble and the content without changing other margins, or changing the structure of the HTML.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
   select a result, and enter to follow the selected link. (@EmileTrotignon, #1088)
 - OCaml 5.2.0 compatibility (@Octachron, #1094, #1112)
 - New driver package (@jonludlam, #1121)
+- Fix a big gap between the preamble and the content of a page (@EmileTrotignon, #1147)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


EmileTrotignon:
This is a follow-up to PRs #1143 and #1146. It aims to fix issue #1111.

panglesd:
There is still a gap between preamble and content:

image

This is due to [that](https://github.com/ocaml/odoc/pull/1147/files#diff-0a68a20ade80407afe9aa537eed792ed368120048b57da28bdd32b7fc91d7b4dL321-L323).

Even if smaller than before, this gap will be unexpected to most people who do not know there is a notion of preamble in odoc. Now that I understood that, I think it would be better not to remove the gap, but to make it understandable:

image

That being said, go with what you prefer.

EmileTrotignon:
As far as I understand, the gap was there before we switched to grid, and its quite reasonable to have a gap here. 
If I am mistaken and this PR still has a larger gap than odoc pre-grid layout, I am happy to tune it, otherwise I think we should merge.

panglesd:
Indeed, this PR reverts the introduction of a larger gap from #999.

Feel free to merge.

PR #1150, opened 2024-06-21
Path-references lookup to pages

Julow:
This implements the resolving of some path-references using the new `Named_roots` mechanism recently added.

This is work in progress. Currently, current-package and absolute references are resolved from pages. On top of https://github.com/ocaml/odoc/pull/1142

### DIFF
--- a/src/odoc/resolver.ml
+++ b/src/odoc/resolver.ml
@@ -422,23 +422,32 @@ let lookup_path _ap ~pages ~libs:_ (kind, tag, path) =
     | [] -> []
     | name :: rest -> List.rev (("page-" ^ name ^ ".odoc") :: rest)
   in
-  let find_by_path named_roots path =
-    match Named_roots.find_by_path named_roots ~path with
+  let find_by_path ?root named_roots path =
+    match Named_roots.find_by_path ?root named_roots ~path with
     | Ok x -> x
     | Error (NoPackage | NoRoot) -> None
   in
+  let find_page ?root path =
+    ( pages >>= fun pages ->
+      find_by_path ?root pages path >>= fun path ->
+      load_unit_from_file path >>= function
+      | Odoc_file.Page_content page -> Some page
+      | _ -> None )
+    |> function
+    | Some page -> Env.Path_page page
+    | None -> Env.Path_not_found
+  in
   match (kind, tag) with
-  | `Page, `TCurrentPackage -> (
+  | `Page, `TCurrentPackage ->
       (* [path] is within the current package root. *)
       let path = Fs.File.of_segs (page_path_to_path path) in
-      ( pages >>= fun pages ->
-        find_by_path pages path >>= fun path ->
-        load_unit_from_file path >>= function
-        | Odoc_file.Page_content page -> Some page
-        | _ -> None )
-      |> function
-      | Some page -> Env.Path_page page
-      | None -> Env.Path_not_found)
+      find_page path
+  | `Page, `TAbsolutePath -> (
+      match path with
+      | root :: path ->
+          let path = Fs.File.of_segs (page_path_to_path path) in
+          find_page ~root path
+      | [] -> Env.Path_not_found)
   | _ -> Env.Path_not_found
 
 type t = {
--- a/src/odoc/resolver.ml
+++ b/src/odoc/resolver.ml
@@ -206,16 +206,19 @@ let unit_name
   root_name root
 
 (** TODO: Propagate warnings instead of printing. *)
+let load_unit_from_file path =
+  match Odoc_file.load path with
+  | Ok u -> Some u.content
+  | Error (`Msg msg) ->
+      let warning =
+        Odoc_model.Error.filename_only "%s" msg (Fs.File.to_string path)
+      in
+      prerr_endline (Odoc_model.Error.to_string warning);
+      None
+
 let load_units_from_files paths =
   let safe_read file acc =
-    match Odoc_file.load file with
-    | Ok u -> u.content :: acc
-    | Error (`Msg msg) ->
-        let warning =
-          Odoc_model.Error.filename_only "%s" msg (Fs.File.to_string file)
-        in
-        prerr_endline (Odoc_model.Error.to_string warning);
-        acc
+    match load_unit_from_file file with Some u -> u :: acc | None -> acc
   in
   List.fold_right safe_read paths []
 
@@ -410,6 +413,34 @@ let add_unit_to_cache u =
   in
   Hashtbl.add unit_cache target_name [ u ]
 
+let lookup_path _ap ~pages ~libs:_ (kind, tag, path) =
+  let module Env = Odoc_xref2.Env in
+  let ( >>= ) x f = match x with Some x' -> f x' | None -> None in
+  let page_path_to_path path =
+    (* Turn [foo/bar] into [foo/page-bar.odoc]. *)
+    match List.rev path with
+    | [] -> []
+    | name :: rest -> List.rev (("page-" ^ name ^ ".odoc") :: rest)
+  in
+  let find_by_path named_roots path =
+    match Named_roots.find_by_path named_roots ~path with
+    | Ok x -> x
+    | Error (NoPackage | NoRoot) -> None
+  in
+  match (kind, tag) with
+  | `Page, `TCurrentPackage -> (
+      (* [path] is within the current package root. *)
+      let path = Fs.File.of_segs (page_path_to_path path) in
+      ( pages >>= fun pages ->
+        find_by_path pages path >>= fun path ->
+        load_unit_from_file path >>= function
+        | Odoc_file.Page_content page -> Some page
+        | _ -> None )
+      |> function
+      | Some page -> Env.Path_page page
+      | None -> Env.Path_not_found)
+  | _ -> Env.Path_not_found
+
 type t = {
   important_digests : bool;
   ap : Accessible_paths.t;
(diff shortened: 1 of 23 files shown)


src/xref2/ref_tools.ml
```diff
@@ -174,6 +175,37 @@ let type_lookup_to_class_signature_lookup =
...
+
+  let expect_page ~tag ~path = function
+    | `Unit _ -> Error (`Path_error (`Wrong_kind ([ `Page ], `Unit), tag, path))
+    | `Page p -> Ok p
+
+  let page_in_env env page_path : page_lookup_result ref_result =
+    let tag, path = unpack_page_path [] page_path in
+    handle_lookup_errors ~tag ~path (Env.lookup_path (`Page, tag, path) env)
+    >>= expect_page ~tag ~path
+    >>= fun p -> Ok (`Identifier p.name, p)
```
panglesd, line 199: Slightly more readable?
```suggestion
  let page_in_env env page_path : page_lookup_result ref_result =
    let tag, path = unpack_page_path [] page_path in
    Env.lookup_path (`Page, tag, path) env
    |> handle_lookup_errors ~tag ~path
    >>= expect_page ~tag ~path
    >>= fun p -> Ok (`Identifier p.name, p)
```

src/xref2/env.mli
```diff
@@ -12,16 +12,24 @@ type lookup_page_result = Lang.Page.t option
...
+  | Path_directory
+  | Path_not_found
+
 type root =
-  | Resolved of
-      (Root.t * Odoc_model.Paths.Identifier.Module.t * Component.Module.t)
+  | Resolved of (Root.t * Identifier.Module.t * Component.Module.t)
   | Forward
 
+type path_query = [ `Page | `Unit ] * Reference.tag_page_path * string list
```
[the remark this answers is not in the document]
panglesd: (this is not a comment asking for any change. I was just amused at seeing the input of the function being a tuple)

src/odoc/resolver.ml
```diff
@@ -465,40 +542,64 @@ type roots = {
...
-    | Some { page_roots; lib_roots; current_lib; current_package } ->
-        ( Some (Named_roots.create ~current_root:current_package page_roots),
-          Some (Named_roots.create ~current_root:current_lib lib_roots) )
+    | None -> (None, None, None)
+    | Some { page_roots; lib_roots; current_lib; current_package; current_dir }
+      ->
+        let pages = Named_roots.create ~current_root:current_package page_roots
+        and libs = Named_roots.create ~current_root:current_lib lib_roots in
+        let hierarchy =
+          match Named_roots.find_by_path pages ~path:(Fpath.v ".") with
```
panglesd, line 492: Thanks, by using a bug you uncovered its existence!

As I though it, `find_by_path` was supposed to only return files (and not directories).
I think it is fine for it to return directories, but in this case we need to check that the output is a file (using `Sys.is_regular_file f` or `not (Sys.is_directory f)`)

src/odoc/resolver.ml
```diff
@@ -410,12 +439,60 @@ let add_unit_to_cache u =
...
+    | Some path -> load_page path
+    | None -> Env.Path_not_found
+  in
+  let find_page_in_hierarchy path =
+    match hierarchy with
+    | Some hierarchy -> (
+        match Hierarchy.resolve_relative hierarchy path with
+        | Ok path -> load_page path
+        | Error `Escape_hierarchy ->
+            Env.Path_not_found (* TODO: propagate more information *))
```
panglesd, line 409: Writing a comment here not to forget this TODO.

src/odoc/resolver.ml
```diff
@@ -410,12 +439,60 @@ let add_unit_to_cache u =
...
+  let find_page ?root path =
+    (pages >>= fun pages -> find_by_path ?root pages path) |> function
+    | Some path -> load_page path
+    | None -> Env.Path_not_found
+  in
+  let find_page_in_hierarchy path =
+    match hierarchy with
+    | Some hierarchy -> (
+        match Hierarchy.resolve_relative hierarchy path with
+        | Ok path -> load_page path
```
panglesd, line 407:
```
if Sys.is_directory path then Env.Path_directory else load_page path
```
(this check should probably be in `load_page` maybe)

src/odoc/resolver.ml
```diff
@@ -413,17 +444,23 @@ let build_compile_env_for_unit
   let imports_map = build_imports_map m.imports in
   let lookup_unit = lookup_unit ~important_digests ~imports_map ~libs ap
   and lookup_page = lookup_page ~pages ap
```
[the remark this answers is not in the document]
Julow: Yes! I initially wanted to wait until the "by name" lookup is implemented but I think that's not necessary. Removed.

src/xref2/ref_tools.ml
```diff
@@ -179,17 +179,29 @@ let type_lookup_to_class_signature_lookup =
...
+    | Path_not_found -> Error (`Path_error (`Not_found, tag, path))
+
+  let expect_page ~tag ~path = function
+    | `Unit _ -> Error (`Path_error (`Wrong_kind ([ `Page ], `Unit), tag, path))
+    | `Page p -> Ok p
+
+  let page_in_env env (tag, path) : page_lookup_result ref_result =
+    handle_lookup_errors ~tag ~path (Env.lookup_path (`Page, tag, path) env)
+    >>= expect_page ~tag ~path
+    >>= fun p -> Ok (`Identifier p.name, p)
```
panglesd, line 197: Sorry, I don't like mixing having normal and "inverted" function applications! (even if that ends up one line longer)

```suggestion
    Env.lookup_path (`Page, tag, path) env
    |> handle_lookup_errors ~tag ~path
    >>= expect_page ~tag ~path
    >>= fun p -> Ok (`Identifier p.name, p)
```

src/odoc/resolver.ml
```diff
@@ -413,17 +444,23 @@ let build_compile_env_for_unit
...
-  and lookup_impl = lookup_impl ap in
-  let resolver = { Env.open_units; lookup_unit; lookup_page; lookup_impl } in
+  and lookup_impl = lookup_impl ap
+  (* Do not implement [lookup_path] in compile mode, as that might return
+     different results depending on the compilation order. *)
```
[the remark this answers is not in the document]
Julow: Yes but they are implemented in compile mode. Hopefully not used.

panglesd (approved):
Very nice!

src/xref2/ref_tools.ml
Julow: Added.

src/odoc/resolver.ml
Julow: Here we want a directory. This requires a path that will be used to resolve relative references later, it is not expected to be a page.

Elsewhere, the lookup function loads the page/unit files before returning a positive result so no need to check that at any place. The lookup can't succeed on a directory.

src/odoc/resolver.ml
Julow: Yes!

Though this would be detected while loading the page, but the error is not propagated and is instead printed as a warning outside of the usual warning mechanisms.

Julow: No! I think this should be reported by the `load_page` function. `Sys.is_directory` is just an other point where an exception can be raised and an other inconsistency.
That code have a TODO that I would leave for later.

src/xref2/ref_tools.ml
Julow: Done.

src/odoc/resolver.ml
Julow: This will be part of improving error propagation in the resolver (cf comment https://github.com/ocaml/odoc/pull/1150#discussion_r1673755882) that I would prefer to think about later.

src/odoc/resolver.ml
panglesd: I can't see a TODO yet on this specific error...

The problem is that if you have a directory with the specific name `page-directory.odoc`, then you might have an uncaught exception when resolving `{!./page-directory}`:
```
Error while unmarshalling "h/pkg/doc/page-directory.odoc": Sys_error("Is a directory")
```

I'm happy to leave a better error reporting for later, but I would really prefer to merge without the risk of uncaught exceptions when resolving references... (note that when we lookup for assets, the right of having a directory referenced will be much higher).

src/odoc/resolver.ml
```diff
@@ -389,53 +418,125 @@ let add_unit_to_cache u =
   in
   Hashtbl.add unit_cache target_name [ u ]
 
+let lookup_path _ap ~pages ~libs:_ ~hierarchy (kind, tag, path) =
+  let module Env = Odoc_xref2.Env in
+  let ( >>= ) x f = match x with Some x' -> f x' | None -> None in
```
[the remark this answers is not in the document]
Julow: I just realized how ridiculous the interface is and how hard it makes error propagation. I'll revamp the interface.

src/odoc/resolver.ml
Julow: That's not an uncaught exception but a very badly reported expected error.

I agree this should be checked eventually but without a way to propagate the errors correctly, this will just slightly improve a very bad error message.
In any case, I don't want to spam the code with this kind of checks. This check must happen just before loading the file, not elsewhere.

panglesd: Oh indeed, that's not an uncaught exception. It was obvious from the message and I missed the "catch all exceptions" in `load_`...

Thanks for doing the refactoring!

Julow: I changed the interface of the resolver completely in order to pass the errors. Which I leave for later.

Julow:
CI was not happy.

PR #1151, opened 2024-06-26
Add the '--current-package' option

Julow:
This adds the `--current-package` option, which is needed to resolve pages when linking modules.

The resolver is changed to keep track of the current library and the current package separately, instead of the "current root" which could be inconsistent. These are also made optional to improve error messages.

A cram test showcases all the checks that are done on the link options.


Julow:
This allowed to resolve more references in https://github.com/ocaml/odoc/pull/1150: https://github.com/ocaml/odoc/pull/1150/commits/7d76341474c41a54c35398c1163eebda8cf2dcd1

src/odoc/resolver.ml
```diff
@@ -105,14 +110,19 @@ end = struct
...
   let find_by_name ?root { table = cache; current_root } ~name =
-    let package = match root with None -> current_root | Some pkg -> pkg in
-    match hashtbl_find_opt cache package with
-    | Some { flat = Visited flat; _ } -> Ok (Hashtbl.find_all flat name)
-    | Some ({ flat = Unvisited root; _ } as p) ->
-        let flat = populate_flat_namespace ~root in
-        Hashtbl.replace cache package { p with flat = Visited flat };
-        Ok (Hashtbl.find_all flat name)
-    | None -> Error NoPackage
+    let package =
```
panglesd, line 113: Same comment here! (Wouldn't it make it easier to read with one level less of indentation using a result monad?)

panglesd (approved):
Looks good to me!

src/odoc/resolver.ml
```diff
@@ -75,18 +75,23 @@ end = struct
...
-        match hashtbl_find_opt cache path with
-        | Some x -> Ok (Some x)
-        | None ->
-            let full_path = Fpath.( // ) (Fs.Directory.to_fpath root) path in
-            if Fs.File.exists full_path then (
-              Hashtbl.add cache path full_path;
-              Ok (Some full_path))
-            else Ok None)
-    | None -> Error NoPackage
+    let root = match root with None -> current_root | Some _ as pkg -> pkg in
```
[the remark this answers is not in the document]
Julow: Makes sense, changed.

Julow:
I've rebased and updated the driver.

src/driver/compile.ml
```diff
@@ -16,6 +16,7 @@ type compiled = {
   include_dirs : Fpath.Set.t;
   impl : impl option;
   pkg_args : pkg_args;
+  current_package : string;
```
panglesd, line 19: This change has already been made in various flavor in two other PRs: [here](https://github.com/ocaml/odoc/pull/1084/files#diff-aab3f7bebea0819961f37ab5de1d58bef4863e48e77cd55d26b914a3b63ef87dR19) and [there](https://github.com/ocaml/odoc/pull/1145/files#diff-aab3f7bebea0819961f37ab5de1d58bef4863e48e77cd55d26b914a3b63ef87dR19)!

We'll see which one is kept during the rebase.

panglesd (approved):

PR #1153, opened 2024-07-02
Update ocaml/setup-ocaml to v3

smorimoto:


panglesd:
If I understand correctly the setup-ocaml release changelog, this setups ocaml with opam 2.2 for windows support?
Would that enable windows CI testing? That would be great!

hhugo:
You shouldn't use Fpath.to_string when generating path in dune files.

.github/dependabot.yml
panglesd, line 1: Whats is this file for?

panglesd: Ah, I guess it is for dependabot to open a PR with a version bump whenever setup-ocaml is upgraded?

If that's it, that's neat!

panglesd (approved):
I think we are not ready yet to add windows testing on our CI. However, it is very nice to be able to add it so easily!

I think this is good to merge as soon as the windows-triggering line is commented out. We'll uncomment it after fixing the windows issues.

Thanks a lot for your work!

.github/dependabot.yml
smorimoto: Yes!

smorimoto:
Ready to go!

dune-project
```diff
@@ -1,4 +1,4 @@
-(lang dune 3.7)
+(lang dune 3.16)
```
panglesd, line 1: I think this should raise the minimal version in `odoc.opam` file!

.ocamlformat
```diff
@@ -1,3 +1,3 @@
 module-item-spacing=preserve
-version=0.26.1
```
panglesd, line 2: We should do that at some point but in a separate PR

.github/workflows/build.yml
```diff
@@ -1,61 +1,62 @@
...
-          - 4.04.x
-          - 4.06.x
-          - 4.08.x
-          - 4.10.x
-          - 4.12.x
-          - 4.14.x
-          - 5.0.x
-          - 5.1.x
+          - 4
+          - 5
```
panglesd, line 18: If I understand correctly, this runs odoc's test in 4.02 on ubuntu latest (due to the `include` below), and on 4.14 and the latest 5 version for linux, windows and mac.

I think it makes sense for odoc to continue checking each individual versions (4.07, 5.0 and 5.1, ...). Indeed, odoc has a lot of compatibility code for instance for the different ASTs or for the shapes, that can easily break so that we want to continue testing. (This compatibility code might work for 4.02 and 4.14 but not for an intermediate versions).

.github/workflows/build.yml
```diff
@@ -1,61 +1,62 @@
...
-          opam-pin: false
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
-          opam-local-packages: |
-            odoc.opam
+          # Do not pin odoc to not break Mdx installation
+          opam-pin: ${{ matrix.run-mdx != true }}
 
       - name: Install dependencies
-        run: opam install -y --deps-only -t ./odoc.opam ./odoc-parser.opam
+        run: opam install --deps-only --with-test odoc odoc-parser
```
panglesd, line 50: Won't that install deps from the odoc from opam repository in case where `run-mdx` is true, breaking if we modify the dependencies without releasing?

panglesd (commented):
Thanks and sorry for the slow review: summer if full of holidays!

panglesd (commented):
Being able to run windows test without failing the CI is nice, but a little bit of a resource waste until we start fixing it. Maybe it would be better to disable it until we start working on #1169 ?

.github/workflows/build.yml
smorimoto: Indeed.

dune-project
smorimoto: Reverted

.ocamlformat
smorimoto: Reverted

.github/workflows/build.yml
smorimoto: Reverted

.github/workflows/build.yml
smorimoto: Fixed

smorimoto:
@panglesd Ready to go!

smorimoto:
OK!

PR #1154, opened 2024-07-04
Line editing Contributing.md

christinerose:
In li.29, I get the following error:

```
[ERROR] Package conflict!
  * Missing dependency:
    - odoc-parser >= dev
    no matching version
```

I followed the code from the top of this page. I'm I doing something wrong, or is there a missing step in here?

In li.84, is that version still correct or is there a more recent version now?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ Here's a step-by-step guide to get you set up for HTML contribution:
    ```
    git clone https://github.com/ocaml/odoc.git
    cd odoc
-   opam pin add --no-action `odoc` .
+   opam pin add --no-action odoc .
    opam install --with-test --deps-only odoc
    ```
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -13,19 +13,19 @@ Contributing to the HTML is slightly more complex. The main HTML generator is in
...
      - On Debian / Ubuntu: `sudo apt-get install jq`
 
 2. **Set up for development:**
-   Clone the odoc repository, navigate into the new directory, and install dependencies with these commands:
+   Clone the `odoc` repository, navigate into the new directory, and install dependencies with these commands:
    ```
    git clone https://github.com/ocaml/odoc.git
    cd odoc
-   opam pin add --no-action odoc .
+   opam pin add --no-action `odoc` .
```
Julow, line 28: No quote here, it's a code block.

Julow, line 28: You need to add `opam pin add --no-action odoc-parser .` before this line.

christinerose: Oh nice catch! I missed that.

christinerose: I added this and ran that sequence again, but I'm still getting
```
[ERROR] Package conflict!
  * Missing dependency:
    - odoc-parser >= dev
    no matching version

No solution found, exiting
```

Julow: I would just remove both `opam pin add` commands and change the `opam install` command to:

```
opam install --with-test --deps-only ./odoc-parser.opam ./odoc.opam
```

christinerose: Done! No longer getting that error. Thanks!

panglesd (approved):
Thanks!

That reminds me that at some point, we will need to add some contributing notes on running the reference driver (mostly how to install sherlodoc).

PR #1149, opened 2024-06-21
parser: Don't accept code block results without delim

Julow:
This fixes https://github.com/ocaml/odoc/issues/1137 by implementing @panglesd's suggestion.

The syntax for code blocks with a result section is now:

    {delim@lang[ code ]delim[ result ]}

Note: The terminating ]} do not contain the delimiter.

PR #1156, opened 2024-07-05
Fix fmt for parser test

panglesd:
I merged #1149 and realized only after that the formatting CI did not pass...

Fixing that here.


src/parser/test/test.ml
```diff
@@ -2903,7 +2903,7 @@ let%expect_test _ =
         {|{delim@ocaml[ let x = ]delim[ {err@mdx-error[ here's the error ]} ]err}
         ]delim}|};
       [%expect
-        "
+        {|
```
Julow, line 2906: The escaping within the string also need to change.

Julow (approved):
Let's merge!

PR #1157, opened 2024-07-10
add odoc package as a link deps to the driver

EmileTrotignon:
Add the odoc package as a deps to the driver, because it is one.

Now when you do `dune exec odoc_driver -- -p ...` before doing `dune build`, it works anyway !

### DIFF
--- a/src/driver/dune
+++ b/src/driver/dune
@@ -1,6 +1,8 @@
 (executable
  (public_name odoc_driver)
  (package odoc-driver)
+ (link_deps
+  (package odoc))
  (libraries
   cmdliner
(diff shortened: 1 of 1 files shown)


jonludlam:
This has caught me out before! `link_deps` seems a bit wierd though, but I can't immediately spot a better solution.

EmileTrotignon:
I think thats just how you do this @jonludlam

jonludlam:
Perhaps the dune docs need updating then! In any case, this fixes a problem in a quite reasonable way, so let's get it in. Thanks!

PR #1159, opened 2024-07-11
Fix interaction between search and side bars

EmileTrotignon:
There are now a lot of bars in odoc, which can be side, search or absent or present.

This fixes weird css interactions in with some combinations in that were merged before we had a chance to fix them in the sidebar pr.

### DIFF
--- a/src/html_support_files/odoc.css
+++ b/src/html_support_files/odoc.css
@@ -863,7 +863,7 @@ td.def-doc *:first-child {
 /* When a search bar is present, we need the sticky sidebar to be a bit lower,
    so `top` is higher  */
 
-.odoc-search + * + .odoc-tocs .odoc-toc {
+body.odoc:has( .odoc-search) .odoc-toc {
   --toc-top: calc(var(--search-bar-height) + 2 * var(--search-padding-top));
   max-height: calc(100vh - 2 * var(--toc-top));
   top: var(--toc-top)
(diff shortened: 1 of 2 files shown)


src/html_support_files/odoc.css
```diff
@@ -1253,7 +1274,7 @@ td.def-doc *:first-child {
   }
   .odoc-toc {
       position: unset;
-      max-height: unset;
+      max-height: unset !important;
```
panglesd, line 1277: Maybe it would be better to have the same priority on both the rule setting the `max-height` and the one inside the media block?

EmileTrotignon:
In the presence of a global sidebar the layouts are the following : 

large screens: 
image
image

medium screens:
image
image

small screens:
image
image

src/html_support_files/odoc.css
EmileTrotignon:
```css
/* When a search bar is present, we need the sticky sidebar to be a bit lower,
   so `top` is higher  */

.odoc-search + * + .odoc-tocs .odoc-toc {
  --toc-top: calc(var(--search-bar-height) + 2 * var(--search-padding-top));
  max-height: calc(100vh - 2 * var(--toc-top));
  top: var(--toc-top)
}
```
Here is it, the high priority is built in the rule.

EmileTrotignon:
The screenshot are produced with the "final fixes" ([a1af33c](https://github.com/ocaml/odoc/pull/1159/commits/a1af33cbc49b7f374e50b8a89bfd2d768a42f847)) commit

jonludlam:
Isn't the use of `!important` usually considered to be an anti-pattern?

panglesd:
Yes, it is an antipattern. I think it would be better to give a specific class to `body` when the search bar is present rather than relying on some brittle rule (`.odoc-search + * + .odoc-tocs .odoc-toc`) and antipattern (`!important`).

EmileTrotignon:
> Yes, it is an antipattern. I think it would be better to give a specific class to `body` when the search bar is present rather than relying on some brittle rule (`.odoc-search + * + .odoc-tocs .odoc-toc`) and antipattern (`!important`).

I think the solution you mention is the right one, but I am not sure this should be part of this PR. The brittle rule is present since 2.4.

src/html_support_files/odoc.css
```diff
@@ -1228,55 +1249,33 @@ td.def-doc *:first-child {
...
       flex-direction : column;
       gap: 20px;
-  }
-  .odoc-toc {
-      position: unset;
-      max-height: unset;
+      .odoc-toc {
+        position: unset;
+        max-height: unset;
+      }
```
panglesd, line 1263: What is this rule doing in another one? Am I too old and this is valid?

src/html_support_files/odoc.css
```diff
@@ -1285,29 +1284,40 @@ td.def-doc *:first-child {
...
-  .odoc-search {
-    position: relative;
-    padding-bottom: 0;
-    height: calc(var(--search-bar-height) + var(--search-padding-top));
-  }
-
-  .odoc-tocs {
+    &:has(> .odoc-search:focus-within) {
+      /* This is the same as when there is no focus on the search bar, this is
+      just to prevent the full screen rule from changing anything. */
```
panglesd, line 1289: What do you mean with "fullscreen rule"?

src/html_support_files/odoc.css
panglesd: Yes, it is valid, and quite cool: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting
Thanks for teaching me that!

However, [caniuse](https://caniuse.com/?search=css%20nesting) reports only 82% of used browser with full support. Since this was not used before, is it really worth introducing it?

EmileTrotignon: I was used before, the whole media rule was nested.

src/html_support_files/odoc.css
EmileTrotignon: I meant the large screen/default rule

src/html_support_files/odoc.css
panglesd: Media queries are not CSS nesting... They have 98% support on [caniuse](https://caniuse.com/css-mediaqueries).

src/html_support_files/odoc.css
panglesd:
```suggestion
      just to prevent the "wide layout" rule from changing anything. */
```

panglesd (approved):
This indeed fixes interaction between search and sidebars.

PR #1163, opened 2024-07-15
Disable "big search path" test.

panglesd:
Was too brittle: too many false positive in CI...
(I rebuilded the debian-12-5.2_s390x_opam-2.1 ocaml-CI job on several PR).

I kept the test as a reminder to add it as part of a potential benchmark CI.


Julow (approved):
Looks good. We'll hopefully test this as part of our bigger test cases like ocaml.org or a big dune workspace.

PR #1166, opened 2024-07-19
line editing for odoc_for_authors.mld

christinerose:
Mostly formatting and other line editing, but there are a few questions below for your review @panglesd.

### DIFF
--- a/doc/odoc_for_authors.mld
+++ b/doc/odoc_for_authors.mld
@@ -47,11 +47,11 @@ module, module type, class, and class type should be documented, keeping in mind
 that it will end up as a single page with a table of contents. 
 
 For each module, [odoc] will describe all of the values, types, module types,
-classes, modules, and all other elements found in module signatures, documenting
-the details relevant to the OCaml type system along with any documentation
-written in specially formatted comments in the source code. For any element that
-references any other, for example a [value] that has a [type], [odoc] will
-create links such that clicking them will link to the definitions.
+classes, modules, and other elements found in module signatures. Their types or
+descriptions are documented along with any comments delineated by [(** ... *)],
+known as docstrings, found in the source code. For any element that references
+another, for example a [value] that has a [type], [odoc] will create links that
+point to the definitions.
 
 Therefore, the job of the library author is to organise the module interface
 in a logical manner and write comments explaining each type, value, exception,
--- a/doc/odoc_for_authors.mld
+++ b/doc/odoc_for_authors.mld
@@ -523,7 +523,7 @@ i.e., there shouldn't be any [odoc] markup.
 {3 Block Tags}
 
 These tags have a block of potentially marked-up text associated with them,
-and occasionally some more data too. The block of text is ended by the end
+and occasionally some more data too. The block of text concludes with the end
 of the comment or by another tag. They are:
 
 - [@deprecated <text>] - marks the element as deprecated. [text] should describe
--- a/doc/odoc_for_authors.mld
+++ b/doc/odoc_for_authors.mld
@@ -1,21 +1,21 @@
-{0:top [odoc] for authors}
+{0:top [odoc] for Authors}
 
-This manual describes the features that are available and recommended for users
-to write great documentation using [odoc]. See also the page listing the
+This manual describes the features available and recommended for users
+in order to write great documentation with [odoc]. See also the page listing the
 {{!page-features}language features} [odoc] understands.
 
-+ {{!getting_started}Getting started}
-+ {{:#interfaces} Documenting your interfaces}
-+ {{:#doc-pages} Writing documentation pages}
++ {{!getting_started}Getting Started}
++ {{:#interfaces} Documenting Your Interfaces}
++ {{:#doc-pages} Writing Documentation Pages}
 
-{1:getting_started Getting started}
+{1:getting_started Getting Started}
 
 To generate documentation for your project, you will almost always be using
-odoc indirectly rather than executing it yourself. There are currently several
-'drivers' of odoc at time of writing, each with their own strengths and
+[odoc] indirectly rather than executing it yourself. There are currently several
+'drivers' of [odoc] at time of writing, each with their own strengths and
 weaknesses:
 
-- {{:https://dune.build/}Dune} is the best way to view docs while you're
+- {{:https://dune.build/}Dune} is the best way to view docs while
   developing your project. We have a {{!page-dune}page} describing
   how to use it, but the short version is:
 {v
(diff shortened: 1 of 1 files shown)


doc/odoc_for_authors.mld
```diff
@@ -32,36 +32,36 @@ v}
...
+module, module type, class, and class type should be documented, keeping in mind
 that it will end up as a single page with a table of contents. 
 
 For each module, [odoc] will describe all of the values, types, module types,
-classes, modules and all other elements found in module signatures, documenting
+classes, modules, and all other elements found in module signatures, documenting
 the details relevant to the OCaml type system along with any documentation
 written in specially formatted comments in the source code. For any element that
 references any other, for example a [value] that has a [type], [odoc] will
 create links such that clicking them will link to the definitions.
```
christinerose, line 54:
```suggestion
classes, modules, and all other elements found in module signatures. These record
the details relevant to the OCaml type system along with any documentation
written in specially-formatted comments in the source code. For any element that
references any other, for example a [value] that has a [type], [odoc] will
create links such that clicking them will link to the definitions.
```

These are both long sentences. Could we possibly turn them into 3 sentences for readability like this?

doc/odoc_for_authors.mld
```diff
@@ -487,15 +487,17 @@ the {{!page-driver}driver}.
...
+There are three types of tags. Those with:
+- no associated data (simple tags),
+- a single line of text (line tags), and 
+- a block of
 marked-up text (block tags).
 
 {3 Simple Tags}
 
-The three tags with no data are hints to the HTML renderer to do with [include]s.
+The three tags without data are hints to the HTML renderer to do with [include]s.
```
christinerose, line 500:
```suggestion
The three tags without data are hints to the HTML renderer involving [include]s.
```

Is this still accurate?

doc/odoc_for_authors.mld
```diff
@@ -508,15 +510,15 @@ the signature.
...
+- [@since string] - declares from which version the element has been available
+- [@version string] - declares the version of the element itself
+- [@canonical string] - declares the path to this element's canonical instance.
+It can be applied to modules, module types, and types. See the {{!page-features.canonical}
+Language Features} page for more details
 
-The content of the tag is then the rest of the line, and it is uninterpreted,
-i.e., there shouldn't be any odoc markup.
+The tag's content is then the rest of the line. It is also uninterpreted,
+i.e., there shouldn't be any [odoc] markup.
```
christinerose, line 521:
```suggestion
The tag's content fills the rest of the line. It is also uninterpreted,
i.e., there shouldn't be any [odoc] markup.
```
Is this still accurate?

christinerose:
In line 660, is Stop as a special comment better in monospace?

(** This value appears in the documentation, since the `Stop` special comment
      in the class does not affect the parent module of the class.*)

doc/odoc_for_authors.mld
Julow: I think the intent is "it gives hints to the "renderer" on what to do with "include" elements.", So I say "involving" is not very accurate.

doc/odoc_for_authors.mld
Julow: I think "documenting" was saying that these elements are added to the generated documentation, not just recorded. 
Also, "The details relevant to the OCaml type system" seems too abstract for no good reason.
I'd suggest something like that:

"The types or descriptions for these elements are documented along with any documentation comments present in the source code."

doc/odoc_for_authors.mld
Julow: It talks about how it interprets the tags in the source code so it's more like "The tag's content is read from the rest of the line."

doc/odoc_for_authors.mld
```diff
@@ -487,15 +487,17 @@ the {{!page-driver}driver}.
...
-They should appear on their own lines, with nothing but whitespace before them.
+They should appear on their own lines with nothing but whitespace before them.
 
-There are three types of tag: those with no associated data (simple tags),
-those with a single line of text (line tags) and those with a block of
+There are three types of tags. Those with:
+- no associated data (simple tags),
+- a single line of text (line tags), and 
+- a block of
 marked-up text (block tags).
```
Julow, line 496: The line break should probably be removed.

doc/odoc_for_authors.mld
```diff
@@ -773,12 +775,12 @@ Note: A comment attached to a declaration shouldn't contain any heading.
 
 {2:synopsis Synopsis}
 
-The synopsis of a module (a module type, a class, etc..) is the first
+A module's synopsis (a module type, a class, etc..) is the first
```
Julow, line 778: The part in parentheses are alternatives to "module". So I suggest dropping it and adding a sentence saying "This also applies to module types, classes, etc..".

Julow (commented):
On line 660, this is already in a monospace block and I don't think it's useful to emphasis "Stop" more. I'd suggest "since the special stop-comment in the ...".

Thanks :)

doc/odoc_for_authors.mld
christinerose: Would "Their types or descriptions are added to the documentation comments in the source code" work?

Is it clear that "their" refers to the elements mentioned in the previous sentence?

Julow: I think it's confusing to say that the "types or descriptions" are added to the "comments", the documentation is made from the "types or descriptions" and the "comments" at the same level (not one into the other).
The "Their" is fine.

christinerose: Okay! How about: 
"Their types or descriptions are documented along with any comments in the source code."
This suggests the comments are documented, too.

Does this say the same thing without using Document(ation) twice? Or should we stick with: "Their types or descriptions are documented along with any documentation comments present in the source code."

Julow: I think it's fine to say "comments" instead of "documentation comments" but it adds a slight ambiguity as there's two types of comments (documentation and non-documentation).

christinerose: Oh! Interesting! Do you mean the difference between `**` (documentation) and `*` (non-documentation), @Julow?

jonludlam: Yes that's correct. Comments that begin with a double asterisk - `(** ... *)` - are intended to end up in the published API documentation. Those that start with a single asterisk - `(* ... *)` are ignored by odoc.

Perhaps we could write "Their types or descriptions are documented along with any explicit documentation comments present in the source code."

christinerose: Would it be more clear to put in the `(** ... *)` as a reminder those are the documentation comments? 

"Their types or descriptions are documented along with any `(** ... *)` comments present in the source code.

(I'm looking for a way to avoid saying document(ation) twice in the same sentence, but if none of these work, then we'll stick with both. :-)

panglesd: Let's resolve this and merge this PR! We have a few propositions:

> Their types or descriptions are documented along with any docstring found in the source code.

or

> Their types or descriptions are documented along with any (** ... *) comments found in the source code.

or

> Their types or descriptions are documented along with any explicit documentation comments present in the source code.

or

> The types or descriptions for these elements are documented along with any documentation comments present in the source code.

I'm fine with all of them.

christinerose: > The "documentation comments" are often called "docstrings". Would it be OK to use that to avoid the "documentation" repetition?
> 
> > "Their types or descriptions are documented along with any docstring found in the source code."

I think docstrings really makes it more clear, as long as the reader knows that `docstring` = `(** )` comments

christinerose, line 54:
```suggestion
the details relevant to the OCaml type system along with any documentation
written in specially formatted comments in the source code. These comments, delineated by `(** *)`, are called docstrings. For any element that
references another, for example a [value] that has a [type], [odoc] will
create links that point to the definitions.
```

christinerose: So many changes... the entire section with proposed changes:

"For each module, [odoc] will describe all of the values, types, module types,
classes, modules, and other elements found in module signatures. Their types or descriptions are documented along with any comments delineated by `(** ... *), known as docstrings, found in the source code. For any element that references another, for example a [value] that has a [type], [odoc] will
create links that point to the definitions."

doc/odoc_for_authors.mld
```diff
@@ -32,36 +32,36 @@ v}
...
+class, and class type. This includes any submodules, so if a module [A] contains
+module [B] then two HTML pages will be created. [A/index.html] will have a
 link to [A/B/index.html] at the point that module [B] is defined. So each
-module, module type, class and class type should be documented keeping in mind
+module, module type, class, and class type should be documented, keeping in mind
 that it will end up as a single page with a table of contents. 
 
 For each module, [odoc] will describe all of the values, types, module types,
-classes, modules and all other elements found in module signatures, documenting
+classes, modules, and all other elements found in module signatures, documenting
```
christinerose, line 50:
```suggestion
classes, modules, and other elements found in module signatures, documenting
```

christinerose, line 50:
```suggestion
classes, modules, and other elements found in module signatures. Their types
```

doc/odoc_for_authors.mld
christinerose, line 54:
```suggestion
or descriptions are documented along with any comments delineated by [(** ... *)], known as docstrings, found in the source code. For any element that references another, for example a [value] that has a [type], [odoc] will
create links that point to the definitions.
```

panglesd:
Thanks @christinerose !

PR #1168, opened 2024-07-21
Voodoo odoc driver

jonludlam:
adds two new modes to the driver for different inputs - one to handle voodoo-style layout, and the second to handle dune `_build` style layout.

### DIFF
--- a/src/driver/compile.ml
+++ b/src/driver/compile.ml
@@ -70,32 +70,35 @@ open Eio.Std
 
 type partial = (string * compiled) list * Packages.modulety Util.StringMap.t
 
-let unmarshal filename =
+let unmarshal filename : partial =
   let ic = open_in_bin (Fpath.to_string filename) in
-  let (v : partial) = Marshal.from_channel ic in
-  close_in ic;
-  v
+  Fun.protect
+    ~finally:(fun () -> close_in ic)
+    (fun () -> Marshal.from_channel ic)
 
 let marshal (v : partial) filename =
   let p = Fpath.parent filename in
   Util.mkdir_p p;
   let oc = open_out_bin (Fpath.to_string filename) in
-  Marshal.to_channel oc v [];
-  close_out oc
+  Fun.protect
+    ~finally:(fun () -> close_out oc)
+    (fun () -> Marshal.to_channel oc v [])
 
 let find_partials odoc_dir =
   let tbl = Hashtbl.create 1000 in
   let hashes_result =
-    Bos.OS.Dir.fold_contents ~dotfiles:false
+    Bos.OS.Dir.fold_contents ~dotfiles:false ~elements:`Dirs
       (fun p hashes ->
-        if Fpath.filename p = "index.m" then (
-          let tbl', hashes' = unmarshal p in
-          List.iter
-            (fun (k, v) ->
-              Hashtbl.replace tbl k (Promise.create_resolved (Ok v)))
-            tbl';
-          Util.StringMap.union (fun _x o1 _o2 -> Some o1) hashes hashes')
-        else hashes)
+        let index_m = Fpath.( / ) p "index.m" in
+        match Bos.OS.File.exists index_m with
+        | Ok true ->
+            let tbl', hashes' = unmarshal index_m in
+            List.iter
+              (fun (k, v) ->
+                Hashtbl.replace tbl k (Promise.create_resolved (Ok v)))
+              tbl';
+            Util.StringMap.union (fun _x o1 _o2 -> Some o1) hashes hashes'
+        | _ -> hashes)
       Util.StringMap.empty odoc_dir
   in
   match hashes_result with
--- a/src/driver/compile.ml
+++ b/src/driver/compile.ml
@@ -2,7 +2,7 @@
 
 type ty = Module of Packages.modulety | Mld of Packages.mld
 
-type impl = { impl_odoc : Fpath.t; impl_odocl: Fpath.t; src : Fpath.t }
+type impl = { impl_odoc : Fpath.t; impl_odocl : Fpath.t; src : Fpath.t }
 
 type pkg_args = {
   docs : (string * Fpath.t) list;
@@ -69,7 +69,7 @@ let init_stats (pkgs : Packages.t Util.StringMap.t) =
 
 open Eio.Std
 
-type partial = 
+type partial =
   (string * compiled) list * (string * Packages.modulety) Util.StringMap.t
 
 let unmarshal filename =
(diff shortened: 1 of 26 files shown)


jonludlam:
The voodoo mode requires https://github.com/jonludlam/voodoo/tree/odoc-driver - and the pair have been integrated into docs-ci here: https://github.com/jonludlam/docs-ci/tree/next-gen

src/driver/library_names.ml
```diff
@@ -0,0 +1,68 @@
+(** To extract the library names for a given package, without using dune, we
+
+    1. parse the META file of the package with ocamlfind to see which libraries
+    exist and what their archive name (.cma filename) is.
+
+    2. use ocamlobjinfo to get a list of all modules within the archives. EDIT:
```
panglesd, line 6: I am not 100% sure but it seems this comment is outdated, and the ocamlobjinfo step has been removed.

panglesd (approved):
I pushed some mostly cosmetic changes that hopefully makes it easier to read.
I also left some comments.

I have not reviewed in details every part but I think it is good to be merged as is and improved later.

src/driver/compile.mli
```diff
@@ -2,20 +2,24 @@ type compiled
...
 
-val compile : Fpath.t -> Packages.set -> compiled list
+val compile :
+  ?partial:Fpath.t ->
+  output_dir:Fpath.t ->
+  ?linked_dir:Fpath.t ->
+  Packages.set ->
+  compiled list
+(** Use [partial] to reuse the output of a previous call to [compile]. Useful in
+    the voodoo context.
```
Julow, line 12: It's quite unclear what `partial` means and how to use it. Is the `index.m` file really needed? It seems that it could work without it, at the cost of opening the files in `output_dir`.

Why does the caller have to specify the partial file path ? The compile function could decide where to put it, as it already decides `index.m` which files to read.

src/driver/compile.ml
```diff
@@ -67,20 +69,62 @@ let init_stats (pkgs : Packages.t Util.StringMap.t) =
...
+  let p = Fpath.parent filename in
+  Util.mkdir_p p;
+  let oc = open_out_bin (Fpath.to_string filename) in
+  Marshal.to_channel oc v [];
+  close_out oc
+
+let find_partials odoc_dir =
+  let tbl = Hashtbl.create 1000 in
+  let hashes_result =
+    Bos.OS.Dir.fold_contents ~dotfiles:false
```
Julow, line 91: This could just iterate on directories and check whether they contain a `index.m` file with `` ~elements:`Dirs ``.

src/driver/compile.ml
```diff
@@ -11,12 +11,14 @@ type pkg_args = {
 
 type compiled = {
   m : ty;
-  output_dir : Fpath.t;
-  output_file : Fpath.t;
+  odoc_output_dir : Fpath.t; (* e.g. "_odoc/base/lib/base/" *)
+  odoc_file : Fpath.t; (* Full path to odoc file *)
+  odocl_file : Fpath.t;
```
Julow, line 16: What is the goal of the `odocl_file` field ? It seems to be read only in the `link` function, which would be the right place to construct this path.

src/driver/compile.ml
```diff
@@ -11,12 +11,14 @@ type pkg_args = {
...
   include_dirs : Fpath.Set.t;
   impl : impl option;
   pkg_args : pkg_args;
-  package_name : string;
+  pkg_name : string;
+  pkg_dir : Fpath.t;
```
Julow, line 21: In many places in this module, both of these values are computed from different sources. It would be nice to ensure that they are consistent by always computing them from a single source of truth:
```suggestion
  pkg : string * Fpath.t;
```

src/driver/packages.mli
```diff
@@ -36,16 +38,18 @@ type modulety = {
   m_intf : intf;
   m_impl : impl option;
   m_hidden : bool;
+  m_pkg_dir : Fpath.t;
```
Julow, line 41: This `m_pkg_dir` is used in contextes where the pkg name is probably consistent but we don't need to take the risk.
```suggestion
  m_pkg_dir : string * Fpath.t;
```

src/driver/dune_style.mli
```diff
@@ -0,0 +1 @@
+val of_dune_build : Fpath.t -> Packages.set
```
Julow, line 1: The module name is enigmatic enough that I think this deserves some documentation.

src/driver/dune_style.ml
```diff
@@ -0,0 +1,63 @@
+(* Dune build tree *)
+
+let of_dune_build dir =
```
Julow, line 3: Aren't we supposed to build doc from Dune's `install` folder ?

src/driver/odoc.ml
```diff
@@ -1,12 +1,20 @@
...
+let fpath_of_id id = id
+let id_of_fpath id = id
+
+let index_filename = "index.odoc-index"
+
 type compile_deps = { digest : Digest.t; deps : (string * Digest.t) list }
 
-let odoc = Cmd.v "./_build/default/src/odoc/bin/main.exe"
 (* This is the just-built odoc binary *)
+let default = "./_build/default/src/odoc/bin/main.exe"
```
Julow, line 13: When we run the driver with `dune exec` I think we should be able to call odoc as `odoc`. The reference below could be removed also, as `odoc` should be available in every contextes.

src/driver/odoc_driver.ml
```diff
@@ -497,33 +497,56 @@ let render_stats env nprocs =
       in
       inner (0, 0, 0, 0, 0, 0, 0, 0))
 
-let run libs verbose odoc_dir html_dir stats nb_workers =
+let run libs verbose packages_dir odoc_dir odocl_dir html_dir stats nb_workers
+    odoc_bin voodoo package_name blessed dune_style =
```
Julow, line 501: This `run` function is too complicated and it's unclear how the CLI should be used. To reduce maintenance and confusion, I suggest this is broken into sub-commands: `dune`, `opam` and `voodoo`.

panglesd: I though of that too.

However, in this case https://github.com/jonludlam/docs-ci/tree/next-gen should be updated. I suggest to do it in a later PR.

src/driver/odoc.ml
Julow: Fixed.

src/driver/dune_style.ml
Julow: No we don't.

src/driver/compile.ml
Julow: I've made the changes.

src/driver/packages.mli
Julow: Fixed as well.

src/driver/odoc_driver.ml
Julow: Agree!

Julow (approved):
Looks good to me :)

src/driver/library_names.ml
jonludlam: I should have made it clearer, this module has been imported from https://github.com/ocaml-doc/voodoo/blob/main/src/voodoo/library_names.ml but is being used a bit differently here!

src/driver/odoc_driver.ml
jonludlam: Don't delay on requiring updates to the docs-ci branch, that's very much an experimental one and if it requires changes I'm happy to update it as required :-)

jonludlam: and I'm definitely in favour of the sub-commands idea

src/driver/packages.ml
```diff
@@ -1,5 +1,7 @@
 (* Packages *)
 
+type pkgdir = string * Fpath.t
```
jonludlam: I'm not keen on the name `pkgdir` to refer to something that isn't just a `dir`

jonludlam: In fact, I'm not keen on pairing the two things together. They're used for quite different purposes - the name is only there because we need the package name for the 'current-package' argument, while the directory is used in many other places. I'll try to separate them back out again.

jonludlam:
Thanks everyone!

PR #1170, opened 2024-07-23
Asset cli

EmileTrotignon:
This is a draft for asset cli. It includes ability to write an asset to a file, but no way to load it afterward.


Julow (approved):
LGTM!

PR #1172, opened 2024-07-24
line edit features.mld

christinerose:
Line editing for `features.mld`, mostly formatting, grammar, syntax, etc.

I changed several module type expressions to module-type expressions, as it seems "module type" was the adjective for "expression," based on the header Module Types. If it's "module" "type expressions," where "module" is modifying "type expressions," I'll change them back to "module type expressions."

### DIFF
--- a/doc/features.mld
+++ b/doc/features.mld
@@ -2,13 +2,13 @@
 
 [odoc] works by taking module interfaces, processing them to make them more
 useful, and turning them into documentation. This processing is largely {{!hiding}hiding}, handling of {{!canonical}
-canonical references}, {{!expansion}expansion} and {{!resolution}resolution}.
+canonical references}, {{!expansion}expansion}, and {{!resolution}resolution}.
 This document explains the features of these processes in detail.
 
 {1:hiding Hiding}
 
 Some items are not intended to be used directly but are present as
-implementation detail, e.g., for testing, for implementing Dune's namespacing, or other reasons. 
+implementation detail, e.g., for testing, implementing Dune's namespacing, or other reasons. 
 
 There are two mechanisms for explicitly hiding elements from the final
 output. The first is to use {{:https://ocaml.org/manual/ocamldoc.html#ss:ocamldoc-stop}documentation stop comments},
@@ -20,15 +20,15 @@ mechanism is only used for modules and is a naming convention. If any module has
 Occasionally it’s useful to declare an item in one place and expose
 it in the public interface in another. In order to prevent unwanted occurrences
 of the actual definition site, [odoc] has a feature whereby the 'canonical'
-location of a module, module type or type can be specified.
+location of a module, module type, or type can be specified.
 
-The biggest user of module aliases is Dune's {{:https://dune.readthedocs.io/en/stable/dune-files.html#library}wrapped libraries}.
+The biggest user of module aliases is Dune's {{:https://dune.readthedocs.io/en/stable/reference/dune/library.html#library}wrapped libraries}.
 This feature allows Dune to produce a library whose interface is exposed entirely
 though a single top-level module. It does this by mangling the names of the
-real implementation modules and generating the single top-level module that
+real implementation modules. Then it generates the single top-level module that
 simply contains aliases to the these implementation modules. With [odoc]'s
-canonical modules feature all references to the implementation modules are
-rewritten to point at the aliases in the top-level module instead. Please
+canonical modules feature, all references to the implementation modules are
+rewritten to point at the top-level module aliases instead. Please
 see the section on Dune's 
 {{!page-dune.library_wrapping}library wrapping} for more detail.
 
@@ -64,7 +64,7 @@ in which [odoc] has to treat specially.
 
 {2:expansion_aliases Aliases}
 
-In general [odoc] doesn’t expand module aliases unless they are an
+In general, [odoc] doesn’t expand module aliases unless they are an
 alias to a hidden module. If this is the case, the right-hand side of
 the declaration is dropped and replaced with [sig ... end], and
 the expansion is created.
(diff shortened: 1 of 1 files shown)


doc/features.mld
```diff
@@ -106,13 +106,13 @@ end
 module type B = A
 ]}
 
-For these simple module type declarations, where the right-hand side is just a
+For these simple module-type declarations, where the right-hand side is just a
 path, [odoc] treats them as module aliases and doesn’t produce an expansion.
 This example is rendered {{!Odoc_examples.Expansion.ModuleTypeAliases}here}.
 
 When strengthening, OCaml turns modules into aliases to the original
 module, but nothing is done to module types. In contrast, [odoc] replaces
```
christinerose, line 114:
```suggestion
When strengthening, OCaml turns modules into aliases of the original
module, but nothing is done to module types. In contrast, [odoc] replaces
```
Should this work be "of" instead of "to"? 
Is OCaml make an alias of the original module, perhaps "OCaml creates alias modules from the original module," or "OCaml takes the original module and turns it into an alias."

Let's clarify this a little more. I suspect the same will go for the next in li 114-115
"In contrast, `odoc` replaces module types with 'aliases' of the originals," rather than "to" the originals.

doc/features.mld
```diff
@@ -20,15 +20,15 @@ mechanism is only used for modules and is a naming convention. If any module has
...
 of the actual definition site, [odoc] has a feature whereby the 'canonical'
-location of a module, module type or type can be specified.
+location of a module, module type, or type can be specified.
 
-The biggest user of module aliases is Dune's {{:https://dune.readthedocs.io/en/stable/dune-files.html#library}wrapped libraries}.
+The biggest user of module aliases is Dune's {{:https://dune.readthedocs.io/en/stable/reference/dune/library.html#library}wrapped libraries}.
 This feature allows Dune to produce a library whose interface is exposed entirely
 though a single top-level module. It does this by mangling the names of the
-real implementation modules and generating the single top-level module that
+real implementation modules. Then it generates the single top-level module that
```
panglesd, line 28: Changing this to two sentences with a different tense makes it look like it has finished "doing so" after "mangling the names". Also, there is no notion of "things being done one after the other" that the word "then" implies!

```suggestion
real implementation modules ; and generating the single top-level module that
```

doc/features.mld
```diff
@@ -106,13 +106,13 @@ end
 module type B = A
 ]}
 
-For these simple module type declarations, where the right-hand side is just a
+For these simple module-type declarations, where the right-hand side is just a
```
panglesd, line 109: I don't think OCaml ever uses dash between module and types: https://ocaml.org/manual/5.2/modtypes.html

```suggestion
For these simple module type declarations, where the right-hand side is just a
```

However, I might not know an English rule that makes this dash correct!

doc/features.mld
```diff
@@ -350,7 +350,7 @@ have no destination. This example is rendered {{!Odoc_examples.Expansion.TypeSub
 The OCaml construct [module type of] allows the type of a module to be
 recovered. As usual, when OCaml performs this operation, it only retains
 the simplified signature, stripped of comments, includes, and more
-complex module type expressions. As with the previous sections, [odoc]
+complex module-type expressions. As with the previous sections, [odoc]
```
panglesd, line 353: It is "module type" expression and not module "type expression" but I still thinks that it should not have a dash as per https://ocaml.org/manual/5.2/modtypes.html !

panglesd (commented):
Thanks! I left some comments in the text.

doc/features.mld
christinerose: Thank you for the clarification!

doc/features.mld
christinerose: Normally there wouldn't be a `-` between module and type, but in instances like these, a `-` is put between two words like this acting as a single adjective; however, I think I see my confusion. 

It's not a type of module definition (which would be module-type), it's a `module type` definition. Correct? The definition of the expression `module type`. Is this what you mean here?

doc/features.mld
```diff
@@ -176,7 +176,7 @@ The above functor can be seen rendered {{!Odoc_examples.Expansion.Functors.Funct
 If part of your module signature comes from an include of another
 module or module type, [odoc] keeps track of this and can render
 the included items in a clearly delimited and collapsible way. 
-For example, given the following:
+For example:
```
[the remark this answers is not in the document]
christinerose: Ah! I see what you mean! Thanks for the clarification.

doc/features.mld
```diff
@@ -411,7 +411,7 @@ module type S =
   end
 ]}
 
-whereas [odoc] preserves the fact that [Y] and [Z] are calculated from [X]. If the 
+Whereas [odoc] preserves the fact that [Y] and [Z] are calculated from [X]. If the
```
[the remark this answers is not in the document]
christinerose: It's not mandatory. It's best in technical docs to have shorter sentences when possible, and sentences that have a code block in the middle can be confusing. 

How about: "On the other hand, `odoc` preserves the fact.... "
As you're saying OCaml does X, but `odoc` does Y. Correct?

If you don't like that, we can go back to the original. You're right that it's the same sentence.

IRT li 427, I think it should be a new sentence starting with "The signature..."
How would this be:

"....the results would be different. For example: 
`code block`
 The signature of `T`, as calculated by OCaml, will be 
`code block`"

Although in a previous note you said that "For example:" doesn't work because it's not the entire example, so perhaps we keep it as it is, but it still needs to start with "the signature....will then be" -- "then" is in a weird place along with the code block.

doc/features.mld
jonludlam: Correct - a "module type" is a different category of thing than a "module". I'd be happy with monospace or just simply "module type definition" - I don't think hyphenating is right in this context.

doc/features.mld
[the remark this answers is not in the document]
panglesd: For line 427, your suggestion:

> "....the results would be different. For example:
`code block`
The signature of T, as calculated by OCaml, will be
`code block`"

looks good I think!

Yes, I agree that sentences with code blocks in the middle is confusing and should be avoided! For some reason, "Whereas, ..." does not a complete sentence, _whereas_ "On the other hand, ..." does look to be a complete sentence!
But I'll leave you the judge on this one!

panglesd:
Thanks!

PR #1173, opened 2024-07-25
Fix the type_expr loader in OCaml 4.14 and later

Octachron:
Currently, generating documentation from ml files
```ocaml
(* test.ml *)
let repeat x y = x ,y, x, y
```
makes odoc misprint equal type variables as distinc type variables:
```ocaml
val repeat : 'a -> 'b -> 'c * 'd * 'e * 'f
```
starting from OCaml 4.14 .

This issue stems from the uses of physical equalities on non-representative type expressions in the name map of the type expression loader of odoc. Consequently, different type nodes are attributed different names by odoc, even if they belong to the same class of equal-after-unification type nodes. 

This issue typically doesn't arise when using mli files, because then we read source type expressions directly, without doing any unification (in most cases?).

This PR fixes this issue by ensuring that the naming maps only use representative type expressions.
Fortunately, this requires mostly tweaks to the  `Compat` module.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -56,6 +56,9 @@
 - Fix wrong links to standalone comments in search results (#1118, @panglesd)
 - Remove duplicated or unwanted comments (@Julow, #1133)
   This could happen with inline includes.
+- Fix misprinting of type variables from ml files for OCaml 4.14 and later
+  (multiple occurences of the same type variable could be named differently)
+  (@octachron, #1173)
 
 
 # 2.4.0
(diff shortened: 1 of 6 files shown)


Julow (approved):
Thanks! I've added a test case in https://github.com/Octachron/odoc/pull/1 if you agree with it.

I think this deserves a change entry.

Octachron:
I have merged the test and added a changes entry, thanks !
(I wasn't sure were to add a test since it not strictly speaking a generator test).

Julow (approved):
Thanks !

PR #1174, opened 2024-07-26
Remove ClassName and ClassTypeName

Julow:
They are replaced by TypeName as they are little risk of mixing them. This is extracted from Jon's `parameterised-paths` branch.


jonludlam:
Thanks!

PR #1175, opened 2024-07-29
line editing driver.mld

christinerose:
Minor line editing for grammar, punctuation, capitalisation, etc. Some questions for clarification are also in the comments.

### DIFF
--- a/doc/driver.mld
+++ b/doc/driver.mld
@@ -93,7 +93,7 @@ Using [odoc] is a three-phase process:
 {ol {- Compilation: odoc compile
 }}
 
-This takes as input either [.mld] files containing pure [odoc] markup, or the output from the compiler in the form of [.cmti], [.cmt], or [.cmi] files (in order of preference). For [.mld] files, this step simply translates them into [odoc]'s internal format and writes the corresponding file. For example, given the input [foobar.mld], [odoc] will output [page-foobar.odoc]. There are no dependencies for compiling [.mld] files beyond the parent as outlined above.
+This takes as input either [.mld] files containing pure [odoc] markup or the output from the compiler in the form of [.cmti], [.cmt], or [.cmi] files (in order of preference). For [.mld] files, this step simply translates them into [odoc]'s internal format and writes the corresponding file. For example, given the input [foobar.mld], [odoc] will output [page-foobar.odoc]. There are no dependencies for compiling [.mld] files beyond the parent as outlined above.
 
 For modules, compilation is the point where [odoc] performs some initial expansion and resolution operations, a process that usually introduces dependencies. For a given input [/path/to/file.cmti], it will output the file [/path/to/file.odoc] unless the [-o] option is used to override the output file. If there were [.cmi] dependencies required for OCaml to compile a particular module, then there will be equivalent [.odoc] dependencies needed for the [odoc compile] step. [odoc] will search for these dependencies in the paths specified with the [-I] directive on compilation. [odoc] provides a command to help with this: [odoc compile-deps].
 
--- a/doc/driver.mld
+++ b/doc/driver.mld
@@ -807,7 +807,7 @@ let index_generate ?(ignore_output = false) occurrence_file =
 We turn the JSON index into a JavaScript file. In order to never block the UI, this file will be used as a web worker by [odoc] to perform searches:
 
 - The search query will be sent as a plain string to the web worker, using the standard mechanism of message passing.
-- The web worker has to send back the result as a message to the main thread, containing the list of result. Each entry of this list must have the same form it had in the original JSON file.
+- The web worker has to send back the result as a message to the main thread, containing the list of results. Each entry of this list must have the same form it had in the original JSON file.
 - The file must be given to the [odoc-support] URI.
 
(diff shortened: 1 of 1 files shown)


christinerose:
For li.476:

"Many of the units will be 'hidden', meaning that Dune will mangle their name
in order to namespace them. This is achieved by prefixing the namespace module and
a double underscore, so we can tell a module is intended to be hidden. The following predicate tests for that condition:"

Does this read correctly? (i.e., is this phrase "the existence of a double underscore that" unnecessary?)

doc/driver.mld
```diff
@@ -511,7 +511,7 @@ type unit = {
 
 For [odoc] libraries, we infer the implementation and interface source file path
 from the library name. We list them in a file, passed to [odoc source-tree], to
-generate [src-source.odoc]. This file contains the source hierarchy, and will be
+generate [src-source.odoc]. This file contains the source hierarchy. It will be
 linked and passed to [html-generate] just as other pages and compilation units.
 
 It is used as the [source-parent] for all units for which we could provide
```
christinerose, line 517:
```suggestion
It is used as the [source-parent] for all units with provided
```

Is this still accurate?

doc/driver.mld
```diff
@@ -675,7 +675,7 @@ let compile_mlds () =
...
+Now we get to the compilation phase. For each unit, we query its dependencies, then recursively call to compile these dependencies. Once completed, we compile the unit itself. If the unit has already been compiled, we don't do anything. Note that we aren't checking the hashes of the dependencies which a build system should do to ensure that the module being compiled is the correct one. Again we benefit from the fact that we're creating the docs for one leaf package and that there must be no module name clashes in its dependencies. The result of this function is a list of the resulting [odoc] files.
```
christinerose, line 678:
```suggestion
Now we get to the compilation phase. For each unit, we query its dependencies, then recursively call to compile these dependencies. Once completed, we compile the unit itself. If the unit has already been compiled, we don't do anything. Note that we aren't checking the dependencies' hashes, which a build system should do to ensure that the module being compiled is the correct one. Again, we benefit from the fact that we're creating the docs for one leaf package and that there must be no module name clashes in its dependencies. The result of this function is a list of the resulting [odoc] files.
```

christinerose:
For hard-code / hardcode / hard code, i've found three different ways to write this in my research. Microsoft says always hyphenate. Others say it's two words: "We're going to hard code it." ... unless it's an adjective "The hard-coded file" -- or should it be "We're going to hardcode it" and "The hardcoded file"?

panglesd:
> For li.476:
>
> "Many of the units will be 'hidden', meaning that Dune will mangle their name
in order to namespace them. This is achieved by prefixing the namespace module and
a double underscore, so we can tell a module is intended to be hidden. The following predicate tests for that condition:"
>
> Does this read correctly? (i.e., is this phrase "the existence of a double underscore that" unnecessary?)

Yes, your proposition is correct, but the "by the existence of a double underscore" part makes it more precise, so I would be in favor of keeping it (we've had people confused by not knowing the convention that double underscore modules are meant to be hidden).

> For hard-code / hardcode / hard code, i've found three different ways to write this in my research. Microsoft says always hyphenate. Others say it's two words: "We're going to hard code it." ... unless it's an adjective "The hard-coded file" -- or should it be "We're going to hardcode it" and "The hardcoded file"?

I don't know, sorry, but I'm happy with any consistent solution you prefer!

doc/driver.mld
panglesd: Yes!

doc/driver.mld
panglesd: Not sure if it is correct, but my low English level makes me unsure.

Here is is not that dependencies have hashes and we can take them. It is that we compute a "hash" function to get the hashes of the dependencies. Does that makes sense?

christinerose: I put it back the way it was. Thanks for the clarification!

christinerose:
> I don't know, sorry, but I'm happy with any consistent solution you prefer!

Wikipedia shows all three! Cambridge Dictionary shows one word: hardcoded, and I like the way that looks, so let's go with that!

christinerose:
> Yes, your proposition is correct, but the "by the existence of a double underscore" part makes it more precise, so I would be in favor of keeping it (we've had people confused by not knowing the convention that double underscore modules are meant to be hidden).

Thanks for your input. I'll leave it alone. 🙂

christinerose:
What happens to this PR since I inadvertently included this doc [in the #1178 PR](https://github.com/ocaml/odoc/pull/1178), @panglesd? Do I make all these changes there instead and delete this one? Or will the both merge without conflicts (or at least too many conflicts)?

panglesd:
The problem is that this PR has received some new commits, which are not included in #1178, and reciprocally for #1178, which makes the review of each of them hard: if you see a problem in one PR, maybe it has already been fixed in the other PR!
What I can do is pick the relevant commits from this PR and add them to #1178, so we can close this PR and focus on #1178. Does that sound good to you?

panglesd:
Merged in #1178.

PR #1176, opened 2024-07-29
Simplify idents

Julow:
Use 'type_' idents instead of 'class_' and 'class_type'. Use 'module_' instead of 'functor_parameter' and 'result'. This allows to remove some code.
This is extracted from Jon's `parameterised-paths` branch


jonludlam:
Hrm, the failure in CI actually looks a bit bad, even though it's in the docs build!

PR #1177, opened 2024-07-29
 Fewer conversions between Names and strings

Julow:
The typed `` `Dot `` constructors allow to propagate the typed names further. The `Find` module is changed to avoid converting names to strings in many cases.
This is extracted from Jon's `parameterised-paths` branch


jonludlam:
Looks good I think. As I'm paranoid though, I'd like to check this by running it on some JS libs before and after this PR to confirm there are no changes. The driver ought to make this simple enough to try.

PR #1179, opened 2024-07-29
Driver: Separate types for "package" and "odoc unit".

panglesd:
The original reason for this is to allow adding odoc units that are not part of a package (eg pages that list packages, ...).

But I think it also makes it more readable:
- It factors path computations from the various "package extractors"[^1], in the end they have less things to do,
- The paths for an odoc unit are computed once, all at the same place, in a place which does nothing else,
- Understanding the code that runs commands is much simpler, since it has less things to do (run every command once and in the right order).

Indexes yet have to be implemented, so this PR is not finished yet.

[^1]: By "package extractors" I mean the `of_libs` function (creating packages from the opam switch), the `of_dune_build` function (creating packages from `_build/default/...`) and the `of_voodoo` function.

### DIFF
--- a/src/driver/compile.ml
+++ b/src/driver/compile.ml
@@ -10,27 +10,30 @@ let mk_byhash (pkgs : Odoc_unit.intf Odoc_unit.unit list) =
     Util.StringMap.empty pkgs
 
 let init_stats (units : Odoc_unit.t list) =
-  let total, total_impl, non_hidden, mlds =
+  let total, total_impl, non_hidden, mlds, indexes =
     List.fold_left
-      (fun (total, total_impl, non_hidden, mlds) (unit : Odoc_unit.t) ->
+      (fun (total, total_impl, non_hidden, mlds, indexes) (unit : Odoc_unit.t) ->
         let total = match unit.kind with `Intf _ -> total + 1 | _ -> total in
         let total_impl =
           match unit.kind with `Impl _ -> total_impl + 1 | _ -> total_impl
         in
+        let indexes = Fpath.Set.add unit.index.output_file indexes in
         let non_hidden =
           match unit.kind with
           | `Intf { hidden = false; _ } -> non_hidden + 1
           | _ -> non_hidden
         in
         let mlds = match unit.kind with `Mld -> mlds + 1 | _ -> mlds in
-        (total, total_impl, non_hidden, mlds))
-      (0, 0, 0, 0) units
+        (total, total_impl, non_hidden, mlds, indexes))
+      (0, 0, 0, 0, Fpath.Set.empty)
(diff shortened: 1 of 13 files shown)


src/driver/compile.ml
```diff
@@ -107,252 +77,173 @@ let find_partials odoc_dir =
...
+  let mod_results = Fiber.List.map compile_mod to_build in
+  let compile_mld (unit : Odoc_unit.mld Odoc_unit.unit) =
+    let includes = Fpath.Set.of_list unit.include_dirs in
+    Odoc.compile ~output_dir:unit.output_dir ~input_file:unit.input_file
+      ~includes ~parent_id:unit.parent_id;
+    Atomic.incr Stats.stats.compiled_mlds
   in
-  let mods =
-    List.filter_map (function Ok x -> Some x | Error _ -> None) mod_results
+  let _compiled_mlds = Fiber.List.map compile_mld mld_units in
```
Julow, line 152: That's a `unit list`. Do you mean to use `Fiber.List.iter` ?

src/driver/odoc_unit.ml
```diff
@@ -0,0 +1,231 @@
...
+    let pages = List.concat pages in
+    let libs = List.concat libs in
+    { pages; libs }
+  in
+  let index_of pkg =
+    let pkg_args = pkg_args_of pkg in
+    let output_file = Fpath.(index_dir / pkg.name / Odoc.index_filename) in
+    { pkg_args; output_file; json = false; search_dir = pkg.pkg_dir }
+  in
+  let rec of_intf hidden pkg libname (intf : Packages.intf) : intf unit =
```
Julow, line 93: There's too much duplicated code in these 3 functions. The code that differ between them is surprising small: `kind`, `odoc_file` and `odocl_file` differ between `of_intf` and `of_impl`. `of_mld` is very similar too.

Julow (approved):
Awesome! I love that this removes code from the most complex modules.

src/driver/odoc_unit.ml
panglesd: Unfortunately, the types makes it a bit difficult to factor all of those into a single `make_unit` function (`of_intf` returns an `intf unit` while `of_impl` returns an `impl unit`). I'll try!

panglesd: In the end the `make_unit` function returns a "`unit/2 unit`" (shadowing the unit type is not great...) and the extra part is added outside of `make_unit`.

src/driver/compile.ml
panglesd: Yes, thanks!

Julow:
Let's merge!

PR #1180, opened 2024-07-31
Line editing for dune.mld

christinerose:
Mostly minor formatting. Please review to ensure all is still accurate. 🙂

### DIFF
--- a/doc/dune.mld
+++ b/doc/dune.mld
@@ -100,13 +100,16 @@ contained within the module [Lib].
 {2 Hand-Written Top-Level Module}
 
 In some cases it's desirable to hand-write the top-level library module. This
-is usually done because some modules in the library are intended to
-be internal only and not exposed. Dune will notice that a module exists with
-the library's name ([lib.ml] in this case), so instead it will create the
-file [lib__.ml]. Its contents are identical to the previous section, with aliases for all modules. Like before, the canonical tags on these aliases are references to [Lib.A] and [Lib.B]. These module should be present in [lib.ml] as module aliases. If these are {e not} there, [odoc] won't be able to
-resolve the canonical references, and any items from these modules that are
-exposed elsewhere will be hidden. If the items are type aliases, they can be
-replaced, but otherwise they'll be rendered as unresolved links.
+is usually done because some modules in the library are intended to be internal
+only and not exposed. Dune will notice that a module exists with the library's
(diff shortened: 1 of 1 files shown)


doc/dune.mld
```diff
@@ -94,21 +94,21 @@ have double underscores meaning they are hidden. Only the non-hidden module
...
-is usually done because some of the modules within the library are intended to
+is usually done because some modules in the library are intended to
 be internal only and not exposed. Dune will notice that a module exists with
-the name of the library ([lib.ml] in this case), so instead it will create the
-file [lib__.ml]. The contents of this are identical to the previous section, 
+the library's name ([lib.ml] in this case), so instead it will create the
+file [lib__.ml]. Its contents are identical to the previous section, 
 with aliases for all modules. The canonical tags on the aliases are, as before,
-to [Lib.A] and [Lib.B]. These are references to module aliases that should be
+[Lib.A] and [Lib.B]. These are references to module aliases that should be
```
panglesd, line 108: `to` should likely be kept, or maybe replacing "are" with "point":

> The canonical tags on the aliases, as before, point to [Lib.A] and[Lib.b]

panglesd (approved):
Only one small comments. Thanks!

doc/dune.mld
panglesd: Another possibility:

> The canonical tags on the aliases are, as before, references to [Lib.A] and[Lib.B].

Is that better?

christinerose: Yes! The second one is more clear. 🙂

christinerose: Its contents are identical to the previous section, 
with aliases for all modules. Like before, the canonical tags on these aliases
are references to [Lib.A] and [Lib.B]. These references should be present in [lib.ml].

It's less repetitive, but is it still accurate?

panglesd: Unfortunately, not perfectly accurate :sweat_smile: It is not the references which are present in [lib.ml], but the target of the references...

I believe we are close to something good and accurate! I think this is correct:
> ```
> Its contents are identical to the previous section, with aliases for all modules. Like before, the canonical tags on these aliases are references to [Lib.A] and [Lib.B]. These module should be present in [lib.ml] as module aliases.
> ```

christinerose: See if li 108 reads better now.

PR #1182, opened 2024-07-31
Line editing interface.mld

christinerose:
Mostly formatting, but there was also some syntax tightening (removing unnecessary words), so please let me know if anything is inaccurate.

### DIFF
--- a/doc/interface.mld
+++ b/doc/interface.mld
@@ -13,7 +13,7 @@ We have a separate page describing the {{!page-ocamldoc_differences}markup diffe
 {2 CLI Interface}
 
 The way in which we invoke the [odoc] CLI is not trivial, and it requires careful
-ordering and accuarate arguments to produce correct linked documentation. It’s not expected that
+ordering and accurate arguments to produce correct linked documentation. It’s not expected that
 end users will invoke [odoc] by hand, but rather it will be driven by separate tools. As a consequence, it’s important to preserve the tools’ ability to create good documentation with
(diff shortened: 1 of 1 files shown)


doc/interface.mld
```diff
@@ -1,40 +1,40 @@
...
 We have a separate page describing the {{!page-ocamldoc_differences}markup differences from OCamldoc}.
 
 {2 CLI Interface}
 
-The way in which the [odoc] CLI is invoked is not trivial, and it requires careful
-ordering and correct arguments to produce correctly linked documentation. It’s not expected that
-end-users will invoke [odoc] by hand, but rather it will be driven by separate tools. As a consequence, it’s important to preserve the tools’ ability to create good documentation with
-each release of [odoc], so we’ll ensure backward compatibility of the CLI as much as possible.
+The way in which we invoke the [odoc] CLI is not trivial, and it requires careful
+ordering and accuarate arguments to produce correct linked documentation. It’s not expected that
```
panglesd, line 16: Not sure if "accurate" is an adjective that applies to CLI arguments, but I think we understand the idea and it avoids the repetition! However, there is a typo:
```suggestion
ordering and accurate arguments to produce correct linked documentation. It’s not expected that
```

"correct link documentation" and "correctly linked documentation" feels different and the latter feels ~correcter~ more correct to me.

panglesd (approved):
One small comment!

doc/interface.mld
christinerose: Nice catch!!

panglesd:
Thanks!

PR #1185, opened 2024-08-01
Add ability to call generate commands on asset units

panglesd:
The pipeline for asset was missing the generate command. This PR simply adds that!

The driver will need to be updated to follow the convention for installed asset described in https://github.com/ocaml/odoc/discussions/1113 (section: "Assets for installed packages").


panglesd:
(I'm not very proud of the name of the CLI option: `--asset-path`)

src/odoc/bin/main.ml
```diff
@@ -935,11 +935,11 @@ end = struct
           exit 1
 
     let generate extra _hidden output_dir syntax extra_suffix input_file
-        warnings_options source_file source_root sidebar =
+        warnings_options source_file source_root sidebar asset_path =
       let source = source_of_args source_root source_file in
       let file = Fs.File.of_string input_file in
       Rendering.generate_odoc ~renderer:R.renderer ~warnings_options ~syntax
-        ~output:output_dir ~extra_suffix ~source ~sidebar extra file
+        ~output:output_dir ~extra_suffix ~source ~sidebar ~asset_path extra file
```
Julow, line 942: I'm not a fan of adding more arguments to this command. This argument makes no sense in most cases and needs to be checked. Similarly, `Document.Asset` seems a bit out of place, as it interacts with nothing else, it could be out of the document thing entirely.

What about adding a different command for generating assets ? `html-generate-asset`, `latex-generate-latex`, etc..

panglesd: The same is true for `--source` and `--source-root`, but OK, I'll do that!

panglesd: (the `Document.Asset` will be inconvenient to remove in my opinion, unless we remove the old way of attaching assets, which would be a good thing I think)

panglesd:
I have added support in the driver for the asset pipeline, following the convention.

This means, we now have images working for packages following the odig convention!

image

src/driver/dune_style.ml
```diff
@@ -53,6 +53,7 @@ let of_dune_build dir =
                       version = "1.0";
                       libraries = [ lib ];
                       mlds = [];
+                      assets = [] (* TODO *);
```
Julow, line 56: What could we do ?

src/driver/opam.ml
```diff
@@ -135,6 +135,10 @@ let pkg_to_dir_map () =
               | "doc" :: _pkg :: "odoc-pages" :: _ ->
                   Logs.debug (fun m -> m "Found odoc page: %a" Fpath.pp fpath);
 
+                  (Fpath.Set.add Fpath.(v prefix // fpath) odoc_pages, others)
+              | "doc" :: _pkg :: "odoc-assets" :: _ ->
```
Julow, line 139: Unrelated to this PR but pages and assets are taken from the list of installed files for a package. I find that unnecessary and potentially buggy. A package could install doc in an other package or this file could be outdated (`dune install`), which might crash the driver.
I think the `doc/pkgname` folder should be scanned independently of `pkg_contents`. I think the same for libraries but there might be cases that require this?

Julow (approved):
Looks good :)

Julow:
There's a broken asset in `vg.lib.vg.Vg.P` because it's using an hardcoded path but I couldn't make the reference syntax work. Asset references from libraries might not work.

src/driver/dune_style.ml
panglesd: When dune has a notion of asset, we will be able to do something. Right now, nothing. I'll update the comment.

(However, I think something could be done for mld files. I'm not sure we need to spend time on that though)

panglesd:
This has been rebased on top of #1188. `html-generate` and `html-generate-asset` have been split. The old asset pipeline has been removed: `--child asset-...` for compile and `--asset` for `html-generate`.

src/document/renderer.ml
```diff
@@ -24,7 +24,6 @@ type input =
 type 'a t = {
   name : string;
   render : 'a -> Types.Block.t option -> Types.Document.t -> page list;
-  extra_documents : 'a -> input -> Types.Document.t list;
```
Julow, line 27: Awesome !

src/odoc/rendering.ml
```diff
@@ -128,22 +115,29 @@ let generate_source_odoc ~syntax ~warnings_options ~renderer ~output
   | Page_content _ | Unit_content _ | Asset_content _ ->
       Error (`Msg "Expected an implementation unit")
 
+let generate_asset_odoc ~warnings_options:_ ~renderer ~output ~asset_file
+    ~extra_suffix extra file =
+  Odoc_file.load file >>= fun unit ->
+  match unit.content with
+  | Odoc_file.Asset_content unit ->
+      let doc = Renderer.document_of_asset asset_file unit in
+      render_document renderer ~output ~sidebar:None ~extra_suffix ~extra doc;
```
Julow, line 124: There was a lot of simplification in this PR, why not continue and remove assets from the document and the renderer ?
`document_of_asset` seems very orthogonal to the rest of the document (it's just the `Asset` constructor which shares nothing with the other document types) and because of `render_document` and the `renderer` type, the content of the asset file has to go through a `Format` buffer, which seems totally unnecessary.
I think all of this could be removed.

panglesd: True! (I started that and decided to stop, thinking you would not see it, so I stashed everything. Time to stash pop!)

Julow (commented):
I think the driver is broken on main as it calls `odoc html-generate-impl` but I could test this a bit.

src/odoc/fs.ml
```diff
@@ -92,6 +104,27 @@ module File = struct
...
+    try
+      with_ic src (fun ic ->
+          mkdir_p (dirname dst);
+          with_oc dst (fun oc ->
+              let len = 1024 in
+              let buf = Bytes.create len in
+              let rec loop () =
+                let read = input ic buf 0 len in
+                output oc buf 0 read;
+                if read = len then loop ()
```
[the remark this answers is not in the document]
panglesd: Indeed, thanks!

panglesd:
> I think the driver is broken on main as it calls odoc html-generate-impl but I could test this a bit.

I'm being pragmatic and fixing this in this PR.

Julow (approved):
It's perfect now, let's merge !

PR #1187, opened 2024-08-19
Add a frontmatter to mld pages

panglesd:
This PR adds a frontmatter to mld pages.

The frontmatter consists in this PR to a list of string key values. Code blocks of the form:
```
{@meta[
key1: value1
key2: value2
]}
```
are removed from the content and the key-value pairs are added to the frontmatter of the page.

The frontmatter is added to the odoc file, including the page's "root", allowing to access it quickly without loading the whole content.

---

The idea of having a frontmatter originates from https://github.com/ocaml/odoc/discussions/1097 (see the "Metadata block" section).
In this discussion, the metadata provided by the frontmatter is:
- The package and library dependencies of the page (to be able to do cross-package referencing)
- The order in the table of content, for index pages.

The parsing of the data is not included in this PR. (Also, I am not sure which other data than the two items above could fit in the metadata block.)

About this: @dbuenzli wrote:

> As a driver author I don't want to parse this metadata block. If the driver will need that info odoc should be able to extract it for me in a simple from like e.g. we have for deps (I don't mind lines based, json or sexp).

I agree. However, this is not part of this PR (it will be in the PR defining the syntax for libs and pkgs dependencies, I guess).

### DIFF
--- a/src/model/comment.ml
+++ b/src/model/comment.ml
@@ -145,19 +145,16 @@ let extract_frontmatter docs : _ =
       (fun line -> Astring.String.cut ~sep:":" line)
       lines
   in
-  let extracted =
-    let rec aux acc = function
-      | [] -> None
-      | doc :: l -> (
+  let fm, content =
+    let fm, rev_content =
+      List.fold_left
+        (fun (fm_acc, content_acc) doc ->
           match doc.Location_.value with
           | `Code_block (Some "frontmatter", content, None) ->
-              Some
-                ( parse_frontmatter content.Location_.value,
-                  List.rev_append acc l )
-          | _ -> aux (doc :: acc) l)
+              (parse_frontmatter content.Location_.value :: fm_acc, content_acc)
+          | _ -> (fm_acc, doc :: content_acc))
+        ([], []) docs
     in
-    aux [] docs
+    (List.concat fm, List.rev rev_content)
   in
-  match extracted with
-  | None -> (None, docs)
-  | Some (fm, docs) -> (Some fm, docs)
+  (fm, content)
--- a/src/model/comment.ml
+++ b/src/model/comment.ml
@@ -150,7 +150,7 @@ let extract_frontmatter docs : _ =
       List.fold_left
         (fun (fm_acc, content_acc) doc ->
           match doc.Location_.value with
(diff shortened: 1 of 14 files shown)


src/model/comment.ml
```diff
@@ -130,15 +130,31 @@ and link_content_of_inline_elements l =
   l |> List.map link_content_of_inline_element |> List.concat
 
 let find_zero_heading docs : link_content option =
-  let rec find_map f = function
-    | [] -> None
-    | x :: l -> (
-        match f x with Some _ as result -> result | None -> find_map f l)
-  in
-  find_map
+  Odoc_utils.List.find_map
```
Julow, line 133: Let's open `Odoc_utils` at the top of the module.

src/model/comment.ml
```diff
@@ -130,15 +130,31 @@ and link_content_of_inline_elements l =
...
+let extract_frontmatter docs : _ =
+  let parse_frontmatter s =
+    let lines = Astring.String.cuts ~sep:"\n" s in
+    Odoc_utils.List.filter_map
+      (fun line -> Astring.String.cut ~sep:":" line)
+      lines
+  in
+  let fm, content =
+    let fm, rev_content =
+      List.fold_left
```
Julow, line 150: One advantage of the frontmatter is that it's immediately recognizable. With this syntax, it can be confused for a code block and present several times all over the document.
This doesn't sound much better than `@` tags, which would be shorter to type and more easily recognized.

I think a proper frontmatter should be a new syntax and expected at the beginning of the page only. For example using the markdown syntax (but without the yaml):
```
---
foo: bar
---
{0 Title}
```

jonludlam: We've agreed to consider this, but not to hold up the PR until it's decided, as several things depend upon this. We'll revisit the question before releasing 3.0.

panglesd:
Disregarding the syntax of the frontmatter, this PR is ready for review, as it is a base for other work (eg order for pages).
(@Julow you did not approve, I don't know if it was because of the syntax or because of a not finished review)

Julow (approved):
I do not approve the syntax indeed but we agreed to change it before releasing. Let's merge !

PR #1188, opened 2024-08-20
Separate commands for html generation of page+units and implementation

panglesd:
This was suggested in https://github.com/ocaml/odoc/pull/1185#discussion_r1701861980

The benefit is cleaner man pages, safer CLI validation.

This PR also takes this opportunity to do some cleanup, by removing the old and never used way of rendering sources:
  - Remove the "source dir" identifier. We have enough identifier types!
  - Remove the source_tree odoc unit.
  - Remove the `--source-root` option which is not anymore usable in odoc 3.

Once this is merged, I'll rebase #1185 on top and apply @Julow's comment!


Julow (approved):
Awesome! All this code that we didn't need is gone :)

jonludlam:
Ooh I do like a good cleanup :-)

jonludlam:
Just want to check though: should it be `odoc html-generate-source` or `src` rather than `odoc html-generate-impl`? Asking for opinions, I don't have a strong opinion.

jonludlam:
Also we should have some negative tests - what happens when you do `odoc html-generate` on an impl file and vice-versa.

panglesd:
In the odoc dev meeting, we decided to rename it to `compile-source` and use the source file as positional argument while the implementation file is passed as an option. It'll make it easier when including dune or other file in source rendering.

src/odoc/rendering.ml
```diff
@@ -184,3 +145,38 @@ let targets_odoc ~resolver ~warnings_options ~syntax ~renderer ~output:root_dir
...
+          Renderer.traverse pages ~f:(fun filename _content ->
+              let filename =
+                Fpath.normalize @@ Fs.File.append root_dir filename
+              in
+              Format.printf "%a\n" Fpath.pp filename))
+        docs;
+      Ok ()
+  | Page_content _ | Unit_content _ | Asset_content _ ->
+      Error (`Msg "Expected an implementation unit")
+(* let docs = *)
```
Julow, line 168: Brand new commented-out code :)

Julow (approved):
I think this is ready to be merged :) I've left some comments but I think it's fine to discuss them after this is merged.

src/odoc/bin/main.ml
```diff
@@ -1038,6 +945,34 @@ end = struct
   end
 
   let targets ~docs = Targets.(cmd, info ~docs)
+
+  module Targets_source = struct
```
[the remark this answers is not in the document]
Julow: In fact we continued to maintain this command in Odoc 2 and there's a high chance that Odig uses it. Let's keep `-targets-sources`.

src/odoc/rendering.ml
panglesd: RIP

PR #1189, opened 2024-08-21
Add the ability to remap identifiers during HTML generation

jonludlam:
This is to allow the publishing of 'partial docsets' - the idea being that the driver proceeds as usual up until HTML generation, and at that point only generates HTML pages for the packages you wish to publish on your site, and remap links to other packages to ocaml.org or other site.

### DIFF
--- a/src/html/link.ml
+++ b/src/html/link.ml
@@ -15,16 +15,17 @@ module Path = struct
 
   let remap config f =
     let l = String.concat "/" f in
-    match
-      List.find_opt
-        (fun (prefix, _replacement) -> Astring.String.is_prefix ~affix:prefix l)
-        (Config.remap config)
-    with
-    | None -> None
-    | Some (prefix, replacement) ->
-        let len = String.length prefix in
-        let l = String.sub l len (String.length l - len) in
-        Some (replacement ^ l)
+    try
+      let prefix, replacement =
+        List.find
+          (fun (prefix, _replacement) ->
+            Astring.String.is_prefix ~affix:prefix l)
+          (Config.remap config)
+      in
+      let len = String.length prefix in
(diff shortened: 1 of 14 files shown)


panglesd (commented):
Implementation looks good to me. This put a lot of pressure on the targeted website (likely ocaml.org) not changing its url format!

It would be nice to be able to test it by having the odoc_driver rewrite some links to ocaml.org.

src/html/link.ml
```diff
@@ -48,46 +67,55 @@ let rec drop_shared_prefix l1 l2 =
...
+          in
+          let relative_target =
+            if Config.semantic_uris config then
+              remove_index_html relative_target
+            else relative_target
+          in
+          match (relative_target, anchor) with
+          | [], "" -> "#"
+          | page, "" -> String.concat "/" page
+          | page, anchor -> String.concat "/" page ^ "#" ^ anchor))
```
panglesd, line 121: Feel free to ignore: We can save one line by reusing `add_anchor` :)
```diff
-           | page, "" -> String.concat "/" page
-           | page, anchor -> String.concat "/" page ^ "#" ^ anchor))
+           | page, _ -> add_anchor @@ String.concat "/" page))
```

jonludlam: Good idea!

jonludlam:
> Implementation looks good to me. This put a lot of pressure on the targeted website (likely ocaml.org) not changing its url format!

I think we're already in that situation. We discussed leaving redirects from the old paths to the new ones when we change to the odoc3 layout.

> It would be nice to be able to test it by having the odoc_driver rewrite some links to ocaml.org.

I thought about that, but right now that'd break all the source links as we don't have that on ocaml.org yet, so I think we should hold off until we've got a good way to test it.

PR #1190, opened 2024-08-26
Show packages and libraries in the breadcrumbs

Julow:
This is still work in progress as pages that are inside a library are not recognized and the library name might be inconsistent with the one passed to the link command.

This makes the breadcrumbs more readable by removing uninterested components and make the beginning of the module path more recognizable.
The `` `Library `` constructor is added to identifiers to propagate this information.


panglesd:
This change requires a change in sherlodoc, likely due to the change in the identifier type.

Do you have a branch with the required modification, so that I can test this PR?

Julow:
Oops I forgot about that. Here's the fix: https://github.com/jonludlam/sherlodoc/pull/1 (branch [`odoc3_breadcrumbs_library`](https://github.com/Julow/sherlodoc/tree/odoc3_breadcrumbs_library))

Julow:
Closing as this is probably not the right approach. Maybe https://github.com/ocaml/odoc/pull/1196 is better.

PR #1191, opened 2024-08-26
Consistent output file names in all backends

Julow:
The latex and man backends were prefixing names differently than the HTML backend. This can create naming conflicts.

This removes code that was caused problems while working on https://github.com/ocaml/odoc/pull/1190


src/document/url.mli
```diff
@@ -27,6 +27,9 @@ module Path : sig
 
   val string_of_kind : kind -> string
 
+  val pp_kind_prefix_for_output : Format.formatter -> kind -> unit
+  (** Print the ["kind-"] prefix used in output files. *)
```
panglesd, line 31: This seems to be for output file in html flat mode, but if I understand correctly used as labels in latex mode (and not as filenames).

Maybe:

```suggestion
  val pp_disambiguating_prefix : Format.formatter -> kind -> unit
  (** Print the ["kind-"] prefix used to disambiguate urls in "flat modes": e.g. latex labels and output files in [--flat] HTML output *)
```

panglesd (commented):
I left a small comments, otherwise the implementation looks good to me.

Note that this changes labels in latex output. Do we have any guarantee that those won't change, just like we try not modifying URLs in HTML output? Can there be any handwritten  latex referencing odoc labels?

src/document/url.mli
Julow: And the man output too. Thanks!

CHANGES.md
```diff
@@ -42,6 +42,8 @@
 - Allow `][` in code blocks (@Julow, #1149)
   This was interpreted as "code blocks with result", which now mandate a delimiter:
   `{delim@lang[ code ]delim[ result ]}`
+- Output file paths in the man and latex backends changed to avoid name clashes
```
panglesd, line 45:
```suggestion
- Output file paths and labels in the man and latex backends changed to avoid name clashes
```
Just so that if someone is using the latex backend and handwriting labels, they can find the cause of their new problem in the changelog.

panglesd (approved):
LGTM (apart from the CI)!

src/html/generator.ml
```diff
@@ -476,7 +476,7 @@ end
 
 module Page = struct
   let on_sub = function
-    | `Page _ -> None
+    | `Pge _ -> None
```
panglesd, line 479: Is this change intended?

Julow: Fixed, thanks

PR #1192, opened 2024-08-27
Careful library deps

jonludlam:
Make sure the set of libraries provided during linking only comes from the dependency libraries of the library being linked.

Also ensure that the modules in the libraries passed are in the normal include path.

### DIFF
--- a/src/driver/dune_style.ml
+++ b/src/driver/dune_style.ml
@@ -34,11 +34,14 @@ let of_dune_build dir =
             let cmtidir =
               Fpath.(path / Printf.sprintf ".%s.objs" libname / "byte")
             in
+            let all_lib_deps = Util.StringMap.empty in
+            (* TODO *)
             let pkg_dir = Fpath.rem_prefix dir path |> Option.get in
             ( pkg_dir,
               Packages.Lib.v
                 ~libname_of_archive:(Util.StringMap.singleton libname libname)
-                ~pkg_name:libname ~dir:path ~cmtidir:(Some cmtidir) ))
+                ~pkg_name:libname ~dir:path ~cmtidir:(Some cmtidir)
+                ~all_lib_deps ))
           libs
       in
       let packages =
--- a/src/driver/odoc_unit.ml
+++ b/src/driver/odoc_unit.ml
@@ -43,58 +43,53 @@ let of_packages ~output_dir ~linked_dir ~index_dir (pkgs : Packages.t list) :
   in
   let index_dir = match index_dir with None -> output_dir | Some dir -> dir in
   (* This isn't a hashtable, but a table of hashes! Yay! *)
-  let hashtable =
+  let hashtable, lib_dirs =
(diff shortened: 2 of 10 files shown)


src/driver/odoc_unit.ml
```diff
@@ -195,6 +194,7 @@ let of_packages ~output_dir ~linked_dir ~index_dir (pkgs : Packages.t list) :
     let name = mld_path |> Fpath.rem_ext |> Fpath.basename |> ( ^ ) "page-" in
     let unit =
       make_unit ~name ~kind ~rel_dir ~input_file:mld_path ~pkg ~include_dirs
+        ~lib_deps:[]
```
panglesd, line 197: Is there anything bigger but still sensible we could use here?

- We likely cannot add the union of the the `lib_deps` from the libraries of the package: I think there might be different deps with the same name? (not sure here).

I know we will provide a way to extend this, but having a bigger default would be nice!

src/odoc/bin/main.ml
```diff
@@ -688,6 +688,7 @@ end = struct
           current_dir;
         }
     in
+    let directories = directories @ List.map ~f:snd lib_roots in
```
panglesd, line 691: I was about to say that this requires an update to the man for the `-L` option, but it was actually already saying that!

So whenever we can write `{!/libname/Module}` we can also write `{!Module}`.

src/driver/dune_style.ml
```diff
@@ -34,11 +34,14 @@ let of_dune_build dir =
             let cmtidir =
               Fpath.(path / Printf.sprintf ".%s.objs" libname / "byte")
             in
+            let all_lib_deps = Util.StringMap.empty in
+            (* TODO *)
```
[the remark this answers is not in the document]
jonludlam: I was putting it off, it's a bit of a PITA, especially for the dune mode!

src/driver/odoc_unit.ml
```diff
@@ -43,58 +43,53 @@ let of_packages ~output_dir ~linked_dir ~index_dir (pkgs : Packages.t list) :
   in
   let index_dir = match index_dir with None -> output_dir | Some dir -> dir in
   (* This isn't a hashtable, but a table of hashes! Yay! *)
-  let hashtable =
+  let hashtable, lib_dirs =
```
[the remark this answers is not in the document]
jonludlam: Yes, let's do that

src/driver/odoc_unit.ml
jonludlam: I think the libraries in the package is what we decided before.

src/odoc/bin/main.ml
jonludlam: I wasn't sure whether adding the paths to the include path was the best approach, or whether we should do something else elsewhere?

jonludlam:
Superseded by #1208

PR #1193, opened 2024-08-30
Specify children order in frontmatter

panglesd:
Now that we have a convention for hierarchical pages, we need a way for the author to specify the order.

This PR does that by using the frontmatter, using this syntax:

```
children: child1 dir1/ child2 dir2/
```

The trailing `/` for directories is needed to distinguish them from files.

This PR is mostly ready, but:
- TODO warnings:
  - [x] warn when a children field is given in the frontmatter of a non index page,
  - [x] warn when an entry of the `children` list is not found (should it happen during `linking` of the page or during the `compile-index` command?)
  - [x] Test those warnings
- Decide what to do when a children is omitted from the list: 
   - :x: do not include it in the toc (I'm in favor, this is what is implemented)? 
   - [X] :heavy_check_mark:  Include it at the end of the list (I'm not in favor)?
   - [X] :heavy_check_mark: Warn the user?
-  Decide what to do when there is no `children` list in an index page:
   - :x: Warn the user? (not in favor!)
   - [X] :heavy_check_mark: Use alphabetical order on unit names (in favor, this is what the current PR does)
   - :x: Use the alphabetical order on page main title (not in favor!)
- [x] Clean up/factor the implementation of `PageToc`.
- [x] Fix compat CI

### DIFF
--- a/src/model/comment.ml
+++ b/src/model/comment.ml
@@ -147,20 +147,16 @@ let find_zero_heading docs : link_content option =
     docs
 
 let extract_frontmatter docs : _ =
-  let parse_frontmatter s =
-    let lines = Astring.String.cuts ~sep:"\n" s in
-    List.filter_map (fun line -> Astring.String.cut ~sep:":" line) lines
-  in
   let fm, content =
     let fm, rev_content =
       List.fold_left
         (fun (fm_acc, content_acc) doc ->
           match doc.Location_.value with
           | `Code_block (Some "meta", content, None) ->
-              (parse_frontmatter content.Location_.value :: fm_acc, content_acc)
+              (content.Location_.value :: fm_acc, content_acc)
           | _ -> (fm_acc, doc :: content_acc))
         ([], []) docs
     in
-    (List.concat fm, List.rev rev_content)
+    (fm |> String.concat "\n" |> Frontmatter.parse, List.rev rev_content)
   in
   (fm, content)
--- a/src/model/comment.ml
+++ b/src/model/comment.ml
@@ -147,16 +147,13 @@ let find_zero_heading docs : link_content option =
     docs
 
 let extract_frontmatter docs : _ =
-  let fm, content =
-    let fm, rev_content =
-      List.fold_left
-        (fun (fm_acc, content_acc) doc ->
-          match doc.Location_.value with
-          | `Code_block (Some "meta", content, None) ->
-              (content.Location_.value :: fm_acc, content_acc)
-          | _ -> (fm_acc, doc :: content_acc))
-        ([], []) docs
-    in
-    (fm |> String.concat "\n" |> Frontmatter.parse, List.rev rev_content)
+  let fm, rev_content =
+    List.fold_left
+      (fun (fm_acc, content_acc) doc ->
+        match doc.Location_.value with
+        | `Code_block (Some "meta", content, None) ->
+            (Frontmatter.parse content, content_acc)
+        | _ -> (fm_acc, doc :: content_acc))
+      (Frontmatter.empty, []) docs
   in
-  (fm, content)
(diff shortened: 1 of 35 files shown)


src/driver/odoc_driver.ml
```diff
@@ -582,6 +582,20 @@ let run libs verbose packages_dir odoc_dir odocl_dir html_dir stats nb_workers
       (fun () -> render_stats env nb_workers)
   in
 
+  List.iter
```
Julow, line 585: Is all of this just debug ? Otherwise, it seems too verbose and opaque.

src/model/comment.ml
```diff
@@ -147,20 +147,16 @@ let find_zero_heading docs : link_content option =
...
-    List.filter_map (fun line -> Astring.String.cut ~sep:":" line) lines
-  in
   let fm, content =
     let fm, rev_content =
       List.fold_left
         (fun (fm_acc, content_acc) doc ->
           match doc.Location_.value with
           | `Code_block (Some "meta", content, None) ->
-              (parse_frontmatter content.Location_.value :: fm_acc, content_acc)
+              (content.Location_.value :: fm_acc, content_acc)
```
Julow, line 156: The previous code was more logical (parse each blocks and concatenate the values rather than concatenate the blocks and parse something no one wrote) and allowed more graceful handling of errors with locations.
Anyway, that's a temporary syntax.

src/model/sidebar.ml
```diff
@@ -0,0 +1,150 @@
...
+    let dir = empty_t None in
+    List.iter (add dir) l;
+    dir
+end
+
+type toc = PageToc.t
+
+type library = { name : string; units : Paths.Identifier.RootModule.t list }
+
+type page_hierarchy = { hierarchy_name : string; pages : toc }
```
Julow, line 148: I didn't expect the page hierarchy to have a name and be an extra level on top of a `toc`. What if this hierarchy contains an `index.mld` and a children order ? Is here an extra indentation with the name repeated ?

src/driver/odoc_driver.ml
panglesd: Yes, this is only showing filtered logs for debugging. I'll remove it.

src/model/comment.ml
panglesd: Indeed, we are going to have the frontmatter consistently at the top, so there won't be any distinction.
So, I suggest leaving it as is.

src/model/sidebar.ml
panglesd: This "hierarchy_name" corresponds to the name of the library's package: the one passed in the `-P` option of `odoc compile-index`.
It is currently [turned](https://github.com/ocaml/odoc/blob/master/src/document/sidebar.ml#L126) into a `"{hierarchy_name}'s Pages"` title:

```
# Package1's Pages

- The best way to do foo
  - Guide
  - example
  - tutorial 

# Package2's Pages

- Blazingly fast bar
  - Guide
  - example
  - tutorial 

# Library1's modules

- Foo
- Bar
```

We could possibly remove this `hierarchy_name`, but I would argue that it is better done in another PR (this PR is not introducing the behavior).

(Maybe:
```
# Pages

- The best way to do foo
  - Guide
  - example
  - tutorial 
- Blazingly fast bar
  - Guide
  - example
  - tutorial 

# Libraries

- library1
  - Foo
  - Bar
```
)

Julow: Makes total sense! I think the hierarchy name is something we'd want in the breadcrumbs and so for clarity, we should also have it in the sidebar.

panglesd:
All warnings have been implemented, and the code has been cleaned up.

Only remains a review for the implementation details, on my side this is ready to be merged.

PR #1201, opened 2024-09-30
opam: Add strict constraints on OCaml < 5.3

Julow:
https://github.com/ocaml/odoc/pull/1200


panglesd:
For people confused like me: this one is #1200 but for the `v2.4` branch.

PR #1203, opened 2024-09-30
CI: Run the driver

Julow:
This builds the doc of `base` in CI to ensure that the driver works. This will also check that the pinned version of sherlodoc is compatible.


Julow:
Better done in https://github.com/ocaml/odoc/pull/1207

PR #1206, opened 2024-10-07
Count-occurrences: CLI consistency and driver support

panglesd:
Made the count-occurrences command a bit more consistent:
- output names are now `name.odoc-occurrences` instead of `occurrences-name.odoc` (similar to "index" artifacts)
- input dirs are positional arguments instead of `-I` (which are for search path).

Also, implemented support for counting occurrences, and passing that to the search indexing, in the reference driver.


src/driver/compile.ml
```diff
@@ -225,7 +225,7 @@ let sherlodoc_index_one ~output_dir (index : Odoc_unit.index) =
   Sherlodoc.index ~format:`js ~inputs ~dst ();
   rel_path
 
-let html_generate output_dir linked =
+let html_generate ~occurrence_file output_dir linked =
```
Julow, line 228: The index is set to be used for other purpose in the future (sidebar) so it doesn't make sense to build it from inside this function.
I think the `compile_index` code should be moved out of this function, `html_generate` should take `~db_path` instead.

Julow (approved):
Looks good to me :)

src/driver/compile.ml
jonludlam: That this function calls `compile_index` was already true before this PR and isn't changed. Agreed that it'd be nicer to factor it out but let's not hold up this PR for that.

PR #1211, opened 2024-10-10
Reduce lower bounds on Dune

Julow:
The CI caught that this test increased the lower bounds on Dune.


jonludlam:
Thanks!

Julow:
The CI caught an other lower bound problem, probably one similar to https://github.com/ocaml/odoc/pull/994, I'll fix it in an other PR.

Julow:
Next PR: https://github.com/ocaml/odoc/pull/1212

PR #1212, opened 2024-10-10
opam: Update lower bounds on yojson

Julow:
The output format of older yojson was unstable in some cases.

### DIFF
--- a/odoc.opam
+++ b/odoc.opam
@@ -51,7 +51,7 @@ depends: [
   "fmt"
 
   "ocamlfind" {with-test}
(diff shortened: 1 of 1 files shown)


Julow:
The snippet of JSON in test `hidden_modules.t` changes formatting several times between yojson 1.6.0 and 2.1.0 and has happened in the past https://github.com/ocaml/odoc/pull/994
If this happens again, we might want to pipe our JSON outputs through `jq`.

Julow:
The lower-bound check is passing :)

jonludlam:
nice! thanks!

PR #1228, opened 2024-10-30
Define Tree and Forest modules for sidebars and index

gpetiot:
Reduce diff of #1220

Probably more could be put in this branch but my goal is to split #1220 as much as possible to make the diff digestable.

cc @panglesd for review (I don't have the repo rights to add reviewers anymore)

### DIFF
--- a/src/utils/tree.ml
+++ b/src/utils/tree.ml
@@ -1,0 +1,51 @@
+type 'a tree = { node : 'a; children : 'a forest }
+and 'a forest = 'a tree list
+
+module type S = sig
+  type 'a t
+
+  val fold_left : f:('acc -> 'a -> 'acc) -> 'acc -> 'a t -> 'acc
+  val iter : f:('a -> unit) -> 'a t -> unit
+  val map : f:('a -> 'b) -> 'a t -> 'b t
+end
+
+type 'a t = 'a tree
+
+let leaf node = { node; children = [] }
+
+let rec fold_left ~f acc { node; children } =
+  let acc = f acc node in
+  fold_left_forest ~f acc children
+
+and fold_left_forest ~f acc forest = List.fold_left (fold_left ~f) acc forest
+
+let rec iter ~f { node; children } =
+  let () = f node in
+  iter_forest ~f children
+
+and iter_forest ~f forest = List.iter (iter ~f) forest
+
+let rec map ~f { node; children } =
+  let node = f node in
+  let children = map_forest ~f children in
+  { node; children }
+
+and map_forest ~f forest = List.map (map ~f) forest
+
+let rec filter_map ~f { node; children } =
+  match f node with
+  | None -> None
+  | Some node ->
+      let children = filter_map_forest ~f children in
+      Some { node; children }
+
+and filter_map_forest ~f forest = List.filter_map (filter_map ~f) forest
+
+module Forest = struct
+  type 'a t = 'a forest
+
+  let fold_left = fold_left_forest
+  let iter = iter_forest
+  let map = map_forest
+  let filter_map = filter_map_forest
+end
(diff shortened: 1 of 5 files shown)


src/utils/tree.ml
```diff
@@ -0,0 +1,51 @@
+type 'a tree = { node : 'a; children : 'a forest }
+and 'a forest = 'a tree list
+
+module type S = sig
+  type 'a t
+
+  val fold_left : f:('acc -> 'a -> 'acc) -> 'acc -> 'a t -> 'acc
+  val iter : f:('a -> unit) -> 'a t -> unit
+  val map : f:('a -> 'b) -> 'a t -> 'b t
```
gpetiot, line 9: I'm using the `base` convention of labelling the function `f`, as I find it cleaner for this kind of functions, but if you want to keep the existing odoc conventions that's fine.

src/utils/tree.ml
```diff
@@ -0,0 +1,51 @@
...
+and map_forest ~f forest = List.map (map ~f) forest
+
+let rec filter_map ~f { node; children } =
+  match f node with
+  | None -> None
+  | Some node ->
+      let children = filter_map_forest ~f children in
+      Some { node; children }
+
+and filter_map_forest ~f forest = List.filter_map (filter_map ~f) forest
```
panglesd, line 42: To be able to use such advanced function such as `List.filter_map`, for compatibility reasons you need to use the defined version from `Odoc_utils.List`. But, to be able to do that without circular dependency, you need to extract `Odoc_utils.List` in its own compilation unit.

See [the file](https://github.com/ocaml/odoc/pull/1220/files#diff-aecf42552bb3fad3ba10e9d479458936c0fc797359e217daa2d84e77b39e0e9f) and [the](https://github.com/ocaml/odoc/pull/1220/files#diff-b2cf264f6cf17ee07a34ed13c3907e9278b67efda5efbaabb124d14e25581799R48) [use](https://github.com/ocaml/odoc/pull/1220/files#diff-7499e1ac2a9242dfdbf9a5cf18e2127eb099cf8091fa336fa0b530e5af2feb66R1) in the original PR.

src/utils/tree.ml
panglesd: I'm fine with either. We already have a mix. `f` is never really enlightening nor get punned, but it's sometimes nice to choose the order of the args (`map x ~f:(fun x -> very long function body)` or `x |> f1 |> f2 |> map ~f:apply`).

src/utils/tree.mli
```diff
@@ -0,0 +1,20 @@
...
+
+val leaf : 'a -> 'a tree
+
+module type S = sig
+  type 'a t
+
+  val fold_left : f:('acc -> 'a -> 'acc) -> 'acc -> 'a t -> 'acc
+  val iter : f:('a -> unit) -> 'a t -> unit
+  val map : f:('a -> 'b) -> 'a t -> 'b t
+end
```
panglesd, line 12: Why not include `filter_map` in the signature, since it is defined for both trees and forests?

panglesd (approved):
Thanks for opening this, this is cleaner and atomic.

I left minor comments and a more important compat issue. Once this one is resolved, it can be merged!

src/utils/tree.mli
gpetiot: Because the signature was different for the trees, and sknce it wasn’t used in the original PR I didn’t spend time trying to fit the same signature

panglesd: Indeed, filter_map on a tree may return None if the function decides to remove it. Fine not to include it.

src/utils/tree.ml
gpetiot: Good catch! I did that in my last commit.

jonludlam:
Thanks!

PR #1231, opened 2024-10-31
Docs lib deps

jonludlam:
A few minor fixes


panglesd (approved):
Thanks! The broken links on pages were always something I wanted to fix but postponed.

Looks good to me!

PR #1232, opened 2024-10-31
Represent sidebar TOC as a Tree

gpetiot:
Follow-up on #1228 
Reduce diff of #1220

@panglesd I made the changes minimal, to keep the implementation of Page_hierarchy and Skeleton separate, let me know what you think.


panglesd (approved):
This PR uses the `Tree.t` representation for page hierarchies. Looks good to me!

> to keep the implementation of Page_hierarchy and Skeleton separate

I'm not sure, what do you mean here?

gpetiot:
I think I just meant that I didn’t want to implement changes that required the new code in Skeleton or Page hierarchy. That would be made in a future PR.

gpetiot:
By the way I tried to bridge the gap between this PR and #1220 but I just ended up reimplementing the whole #1220, ie I couldn’t further split the PR in smaller changes

panglesd:
I can try. I think when we discussed together there was:
- Gathering the unrelated changes (eg [this one](https://github.com/ocaml/odoc/pull/1220/files#diff-d955f7dbfb34dd3b6f73f9c626306d7240006429f385bb1de02bc55183087595)) in a commit,
- Moving the relevant files in the `index` folder (without changing them) in a commit,
- Actually changing them in another commit

This allows to better see how things are modified: the file renaming + lots of modification prevented git to detect the renaming. Splitting it in two commits would help with that.

(I've not tried though, maybe the items 2 and 3 are not reasonably doable!)

src/document/sidebar.ml
```diff
@@ -17,41 +17,14 @@ end = struct
...
-      in
-      { Tree.node = payload; children = entries }
-    in
-    of_lang ~parent_id:None dir
+    Tree.map dir ~f:(function
+      | None -> None
+      | Some (parent_id, title) ->
+          let path = Url.Path.from_identifier parent_id in
+          let content = Comment.link_content title in
+          Some (path, sidebar_toc_entry parent_id content))
```
gpetiot, line 27: @panglesd I simplified this code in regards to what is currently done on master, not on #1220, if this is over-factorized and makes the rebase harder for #1220 later feel free to revert this commit before merging.

panglesd: Looks good!

src/odoc/odoc_file.mli
```diff
@@ -30,6 +30,11 @@ type content =
 
 type t = { content : content; warnings : Error.t list }
 
+type 'a index = {
+  sidebar : Odoc_index.Sidebar.t;
+  index : 'a Paths.Identifier.Hashtbl.Any.t;
+}
+
```
gpetiot, line 37: This is temporary to deal with the deletion of `Lang.Index`

panglesd: It would be more logic to have this in `src/index/odoc_index.ml` (which should then not be removed, contrary to my previous comment).

Indeed, that's where it is going to be replaced by the new index type.

src/document/sidebar.ml
```diff
@@ -1,9 +1,18 @@
...
+module Id = Odoc_model.Paths.Identifier
 
 let sidebar_toc_entry id content =
-  let href = id |> Url.Path.from_identifier |> Url.from_path in
-  let target = Target.Internal (Resolved href) in
+  let target =
+    match
+      (id :> Odoc_model.Paths.Identifier.t)
+      |> Url.from_identifier ~stop_before:false
+    with
```
jonludlam, line 10: What are we doing with aliases in the sidebar?

src/document/sidebar.ml
```diff
@@ -18,27 +18,33 @@ let sidebar_toc_entry id content =
...
-            with
-            | Ok r -> r
-            | Error _ -> assert false
-            (* This error case should never happen since [stop_before] is false, and even less since it's a page id *)
-          in
-          let content = Comment.link_content title in
-          Some (path, sidebar_toc_entry parent_id content))
+  let of_page_hierarchy (dir : Odoc_index.Page_hierarchy.t) =
+    let f index =
+      let payload =
```
jonludlam, line 30: What's with the extra `let payload = ... in payload`?

gpetiot: Fixed

src/document/sidebar.ml
gpetiot: @panglesd what do you think about it? It looks to me that we don't lose anything from the previous implementation, that's something that could be improved in a future PR.

gpetiot:
This PR should be reviewed commit-by-commit.
I'm aiming at catching up #1220 but it can be merged at any point if the reviews are done and we decide it's better to switch to a fresh PR for the next commits.

src/document/sidebar.ml
panglesd: I don't know how much you have taken from the original PR, but in the original PR we keep track of which module have an expansion.
If it has one, we use `~stop_before:false`, if it has none we use `stop_before:true`. So, aliases appear in the sidebar.

jonludlam: That's good, we need to keep that behaviour!

panglesd: The `has_expansion` bool is recorded [here](https://github.com/ocaml/odoc/pull/1220/files#diff-1304c513720e9777b80d9e0c31d217f5b41a5e025b50f7fbd996a534c4ef3c15R41) and used [here](https://github.com/ocaml/odoc/pull/1220/files#diff-d44f4d07a7ce1288d4c91f2a6a38230bb446b85e91fa5d2536f93a3da13eaa0dR108).

I think there is some refactoring to do though, in the final version: `sidebar_toc_entry` is [still there](https://github.com/ocaml/odoc/pull/1220/files#diff-d44f4d07a7ce1288d4c91f2a6a38230bb446b85e91fa5d2536f93a3da13eaa0dR4), but [used only for _pages_ sidebar](https://github.com/ocaml/odoc/pull/1220/files#diff-d44f4d07a7ce1288d4c91f2a6a38230bb446b85e91fa5d2536f93a3da13eaa0dR45). This should be made explicit!

panglesd: > I don't know how much you have taken from the original PR, but in the original PR we keep track of which module have an expansion.

I've checked: currently, what has been taken is mostly refactoring: introducing a tree type and using it in each tree structure, and moving things in the right directory. So, in the current state of this PR the sidebar is only showing root modules (as before), so no need to handle aliases!

panglesd (approved):
Looks good to me!

PR #1233, opened 2024-11-04
Allow more hierarchy to be compatible with `-L` and `-P`.

panglesd:
In the initial implementation of `-L` and `-P` we added too many constraints: the trees below the roots (both `-L` and `-P`) had to all be disjoint.

This prevented some hierarchy, such as the one where pages are put in `pkgname/` and libraries in `pkgname/libname/`: indeed, `-L pkgname/libname` would not be disjoint from `-P pkgname/`.

This PR relax the constraint, `-L` roots have to be disjoints, and `-P` roots have to be disjoint, but there is no constraint between a `-P` and a `-L` root.

It also relaxes the constraint the when linking modules, the module has to be below a `-L`.

Tests are updated to use the simpler hierarchy.

I'll update the driver in another PR!


src/odoc/bin/main.ml
```diff
@@ -467,11 +467,12 @@ module Indexing = struct
...
-       not
-         (Antichain.check
-            (List.rev_append lib_roots page_roots |> List.map ~f:snd))
-     then Error (`Msg "Paths given to all -P and -L options must be disjoint")
+    (if not (Antichain.check (page_roots |> List.map ~f:snd)) then
+       Error (`Msg "Paths given to all -P options must be disjoint")
+     else Ok ())
+    >>= fun () ->
+    (if not (Antichain.check (lib_roots |> List.map ~f:snd)) then
+       Error (`Msg "Paths given to all -L options must be disjoint")
```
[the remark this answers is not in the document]
panglesd: Yes, that was not very readable, hopefully I've improved it in the last commit.

test/integration/link_opts.t/run.t
```diff
@@ -1,50 +1,55 @@
...
-  [1]
+ For pages:
+
+  $ odoc link --current-package pkg -P pkg:h/pkg2 h/pkg/page-page.odoc
+  $ odoc link -P pkg:h/pkg h/pkg/page-page.odoc
+
+It is not required to be below a `-L`, even for modules:
 
-Specified current package is wrong:
+  $ odoc link -P pkg:h/pkg -L otherlib:h/otherpkg h/pkg/libname/test.odoc
```
jonludlam, line 28: nit: do you want `-L otherlib:h/otherpkg/otherlib` here instead?

panglesd: Yes. I've cleaned up the test a bit (some options were not really necessary for the test and made it less readable. Hopefully it is better now!

src/odoc/bin/main.ml
```diff
@@ -601,81 +601,49 @@ end = struct
...
-            ~f:(fun (pkg, path) ->
-              if Fpath.is_prefix path o then Some pkg else None)
-            l
-        with
-        | Some _ as r -> Ok r
-        | None -> Error `Not_found)
+    | [] -> None
+    | _ ->
+        find_map
+          ~f:(fun (pkg, path) ->
```
jonludlam, line 607: These names are slightly misleading as `pkg` might actually be a lib. I know that was also the case before, but we might as well fix that now.

panglesd:
I have made more tests with the new (doc-lib)less hierarchy, and there is a small issue.
When we have `-P pkgname/ -L pkgname/libname`, we don't want the content of `pkgname/libname` to be added to the accessible pages for pkgname (at least, not in the sidebar: the libraries are already included in the sidebar).
(The problem is shown in [this commit](https://github.com/ocaml/odoc/pull/1233/commits/fc85b78c1f997ba70a157f1879e4328c466a57db))

<details><summary>An image is worth a thousand words</summary>
<p>

The sidebar for odoc, suffering from this:

image


</p>
</details> 

I have solved this by excluding `-L` subdirectories from `-P` roots ([a5f767c](https://github.com/ocaml/odoc/pull/1233/commits/a5f767c7030f49a866fac832e71ba343e2fd597b)). Another option would have to add a metadata tag `@sidebar omit` which tells odoc to omit this subhierarchy from the sidebar. I think this second option is better, otherwise the name clashes are more likely.

jonludlam:
That `@sidebar omit` is a little odd as it's only omitting it as a subdirectory of another 'sidebar root'. It doesn't seem odd to me that things that are configured as 'sidebar roots' are never permitted to be subdirs of another 'sidebar root' so I think your current fix isn't unreasonable at all.

src/odoc/bin/main.ml
```diff
@@ -601,81 +601,49 @@ end = struct
           match f x with Some _ as result -> result | None -> find_map ~f l)
```
gpetiot, line 601: By the way `find_map` is in `Odoc_utils.List`

src/odoc/bin/main.ml
```diff
@@ -601,81 +601,49 @@ end = struct
...
-    let is_page_or_asset = is_page input || is_asset input in
-    (if is_page_or_asset then Ok None
-     else current_library_of_input lib_roots input)
-    >>= fun current_lib ->
-    (if is_page_or_asset then
-       current_package_of_page ~current_package page_roots input
-     else validate_current_package page_roots current_package)
+    Antichain.check (page_roots |> List.map ~f:snd) ~opt:"-P" >>= fun () ->
+    Antichain.check (lib_roots |> List.map ~f:snd) ~opt:"-L" >>= fun () ->
+    let current_lib = current_library_of_input lib_roots input in
```
gpetiot, line 645: We might as well just use ` find_root_of_input` (clearer imho) if `current_library_of_input` is just an alias

gpetiot:
> I have made more tests with the new (doc-lib)less hierarchy, and there is a small issue. When we have `-P pkgname/ -L pkgname/libname`, we don't want the content of `pkgname/libname` to be added to the accessible pages for pkgname (at least, not in the sidebar: the libraries are already included in the sidebar). (The problem is shown in [this commit](https://github.com/ocaml/odoc/pull/1233/commits/fc85b78c1f997ba70a157f1879e4328c466a57db))

Is there a previous discussion about this? I'm interested in having more context before forming an opinion, because naively it looks to me that it's something that should be handled in the driver (meaning that redundancies in the arguments would be intentional and should be preserved in the sidebar). But I don't have all the context :)

panglesd:
I think some context can be understood by reading the discussion starting from [this comment](https://github.com/ocaml/odoc/pull/1221#issuecomment-2425880065).

We can also do a quick call if you'd like.

jonludlam:
The alternate suggestions for how to create the perfect sidebar can happen in a later PR - this one is pretty foundational for some other work so let's get this in to unblock the other PRs.

PR #1235, opened 2024-11-05
Driver: uniformize names for directories

panglesd:
`odoc_dir` was sometimes used to refer to the root of the directory containing .odoc files (`_odoc` by default), and sometimes used to refer to the direct parent of a `.odoc` file.

The root of the directory containing `.odoc` files was referred as `odoc_dir`, `output_dir`.

The root of the directory containing `.odocl` files was referred as `odocl_dir`, `link_dir`, `linked_dir`.

This commit cleans that. The root of the directory for `.odoc` files is consistently named `odoc_dir`, the root for `.odocl` files is `odocl_dir`.

It has the nice effect of removing code with punning working, and fixing some misunderstanding wrt to this in the `Landing_page` module.

### DIFF
--- a/src/driver/compile.ml
+++ b/src/driver/compile.ml
@@ -92,7 +92,7 @@ let find_partials odoc_dir :
   | Ok h -> (h, tbl)
   | Error _ -> (* odoc_dir doesn't exist...? *) (Util.StringMap.empty, tbl)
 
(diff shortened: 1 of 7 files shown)


jonludlam:
Much nicer, thanks!

PR #1236, opened 2024-11-06
Driver: New hierarchy 

panglesd:
Based on #1233.

Implement the new hierarchy in the driver. This time, properly split in commits!


src/driver/odoc_driver.ml
```diff
@@ -614,6 +614,8 @@ let run libs verbose packages_dir odoc_dir odocl_dir html_dir stats nb_workers
     if voodoo then
       match Util.StringMap.to_list all with
       | [ (_, p) ] ->
+          (* TODO: check whether the [doc] layer should be removed (due to new
```
jonludlam, line 617: This should be fine.

jonludlam:
This looks good, except that there's some code in the voodoo module to determine libraries based on their paths which we need to figure out: https://github.com/ocaml/odoc/blob/master/src/driver/voodoo.ml#L277-L280

jonludlam:
I'm working on a fix for voodoo mode.

panglesd:
Then, I pushed only the fix regarding your other comment.

Is there anything more to do than removing the "lib" intermediate directory in your linked file? I'll wait and see your fix!

jonludlam:
Seems voodoo mode is broken for more reasons, let's just merge this and I'll fix it in another PR.

PR #1239, opened 2024-11-08
Implicitely ended tags.

panglesd:
This is an attempt to solve #1138.

The solution I went for is to end tags implicitly, in the same way as shorthand list syntax is implicitly ended: by a new line, a header or a tag. Indeed, otherwise the implicit ending of a tag containing a shorthand list item is very unclear, see https://github.com/ocaml/odoc/issues/1138#issuecomment-2463996122.


jonludlam:
s/implicitely/implicitly/g

gpetiot (approved):
Looks good!

Julow (commented):
Tags containing a `block_element_list` do not end at a heading. For example, here the heading is turned into a paragraph and part of the tag:

     @deprecated foo

     {1 heading}

Is this intended ? Perhaps `` `After_tag `` in type `where_in_line` should be removed ?

jonludlam:
I'm also not seeing the behaviour that @Julow observed.

I'd like to have a documentation page that describes these behaviours - including those of lists that I don't believe are documented anywhere. We need that before release, but not before we merge this PR :-)

jonludlam:
Thanks!

PR #1240, opened 2024-11-08
Fix duplicate indexing

panglesd:
Pages and modules found under the page and library roots are already indexed, no need to add them to the include_rec flag.

Otherwise, we risk to hit https://github.com/art-w/sherlodoc/pull/45


panglesd:
Actually, this is too early, I'll include this commit in the last PR part of #1220

PR #1241, opened 2024-11-08
Voodoo fixes following the change in layout

jonludlam:
Combined with the splitting of the compile/link phases so we can link to more than just the dependencies of your package. This is a building block to allow links in the documentation of package A to go to package B that isn't a direct dependency - or even to package C that depends on package A!

### DIFF
--- a/src/driver/library_names.ml
+++ b/src/driver/library_names.ml
@@ -29,8 +29,7 @@ let read_libraries_from_pkg_defs ~library_name pkg_defs =
     in
 
     let deps_str = Fl_metascanner.lookup "requires" [] pkg_defs in
-    let deps = Astring.String.fields deps_str in
-    let deps = List.filter (fun x -> String.length x > 0) deps in
+    let deps = Astring.String.fields ~empty:false deps_str in
     let dir =
       List.find_opt (fun d -> d.Fl_metascanner.def_var = "directory") pkg_defs
     in
--- a/src/driver/library_names.ml
+++ b/src/driver/library_names.ml
@@ -30,6 +30,7 @@ let read_libraries_from_pkg_defs ~library_name pkg_defs =
 
     let deps_str = Fl_metascanner.lookup "requires" [] pkg_defs in
     let deps = Astring.String.fields deps_str in
+    let deps = List.filter (fun x -> String.length x > 0) deps in
     let dir =
       List.find_opt (fun d -> d.Fl_metascanner.def_var = "directory") pkg_defs
     in
--- a/src/driver/odoc_driver.ml
+++ b/src/driver/odoc_driver.ml
@@ -149,12 +149,16 @@ let run mode
         let all = Voodoo.of_voodoo p ~blessed in
         let extra_paths = Voodoo.extra_paths odoc_dir in
         (all, extra_paths, actions)
-    | Dune { path } -> (Dune_style.of_dune_build path, Voodoo.empty_extra_paths, All)
+    | Dune { path } ->
+        (Dune_style.of_dune_build path, Voodoo.empty_extra_paths, All)
     | OpamLibs { libs } ->
         ( Packages.of_libs ~packages_dir:None (Util.StringSet.of_list libs),
-          Voodoo.empty_extra_paths, All )
+          Voodoo.empty_extra_paths,
+          All )
     | OpamPackages { packages } ->
-        (Packages.of_packages ~packages_dir:None packages, Voodoo.empty_extra_paths, All)
+        ( Packages.of_packages ~packages_dir:None packages,
+          Voodoo.empty_extra_paths,
+          All )
   in
 
   let virtual_check =
--- a/src/driver/odoc_driver.ml
+++ b/src/driver/odoc_driver.ml
@@ -111,8 +111,10 @@ let remap_virtual_interfaces duplicate_hashes pkgs =
       })
     pkgs
 
+type action_mode = CompileOnly | LinkAndGen | All
+
 type mode =
-  | Voodoo of { package_name : string; blessed : bool }
+  | Voodoo of { package_name : string; blessed : bool; actions : action_mode }
   | Dune of { path : Fpath.t }
   | OpamLibs of { libs : string list }
   | OpamPackages of { packages : string list }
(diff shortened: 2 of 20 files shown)


jonludlam:
The 'ensure pages' commit has fixed some errors reported in one of the test cases, though there are still some errors present. I think there are still some references in there that look valid that aren't yet resolving though.

src/driver/voodoo.ml
```diff
@@ -259,24 +277,66 @@ let of_voodoo pkg_name ~blessed =
       in
       let packages = List.filter_map (fun x -> x) (last :: packages) in
       let packages = List.map process_package packages in
+      let pkg = List.hd packages in
+      Logs.debug (fun m -> m "Package: %a\n%!" Packages.pp pkg);
```
gpetiot, line 281: Is the first package always the "main" one? Maybe we should just print`pkg_name` here

src/driver/odoc_driver.ml
```diff
@@ -551,8 +551,10 @@ let remap_virtual_interfaces duplicate_hashes pkgs =
       })
     pkgs
 
+type action_mode = CompileOnly | LinkAndGen | All
```
gpetiot, line 554: I would find something like `type action_mode = Compile | Link | Gen` more future-proof, and we don't have too many actions so it's not too tedious to enumerate them all in a pattern-matching anyway.
The cli arg would be something like:
- `--actions compile`
- `--actions link,gen`
- `--actions compile,link,gen` (maybe use an alias `--actions all` for the full list)

Would that be problematic if users specify an invalid combination like `--actions compile,gen` ? That could be filtered when parsing the arg anyway.

src/driver/voodoo.mli
```diff
@@ -3,4 +3,6 @@ val find_universe_and_version :
 
 val of_voodoo : string -> blessed:bool -> Packages.set
 
-val extra_libs_paths : Fpath.t -> Fpath.t Util.StringMap.t
+val extra_paths : Fpath.t -> Fpath.t Util.StringMap.t * Fpath.t Util.StringMap.t
```
gpetiot, line 6: Can we return a record instead of a tuple here?

src/driver/odoc_units_of.ml
```diff
@@ -1,6 +1,7 @@
 open Odoc_unit
 
-let packages ~dirs ~extra_libs_paths (pkgs : Packages.t list) : t list =
+let packages ~dirs ~extra_paths:(extra_pkg_paths, extra_libs_paths)
```
gpetiot, line 3: Could we split the tuple: `~extra_pkg_paths ~extra_lib_paths` here?

src/driver/voodoo.ml
```diff
@@ -12,6 +12,9 @@ type pkg = {
 
 let prep_path = ref "prep"
 
+let lib_marker = ".odoc_lib_marker"
+let pkg_marker = ".odoc_pkg_marker"
```
gpetiot, line 16: This could use some comments.

src/driver/odoc_driver.ml
jonludlam: I don't anticipate users using this feature at all - it's really only here so the docs CI can drive this. As such, I don't really worry so much about future-proofing this, as whatever we do will be very tightly coupled with the docs CI anyway.

src/driver/voodoo.ml
jonludlam: yes, fair enough :-)

gpetiot:
I have a more generic question: what are the reasons for using many lib marker files instead of using a single file to track the lib markers? e.g. something like:
```
- A:
  - pkg-paths:
    - xx
    - xx
  - lib-paths:
    - xx
    - xx
- B:
...
```

jonludlam:
Because each package is built separately, so there will _have_ to be at least as many files as there are packages. Since we're trying to locate directories, I thought I might as well have a marker in each directory - at least we know it's definitely there in the filesystem then!

src/driver/odoc_driver.ml
```diff
@@ -633,27 +636,37 @@ let run mode
...
+          match actions with
+          | LinkAndGen -> units
+          | _ -> Compile.compile ?partial ~partial_dir:odoc_dir units
         in
-        let () = Compile.html_generate ~occurrence_file html_dir linked in
-        let _ = Odoc.support_files html_dir in
-        ())
+        (match mode with
+        | Voodoo _ -> Voodoo.write_lib_markers odoc_dir all
+        | _ -> ());
```
panglesd, line 655: Nitpick: `let () = ... in` is more readable in this case I think.

src/driver/odoc_driver.ml
```diff
@@ -633,27 +636,37 @@ let run mode
...
-          let output =
-            Fpath.( / ) odoc_dir "occurrences-all.odoc-occurrences"
-          in
-          let () = Odoc.count_occurrences ~input:[ odoc_dir ] ~output in
-          output
+        Compile.init_stats units;
+        let compiled =
+          match actions with
+          | LinkAndGen -> units
+          | _ -> Compile.compile ?partial ~partial_dir:odoc_dir units
```
panglesd, line 651: Even though we might not need to be extra future proof, an exhaustive check here is cheap and useful!

src/driver/odoc_driver.ml
```diff
@@ -633,27 +636,37 @@ let run mode
...
         in
-        let () = Compile.html_generate ~occurrence_file html_dir linked in
-        let _ = Odoc.support_files html_dir in
-        ())
+        (match mode with
+        | Voodoo _ -> Voodoo.write_lib_markers odoc_dir all
+        | _ -> ());
+        match actions with
+        | CompileOnly -> ()
+        | _ ->
```
panglesd, line 658: Same, I'd prefer an exhaustive check here!

src/odoc/resolver.ml
```diff
@@ -580,13 +579,21 @@ let create ~important_digests ~directories ~open_modules ~roots =
               { Named_roots.name; dir; omit })
             roots
         in
+        let directories =
+          match current_package with
+          | None -> directories
+          | Some pkg -> (
+              try List.assoc pkg page_roots :: directories
+              with _ -> directories)
```
panglesd, line 587: It would be better to catch only `Not_found` exceptions.

panglesd, line 587: This is a change for drivers... If I understand correctly:
- Any `-L _:<path>` also adds `<path>` to the search path
- A `-P pkg:<path>` adds `<path>` to the search path _when_ `pkg` is the current package

I think this should be added to the odoc manpage and the driver doc page, now that it is merged!

panglesd: In fact, the `Not_found` exception cannot be raised: There must be a directory associated to `current_package` it is checked before.
I think either remove the catch and add a comment, or keep the "`current_package_dir`" when computing the `current_package`?

panglesd (approved):
Looks good!

src/driver/library_names.ml
```diff
@@ -30,6 +30,7 @@ let read_libraries_from_pkg_defs ~library_name pkg_defs =
 
     let deps_str = Fl_metascanner.lookup "requires" [] pkg_defs in
     let deps = Astring.String.fields deps_str in
+    let deps = List.filter (fun x -> String.length x > 0) deps in
```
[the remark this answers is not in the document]
jonludlam: cool thanks!

src/odoc/resolver.ml
jonludlam: Yes, that's the behaviour I intended. What I wasn't sure about was whether this is the right mechanism to do this?

jonludlam: In pulling this particular thread a bit more unravelled than I'd expected. @panglesd could you please check carefully commit https://github.com/ocaml/odoc/pull/1241/commits/fa215e0f8f0fba7f8eacc641589720d030925280 to check I've not made any silly assumptions?

src/odoc/resolver.mli
```diff
@@ -26,8 +26,10 @@ type t
 type roots = {
   page_roots : (string * Fs.Directory.t) list;
   lib_roots : (string * Fs.Directory.t) list;
-  current_lib : string option;  (** Name of the current [-L]. *)
-  current_package : string option;  (** Name of the current [-P]. *)
+  current_lib : (string * Fs.Directory.t) option;
```
panglesd, line 29: This could be a `named_root` type, given how often it is used!

```ocaml
type named_root = { name : string; root : Fs.Directory.t }
```

panglesd (approved):
Looks good to me!

src/odoc/resolver.ml
panglesd: Seems reasonable... Not adding anything to the search path would be another reasonable approach (that would give more work, but more freedom, to the driver).

PR #1242, opened 2024-11-12
Remove core types and exceptions from identifiers

panglesd:
Identifiers are values that can be turned into a link. Core types and values cannot, so they should not have an ID.

This PR fixes this. Core types can appear in unresolved path, and that's all. This better representation allows to:
- Have `Url.from_identifier` not return a result type (which removes many `assert false` from the code!)
- Remove the `Predefined` module, which did not made sense.


gpetiot (approved):
This looks good to me, the code simplification is impressive!
Do you have an idea what causes the mdx issue in the tests?

panglesd:
Yes: the mdx tests are only run on `dune build @runmdx`, and I forgot to do that. The new output is legit, I promoted them.

jonludlam:
I'm not convinced this is the right approach, though it's not without its merits. Perhaps we can try removing core types only from identifiers instead?

panglesd:
I tried your suggested approach: moving "path to core types" from identifiers to resolved paths. So paths to core type are represented by:
- `` `Resolved (`Identifier (`CoreType name)) `` before this PR
- `` `CoreType name `` before @jonludlam's suggestion
- `` `Resolved (`CoreType name) `` after the suggestion

As a consequence of considering core types path as resolved, taking an identifier out of a path type returns an option.
However, this is likely better if we, one day, want to add canonical references to core types.

gpetiot:
> However, this is likely better if we, one day, want to add canonical references to core types.

You mean so that we can link to e.g. https://github.com/ocaml/ocaml/blob/trunk/stdlib/bool.mli#L22 each time there is a bool, or is there other benefits to having these references? is it useful for the search?

gpetiot (approved):
The new diff looks good by the way.

panglesd:
> You mean so that we can link to e.g. https://github.com/ocaml/ocaml/blob/trunk/stdlib/bool.mli#L22 each time there is a bool, or is there other benefits to having these references? is it useful for the search?

Sorry, I did not meant references but paths.

```ocaml
module Int : sig
  type t = int
  (** @canonical int *)
end

val plus_one : Int.t -> Int.t
```

If the feature I was mentioning were implemented, it would make the rendered signature of `plus_one` use the core type `int` instead of the alias `Int.t`.

PR #1243, opened 2024-11-12
Use a tag to declare the order of children

panglesd:
Currently, it is:

```
@children_order foo    bar dir/ baz
             boo
```

(The tag expect a single paragraph containing only words and whitespace)

Based on #1239.


src/parser/test/test.ml
```diff
@@ -143,6 +143,8 @@ module Ast_to_sexp = struct
           @ List.map (at.at (nestable_block_element at)) es)
     | `Return es ->
         List (Atom "@return" :: List.map (at.at (nestable_block_element at)) es)
+    | `Children_order es ->
+        List (Atom "@return" :: List.map (at.at (nestable_block_element at)) es)
```
gpetiot, line 147:
```suggestion
        List (Atom "@children_order" :: List.map (at.at (nestable_block_element at)) es)
```

src/model/semantics.ml
```diff
@@ -57,6 +70,24 @@ let handle_internal_tags (type a) tags : a handle_internal_tags -> a = function
           None
       | Some ((`Dot _ as p), _) -> Some p
       | None -> None)
+  | Expect_page_tags ->
+      let unparsed_lines =
+        find_tags []
+          (function `Children_order _ as p -> Some p | _ -> None)
```
gpetiot, line 76: I think `find_tags` would be easier to read if `f` was labelled:
```suggestion
          ~filter:(function `Children_order _ as p -> Some p | _ -> None)
```

jonludlam: Now `find_tags` is inconsistent with `find_tag` immediately above it, which still takes an `f` parameter :-P

panglesd: Thanks!

gpetiot (approved):
Looks good now!

Julow (approved):
Looks good :) I particularly like that the frontmatter tags don't end up in the model.

jonludlam:
Thanks!

PR #1244, opened 2024-11-13
Sidebar and index overhaul (2)

gpetiot:
Follow-up on #1232, catching up on #1220


jonludlam:
I'm getting a different sidebar viewing the modules than when viewing the pages - when viewing the modules the whole hierarchy seems to disappear:

Screenshot 2024-11-14 at 19 58 06

vs

Screenshot 2024-11-14 at 19 58 12

PR #1246, opened 2024-11-14
Use a tag to define short titles

panglesd:
Based on #1243.

This adds support for declaring the page's short title using a tag. These short titles are not used yet (this will come after the index work).

The first commits are small refactoring, needed for the last commit.


jonludlam:
Thanks!

PR #1248, opened 2024-11-15
Html remap

jonludlam:
This adds a mode for publishing your package docs on your own site. It will generate HTML for the packages you select, and rewrite links that would have gone to other local packages to their documentation pages on ocaml.org.

### DIFF
--- a/src/driver/odoc.ml
+++ b/src/driver/odoc.ml
@@ -180,7 +180,7 @@ let compile_index ?(ignore_output = false) ~output_file ?occurrence_file ~json
   ignore @@ Cmd_outputs.submit log desc cmd (Some output_file)
 
 let html_generate ~output_dir ?index ?(ignore_output = false)
-    ?(search_uris = []) ?(as_json = false) ~input_file:file () =
+    ?(search_uris = []) ?(remap = None) ?(as_json = false) ~input_file:file () =
   let open Cmd in
   let index =
     match index with None -> empty | Some idx -> v "--index" % p idx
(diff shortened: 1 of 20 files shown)


src/driver/odoc.mli
```diff
@@ -55,6 +55,7 @@ val html_generate :
   ?index:Fpath.t ->
   ?ignore_output:bool ->
   ?search_uris:Fpath.t list ->
+  ?remap:Fpath.t option ->
```
gpetiot, line 58:
```suggestion
  ?remap:Fpath.t ->
```

src/driver/odoc.ml
```diff
@@ -180,7 +180,7 @@ let compile_index ?(ignore_output = false) ~output_file ?occurrence_file ~json
   ignore @@ Cmd_outputs.submit log desc cmd (Some output_file)
 
 let html_generate ~output_dir ?index ?(ignore_output = false)
-    ?(search_uris = []) ?(as_json = false) ~input_file:file () =
+    ?(search_uris = []) ?(remap = None) ?(as_json = false) ~input_file:file () =
```
gpetiot, line 183:
```suggestion
    ?(search_uris = []) ?remap ?(as_json = false) ~input_file:file () =
```

src/driver/packages.ml
```diff
@@ -470,7 +472,28 @@ let of_packages ~packages_dir packages =
             files.docs
           |> Fpath.Set.of_list
         in
-        let enable_warnings = List.mem pkg.name packages in
+        let selected = List.mem pkg.name packages in
+        let remaps =
+          if List.mem pkg.name packages then []
```
gpetiot, line 477:
```suggestion
          if selected then []
```

gpetiot (commented):
It looks good, I only have minor comments so far. Are the remap text files supposed to be used for testing only or will this feature be exposed to users? (I like the default of redirecting to ocaml.org, as long as the urls there don't change in the foreseeable future). If so we could add more checks (and trimming) on the content of the remap files.

jonludlam:
Yes, the intention is to expose this to users.

PR #1247, opened 2024-11-15
odoc_index: Replace -P/-L with --root arguments

gpetiot:
Based on #1244 (see [5a79e19](https://github.com/ocaml/odoc/pull/1247/commits/5a79e1933e4d8d217a48f2f4968e9a46163f981c) only)

PR #1250, opened 2024-11-18
Dedicated sidebar file

panglesd:
This PR adds a `odoc sidebar-generate` command which takes an index (`.odoc-index`) as input and produces a sidebar file (`.odoc-sidebar` or `.json` if `--json` is passed).

This allows to reduce the time spent HTML generating, as sidebar files are much smaller than index files (they don't contain values and other entries that are not shown in the sidebar anyway).
The json format is likely to change with #1247.


src/odoc/sidebar.ml
```diff
@@ -0,0 +1,64 @@
...
+
+let compile_to_json ~output sidebar =
+  let json = sidebar_to_json sidebar in
+  let text = Json.to_string json in
+  let output_channel =
+    Fs.Directory.mkdir_p (Fs.File.dirname output);
+    open_out_bin (Fs.File.to_string output)
+  in
+  let output = Format.formatter_of_out_channel output_channel in
+  Format.fprintf output "%s" text
```
Julow, line 57: It's best practice to close the channel. We have `Fun.protect` in utils to help close properly.

src/odoc/sidebar.ml
```diff
@@ -0,0 +1,64 @@
...
+  `Object [ ("name", `String name); ("pages", Tree.to_json toc_to_json pages) ]
+
+let libs_to_json ({ name; units } : Odoc_document.Sidebar.library) =
+  `Object
+    [
+      ("name", `String name);
+      ("modules", `Array (List.map (Tree.to_json toc_to_json) units));
+    ]
+
+let sidebar_to_json ({ pages; libraries } : Odoc_document.Sidebar.t) =
```
Julow, line 44: Shouldn't this function be in the HTML generator ? That's where the other JSON function are located.
The HTML generator is used in these functions.

src/driver/odoc_unit.ml
```diff
@@ -36,11 +36,14 @@ module Pkg_args = struct
...
 end
 
+type sidebar = { output_file : Fpath.t; json : bool }
+
 type index = {
   pkg_args : Pkg_args.t;
   output_file : Fpath.t;
   json : bool;
   search_dir : Fpath.t;
+  sidebar : sidebar option;
```
Julow, line 46: I found that `sidebar` is never `None`: https://github.com/panglesd/odoc/pull/34

src/odoc/sidebar.ml
```diff
@@ -0,0 +1,64 @@
...
+  let pages = List.map pages_to_json pages in
+  let libraries = List.map libs_to_json libraries in
+  `Object [ ("pages", `Array pages); ("libraries", `Array libraries) ]
```
Julow, line 47: The JSON is slightly confusing with two things called pages. Shouldn't this be called `manuals` ?

src/driver/odoc_unit.ml
panglesd: Thanks!

src/odoc/sidebar.ml
panglesd: Hmm, anyway this changes to a more generic sidebar json in #1251.

Julow (approved):
Perfect, let's merge !

PR #1251, opened 2024-11-19
Uniform treatment of sidebar

panglesd:
This PR is a change in how we handle the sidebar.

The current status is that we give the sidebar the modules and the pages separately. This is good because it allows them to be rendered separately which can be readable, but this is bad because:
- It gives less flexibility in the sidebar/hierarchy (for instance in a monorepo we might want to hierarchize the modules) while letting it be quite uniform in the canonical case.
- The sidebar and the breadcrumbs do not correspond
- There are conflicts whenever the library roots are included in the pages roots
- This does not handle implementations!

This PR changes the sidebar so that it is generated from the hierarchy of both modules, pages, sources, ...
The good things are above. The bad things are:
- It is a bit harder to make it look good.
- The children_order is more important (to have the libraries at the end).

Here is a [preview](https://choum.net/panglesd/ppxlib_global_toc/ppxlib/index.html) for such a sidebar for ppxlib.

(It seems to me that we will need to add the `@toc-expand` tags)

I still need to clean up the history but the code should be ok already, and the discussion can start!

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -614,7 +614,7 @@ module Breadcrumbs = struct
           { Types.current; parents; up_url }
     in
     let escape =
-      match (Config.escape_breadcrumb config, find_parent sidebar) with
+      match (Config.home_breadcrumb config, find_parent sidebar) with
       | true, Some { node; _ } -> (
           match page_parent node.url.page with
           | None -> []
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -1132,12 +1132,12 @@ module Odoc_html_args = struct
       (parser, printer)
   end
 
-  let escape_breadcrumb =
+  let home_breadcrumb =
     let doc =
       "Wether to add a 'Home' breadcrumb to go up the root of the given \
        sidebar."
     in
-    Arg.(value & flag & info ~docv:"escape" ~doc [ "escape-breadcrumb" ])
+    Arg.(value & flag & info ~docv:"escape" ~doc [ "home-breadcrumb" ])
 
   let theme_uri =
     let doc =
@@ -1207,7 +1207,7 @@ module Odoc_html_args = struct
 
   let extra_args =
     let config semantic_uris closed_details indent theme_uri support_uri
-        search_uris flat as_json remap remap_file escape_breadcrumb =
+        search_uris flat as_json remap remap_file home_breadcrumb =
       let open_details = not closed_details in
       let remap =
         match remap_file with
@@ -1228,14 +1228,14 @@ module Odoc_html_args = struct
       in
       let html_config =
         Odoc_html.Config.v ~theme_uri ~support_uri ~search_uris ~semantic_uris
-          ~indent ~flat ~open_details ~as_json ~remap ~escape_breadcrumb ()
+          ~indent ~flat ~open_details ~as_json ~remap ~home_breadcrumb ()
       in
       { Html_page.html_config }
     in
     Term.(
       const config $ semantic_uris $ closed_details $ indent $ theme_uri
       $ support_uri $ search_uri $ flat $ as_json $ remap $ remap_file
-      $ escape_breadcrumb)
+      $ home_breadcrumb)
 end
 
 module Odoc_html = Make_renderer (Odoc_html_args)
--- a/src/html/config.ml
+++ b/src/html/config.ml
@@ -12,12 +12,12 @@ type t = {
   flat : bool;
   open_details : bool;
   as_json : bool;
-  escape_breadcrumb : bool;
+  home_breadcrumb : bool;
 }
 
 let v ?(search_result = false) ?theme_uri ?support_uri ?(search_uris = [])
     ~semantic_uris ~indent ~flat ~open_details ~as_json ~remap
-    ?(escape_breadcrumb = true) () =
+    ?(home_breadcrumb = true) () =
   {
     semantic_uris;
     indent;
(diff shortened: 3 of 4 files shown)


jonludlam:
Thanks for this! It breaks sherlodoc though - do you have a branch of that that's compatible?

src/odoc/indexing.ml
```diff
@@ -55,136 +52,103 @@ let compile_to_json ~output ~occurrences files =
     false
   in
   Format.fprintf output "[";
-  let _ : bool =
+  let _first =
```
jonludlam, line 55: What's the reasoning behind only generating json fragments (and therefore having to pass around this 'first' thing) rather than generating a complete json object and simply outputting that?

src/html/utils.ml
```diff
@@ -1,87 +1,3 @@
 (* Shared utility functions *)
 
 let optional_elt f ?a = function [] -> [] | l -> [ f ?a l ]
```
jonludlam, line 3: I think we can move this somewhere else now :-)

src/model/frontmatter.ml
```diff
@@ -71,5 +85,13 @@ let parse_short_title loc t =
         (Error.make
            "Short titles cannot contain other block than a single paragraph" loc)
 
+let parse_toc_status loc (t : tag_payload) =
+  match t with
+  | [
+   { Location_.value = `Paragraph [ { Location_.value = `Word "open"; _ } ]; _ };
+  ] ->
+      Result.Ok (Location_.at loc (Toc_status `Open))
+  | _ -> Error (Error.make "@toc_status can only take the 'open' value" loc)
```
jonludlam, line 94: Do you have any other values in mind that this might take?

jonludlam (commented):
This is looking good to me. We still need a better order for the sidebar when it's not specified in an index file. Other than that, I'm happy with this!

src/odoc/indexing.ml
panglesd: I think the idea when we did that was not to have the full object in memory. Now, I think that we anyway have it in memory. I'll remove the `first` thing.

src/model/frontmatter.ml
panglesd: The value I had in mind was `hidden` or `private` where it is both open and not rendered as a link.

src/odoc/indexing.ml
panglesd: Done in [1e59b62](https://github.com/ocaml/odoc/pull/1251/commits/1e59b6201b1505008918983f4bfaf3b0bfb6897c)!

src/html/utils.ml
panglesd: Done in [f868577](https://github.com/ocaml/odoc/pull/1251/commits/f868577f5f5f98383300fc98b578acf7495051f7)

jonludlam:
Is `@order_category` ever going to be used by people documenting their packages, or is it more for the default index pages that are generated?

panglesd:
There is a priori no need for `@order_category` for the people documenting their package, since they can already order the pages using `children_order`. Should I move the doc of this tag to the `driver.mld` page? Or make it explicit that it is for auto-generated pages?
In my opinion, it is fine to let people use it if they wish.

jonludlam:
I'd rather make something like this internal only -- in the sense that it's not for 'ordinary' package authors -- for now; anything that people put in their actual repositories we have to support forever. We can document it in `driver.mld` but with some caveats that it might change in future.

src/document/sidebar.ml
```diff
@@ -2,72 +2,54 @@ open Odoc_utils
 open Types
 module Id = Odoc_model.Paths.Identifier
 
-type entry = Url.t option * Inline.one
+type entry = {
+  url : Url.t;
```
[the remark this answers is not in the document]
panglesd: `url` is needed for computing breadcrumbs: You need that to orient yourself inside the provided sidebar, from the root to your current position.

doc/odoc.mld
```diff
@@ -1,8 +1,7 @@
-{0 [odoc]}
+@children_order dune odoc_for_authors cheatsheet features ocamldoc_differences interface parent_child_spec driver
```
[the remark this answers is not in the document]
panglesd: I though that the first thing was "how to build doc" before "how to write docs", which is why I put this page first, but I agree it is not ideal... I'll think about that a bit more.

src/odoc/bin/main.ml
```diff
@@ -1143,6 +1132,13 @@ module Odoc_html_args = struct
       (parser, printer)
   end
 
+  let escape_breadcrumb =
+    let doc =
+      "Wether to add a 'Home' breadcrumb to go up the root of the given \
+       sidebar."
+    in
+    Arg.(value & flag & info ~docv:"escape" ~doc [ "escape-breadcrumb" ])
```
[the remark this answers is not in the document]
panglesd: I went for `home-breadcrumb`

src/html/generator.ml
```diff
@@ -498,24 +500,146 @@ end
...
+      | true, Some { node; _ } -> (
+          match page_parent node.url.page with
+          | None -> []
+          | Some parent ->
+              let href =
+                Some
+                  (Link.href ~config ~resolve:(Current current_url)
+                     (Odoc_document.Url.from_path parent))
+              in
+              [ { href; name = [ Html.txt "🏠" ]; kind = parent.kind } ])
```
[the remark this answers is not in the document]
panglesd: Done

jonludlam:
Thanks!

PR #1252, opened 2024-11-21
Further voodoo

jonludlam:
Minor changes for voodoo mode

### DIFF
--- a/src/driver/compile.ml
+++ b/src/driver/compile.ml
@@ -278,8 +278,11 @@ let html_generate ~occurrence_file ~remaps ~generate_json output_dir linked =
       let sidebar =
         match sidebar with
         | None -> None
-        | Some { output_file; json } ->
+        | Some { output_file; json; pkg_dir } ->
             Odoc.sidebar_generate ~output_file ~json index.output_file ();
+            Odoc.sidebar_generate
+              ~output_file:Fpath.(output_dir // pkg_dir / "sidebar.json")
+              ~json:true index.output_file ();
             Some output_file
       in
       (sherlodoc_index_one ~output_dir index, sidebar)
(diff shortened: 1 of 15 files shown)


src/driver/compile.ml
```diff
@@ -278,8 +278,11 @@ let html_generate ~occurrence_file ~remaps ~generate_json output_dir linked =
       let sidebar =
         match sidebar with
         | None -> None
-        | Some { output_file; json } ->
+        | Some { output_file; json; pkg_dir } ->
             Odoc.sidebar_generate ~output_file ~json index.output_file ();
+            Odoc.sidebar_generate
+              ~output_file:Fpath.(output_dir // pkg_dir / "sidebar.json")
+              ~json:true index.output_file ();
```
Julow, line 285: Where is this json file used ? I find this code a bit brittle because if this file is ever needed somewhere else, nothing will ensure that it continues to be generated.
Any way the caller could set `json = true` ?

src/driver/voodoo.ml
```diff
@@ -222,6 +179,8 @@ let process_package pkg =
     |> List.flatten
   in
   let libraries = meta_libraries @ non_meta_libraries in
+  let pkg_dir = top_dir pkg in
+  let doc_dir = Fpath.(pkg_dir / "doc") in
```
Julow, line 183: The doc directory is back ? Is this to avoid breaking urls ?

Julow (approved):
Looks good.

src/driver/compile.ml
jonludlam: It's used by ocaml.org - this is the sidebar json that served up for the per-package sidebar. The ocaml.org server reads this and renders its own sidebar.

src/driver/voodoo.ml
jonludlam: I chose not to change the layout on ocaml.org - mostly because it'd require a bunch of redesign work, and I don't have the capacity to do that.

PR #1259, opened 2024-12-11
Vendor sherlodoc

Julow:
This vendors sherlodoc to simplify maintaining compatibility using a Git subtree. This is not a complicated concept, sherlodoc's code is simply copied into odoc's source tree and can be modified normally.

The commits made by the subtree command also point to sherlodoc's history. This allows to easily pull changes made to sherlodoc or to easily upstream change made in the vendored copy. Here are the commands I used to pull some unmerged PRs on sherlodoc:
- `git subtree pull -P vendor/sherlodoc/ --squash https://github.com/panglesd/sherlodoc odoc_3_compat_clean` to pull https://github.com/art-w/sherlodoc/pull/46 into the vendored copy.
- `git subtree pull -P vendor/sherlodoc/ --squash https://github.com/Julow/sherlodoc rel_blob_path` to pull https://github.com/art-w/sherlodoc/pull/44

This system might not work for ever and conflicts might be too hard to resolve for git subtree's algorithm. That's not a problem as this has no cost to begin with.

The last commit makes sherlodoc be part of the `odoc-driver` package.


dune
```diff
@@ -23,3 +23,5 @@
   (progn
    (bash "diff doc/driver.mld doc/driver.mld.corrected >&2 || true")
    (cat doc/driver-benchmarks.json))))
+
+(vendored_dirs vendor)
```
panglesd, line 27: I don't think we want to consider sherlodoc as vendored: we want the tests to be run, the warnings to be displayed, ...

Julow:
Closed in favor of https://github.com/ocaml/odoc/pull/1263

PR #1207, opened 2024-10-07
Vendor sherlodoc as a submodule and run the driver in CI

Julow:
This vendors Sherlodoc as a submodule to make running the driver easier locally and in CI.
This allows the CI to check that the driver runs without errors and that sherlodoc continues to build.
Currently, it's checked out to <https://github.com/EmileTrotignon/sherlodoc/pull/1>. It's expected that the remote and branch of the submodule is changed often.
When testing the driver, make sure to have sherlodoc in the workspace with `git submodule update --init`.

PR #1263, opened 2024-12-13
Merge Sherlodoc

Julow:
This replaces https://github.com/ocaml/odoc/pull/1259 and https://github.com/ocaml/odoc/pull/1207

Sherlodoc will now be maintained as part of Odoc, except for its website. The entire history of Sherlodoc is kept for proper attribution using git subtree but the code is not intended to be extracted again. It is based on https://github.com/Julow/sherlodoc/tree/odoc3-merge
Sherlodoc will be released at the same time as Odoc with the same version numbers.


panglesd:
Benchmark CI is failing but it is not checking anything, that's just a infra issue, so it's not an issue for us now.

PR #1264, opened 2024-12-13
Add odoc 3 features in odoc's doc

panglesd:
This adds to the cheatsheet:
- Example of images, one of them being included as an asset,
- Example of cross-package linking, to a module from the `fmt` library and to a page from the `cmdliner` package

This also adds to the index page the "Ode to odoc" song.


doc/cheatsheet.mld
```diff
@@ -66,12 +66,20 @@ Quick reference for the odoc language!
 
       See {{!Odoc_odoc.Compile.compile}this function}.
 
+      See {{!/fmt/Fmt.pf}this function from another library}.
+
+      See {{!/cmdliner/tutorial}this page from another package}.
```
Julow, line 71: This reference doesn't work when building the doc with the driver.

doc/odoc.mld
```diff
@@ -3,6 +3,8 @@
 
 {0 The [odoc] documentation generator}
 
+{audio:https://choum.net/panglesd/odoc.mp3}
```
Julow, line 6: I just want to add this to my playlist.

Julow (approved):
Awesome :)

doc/cheatsheet.mld
panglesd: Hum that is strange, it should...

panglesd: I redid the test from scratch and the reference works. Are you sure you tested it right?

Issue #1253, opened 2024-11-26
No parentheses around word-like operators?


yawaramin:
Odoc generated HTML output seems to be not parenthesizing operators which have word-like characters inside them. Eg https://ocaml.org/p/lwt/latest/doc/Lwt_result/Syntax/index.html

This is formatted as:

```ocaml
val let* : ('a, 'e) t -> ('a, ('b, 'e) t) -> ('b, 'e) t
```

Shouldn't it be:

```ocaml
val ( let* ) : ('a, 'e) t -> ('a, ('b, 'e) t) -> ('b, 'e) t
```

?

PR #1268, opened 2024-12-20
 Add missing parentheses around identifier 'let*' 

Julow:
Fix https://github.com/ocaml/odoc/issues/1253

### DIFF
--- a/src/model/names.ml
+++ b/src/model/names.ml
@@ -1,3 +1,5 @@
+open Odoc_utils
+
 (** Returns [true] on chars that are part of operators. *)
 let operator_char = function
   (* https://ocaml.org/manual/5.2/lex.html#core-operator-char *)
--- a/src/model/names.ml
+++ b/src/model/names.ml
@@ -10,10 +10,11 @@ let operator_char = function
       true
   | _ -> false
 
-let parenthesise name =
-  match name with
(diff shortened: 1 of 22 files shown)


yawaramin:
Does this also handle the `( or )` operator?

Julow:
It did not, thanks !

src/model/names.ml
```diff
@@ -1,19 +1,22 @@
...
+  (* https://ocaml.org/manual/5.2/indexops.html#s:index-operators *)
+  | '(' | ')' | '[' | ']' | '{' | '}' ->
+      true
+  | _ -> false
+
+let is_operator = function
+  | "asr" | "land" | "lor" | "lsl" | "lsr" | "lxor" | "mod" | "or" -> true
+  | name -> String.exists operator_char name
+
+let parenthesise name = if is_operator name then "(" ^ name ^ ")" else name
```
yawaramin, line 19: I would suggest to do `"( " ^ name ^ " )"` because of the ambiguous `(*)`.

yawaramin (approved):

src/model/names.ml
Julow: That's a good point!

Changing this code would change URLs, which is not fine. More work is needed to handle that in the generated code only. It's in `src/document/generator.ml` if you have some time to work on it.

jonludlam:
Thanks!

PR #1270, opened 2025-01-08
Driver: split into three

jonludlam:

PR #1271, opened 2025-01-09
Some minor fixes

jonludlam:
This is on top of #1270

### DIFF
--- a/src/markdown/doc_of_md.ml
+++ b/src/markdown/doc_of_md.ml
@@ -64,11 +64,15 @@ let textloc_to_loc ~locator textloc =
   let point_of_line_and_byte_pos ~locator:(location, line_trim_counts) l pos =
     let line_num, line_pos = l in
     let line = location.Lexing.pos_lnum + line_num - 1 in
-    let column = line_trim_counts.(line_num - 1) + (pos - line_pos) in
-    let column =
-      match line_num with 1 -> comment_col ~location + column | _ -> column
-    in
-    { Loc.line; column }
+    try
+      let column = line_trim_counts.(line_num - 1) + (pos - line_pos) in
+      let column =
+        match line_num with 1 -> comment_col ~location + column | _ -> column
+      in
+      { Loc.line; column }
+    with _ ->
+      (* Presumably this is the above-mentioned bug that's being hit. *)
+      { Loc.line = -1; column = -1 }
   in
   let file = Textloc.file textloc in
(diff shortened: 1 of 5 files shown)


src/markdown/odoc_md.ml
```diff
@@ -53,7 +57,10 @@ let run input_s parent_id_str odoc_dir =
   let page = mk_page input_s id content in
 
   let output =
-    Fpath.(v odoc_dir // v parent_id_str / ("page-" ^ page_name ^ ".odoc"))
+    match parent_id_opt with
+    | None -> Fpath.(v odoc_dir / ("page-" ^ page_name ^ ".odoc"))
+    | Some parent_id_str ->
+        Fpath.(v odoc_dir // v parent_id_str / ("page-" ^ page_name ^ ".odoc"))
```
Julow, line 63: The rhs of the path could be shared between the two cases.

src/markdown/doc_of_md.ml
```diff
@@ -64,11 +64,15 @@ let textloc_to_loc ~locator textloc =
...
-    { Loc.line; column }
+    try
+      let column = line_trim_counts.(line_num - 1) + (pos - line_pos) in
+      let column =
+        match line_num with 1 -> comment_col ~location + column | _ -> column
+      in
+      { Loc.line; column }
+    with _ ->
+      (* Presumably this is the above-mentioned bug that's being hit. *)
+      { Loc.line = -1; column = -1 }
```
Julow, line 75: The only thing that raises a `Invalid_argument` is the array access at line 68. I think this could be handled better. It's also possible that `locator` is not computed right by `massage_comment` as it's so complicated.

src/markdown/odoc_md.ml
```diff
@@ -40,10 +40,14 @@ let mk_page input_s id elements =
     frontmatter;
```
Julow, line 40: The code is not formatted.

src/markdown/doc_of_md.ml
jonludlam: There's quite a lot we can do with the markdown code, but I don't want to spend too much time on it until I'm sure people want it and are using it.

src/markdown/odoc_md.ml
jonludlam: oops yes!

Julow (approved):
Perfect, let's merge !

PR #1275, opened 2025-01-15
More misc fixes

jonludlam:
- The custom-layout option is for 'monorepo mode'
- As is the short-title option
- The duplicates problem is causing bug in trunk right now
- The fix for Cmx_format is only helpful in 'monorepo mode'
- The versions thing and the global_config fixes come up when building in 'normal mode' when in a dune repo

### DIFF
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -687,11 +687,16 @@ end = struct
     validate_current_package ?detected_package page_roots current_package
 
   let link directories page_roots lib_roots input_file output_file
-      current_package warnings_options open_modules =
+      current_package warnings_options open_modules custom_layout =
     let input = Fs.File.of_string input_file in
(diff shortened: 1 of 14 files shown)


src/odoc/bin/main.ml
```diff
@@ -222,7 +222,7 @@ end = struct
 
   let compile hidden directories resolve_fwd_refs dst output_dir package_opt
       parent_name_opt parent_id_opt open_modules children input warnings_options
-      unique_id =
+      unique_id short_title =
```
Julow, line 225: Perfect for making the experience consistent on every packages. The short title can be removed from `doc/index.mld` when the driver start to use this flag.

Julow (approved):
Let's merge!

jonludlam:
Thanks!

PR #1276, opened 2025-01-15
Monorepo mode fixes

jonludlam:


src/driver/odoc_units_of.ml
```diff
@@ -264,11 +276,23 @@ let packages ~dirs ~extra_paths ~remap ~gen_indices (pkgs : Packages.t list) :
...
+        @ mld_units @ asset_units @ md_units)
+    | Custom ->
+      if pkg.name = "pkg" then
+        let others :> t list = Landing_pages.make_custom dirs index_of (List.find (fun p -> p.Packages.name = "pkg") pkgs) in
+        others @ List.concat
+          (mld_units @ asset_units @ md_units @ lib_units)
+      else
+        List.concat
+          ((pkg_index () :: src_index () :: lib_units)
+          @ mld_units @ asset_units @ md_units)
```
Julow, line 292: When `pkg.name = "pkg"` is false, this returns the same thing as in the Normal or Voodoo case. The code could be made clearer.

src/driver/odoc_units_of.mli
```diff
@@ -1,9 +1,14 @@
 open Odoc_unit
 
+type indices_style =
+| Voodoo
+| Normal
+| Custom
```
Julow, line 6: The Custom mode generates index pages in every places it needs. Could it be called Automatic ?

Julow (commented):
This needs to be formatted.

src/driver/monorepo_style.ml
```diff
@@ -67,64 +110,93 @@ let of_dune_build dir =
...
+                let cmtidir =
+                  Fpath.(
+                    append dir
+                      (v lib.source_dir
+                      / Printf.sprintf ".%s.objs" libname
+                      / "byte"))
+                in
+                let id_override =
+                  Fpath.relativize
+                    ~root:Fpath.(v "_build/default")
```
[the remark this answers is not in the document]
jonludlam: ~root has `dir` prepended:

```
let root = Fpath.(dir / "_build" / "default") in
```

PR #1277, opened 2025-01-16
Fix stack overflow when highlighting large source files

Julow:
This is enough to highlight `reason_parser.ml` in `reason.3.8.0`, which is one very large file that Odoc has to process.


jonludlam:
Thanks!

PR #1279, opened 2025-01-17
Fix backward home-breadcrumb logic

jonludlam:


jonludlam:
The breadcrumbs logic is a bit tricky to follow. I'd like to do a better tidy up, but not right now :-)

Julow (approved):
Looks good to me :)

jonludlam:
Thanks!

PR #1280, opened 2025-01-17
Driver: allow overriding main package index

jonludlam:


src/driver/landing_pages.mli
```diff
@@ -1,5 +1,18 @@
+type u = unit
```
Julow, line 1: What is the point of this ?

src/driver/odoc_units_of.ml
```diff
@@ -283,10 +283,31 @@ let packages ~dirs ~extra_paths ~remap ~indices_style (pkgs : Packages.t list) :
...
+  match indices_style with
+  | Normal None ->
+      let gen_indices :> t = Landing_pages.package_list ~dirs ~remap pkgs in
+      gen_indices :: List.concat_map of_package pkgs
+  | Normal (Some index_mld) -> (
+      match Bos.OS.File.read index_mld with
+      | Error (`Msg msg) ->
+          Logs.err (fun m ->
+              m "Failed to read index_mld file '%a': %s" Fpath.pp index_mld msg);
+          []
```
Julow, line 298: Return no pages, not even the pkgs pages ? I think the driver could abort with an exception instead.

src/driver/odoc_units_of.mli
```diff
@@ -1,6 +1,6 @@
 open Odoc_unit
 
-type indices_style = Voodoo | Normal | Automatic
+type indices_style = Voodoo | Normal of Fpath.t option | Automatic
```
Julow, line 3: It would be nice to document Normal's arguments, or perhaps use an inline record.

src/driver/landing_pages.mli
```diff
@@ -1,5 +1,18 @@
...
 
+val make_index :
+  dirs:dirs ->
+  rel_dir:Fpath.t ->
+  ?libs:(Packages.t * Packages.libty) list ->
+  ?pkgs:Packages.t list ->
+  ?index:index ->
+  enable_warnings:bool ->
+  content:(Format.formatter -> u) ->
+  u ->
```
Julow, line 13: The last unit argument is unnecessary.

Julow (commented):
What is the usecase of this page ?

src/driver/landing_pages.mli
jonludlam: Odoc_unit exposes a type 'unit' that's not `()`! Perhaps I should just fix that in this PR, and avoid this mess.

jonludlam:
> What is the usecase of this page ?

This is the main index page - the one that ends up in `_html/index.html`.

src/driver/odoc_units_of.ml
jonludlam: Hmm, I think what I'll do is move the `load` logic into `odoc_driver.ml`. The problem is that by the time we get here we've done a load of work and I didn't want to just exit. Something something sunk cost... :-)

Julow:
> This is the main index page - the one that ends up in _html/index.html.

Sorry, I meant what the users will put there ? If the package list page is not desired, what could users do better in Odoc syntax ? I'd expect people to want to integrate this page in their website, and this isn't helping.

jonludlam:
Well, if you're trying to produce a site documenting multiple packages, like we might well do for odoc, then the autogenerated page is not particularly great. This gives a way to have a front page that can be customized as you see fit.

src/driver/landing_pages.mli
jonludlam: it was there because all the args are either optional or named. I've removed it and made all args named now.

PR #1284, opened 2025-01-22
Simplified search index

jonludlam:
This is for search on ocaml.org.

### DIFF
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -484,8 +484,8 @@ module Indexing = struct
              || Fpath.has_ext "js" (Fpath.v file)) ->
         Error
           (`Msg
-            "When generating a json index, the output must have a .json file \
-             extension")
+            "When generating a json index, the output must have a .json or .js \
+             file extension")
     | Some file, `Marshall when not (Fpath.has_ext "odoc-index" (Fpath.v file))
       ->
         Error
@@ -542,7 +542,8 @@ module Indexing = struct
     in
     let wrap_json =
       let doc =
-        "whether to wrap the json file. Only has an effect in json output mode."
+        "Not intended for general use. Wraps the json output in a JavaScript \
+         variable assignment, and assumes the use of fuse.js"
       in
       Arg.(value & flag & info ~doc [ "wrap-json" ])
     in
--- a/src/odoc/bin/main.ml
+++ b/src/odoc/bin/main.ml
@@ -478,7 +478,10 @@ module Indexing = struct
 
   let output_file ~dst marshall =
     match (dst, marshall) with
-    | Some file, `JSON when not (Fpath.has_ext "json" (Fpath.v file)) ->
+    | Some file, `JSON
+      when not
+             (Fpath.has_ext "json" (Fpath.v file)
+             || Fpath.has_ext "js" (Fpath.v file)) ->
         Error
           (`Msg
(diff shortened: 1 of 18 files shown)


src/odoc/indexing.ml
```diff
@@ -23,21 +23,29 @@ let parse_input_files input =
...
+            Json_search.of_entry ~simplified ?occurrences entry :: acc)
           acc hierarchy)
       [] hierarchies
   in
   Format.fprintf output "%s" (Odoc_utils.Json.to_string (`Array (List.rev all)));
+  if wrap then
+    Format.fprintf output
+      ";\n\
+       const options = { keys: ['name', 'comment'] };\n\
+       var idx_fuse = new Fuse(documents, options);\n";
```
panglesd, line 47: If we are making that the use of Fuse part of odoc (and not part of `odoc_driver`) than I think we should add it to `html_support_files`. Personally, I would prefer to have Fuse use part of `odoc_driver` (although I'm fine with both solutions).

src/odoc/indexing.ml
```diff
@@ -23,21 +23,29 @@ let parse_input_files input =
...
 
-let compile_to_json ~output ~occurrences hierarchies =
+let compile_to_json ~output ~occurrences ~wrap ~simplified hierarchies =
   let output_channel =
     Fs.Directory.mkdir_p (Fs.File.dirname output);
     open_out_bin (Fs.File.to_string output)
   in
   let output = Format.formatter_of_out_channel output_channel in
+  if wrap then Format.fprintf output "let documents = ";
+  Format.fprintf output "[";
```
panglesd: I'm not sure to understand this new `[`. It does not seem to be closed? And, it's independent from the new options: is it fixing a bug?

src/odoc/bin/main.ml
```diff
@@ -478,7 +478,10 @@ module Indexing = struct
...
   let output_file ~dst marshall =
     match (dst, marshall) with
-    | Some file, `JSON when not (Fpath.has_ext "json" (Fpath.v file)) ->
+    | Some file, `JSON
+      when not
+             (Fpath.has_ext "json" (Fpath.v file)
+             || Fpath.has_ext "js" (Fpath.v file)) ->
         Error
           (`Msg
             "When generating a json index, the output must have a .json file \
```
panglesd, line 487: Or `.js` in this PR

src/odoc/indexing.ml
panglesd: Ok, my comment is already outdated :)

jonludlam: Nope, dunno how that got there. It's in the commit on the branch that I thought I was previously testing - but it can't possibly work! A mystery!

src/odoc/indexing.ml
jonludlam: I think this is a temporary grotty hack that I'd like to go away as soon as we've got proper sherlodoc support in ocaml.org. I'll see about hiding the options in the CLI.

panglesd: If it is something we are not going to keep, why not making it generated by the driver, from the output json file? This way, no need to hide it, and we have a guarantee no-one relies on it, if it happens to stay a bit longer before being removed.

panglesd: (This could be done after the beta release, but before the actual release!

PR #1292, opened 2025-01-24
test: Remove 'type_desc_to_yojson' library

Julow:
The library is installable and is built when the tests are not going to be run.
This fails in the lower-bounds CI because of the requirement on yojson is not respected when the dependencies are installed with 'with-test'.

    "yojson" {>= "2.1.0" & with-test}


jonludlam:
Thanks!

Issue #1286, opened 2025-01-22
Give a size to images


panglesd:
It is very nice to be able to include images! However, we might want to give them a size, either in pixel, or percentage, or ...

This requires a new syntax, and likely a general way of giving attributes to nodes...

PR #1297, opened 2025-01-28
Add png, pdf and jpeg image support to the latex backend

Octachron:
This  small PR adds support for images to the latex backed for png, pdf, and jpeg internal assets.

As a side-note, I am slightly surprised that there is no support for width information on the media tags since both image and video are hard to place properly without fixing their size.

### DIFF
--- a/src/latex/generator.ml
+++ b/src/latex/generator.ml
@@ -97,7 +97,7 @@ let elt_size (x : elt) =
   | Code_fragment _ | Tag _ | Break _ | Ligaturable _ ->
       Small
   | List _ | Section _ | Verbatim _ | Raw _ | Code_block _ | Indented _
-  | Description _ ->
+  | Description _ | Image _ ->
       Large
   | Table _ | Layout_table _ -> Huge
 
@@ -167,6 +167,7 @@ let rec pp_elt ppf = function
   | Indented x -> Raw.indent pp ppf x
   | Ligaturable s -> Fmt.string ppf s
   | Tag (s, t) -> tag s ppf t
+  | Image target -> Raw.includegraphics Fpath.pp ppf target
 
 and pp ppf = function
   | [] -> ()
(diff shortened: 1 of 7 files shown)


panglesd:
Thanks a lot for the PR!

Yes, I agree that it is the first thing we should do for images. (#1286)

The main problem is deciding the syntax. There is no existing syntax for attributes so we have to create something new. I can't find issues with the discussion, it must have happened elsewhere, but we could not find a consensus on the syntax and decided to start without it (in the worst case, images themselves can be resized I think).

I'll restart the discussion in #1286 with some propositions soon.

src/latex/generator.ml
```diff
@@ -283,9 +284,24 @@ let rec block ~in_source (l : Block.t) =
...
+      | "" | ".jpeg" | ".png" | ".pdf" ->
+        let fpath = Fpath.v (String.concat Fpath.dir_sep (dir @ [file])) in
+        [ Image fpath ]
+      | _ ->
+        txt ~verbatim:false ~in_source:false [ alt ]
         @ if in_source then [] else [ Break Paragraph ]
+      end
+    | Image (Internal Unresolved, c) | Audio(Internal _, c) | Video (Internal _, c) ->
+        txt ~verbatim:false ~in_source:false [ c ]
+        @ if in_source then [] else [ Break Paragraph ]
```
panglesd, line 299: Nitpick: using a `when` clause removes the duplication of the "alternative text" usage:

```suggestion
    | Image (Internal (Resolved x), alt)
      when List.mem Fpath.(get_ext @@ v file) [ ""; ".jpeg"; ".png"; ".pdf" ] ->
        let dir, file = Link.get_dir_and_file x.page in
        let fpath = Fpath.v (String.concat Fpath.dir_sep (dir @ [ file ])) in
        [ Image fpath ]
    | Image (Internal _, c) | Audio (Internal _, c) | Video (Internal _, c) ->
        txt ~verbatim:false ~in_source:false [ c ]
        @ if in_source then [] else [ Break Paragraph ]
```

panglesd (approved):
Looks good to me!

I'm curious about the choice of extensions: shouldn't `jpg`, and capitalized extensions, be allowed as well?
I suspect the list of supported format depends on the latex implementation.

Maybe the list taken from [here](https://tex.stackexchange.com/questions/469734/which-file-does-includegraphics-select-if-there-are-multiple-files-with-differen) could be be a valid choice:

    .pdf,.png,.jpg,.mps,.jpeg,.jbig2,.jb2,.PDF,.PNG,.JPG,.JPEG,.JBIG2,.JB2

Also I think it needs ocamlformatting! (maybe we should use [this action](https://discuss.ocaml.org/t/ocamlformat-and-github-actions/15464/8)?)

panglesd:
I'm not sure to understand! `object-fit` decides how images are placed in their bounding box.

But if the user can't decide the dimension of the bounding box, it does not help much?

Also, it is hard for odoc to decide on a "good size for all": eg [vg](https://erratique.ch/software/vg/doc/tutorial.html) has small images, but in other contexts a big image makes sense.

Am I missing something?

dbuenzli:
IIRC for responsive images you basically want users to define *one* dimension with a relative unit. If it's horizontal it should be a percentage of the measure. If it's vertical a multiple of the leading, capped so that it doesn't overflow the measure. That's for running text but if you have a table for example then your box may already have constraints at which point (again IIRC) `object-fit` comes handy.

src/latex/generator.ml
panglesd: Sorry, I haven't seen you needed "file" for this...

dbuenzli:
Basically user should work with the assumption that the `css` has:

```
img, picture, video, canvas, svg { max-width: 100%; display: block }
```

src/latex/generator.ml
Octachron: I have lifted the corresponding code to separate auxiliary function to avoid the duplication.

Octachron:
I have added support for all image formats supported by pdftex by default.

panglesd (approved):
Thanks!

I rebased to fix a conflict and put the changelog entry into the "unreleased" section.

Will merge as soon as the CI finishes green.

PR #1299, opened 2025-02-03
Driver: Fix monorepo mode

jonludlam:


src/driver/monorepo_style.ml
```diff
@@ -102,11 +102,15 @@ let of_dune_build dir ~extra_pkgs ~extra_libs =
       let all_lib_deps =
         List.fold_left
           (fun acc (l : library) ->
+            let libs =
+              List.filter_map (fun uid -> Util.StringMap.find_opt uid uid_to_libname)
+              l.requires
+            in
+            let libs =
+              if l.name = "stdlib" then libs else "stdlib"::libs
```
Julow, line 110: Conditionally adding `stdlib` is the only change ?

jonludlam: yep

PR #1304, opened 2025-02-06
Fix suppress_warnings parameter

jonludlam:
It wasn't being propagated to all of the doc strings

### DIFF
--- a/src/model/error.mli
+++ b/src/model/error.mli
@@ -41,7 +41,7 @@ val catch_errors_and_warnings : (unit -> 'a) -> 'a with_errors_and_warnings
 type warnings_options = {
   warn_error : bool;  (** If [true], warnings will result in an error. *)
   print_warnings : bool;  (** Whether to print warnings. *)
-  suppress_warnings : bool;  (** Whether to suppress warnings. *)
+  warnings_tag : string option;  (** Whether to tag references. *)
 }
 
 val handle_warnings :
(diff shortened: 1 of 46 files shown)


src/odoc/bin/main.ml
```diff
@@ -713,8 +713,16 @@ end = struct
     let detected_package = find_root_of_input page_roots input in
     validate_current_package ?detected_package page_roots current_package
 
+  let warnings_tags =
+    let doc =
+      "Warnings tags. Only report warnings in references that have been compiled \
```
panglesd: I think it does report warnings that were compiled without tags.

panglesd: Maybe 

> Filter warnings that were compiled with a tag that is not in the list of --warnings-tags passed?

?

src/model/error.mli
```diff
@@ -41,7 +41,7 @@ val catch_errors_and_warnings : (unit -> 'a) -> 'a with_errors_and_warnings
 type warnings_options = {
   warn_error : bool;  (** If [true], warnings will result in an error. *)
   print_warnings : bool;  (** Whether to print warnings. *)
-  suppress_warnings : bool;  (** Whether to suppress warnings. *)
+  warnings_tag : string option;  (** Whether to tag references. *)
```
panglesd, line 44:
```suggestion
  warnings_tag : string option;  (** Whether to tag references to filter them later. *)
```

panglesd (approved):
Only microscopic nitpicks, looks good to me!

jonludlam:
Thanks!

Issue #610, opened 2021-02-27
html-fragment turns headings into paragraphs


dbuenzli:
```
> echo "{0:sample OCaml 4.11.1 sample package documentation}" > frag.mld
> odoc html-fragment frag.mld
<p>OCaml 4.11.1 sample package documentation</p>
```

This breaks the `odig odoc --index-intro` option used e.g. by https://docs.mirage.io/, https://b0-system.github.io/odig/doc/, https://ocaml.janestreet.com/ocaml-core/latest/doc/index.html or https://binaryanalysisplatform.github.io/bap/api/master/index.html

dbuenzli:
Still an issue 3.0.0~beta1.

PR #1306, opened 2025-02-07
Fix issue #610 - `odoc html-fragment` not producing headings correctly

jonludlam:
@dbuenzli could you please check this works for you? As a quick check, with this I get:

```
$ echo "{0:sample OCaml 4.11.1 sample package documentation}" > frag.mld 
$ dune exec -- odoc html-fragment frag.mld  
<h1 id="sample"><a href="#sample" class="anchor"></a>OCaml 4.11.1 sample package documentation</h1>
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,8 @@
 - Fix incomplete handling of `--suppress-warnings` (now `--warnings-tag`)
   (@jonludlam, #1304)
 - Fix bug in odoc_driver_voodoo related to virtual libraries (@jonludlam, #1309)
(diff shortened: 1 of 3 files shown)


dbuenzli:
Works as you mention.

panglesd (commented):
I was wondering where `Odoc_html.Generator.doc` is used to remove it/find related bugs, and it seems to be used only [here](https://github.com/ocaml/odoc/blob/master/src/search/html.ml#L215-L217). And I think this code has same problem (headings are turned into paragraphs).

jonludlam:
We'll fix any issues with search separately.

PR #1309, opened 2025-02-12
Further fixes for virtual libraries

jonludlam:
So in the docs-ci, we don't have the 'prep' dir (with all the cmts/cmtis) for any package apart from the one being built. This is a problem for implementations of virtual libraries that aren't in the same package as the virtual library itself, as we try to copy the cmti from the virtual library to the implementation. 

The solution here it to copy _all_ cmtis for libraries without archives to the odoc dir. These will then be available to be used.

### DIFF
--- a/src/driver/compile.ml
+++ b/src/driver/compile.ml
@@ -138,8 +138,7 @@ let compile ?partial ~partial_dir (all : Odoc_unit.any list) =
           (Odoc_unit.Pkg_args.compiled_libs unit.pkg_args)
       in
       Odoc.compile ~output_dir:unit.output_dir ~input_file:unit.input_file
-        ~includes ~warnings_tag:unit.pkgname
-        ~parent_id:unit.parent_id;
+        ~includes ~warnings_tag:unit.pkgname ~parent_id:unit.parent_id;
       (match unit.input_copy with
       | None -> ()
       | Some p -> Util.cp (Fpath.to_string unit.input_file) (Fpath.to_string p));
--- a/src/driver/compile.ml
+++ b/src/driver/compile.ml
@@ -147,26 +147,25 @@ let compile ?partial ~partial_dir (all : Odoc_unit.any list) =
     in
     let rec compile_mod : string -> ('a list, [> `Msg of string ]) Result.t =
      fun hash ->
+      let map_units =
+        Fiber.List.map (fun unit ->
+            match
+              Hashtbl.find_opt tbl
+                (hash, Odoc.Id.to_string unit.Odoc_unit.parent_id)
+            with
+            | Some p ->
+                Promise.await p;
+                None
+            | None ->
+                let p, r = Promise.create () in
+                Hashtbl.add tbl (hash, Odoc.Id.to_string unit.parent_id) p;
+                let _result = compile_one compile_mod unit in
+                Promise.resolve r ();
+                Some unit)
+      in
       try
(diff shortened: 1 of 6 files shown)


src/driver/odoc_unit.ml
```diff
@@ -203,12 +199,32 @@ let fix_virtual ~(precompiled_units : intf t list Util.StringMap.t)
...
+                        xs
+                    in
+                    match possibles with
+                    | [] ->
+                        Logs.debug (fun m -> m "Not replacing input file");
+                        unit
+                    | x :: _ ->
+                        Logs.debug (fun m ->
+                            m "Replacing input_file of unit with %a" Fpath.pp x);
+                        { unit with input_file = x })))
```
panglesd, line 228: Useless nitpick: why not use `List.find_map`?

```suggestion
                    let possibles =
                      List.find_map
                        (fun x ->
                          match x.input_copy with
                          | Some x ->
                              if
                                x |> Bos.OS.File.exists
                                |> Result.value ~default:false
                              then Some x
                              else None
                          | None -> None)
                        xs
                    in
                    match possibles with
                    | None ->
                        Logs.debug (fun m -> m "Not replacing input file");
                        unit
                    | Some x ->
                        Logs.debug (fun m ->
                            m "Replacing input_file of unit with %a" Fpath.pp x);
                        { unit with input_file = x })))
```

src/driver/compile.ml
```diff
@@ -111,76 +114,62 @@ let compile ?partial ~partial_dir (all : Odoc_unit.any list) =
...
+                  let p, r = Promise.create () in
+                  Hashtbl.add tbl (hash, Odoc.Id.to_string unit.parent_id) p;
+                  let _result = compile_one compile_mod unit in
+                  Promise.resolve r ();
+                  Some unit)
+            units
+        in
+        Ok (List.filter_map Fun.id r)
+      with Not_found ->
+        Error (`Msg ("Module with hash " ^ hash ^ " not found"))
```
panglesd, line 172: I think it's acceptable to use exceptions here, but the exception handling seems quite for from where it can be raised.

Would it be possible to use `StringMap.find_opt`, turn the option into a result and use a `let*`?

src/driver/odoc_unit.ml
```diff
@@ -74,6 +74,8 @@ let pp_index fmt x =
 type 'a t = {
   parent_id : Odoc.Id.t;
   input_file : Fpath.t;
+  input_copy : Fpath.t option;
+      (* Used to stash cmtis from virtual libraries into the odoc dir for voodoo mode *)
```
panglesd, line 78: Would adding a link to this PR allow for more information while not making it too verbose?

(When you read something like that you might wonder why we need to "stash" those cmti there, and it's definitely possible to find the PR from the commit from a git blame but is not the best experience!)

panglesd (approved):
Looks good.

I think it would be nice to add a mention on how virtual libraries should be treated in the odoc_driver documentation, but that could be in another PR.

src/driver/odoc_unit.ml
jonludlam: Nice thanks!

src/driver/odoc_unit.ml
jonludlam: Hmm, yeah we could do this for now, then when we've got some actual docs we can point there instead.

PR #1312, opened 2025-02-13
Fix indentation bug with verbatim in tags

### LINKED ISSUE #1139: Weird indentation for verbatim blocks within tags
Given something like:
```ocaml
(** Some docs

     @param Start a tag that never stops

     {v
| A table | With columns   |
|----------+---------------------|
| Yes      | Always             |
| No       | Never               |
v}
*)
```

The indentation of the table is messed up because the markup for the `@`tag includes `text-indent: -3ex` and that is being inherited by the `<pre>` block. A simple solution would be to add `text-indent: initial` to the css for `.odoc pre`.

jonludlam:
closes #1139

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@
 
 - Fix bug causing stack overflow in odoc_driver_monorepo (@jonludlam, #1304)
 - Fix backtrace on invalid input in compile-deps (@jonludlam, #1313)
+- Fix bug in our CSS hitting verbatim blocks in tags (@jonludlam, #1312)
 
 # 3.0.0~beta1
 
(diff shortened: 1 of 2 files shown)


panglesd (commented):
Although this "works", we are going to hit the same bug when we add new constructs (eg admonitions).

I'm not a fan of this negative margin... I would prefer to simply remove it for tags.

(I was surprised why tables are not affected by this, it turns out that `text-indent: initial` is part of the user agent's `table` styling in major browsers. Again, a bit brittle!)

jonludlam:
Right, we can look into fixing this in a more sustainable way, but for now this does fix a problem, so in it goes!

Issue #1134, opened 2024-05-26
More fields in `@doc-json`


dlesbre:
As mentioned in [the discuss.ocaml post](https://discuss.ocaml.org/t/odoc-documentation-improvements/14674/3), I would appreciate a few extra fields in the generated JSON files to make integrating them in my website easier. jonludlam suggested I open an issue for these here as well:
- [ ] A `title` field with the page title (currently obtainable from `breadcrumbs[-1].name`
- [ ] A `package` field with the name of the current package (currently can be obtained from the path of the generated doc)
- [ ] A `version` field with the version of the current package (I'm building versioned docs). Currently I can only obtain this through git ref names.
- [ ] An `objects` fields with the list and type (val, module, type, module type...) of all objects defined in the current file
- [ ] A `sherlodoc` field indicating whether or not sherlodoc was present (and if so, the relative path from the current file to the `db.js` file).

AllanBlanchard:
I don't know if it is entirely related, but this news:

https://tarides.com/blog/2024-02-28-two-major-improvements-in-odoc-introducing-search-engine-integration/

Suggests that the DB exists in Json, would it be possible to get it? The main problem with the `db.js` files is that if you have several packages in your project, the search engine does not merge all packages at top-level.

jonludlam:
So, coming back to this, I think we can address a few of these for the v3.0 release, mostly because a lot of these were tricky prior to the existence of `odoc_driver`.

- Title is easy enough
- Package and Version - odoc itself doesn't know anything about these, but odoc_driver certainly does.
- objects - I'm not sure what this will be used for? And by "current file" - do you mean the HTML file (so not including submodes), or the `odocl` file (ie, top-level module and all submodules).

An alternative to putting all of these fields into the existing json output files, is that we coud output in the json the path to the top-level module. Unless you've got a very unusual layout (e.g. in the `monorepo` driver), this would be the 'package root', where we could put a package metadata file owned by `odoc_driver`. This could contain the sherlodoc db path, the name and version of the package, a reference to the sidebar json file, and potentially the json index with all of the 'contents' of the entire package. Would this work for you?

AllanBlanchard:
> where we could put a package metadata file owned by odoc_driver. 

What would be the advantage with respect to having the very same information in the JSON? Having two files means that we have to parse two different files for extracting information (potentially with different formats?).

> This could contain the sherlodoc db path.

Would the DB still be a JS file?

panglesd:
[...]

PR #1314, opened 2025-02-18
Add header field to the json output

panglesd:
Fix part of #1134.

The ocaml.org part should also be fixed to use the header.


Julow (approved):
Looks good to me :)

jonludlam:
Thanks!

PR #1315, opened 2025-02-19
Fix unclosed open file when counting occurrences

panglesd:
As reported by @lukemaurer!

Unfortunately `In_channel.with_open_bin` in only available in 4.14, we are not there yet...

### DIFF
--- a/src/odoc/occurrences.ml
+++ b/src/odoc/occurrences.ml
@@ -60,7 +60,9 @@ let aggregate files file_list ~warnings_options:_ ~dst =
     let files = files @ new_files in
     let from_file file : Odoc_occurrences.Table.t =
       let ic = open_in_bin (Fs.File.to_string file) in
-      Marshal.from_channel ic
+      let res = Marshal.from_channel ic in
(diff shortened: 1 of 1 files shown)


src/odoc/occurrences.ml
```diff
@@ -60,7 +60,9 @@ let aggregate files file_list ~warnings_options:_ ~dst =
     let files = files @ new_files in
     let from_file file : Odoc_occurrences.Table.t =
       let ic = open_in_bin (Fs.File.to_string file) in
-      Marshal.from_channel ic
+      let res = Marshal.from_channel ic in
+      close_in ic;
+      res
```
Julow, line 65: We have `Fun.protect` in `Odoc_utils` that can help.

Julow (commented):
There are other occurrences (hehe) of this in `Indexing.read_occurrences`, `Occurrences.count` and at the end of `Occurrences.aggregate`. I think we need helper functions for reading and writing marshals safely.

panglesd:
Supersed by #1320

PR #1316, opened 2025-02-19
Add test for delimited code block with meta and output

NoahTheDuke:
As a "getting my feet wet", I've added a single test case that wasn't already covered.


jonludlam:
Thanks @NoahTheDuke !

Issue #706, opened 2021-07-21
Incorrect rejection of `{[ ]}` docstring


jberdine:
Code such as:
```ocaml
type sp =
  | Blank  (** {[ ]} *)
```
leads to a warning such as
```
Warning: Invalid documentation comment:
File "lib/Fmt.mli", line 23, characters 15-20:
'{[...]}' (code block) should not be empty.
```
There isn't any indication in the [spec](https://ocaml.org/releases/4.12/htmlman/ocamldoc.html) that the "preformatted source code style" should not be a single space. And for the warning in particular, being preformatted, a single space is not empty.

dbuenzli:
If you need a solution now you can likely use a non-breaking space (U+00A0) to work around (didn't try, but that usually works most of the time when you hit silly things like that).

Julow:
Odoc's spec is not exactly ocamldoc's one and there's no written spec currently. The warning is from Odoc's parser, which strip the spaces.
It is possible to fix this in Odoc by moving this check further in the pipeline (there's already a place for these checks) or removing it completely.

jberdine:
> you can likely use a non-breaking space (U+00A0) to work around

Thanks @dbuenzli, that is helpful.

jberdine:
> Odoc's spec is not exactly ocamldoc's one and there's no written spec currently.

The situation where odoc consciously disagrees with the OCaml manual, and doesn't have a spec itself, is a continuing issue. Are there any plans in this ?

Drup:
> It is possible to fix this in Odoc by moving this check further in the pipeline (there's already a place for these checks) or removing it completely.

Wait, you closed this a a bit quickly. This issue has a point: I don't really see the purpose of this check and it probably shouldn't be there.

Julow:
> Wait, you closed this a a bit quickly
Indeed ! Sorry about that. I thought I was reading an issue on OCamlformat's repository.

asavahista:
is there a usecase for a codeblock containing a single space, given that it's indistinguishable from a blank line? (`[ ]` would be distinguishable from `[]` though)
maybe jberdine wants an inline preformatted code span rather than a code block, which i don't believe exists currently?

jberdine:
[...]

PR #1317, opened 2025-02-20
Fix "box model" for code blocks

panglesd:
In my mental model, code blocks are supposed to follow a "box model", where the actual content of a code block is the smallest box that contains it all:

image

In the example above, the actual content of the code blocks is in lighter blue. As you can see, some whitespace has been stripped before, after, but also at the beginning of each line.

However, (supposing my mental model is the right one), there are two bugs in the implementation.

---

The first one is about lines with only whitespace, but less than the box:

image

Suppose that `_` are spaces (I used `_` to make it visible). I think those spaces are not part of the box and should be stripped, but currently they _are_ part of the content, which would be:

```
    Here is some content
___
With some other content

  Hey!
```

---

The second one is about a corner case of the "box model": when the code block does not start with a new line.

In this case, it is clear what the content of the code block should be:

image

However, in this case it is not! 

image

Of course, `  {[` should not be part of "the box". So, what do we put instead?

One possibility would be to fill it with spaces:

```
     Here is some content

With some other content
```

Another solution would be to keep the first line as it is, in this degenerated case:

```
 Here is some content

With some other content
```

The currently implemented solution is very confusing in my opinion. For the first line, it removes a number of spaces corresponding to the min between the de-indentation level and the number of opening spaces in the first line. In effect it ignores the offset in which the first line start.

The first solution seems the most consistent with the box model, but also probably the most "breaking change". The second solution seems good as:
```
{[a
b
c]}
```
is turned as:
```
a
b
c
```
and not 
```
  a
b
c
```
as it would in the first solution.

Which solution do you think is best?

### DIFF
--- a/src/parser/lexer.mll
+++ b/src/parser/lexer.mll
@@ -124,7 +124,7 @@ let trim_leading_whitespace : first_line_offset:int -> string -> string =
       (* Since blank lines were ignored when calculating
          [least_amount_of_whitespace], their length might be less than the
          amount. *)
-      if String.length line < n then line
+      if String.length line < n then ""
       else String.sub line n (String.length line - n)
     in
     let lines =
--- a/src/parser/lexer.mll
+++ b/src/parser/lexer.mll
@@ -104,15 +104,15 @@ let trim_leading_whitespace : first_line_offset:int -> string -> string =
       | _ -> least_so_far)
   in
 
-  let first_line_max_drop, least_amount_of_whitespace =
+  let least_amount_of_whitespace =
     match lines with
-    | [] -> 0, None
+    | [] -> None
     | first_line :: tl ->
       begin match count_leading_whitespace first_line with
         | Some n ->
-          n, least_amount_of_whitespace (Some (first_line_offset + n)) tl
+          least_amount_of_whitespace (Some (first_line_offset + n)) tl
         | None ->
-          0, least_amount_of_whitespace None tl
+          least_amount_of_whitespace None tl
       end
   in
 
@@ -131,8 +131,9 @@ let trim_leading_whitespace : first_line_offset:int -> string -> string =
       match lines with
       | [] -> []
       | first_line :: tl ->
-        drop (min first_line_max_drop least_amount_of_whitespace) first_line
-        :: List.map (drop least_amount_of_whitespace) tl
+         let first_line_drop = max 0 (least_amount_of_whitespace - first_line_offset) in
+         drop first_line_drop first_line
+         :: List.map (drop least_amount_of_whitespace) tl
     in
     String.concat "\n" lines
 
--- a/src/parser/lexer.mll
+++ b/src/parser/lexer.mll
@@ -131,9 +131,14 @@ let trim_leading_whitespace : first_line_offset:int -> string -> string =
       match lines with
       | [] -> []
       | first_line :: tl ->
-         let first_line_drop = max 0 (least_amount_of_whitespace - first_line_offset) in
-         drop first_line_drop first_line
-         :: List.map (drop least_amount_of_whitespace) tl
+         let first_line =
+           let spaces_to_remove = least_amount_of_whitespace - first_line_offset in
+           if spaces_to_remove >= 0 then
+             drop spaces_to_remove first_line
+           else
+             (String.make (-spaces_to_remove) ' ') ^ first_line
+         in
+         first_line :: List.map (drop least_amount_of_whitespace) tl
     in
     String.concat "\n" lines
 
--- a/src/parser/test/test.ml
+++ b/src/parser/test/test.ml
@@ -2566,7 +2566,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "foo\
+          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  " foo\
                                                                \nbar")))))
          (warnings ()))
         |}]
@@ -2576,7 +2576,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "foo\r\
+          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  " foo\r\
                                                                \nbar")))))
          (warnings ()))
         |}]
@@ -2595,7 +2595,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "  foo\
+          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "   foo\
                                                                \nbar")))))
          (warnings ()))
         |}]
@@ -2614,7 +2614,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (3 6)) (code_block ((f.ml (1 2) (3 4))  "foo\
+          (((f.ml (1 0) (3 6)) (code_block ((f.ml (1 2) (3 4))  " foo\
                                                                \n\
                                                                \nbar")))))
          (warnings ()))
@@ -2625,7 +2625,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (3 7)) (code_block ((f.ml (1 2) (3 5))  "foo\
+          (((f.ml (1 0) (3 7)) (code_block ((f.ml (1 2) (3 5))  "  foo\
                                                                \n\
                                                                \nbar")))))
          (warnings ()))
@@ -2639,7 +2639,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (2 7) (4 4)) (code_block ((f.ml (2 9) (4 2))  "a\
+          (((f.ml (2 7) (4 4)) (code_block ((f.ml (2 9) (4 2))  " a\
                                                                \nb\
                                                                \nc")))))
          (warnings ()))
@@ -2650,7 +2650,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (3 6)) (code_block ((f.ml (1 2) (3 4))  "foo\
+          (((f.ml (1 0) (3 6)) (code_block ((f.ml (1 2) (3 4))  " foo\
                                                                \n  \
                                                                \nbar")))))
          (warnings ()))
@@ -2677,7 +2677,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "foo\
+          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "\tfoo\
                                                                \nbar")))))
          (warnings ()))
         |}]
@@ -2687,7 +2687,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (2 7)) (code_block ((f.ml (1 2) (2 5))  "foo\
+          (((f.ml (1 0) (2 7)) (code_block ((f.ml (1 2) (2 5))  "\tfoo\
                                                                \nbar")))))
          (warnings ()))
         |}]
@@ -2701,7 +2701,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (2 1) (4 3)) (code_block ((f.ml (2 3) (4 1))  " foo\
+          (((f.ml (2 1) (4 3)) (code_block ((f.ml (2 3) (4 1))  "  foo\
                                                                \nbar")))))
          (warnings ()))
         |}]
--- a/src/parser/test/test.ml
+++ b/src/parser/test/test.ml
@@ -2568,7 +2568,8 @@ let%expect_test _ =
         ((output
           (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "foo\
                                                                \nbar")))))
-         (warnings ())) |}]
+         (warnings ()))
+        |}]
 
     let leading_whitespace_two_cr_lf =
       test "{[ foo\r\n bar]}";
@@ -2577,7 +2578,8 @@ let%expect_test _ =
         ((output
           (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "foo\r\
                                                                \nbar")))))
-         (warnings ())) |}]
+         (warnings ()))
+        |}]
 
     let leading_whitespace_two_different_indent =
       test "{[ foo\n   bar]}";
@@ -2595,7 +2597,8 @@ let%expect_test _ =
         ((output
           (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "  foo\
                                                                \nbar")))))
-         (warnings ())) |}]
+         (warnings ()))
+        |}]
 
     let leading_whitespace_two_different_indent_reloc =
       test "{[ foo\n      bar]}";
@@ -2614,7 +2617,8 @@ let%expect_test _ =
           (((f.ml (1 0) (3 6)) (code_block ((f.ml (1 2) (3 4))  "foo\
                                                                \n\
                                                                \nbar")))))
-         (warnings ())) |}]
+         (warnings ()))
+        |}]
 
     let leading_whitespace_with_whitespace_line_short =
       test "{[  foo\n \n  bar]}";
@@ -2627,6 +2631,20 @@ let%expect_test _ =
          (warnings ()))
         |}]
 
+    let bli =
+      test {|
+       {[ a
+ b
+ c]}|};
+      [%expect
+        {|
+        ((output
+          (((f.ml (2 7) (4 4)) (code_block ((f.ml (2 9) (4 2))  "a\
+                                                               \nb\
+                                                               \nc")))))
+         (warnings ()))
+        |}]
+
     let leading_whitespace_with_whitespace_line_long =
       test "{[ foo\n   \n bar]}";
       [%expect
@@ -2635,7 +2653,8 @@ let%expect_test _ =
           (((f.ml (1 0) (3 6)) (code_block ((f.ml (1 2) (3 4))  "foo\
                                                                \n  \
                                                                \nbar")))))
-         (warnings ())) |}]
+         (warnings ()))
+        |}]
 
     let leading_whitespace_leading_newline =
       test "{[\n  foo\n  bar\n]}";
@@ -2660,7 +2679,8 @@ let%expect_test _ =
         ((output
           (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "foo\
                                                                \nbar")))))
-         (warnings ())) |}]
+         (warnings ()))
+        |}]
 
     let leading_tab_two_different_indent =
       test "{[\tfoo\n\t\tbar]}";
@@ -2669,7 +2689,22 @@ let%expect_test _ =
         ((output
           (((f.ml (1 0) (2 7)) (code_block ((f.ml (1 2) (2 5))  "foo\
                                                                \nbar")))))
-         (warnings ())) |}]
+         (warnings ()))
+        |}]
+
+    let leading_whitespace_when_box_model_not_applicable =
+      test {|
+ {[  foo
+ bar
+ ]}
+|};
+      [%expect
+        {|
+        ((output
+          (((f.ml (2 1) (4 3)) (code_block ((f.ml (2 3) (4 1))  " foo\
+                                                               \nbar")))))
+         (warnings ()))
+        |}]
 
     let leading_newline =
       test "{[\nfoo]}";
--- a/src/parser/test/test.ml
+++ b/src/parser/test/test.ml
@@ -2531,28 +2531,31 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (2 5)) (code_block ((f.ml (1 2) (2 3))  "foo\
+          (((f.ml (1 0) (2 5)) (code_block ((f.ml (1 2) (2 3))  "  foo\
                                                                \nbar")))))
-         (warnings ())) |}]
+         (warnings ()))
+        |}]
 
     let cr_lf =
       test "{[foo\r\nbar]}";
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (2 5)) (code_block ((f.ml (1 2) (2 3))  "foo\r\
+          (((f.ml (1 0) (2 5)) (code_block ((f.ml (1 2) (2 3))  "  foo\r\
                                                                \nbar")))))
-         (warnings ())) |}]
+         (warnings ()))
+        |}]
 
     let blank_line =
       test "{[foo\n\nbar]}";
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (3 5)) (code_block ((f.ml (1 2) (3 3))  "foo\
+          (((f.ml (1 0) (3 5)) (code_block ((f.ml (1 2) (3 3))  "  foo\
                                                                \n\
                                                                \nbar")))))
-         (warnings ())) |}]
+         (warnings ()))
+        |}]
 
     let leading_whitespace =
       test "{[ foo]}";
@@ -2566,7 +2569,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  " foo\
+          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "  foo\
                                                                \nbar")))))
          (warnings ()))
         |}]
@@ -2576,7 +2579,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  " foo\r\
+          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "  foo\r\
                                                                \nbar")))))
          (warnings ()))
         |}]
@@ -2595,7 +2598,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "   foo\
+          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "    foo\
                                                                \nbar")))))
          (warnings ()))
         |}]
@@ -2614,7 +2617,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (3 6)) (code_block ((f.ml (1 2) (3 4))  " foo\
+          (((f.ml (1 0) (3 6)) (code_block ((f.ml (1 2) (3 4))  "  foo\
                                                                \n\
                                                                \nbar")))))
          (warnings ()))
@@ -2639,9 +2642,10 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (2 7) (4 4)) (code_block ((f.ml (2 9) (4 2))  " a\
-                                                               \nb\
-                                                               \nc")))))
+          (((f.ml (2 7) (4 4))
+            (code_block ((f.ml (2 9) (4 2))  "         a\
+                                            \nb\
+                                            \nc")))))
          (warnings ()))
         |}]
 
@@ -2650,7 +2654,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (3 6)) (code_block ((f.ml (1 2) (3 4))  " foo\
+          (((f.ml (1 0) (3 6)) (code_block ((f.ml (1 2) (3 4))  "  foo\
                                                                \n  \
                                                                \nbar")))))
          (warnings ()))
@@ -2677,7 +2681,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  "\tfoo\
+          (((f.ml (1 0) (2 6)) (code_block ((f.ml (1 2) (2 4))  " \tfoo\
                                                                \nbar")))))
          (warnings ()))
         |}]
@@ -2701,7 +2705,7 @@ let%expect_test _ =
       [%expect
         {|
         ((output
-          (((f.ml (2 1) (4 3)) (code_block ((f.ml (2 3) (4 1))  "  foo\
+          (((f.ml (2 1) (4 3)) (code_block ((f.ml (2 3) (4 1))  "    foo\
                                                                \nbar")))))
          (warnings ()))
         |}]
@@ -2928,9 +2932,10 @@ let%expect_test _ =
         {|
         ((output
           (((f.ml (1 0) (2 14))
-            (code_block ((f.ml (1 2) (2 12))  "(* foo *)\
+            (code_block ((f.ml (1 2) (2 12))  "  (* foo *)\
                                              \nlet bar = ()")))))
-         (warnings ())) |}]
+         (warnings ()))
+        |}]
 
     let docstring =
       test "{[(** foo *)\nlet bar = ()]}";
@@ -2938,9 +2943,10 @@ let%expect_test _ =
         {|
         ((output
           (((f.ml (1 0) (2 14))
-            (code_block ((f.ml (1 2) (2 12))  "(** foo *)\
+            (code_block ((f.ml (1 2) (2 12))  "  (** foo *)\
                                              \nlet bar = ()")))))
-         (warnings ())) |}]
+         (warnings ()))
+        |}]
 
     let docstring_with_code_block =
       test "{[(** {[foo]} *)\nlet bar = ()]}";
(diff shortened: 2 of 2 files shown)


panglesd:
I think this PR should serve as a basis for @NoahTheDuke to add the layout information necessary to reconstruct the actual layout from the content. Indeed, without a proper "box model", it will be hard to have a nice layout type.

I think this layout could have the following type: 

```ocaml
type code_block_layout = {
  top : string;
  left : int list;
  bottom : string
}
```
And that it should be possible to reconstruct the content with the following function:
```ocaml
let reconstruct layout content =
  let lines = String.split_on_char '\n' content in
  let lines = List.map2 (fun i l -> (String.make ' ' i ^ l) layout.left lines in
  layout.top ^
  String.concat ~sep:'\n' lines
  ^ layout.bottom
```

dbuenzli:
I think you rather want to mostly follow what [commonmark does](https://spec.commonmark.org/0.31.2/#fenced-code-blocks) here slightly adapted for the fact that the content can be on the line delimiting the block `{[`. That would be: 

1. Strip leading and trailing whitespace from the contents of `{[…]}`.
2. The first non whitespace character in `{[]}` defines the left margin column `m`
3. For each line strip whitespace < `m`. 
4. For each line, indent to column `m`, if there is a non-whitespace character before column `m`

dbuenzli:
> That being said, I'm not sure how close this is to what commonmark does. From what I understand, commonmark only looks at the indentation of the opening fence (as mentioned in [this example](https://spec.commonmark.org/0.31.2/#example-131)). In what you suggest, it's not the fence opening which determines the left margin,

Yes it is different: there is never any content on the opening fence in commonmark so you can use the marker as the indentation level.

> Using the first non-whitespace as left-margin does not allow that

You could escapes for that but I'm no longer sure how escapes work in `ocamldoc` it has always been a bit ill-defined an confusing.

That being said if your goal is [this](https://github.com/ocaml/odoc/pull/303#issuecomment-2667845389) then you would be better off keeping a precise location of the start of content and simply keep the raw string of the content in the node.

dbuenzli:
> keep the raw string of the content in the node.

(And not even, if you have precise start/end locations *and the source* you can simply extract the part)

dbuenzli:
> I would be in favor of keeping the current box model.

Note that the problem with the current box model, is that it's not very clear. To this day I always write code fences fully flushed left to make sure I get what I want… [This](https://github.com/ocaml/odoc/pull/1317#issuecomment-2670899826) allows for a clear mental model in my opinion: namely something clear has to define the indentation level in the beginning of the code fence and then everything is simply left aligned on that indentation level with any whitespace before the indentation level being dropped.

NoahTheDuke:
Ocaml's comments are naturally indented, which makes using Commonmark's style harder. It requires deleting indentation that an editor or formatter might try to add.

jonludlam:
> (And not even, if you have precise start/end locations _and the source_ you can simply extract the part)

Mdx does it this way - by getting the locations (which are indeed precise) and splicing the text [from the original file](https://github.com/realworldocaml/mdx/blob/main/lib/mli_parser.ml#L14-L20).

NoahTheDuke:
To be clear, the only benefit to handling this is for display, right? Ocaml doesn't care about indentation, so Mdx and/or an odoc `extract` command would just extract the raw text and slap it into a file. Having the precise location of the text within the `{[...]}` tags is only meaningful for displaying in docs files or hovers.

NoahTheDuke:
I think the box model is the correct way to think about code blocks (and verbatim blocks, while we're at it).

`ocaml/ocaml` doesn't have a consistent way of indenting code blocks right now.

1\) [stdlib/float.mli](https://github.com/ocaml/ocaml/blob/25265d71b5183e65d1c8e4d748efeb22a58ad0ff/stdlib/float.mli#L794-L802) puts the first line against the braces and then indents the following lines to the column where the braces sit:

```ocaml
  (** ...
      For example, consider the following program:
  {[let size = 100_000_000
  let a = Float.Array.make size 1.
  let update a f () =
     Float.Array.iteri (fun i x -> Float.Array.set a i (f x)) a
  let d1 = Domain.spawn (update a (fun x -> x +. 1.))
  let d2 = Domain.spawn (update a (fun x ->  2. *. x +. 1.))
  let () = Domain.join d1; Domain.join d2
  ]}
  ... *)
```

2\) [stdlib/condition.mli](https://github.com/ocaml/ocaml/blob/25265d71b5183e65d1c8e4d748efeb22a58ad0ff/stdlib/condition.mli#L52-L59) has a line break and then indents to the column one past the square brace:

```ocaml
(** ...
   {[
     Mutex.lock m;
     while not P do
       Mutex.unlock m; Mutex.lock m
     done;
     <update the data structure>;
     Mutex.unlock m
   ]}
... *)
```

3\) [stdlib/bytes.mli](https://github.com/ocaml/ocaml/blob/25265d71b5183e65d1c8e4d748efeb22a58ad0ff/stdlib/bytes.mli#L368-L373) has a line break and no indentation:

```ocaml
(** ...
   {[
let string_init len f : string =
  let s = Bytes.create len in
  for i = 0 to len - 1 do Bytes.set s i (f i) done;
  Bytes.unsafe_to_string s
   ]}
... *)
```

4\) Another part of [stdlib/bytes.mli](https://github.com/ocaml/ocaml/blob/25265d71b5183e65d1c8e4d748efeb22a58ad0ff/stdlib/bytes.mli#L755-L763) dedents the braces, has no spaces or line break, and no indentation of the code block:

```ocaml
(** ...
    For example, consider the following program:
{[let size = 100_000_000
let b = Bytes.make size  ' '
let update b f ()  =
  Bytes.iteri (fun i x -> Bytes.set b i (Char.chr (f (Char.code x)))) b
let d1 = Domain.spawn (update b (fun x -> x + 1))
let d2 = Domain.spawn (update b (fun x -> 2 * x + 1))
let () = Domain.join d1; Domain.join d2
]}
... *)
```

5\) [stdlib/arg.mli](https://github.com/ocaml/ocaml/blob/25265d71b5183e65d1c8e4d748efeb22a58ad0ff/stdlib/arg.mli#L21-L37) dedents the braces, and indents the code block two spaces past where the text above sits (an awkward 5 spaces):

```ocaml
(** ...
   This module provides a general mechanism for extracting options and
   arguments from the command line to the program. For example:

{[
     let usage_msg = "append [-verbose] <file1> [<file2>] ... -o <output>"
     let verbose = ref false
     let input_files = ref []
     let output_file = ref ""
... *)
```

6\) [stdlib/nativeint.mli](https://github.com/ocaml/ocaml/blob/25265d71b5183e65d1c8e4d748efeb22a58ad0ff/stdlib/nativeint.mli#L33-L37) has a line break and then indents the code block to align with the square brace instead of 1 past:

```ocaml
(** ...
    {[
     let zero: nativeint = 0n
     let one: nativeint = 1n
     let m_one: nativeint = -1n
    ]}
... *)
```

7\) [stdlib/queue.mli](https://github.com/ocaml/ocaml/blob/25265d71b5183e65d1c8e4d748efeb22a58ad0ff/stdlib/queue.mli#L129-L131) has a line break and then indents to align with the curly brace:

```ocaml
(** ...
    {[
    # let q = Queue.create ()
    val q : '_weak1 Queue.t = <abstr>
... *)
```

8\) Most cursed of all, [stdlib/either.mli](https://github.com/ocaml/ocaml/blob/25265d71b5183e65d1c8e4d748efeb22a58ad0ff/stdlib/either.mli#L28-L29) dedents the braces, has no linebreak and then indents past the code of the initial line as if it was not on the same line as the braces:

```ocaml
(** ...
    For example:

{[List.partition_map:
    ('a -> ('b, 'c) Either.t) -> 'a list -> 'b list * 'c list]}
... *)
```

(This is the only instance of this style in `ocaml/ocaml`, which means we can safely ignore it (or even change it directly to align with other styles).)

All of the code blocks except for 8) are consistent with the below algorithm.

1) If there is any text on the line containing the opening braces, ignore it for now.
2) Find the least indented line in the remaining lines, store the indent as the left edge of the box.
3) Remove the left edge indentation of each remaining line.
4) If there is any text on the line containing the opening braces, trim the whitespace and insert it at the front.

jonludlam:
Just thinking about the implementation, I was wondering about maybe having the raw unaltered text in the AST, then a separate function that applies the formatting rules being suggested in this thread. If we did it that way we'd avoid the need for the original source as currently required by mdx.

NoahTheDuke:
I took a shot at implementing my algorithm above which led to #1318. Can be helpful to compare the two sets of outputs.

dbuenzli:
> Just thinking about the implementation, I was wondering about maybe having the raw unaltered text in the AST, then a separate function that applies the formatting rules being suggested in this thread.

That looks ok to me.

> Ocaml's comments are naturally indented, which makes using Commonmark's style harder. It requires deleting indentation that an editor or formatter might try to add.

Nothing different here, CommonMark also does that. The only difference is that CommonMark's delimiters stand on their own line (and is thus easier to use as the indentation mark).

> To be clear, the only benefit to handling this is for display, right?

No, you need precise locations and layout for compilation errors of the extracted source code to be reported at the right place in the original. You need a [line directive](https://ocaml.org/manual/5.1/lex.html#sss:lex-linedir) followed by the exact, verbatim, code for compilation errors to show up exactly at the right place in the source.

> All of the code blocks except for 9) are consistent with the below algorithm.

Unless I'm mistaken that is the case for my proposal too. Personally I don't see any benefit of using the line with the smallest indent as you suggest (rather than the first line):

1. It makes it harder to spot the indentation point on larger code samples.
2. It more brittle to careless c&p and/or additions. 

Besides the same problem as @panglesd mentioned remains: how do you solve the problem of enforcing an initial indent ? For the latter I have a few ideas but I don't really like them. One would be to simply use the non newline whitespace directly following `{[` to define the indent if the line is blank (made of only whitespace). It is obscure but then we can say that it's likely only used for 0.01% of the cases.

NoahTheDuke:
The reason you use the lowest indent is that in example 9), if it had multiple lines, then you'd account for additional indentation on the second line:

```ocaml
(** ...
    For example:

{[List.partition_map:
    ('a -> ('b, 'c) Either.t) -> 'a list -> 'b list * 'c list

  List.partition_map:
    ('a -> ('b, 'c) Either.t) -> 'a list -> 'b list * 'c list]}
... *)
```
renders as
```ocaml
       [%expect {|
         ((output
           (((f.ml (2 4) (2 16))
             (paragraph
              (((f.ml (2 4) (2 7)) (word For)) ((f.ml (2 7) (2 8)) space)
               ((f.ml (2 8) (2 16)) (word example:)))))
            ((f.ml (4 0) (8 63))
             (code_block
              ((f.ml (4 2) (8 61))
                "List.partition_map:\
               \n  ('a -> ('b, 'c) Either.t) -> 'a list -> 'b list * 'c list\
               \n\
               \nList.partition_map:\
               \n  ('a -> ('b, 'c) Either.t) -> 'a list -> 'b list * 'c list")))))
          (warnings ())) |}]
```

In what scenarios do you want to enforce an initial indent in a code block?

dbuenzli:
> The reason you use the lowest indent

From a usability point of view you turn the problem of understanding how stuff is going to be indented into checking `n` (i.e. every) lines rather than `1`. This becomes particularly acute when your code blocks spans more than one screenful. So I really rather not have that solution.

> In what scenarios do you want to enforce an initial indent in a code block?

Who knows ? There's always the odd case out there and it's enraging the day you need it. But a simple example is ASCII art.

```
{[
     +---+---+---
tape | A | A | …
     +---+---+--- 
]}
```

NoahTheDuke:
Your example would be correctly handled by finding the lowest indent (`tape`), and incorrectly handled by using the first line's indent (`     +`).

Using lowest indent:

```ocaml
     let stdlib_either =
       test "
{[
     +---+---+---
tape | A | A | …
     +---+---+--- 
]}
       ";
       [%expect {|
         ((output
           (((f.ml (2 0) (6 2))
             (code_block
              ((f.ml (2 2) (6 0))
                "     +---+---+---\
               \ntape | A | A | \226\128\166\
               \n     +---+---+--- ")))))
          (warnings ())) |}]
```

Using first line indent:

```ocaml
     let stdlib_either =
       test "
{[
     +---+---+---
tape | A | A | …
     +---+---+--- 
]}
       ";
       [%expect {|
         ((output
           (((f.ml (2 0) (6 2))
             (code_block
              ((f.ml (2 2) (6 0))
                "+---+---+---\
               \n| A | A | \226\128\166\
               \n+---+---+--- ")))))
          (warnings ())) |}]
```

dbuenzli:
Right, now suppose you want to write: 

```
Here is the box:

{[
+---+
| A |
+---+
]}

After shifting one position to the right you get this configuration:

{[
    +---+
    | A |
    +---+
]}
```

dbuenzli:
> We could also use the fence to determine left-margin. But, in
> 
> ```
> {[
>     bli
> ]}
> ```
> the content would become indented.

I think in the end that this is actually the more natural and less confusing overall (and allows to specify any initial indent you wish without fuss). There is the question of what happens when you have:

```
  {[bli
  …
  ]}
```
but you can simply say that this is equivalent to: 
```
  {[
  bli
  …
  ]}
```

So basically the first column of `{[` defines the left margin. Anything that is on the left of that margin is automatically indented to the margin. The rule is simple to understand and allows for any hand-made ASCII layout you wish.

NoahTheDuke:
That's pretty clever. Thanks for thinking it through with us.

panglesd:
Thanks for all the ideas and discussion. Although this PR is "just" a fix for an already established syntax, and not an attempt to change it, it's good to have feedback on it, and we can maybe change it. I'll try to sum up the choice we have, but first:

> Just thinking about the implementation, I was wondering about maybe having the raw unaltered text in the AST, then a separate function that applies the formatting rules being suggested in this thread. If we did it that way we'd avoid the need for the original source as currently required by mdx.

Mdx currently _changes_ the code blocks. But you want it to change them while preserving the layout. For instance:
```
{[
              # let x = 1 ;;
]}
{[
# let x = 1 ;;
]}
```
should be rewritten to 
```
{[
              # let x = 1 ;;
              val x : int
]}
{[
# let x = 1 ;;
val x : int
]}
```
so that eg it does not fight with ocamlformat which is responsible for the layout.

That is why we might need a bit more than the raw string, but a standardized way of going from/to raw content/deindented content.

---

I'll try to sum up the pros and cons of each possibility, at this point:

Here are the nice properties we would like to have:
1. It is easy for the user to "intuitively" know what will be the content of the code block from the raw string
2. It does not break too many existing documentation
3. It allows to have an indentation in the first line
4. Be fit for long code blocks:
    > Be easy to spot the indentation point on larger code samples.
    > Not be brittle to careless c&p and/or additions.

We have several candidates for that:
A. Using the smallest indent.
  1. I think that (with this PR fixing the bugs) it's as intuitive as the others
  2. It's just fixing bugs so should not break any documentation
  3. Does allow to have an indentation as the first line
  4. Not so fit for long code blocks as the indentation depend on it all

B. Using the first indent
  1. I think it's as intuitive as the others. Eg I think @NoahTheDuke got it wrong [here](https://github.com/ocaml/odoc/pull/1317#issuecomment-2675518972): `tape` should be part of the output  which would break the indentation
  2. Breaking impact has to be measured but not that much code starts with an indentation
  3. Does not allow to have an indentation as the first line which is a problem for ascii art-based languages ^^
  4. Fit for long code blocks

C.  Using the opener's indent
  1. I think it's as intuitive as the others.
  2. Breaking impact has to be measured. At least ocamlformat's default formatting makes it non-breaking.
  3. Allow to have an indentation as the first line
  4. Fit for long code blocks

So I'm fine with A. or C. as long as we find out it's not going to break to many code blocks!

I'll try to investigate C's breaking.

dbuenzli:
Btw. I'm just curious are verbatim blocks treated differently from code blocks ? Because with the odoc3. I'm currently forced to write e.g.:

```
    {- [index_properties], the index properties. 0 if not indexed.
{v
7                               0
+---+---+---+---+---+---+---+---+
|   |   |   |   | L | S | T | F |
+---+---+---+---+---+---+---+---+
v}
    }
```

To make sure the verbatim block is not indented. If I write it: 

```
    {- [index_properties], the index properties. 0 if not indexed.
       {v
       7                               0
       +---+---+---+---+---+---+---+---+
       |   |   |   |   | L | S | T | F |
       +---+---+---+---+---+---+---+---+
       v}
    }
```
the block is indented:
Screenshot 2025-02-24 at 12 32 07

dbuenzli:
> So I'm fine with A. or C. as long as we find out it's not going to break to many code blocks!

Note that only C allows you to write [this example](https://github.com/ocaml/odoc/pull/1317#issuecomment-2675548883). The problem is not about being able to indent on the *first line*, it's to be able to indent the whole content as you wish in the code block.

jonludlam:
> That is why we might need a bit more than the raw string, but a standardized way of going from/to raw content/deindented content.

Yes, that's what I'm asking for. 

As an example where we might want the raw text rather than the processed stuff, we could do a better job of error messages than mdx does today by inserting line directives, so that the line numbers matched the source file correctly. However, I don't think there's such a thing as 'column directives', so to get the column numbers to match the source file correctly you could insert each line in exactly from the source, including the indentation.

panglesd:
Some more though about C:
If we use the first character (`{`) of the opening, we get that ocamlformat automatically indent code blocks, that is lots of code blocks are affected by this change:
```ocaml
(** I'm formatted by ocamlformat

    {[
      hello ocamlformat
    ]} *)
```
renders as `  hello ocamlformat`

If we use the last character (`[`), then many things will get de-indented:
```ocaml
(** I'm formatted by ocamlformat

    {@ocaml[
      let x =
        match w with
        | AAAAAAAAAAAAAAAAAAAAAAAA -> 1
        | BBBBBBBBBBBBBBBBBBBBBBBBB -> 2
    ]} *)
```
will render as 
```
let x =
match w with
| AAAAAAAAAAAAAAAAAAAAAAAA -> 1
| BBBBBBBBBBBBBBBBBBBBBBBBB -> 2
```

dbuenzli:
> so to get the column numbers to match the source file correctly you could insert each line in exactly from the source, including the indentation.

Yes that's what you need if you want to use line directives.

> we get that ocamlformat automatically indent code blocks

Well this could be changed in `ocamlformat`. 

In any case I don't think it's a good idea to indent after `{[`, you are usually already quite nested and you still want to preserve horizontal space to write your code blocks.

panglesd:
> Well this could be changed in ocamlformat.

Yes, what I'm saying is more: there are currently many projects using ocamlformat, with automatically indented code blocks. Changing the indentation to be based on the opening `{` will "break" all of them. (with "breaking" meaning adding two spaces in front)

panglesd:
Out of the example given by @NoahTheDuke at https://github.com/ocaml/odoc/pull/1317#issuecomment-2672010251, I think that examples 2, 3, 5 and 6 will get their rendering wrong:
- 2 5 and 6 will be indented too much
- 3 will be deindented

dbuenzli:
Note the whole code block semantics has already been through a few changes and it's perhaps a good idea to review these discussions to what got us to what we have now. 

https://github.com/ocaml/odoc/issues/39
https://github.com/ocaml/odoc/pull/134
https://github.com/ocaml/odoc/issues/133
https://github.com/ocaml/odoc/issues/132

Personally I don't think the right decision was taken at the time (it seems they did not realize that you could not enforce arbitrary layouts) but perhaps it's not worth changing *again*. However I have to say, as far as I'm concerned, the only reliable way of working with code block and verbatims in `ocamldoc` has remained the "ugly workaround" mentioned [here](https://github.com/ocaml/odoc/issues/133#issue-315565731), that is always write all that fully flushed left; so the status quo may not be optimal either.

dbuenzli:
Also if things are changed again to C, then I don't think the unwanted renderings would be catastrophic and for those `ocamlformat` users a release could be coordinated (and even be done before `odoc` changes).

dbuenzli:
Thinking more about it, arbitrary layout is important because of another use case: literate programming. You may want to interrupt a large definition, e.g. a module, by arbitrary documentation content (text, sound, images) and be able to continue the definition without having your indentation resetting (see for example [this](https://erratique.ch/log/2015-04-05/) blog post).

Therefore I think it would be nice for `odoc3` to simply switch both code blocks *and verbatim blocks*[^1] to default to behaviour C. 

If people really think we should provide a way to preserve the current chop minimal indent semantics perhaps a metadata attribute could be provided (`{ legacy-indent=true[`). However note that examples [like this](https://github.com/ocaml/odoc/pull/1317#issuecomment-2678359264) are actually easy to detect (non-blank content before the column of the marker `{[`) and for users to fix (simply move the `{[` marker at the right place). So a good trade-off could be to tackle these cases with a new warning. Then the only remaining problems would be overly indented code (first example [here](https://github.com/ocaml/odoc/pull/1317#issuecomment-2678259684)), but I'd say that those are not a rendering catastrophy.

[^1]: According to what I got from the discussions I linked to, verbatim's behaviour was never changed which explains [this](https://github.com/ocaml/odoc/pull/1317#issuecomment-2678153203)

dbuenzli:
D looks like an excellent proposal to me.

NoahTheDuke:
Agreed, that's a great compromise.

jonludlam:
This one definitely deserves a changelog entry :-)

panglesd:
This PR is ready for review:
- Implements what has been decided here, with the slight changes that single line code blocks (`{[ blibli ]}`) are trimmed and accepted, as discussed with @jonludlam.
- Raises warnings both on "too few indentation" and "multiline code blocks not starting with a new line".
- **Verbatims are treated the same way as code blocks**. It was previously not specified how they are treated in the documentation, and it seems good to have them as code blocks. From what I can understand, everyone agrees with that in the discussion, but there was no discussion focused on that, so let me know if I'm wrong.
- Tests are updated.
- Docs are updated.

dbuenzli:
> * that single line code blocks (`{[ blibli ]}`) are trimmed and accepted,

Is it really worth making an exception here ? It makes the construct more complex to understand and perhaps that some explanations can be written more compactly that way.

For example:
```
In fact the we can simply replace the line 
  {[  let x = List.iter f v  ]}
of the above function by 
  {[  let y = Array.iter f v ]}
and get the result we wish.
```

> * and "multiline code blocks not starting with a new line".

Are you sure about that one ? In the past `odoc` started to warn on a few things that turned out to be rather annoying, I would rather be conservative and not do that.

panglesd:
> It makes the construct more complex to understand and perhaps that some explanations can be written more compactly that way.

I'm a bit confused! The first part of the sentence seems to say it's not worth it, and the second part seem to say that it is worth it :face_with_diagonal_mouth: 
Treating single line code blocks as "exceptional" have three effects:
- It allows to trim at the end of the string (in `{[ hello ]}` we probably don't want the ending space. Currently code blocks trim trailing lines but not trailing space on the last line, I kept that behavior).
- We don't raise a warning.
- We explain it a bit differently. I could not find a way to explain it that was easy to understand both for one-liners and multilines, without explaining those two cases separately. I think it is fine if "multiline content not starting by a new line" is harder to understand.

So, if we trim trailing spaces in code blocks, which maybe we should do, there is no distinction in how multiline/single lines content is handled. But I would still prefer to explain how they are handled "case by case".

> In the past odoc started to warn on a few things that turned out to be rather annoying, I would rather be conservative and not do that.

I ~~think~~ ~~hope~~ maybe that was because odoc was warning about dependencies problems, and that with the new flags which allow to filter warnings the situation will be better, which will allow to fix the warnings as they come.

dbuenzli:
> I'm a bit confused! The first part of the sentence seems to say it's not worth it, and the second part seem to say that it is worth it 🫤

No I think we misunderstood each other 

You said: 

> * Implements what has been decided here, with the slight changes that single line code blocks (`{[ blibli ]}`) are trimmed and accepted, as discussed with @jonludlam.

To me this means `String.trim` so also trimming the initial space which is what I would object (I care a bit less about the trailing thing if you want to make an exception here). If you keep the initial white that's fine with me. 

But then I just find that simply saying that:

```
{|… 
```
is strictly equivalent to 
```
{|
…
```
and then explain how:
```
{|
…
|}
``` 
works is the easiest. You no longer need a notion of single or multi line code block. There are just code blocks.


> I ~think~ ~hope~ maybe that was because odoc was warning about dependencies problems, and that with the new flags which allow to filter warnings the situation will be better, which will allow to fix the warnings as they come.

Not at all. It was warnings about the way I wrote things in `odoc` that suddenly `odoc` had opinions about. It was quite annoying.

panglesd:
> No I think we misunderstood each other

Indeed!

What I had understood from [this comment](https://github.com/ocaml/odoc/pull/1317#issuecomment-2675578021), was that:
```
  {[   text after space
  ...
  ]}
```
was considered as:
```
  {[
  text after space
  ...
  ]}
```
but in fact, I now think you meant that it is turned into:
```
  {[
     text after space
  ...
  ]}
```

So in what I understood version, the beginning was trimmed anyway, which explains why I was confused why you were dissatisfied with trimming the beginning of single line content.

Thinking about it, not trimming makes sense. I'll add a commit to this PR with this behavior.

panglesd:
About the warning, you [wrote](https://github.com/ocaml/odoc/pull/1317#issuecomment-2678643072):
> However note that examples https://github.com/ocaml/odoc/pull/1317#issuecomment-2678359264 are actually easy to detect (non-blank content before the column of the marker {[) and for users to fix (simply move the {[ marker at the right place). So a good trade-off could be to tackle these cases with a new warning.

So in some case, it appears OK to raise a warning if odoc has a wrong opinion on the layout of code blocks! That being said, I'm ok to remove this warning.

dbuenzli:
> So in some case, it appears OK to raise a warning if odoc has a wrong opinion on the layout of code blocks!

No that's a different warning. It's a warning to help people getting rid of a semantics of code blocks we'd like to get rid of.

I'm more talking about warnings/errors like [this one](https://github.com/ocaml/odoc/issues/706) which, while being well-meaning, end up being unhelpful for all sorts of reasons (there were other occurences but I fail to find them). 

Note that if you by:

> Thinking about it, not trimming makes sense.

you meant no initial and trailing trimming that issue may actually be solved by this PR.

panglesd:
In the last changes, I removed the trimming for more than a whitespace only line (as per [this comment](https://github.com/ocaml-doc/octavius/pull/6#issuecomment-336184811), which praises a behavior that was lost...)
I updated the docs.
I removed the warning on "multiline code blocks that do not start with a newline" [^1]

So now, the semantics of code block content is reasonably easy to explain, and it is possible to have code blocks with arbitrary whitespace in it.

I hope the PR is in a final state in term of design!

[^1]: although, I would classify that as a kind of comments we want to get rid of: either we lose the alignment between the first and the second line, as the first line is shifted by 2 (`{[`) or more (`{delim@ocaml[`) to the left; or we lose the ability to start the first line at the beginning of the line.

jonludlam:
I'd still like to have the raw untrimmed code block in the AST to avoid the need to keep a copy of the original source around for MDX-like uses. Or perhaps we should have that in another PR?

doc/odoc_for_authors.mld
```diff
@@ -375,14 +375,95 @@ OCaml code blocks can be written using the enclosing tags [ {[ ... ]} ].
...
+  The following two code blocks have the same content:
+  {[let x = 0]}
+  {[
+  let x = 0]}
+*)
+]delim}
+
+Also note that if there is non-whitespace on the left of the box above, the
+content is deindented only as much as it can without stripping non-whitespace,
+and a warning is raised.
```
dbuenzli, line 422: I think this bit would be in a better place as a point in the `ol` above.

doc/odoc_for_authors.mld
```diff
@@ -375,14 +375,95 @@ OCaml code blocks can be written using the enclosing tags [ {[ ... ]} ].
...
 v}
 
+In other words, [odoc]:
+{ol
+{- Removes the first line if it contains only whitespace,}
+{- Removes the last line if it contains only whitespace,}
+{- Deindents as much as the opening curly brace's column.}}
+
+Note that if there is non-whitespace content in the first new line, it is
+considered starting at the upper [+] in the box above:
```
dbuenzli, line 409: The `it` is a bit ambiguous. It's not clear if you talk about the line or the non-whitespace content.

doc/odoc_for_authors.mld
```diff
@@ -375,14 +375,95 @@ OCaml code blocks can be written using the enclosing tags [ {[ ... ]} ].
...
+  |
+  +-----------------------
+  ]}
 v}
 
+In other words, [odoc]:
+{ol
+{- Removes the first line if it contains only whitespace,}
+{- Removes the last line if it contains only whitespace,}
+{- Deindents as much as the opening curly brace's column.}}
```
dbuenzli, line 406: (Just checking) So if I understand well if you just want one space in the code block like in issue https://github.com/ocaml/odoc/issues/706 you can get one. However not as mentioned there but by writing for example (with ␠ standing for U+0020):

```
␠{[
␠␠
␠]}
```
the `{[␠]}` mentioned there would result in an empty code block. That's ok to me:
1. The rules are simple.
2. They still allow to write any kind of textual content with the whitespace one wishes.
3. They allow to have the code block arbitrarily indented in the source.

panglesd: Yes, exactly.

However, I realized from your comments that it was not so clear, as the explanation says:
> - Removes the first line if it contains only whitespace,

But that is supposed to include the newline character at the end, so what do we do when there is no newline character?
Simply by precising this (the fact that newline chars are included in the removal and that the removal do not happen if there is none, I can make `{[␠]}` represent a "one space code block".

I implemented that and updated the doc. Hopefully, your 3 points above still hold, the specification is more precise and the result of `{[ ]}` is more what we expect.

(Here is the doc [rendered](https://choum.net/panglesd/odoc_cb/cb/odoc_for_authors.html#code_blocks_example))

panglesd:
> I'd still like to have the raw untrimmed code block in the AST to avoid the need to keep a copy of the original source around for MDX-like uses. Or perhaps we should have that in another PR?

I started doing that in another branch!

doc/odoc_for_authors.mld
dbuenzli: Sounds good as far as the semantics goes. 😅 

I'm not sure we have the best explanation for users yet though. Reading the render I have the impression the ASCII art can mislead a bit about what is the first and last line. 

The following explanation mostly avoids talking about lines.

> In other words, odoc, in order:
>
> 1. Does not remove any content if `{[` and `]}` are on the same line.
> 2. Removes the content after `{[` if it contains only whitespace.
> 3. Removes the content before `]}` if it contains only whitespace.
> 4. Removes spaces and tabs (counted as a single spaces) to reach the column of the opening curly brace of the `{[` marker. If there is non-whitespace on the left of this column, the content is deindented only as much as it can without stripping non-whitespace, and a warning is raised.

I think that with that things are clear enough to remove this sentence:

> Note that removal of lines are meant "including the newline character" (so, nothing is removed if there is no newline character: {[ content ]} stays the same, but so do {[ ]}).

dbuenzli: I updated the description it was a bit ambiguous, you do need to talk about lines but a bit differently.

src/parser/lexer.mll
```diff
@@ -223,39 +124,118 @@ let trim_leading_space_or_accept_whitespace input start_offset text =
...
+    section on code blocks and indentation.  *)
+let code_block_content input ~what ~start_offset s =
+  let start_location = input.offset_to_location start_offset in
+  let indent = start_location.column in
+  let rec handle_first_newline index =
+    if index >= String.length s then String.make indent ' ' ^ s
+    else
+      match s.[index] with
+      | ' ' | '\t' | '\r' -> handle_first_newline (index + 1)
+      (* Multiline starting with an empty line *)
```
jonludlam, line 194: These comments are a little confusing, and could do with rewording.

jonludlam (approved):
Looks good, just some comments to change!

jonludlam:
Let's change the comments in your other PR based on this one, to avoid the rebase.

mbarbin:
Hello, as part of upgrading to ocamlformat `0.29.0` I witnessed the change in formatting:
from:
```ocaml
(** Hello code in odoc comments.

    {[
      let f () = ()
    ]} *)
```
to:
```ocaml
(** Hello code in odoc comments.

    {[
    let f () = ()
    ]} *)
```

It looks slightly surprising to see many such changes in the code base, but this could be the new format I will be getting used to. Note however that `ocp-indent` currently on my machine is still on the previous format, which creates a slight conflict during edition. Not sure if there is a release for that yet.

I do not have built sufficient context to really understand all the corner cases, simply reporting the end UX across the latest ocamlformat feels non-intuitive.

I'll note a strategy used in another block extracting app I have encountered previously this may not work for odoc but just in case: The first line of the block is ignored for the purpose of figuring the indentation when it is on the same line as the opening characters (e.g. `{[`). For the rest of the lines, we take the left-most indentation, and call it indentation zero, all the rest of the lines are left-trimmed up to that zero-indent. If the first line of the block was on the `{[` then it is completely left-trimmed for all white space, and then emitted at indentation zero. Under that strategy, for example, both syntax from the examples above would yield an identical result.

I have read a bit through the previous comments but I don't know if this particular strategy would cover odoc's use cases. That sounds like a problem with many non-intuitive counter examples that makes it non-easy to get right. Thank you for the work on that!

panglesd:
Hello @mbarbin !

I think the disadvantage of the solution you describe is that you cannot write indented code. Something like that is therefore impossible:

> ```ocaml
> let f () =
> ```
> Above is the name and arguments of the function
> ```ocaml
>   print_endline "Hello";
>   print_endline "Hello";
> ```
> Above is the body of the function

Note that the body is indented. If you join both code blocks, you get a nicely indented function definition.

Thanks a lot for raising the issue with `ocp-indent`! I'll open an issue there.

mbarbin:
Hello @panglesd !

I am not sure if I understand the use case of indented code correctly. Is the "source" split on two blocks? I wonder how this works with ocamlformat, or ocp-indent for that matter. Aren't the auto formatting tools going to erase the intended indentation on the second block?

Julow:
That's right, the indentation will be erased by OCamlformat. That's on purpose because OCamlformat used to always add this indentation so now every code blocks must be fixed everywhere.

Maybe we could add a flag to code blocks asking OCamlformat not to reformat it ? Something like `{@ocaml noformat[ ... ]}`.

PR #1320, opened 2025-02-25
Safer file handling

Julow:
On top of https://github.com/ocaml/odoc/pull/1315

Remove many misuses of input and output channels.

### DIFF
--- a/src/utils/odoc_utils.ml
+++ b/src/utils/odoc_utils.ml
@@ -80,17 +80,37 @@ module Forest = Tree.Forest
 module Json = Json
 
 module Io_utils = struct
-  let marshal filename v =
-    let oc = open_out_bin filename in
-    Fun.protect
-      ~finally:(fun () -> close_out oc)
-      (fun () -> Marshal.to_channel oc v [])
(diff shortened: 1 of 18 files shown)


src/utils/odoc_utils.ml
```diff
@@ -79,4 +79,44 @@ module Tree = Tree
 module Forest = Tree.Forest
 module Json = Json
 
+module Io_utils = struct
+  let _with_resource res ~close f =
```
panglesd, line 83: Is this starting with an underscore to indicate it's not for use outside of the module?

panglesd (approved):
Thanks! Much better!

src/utils/odoc_utils.ml
Julow: Exactly. This avoids writing a signature to abstract this one function while the others are trivial.

Julow: I just added some documentation to the module to make it easier to read without an interface.

jonludlam:
Thanks @Julow !

PR #1321, opened 2025-02-26
Remove unneeded code and unify "utils" modules

Julow:
This removes things that are no longer needed since the lower bounds got changed to 4.08: Shims for `Option`, `Fun`, etc.. and the `result` dependency.

I then unify the various "utils" modules into `Odoc_utils` to make the code more consistent and remove more unneeded code.


Julow:
I think this can be merged now ?

PR #1322, opened 2025-02-27
Update odoc API documentation links

azzsal:
I am a newcomer to odoc. I have noticed that the API documentation links are broken. I updated them accordingly.
I think the text descriptions around the links might need some change as well, but I don't have the enough experience with odoc to carry that.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,7 +86,7 @@ The tests cover esy and opam builds on Windows, macOS, and Linux. The Linux test
 
 ### Loading
 
-The [`odoc.loader`](https://ocaml.github.io/odoc/odoc/odoc.loader/) library is responsible for converting from the OCaml [`Typedtree`](https://ocaml.org/p/esperanto/0.0.4/doc/Typedtree/index.html) representations to the [internal representation](https://ocaml.github.io/odoc/odoc/odoc.model/Odoc_model/Lang/index.html).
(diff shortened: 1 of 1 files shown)


jonludlam:
Ah, thanks @azzsal ! yes, it does indeed look like those links are incorrect.

CONTRIBUTING.md
```diff
@@ -86,30 +86,30 @@ The tests cover esy and opam builds on Windows, macOS, and Linux. The Linux test
 
 ### Loading
 
-The [`odoc.loader`](https://ocaml.github.io/odoc/odoc_loader/index.html) library is responsible for converting from the OCaml [`Typedtree`](https://ocaml.github.io/odoc/deps/stdlib/Typedtree/index.html) representations to the [internal representation](https://ocaml.github.io/odoc/odoc_model/Odoc_model/Lang/index.html).
+The [`odoc.loader`](https://ocaml.github.io/odoc/odoc/odoc.loader/) library is responsible for converting from the OCaml [`Typedtree`](https://ocaml.org/p/esperanto/0.0.4/doc/Typedtree/index.html) representations to the [internal representation](https://ocaml.github.io/odoc/odoc/odoc.model/Odoc_model/Lang/index.html).
```
jonludlam, line 89: Linking to `esperanto` doesn't look right :-)

azzsal: Yeah sorry. I linked to that because ocaml seems to fail to build [here](https://ocaml.org/p/ocaml/latest/doc/index.html) for some reason.
I think I can link to the on [github](https://github.com/ocaml/ocaml/blob/trunk/typing/typedtree.mli) if that's better.

jonludlam: This is the equivalent of what was there before: https://ocaml.github.io/odoc/ocaml-base-compiler/compiler-libs.common/Typedtree/index.html

PR #1323, opened 2025-03-03
Driver compile warnings

jonludlam:
We currently only output warnings from the Link phase. This PR also outputs warnings from the compile phase, including those that come from the parser. 

This PR also fixes some warnings that this change made apparent.


panglesd (approved):
Looks good to me!

PR #1325, opened 2025-03-04
Store raw content of code blocks in parser's AST

panglesd:
This PR is based on #1317. Only the last commit is new.

It makes the parser AST store raw content for code blocks and verbatim.
It exposes in the parser functions to get the processed content of those blocks.

The consumer are then responsible for processing the content (eg, odoc) or using the raw content (eg a tool building a file with line directives from specific code blocks).

### DIFF
--- a/src/parser/odoc_parser.mli
+++ b/src/parser/odoc_parser.mli
@@ -45,4 +45,27 @@ val position_of_point : t -> Loc.point -> Lexing.position
     that call in addition to the {!Loc.point} being converted. *)
 
 val codeblock_content : Loc.span -> string -> string * Warning.t list
+(** Process the content of a code block, following the rules described
+    {{!/odoc/odoc_for_authors.indentation_code_blocks}here}. To achieve this, it
+    needs the location of the code block (including the separators) and the raw
+    content of the code block. For instance, with the following code block:
+
+    {delim@ocaml[
+      {[
+        hello
+      ]}
+    ]delim}
+
+    We can go from the raw content ["\n    hello\n  "] to the processed content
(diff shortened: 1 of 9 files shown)


src/parser/odoc_parser.mli
```diff
@@ -43,3 +43,6 @@ val position_of_point : t -> Loc.point -> Lexing.position
     the usual representation in the Lexing module. Note that this relies on the
     information passed in {!parse_comment}, and hence requires the result of
     that call in addition to the {!Loc.point} being converted. *)
+
+val codeblock_content : Loc.span -> string -> string * Warning.t list
+val verbatim_content : Loc.span -> string -> string * Warning.t list
```
jonludlam, line 48: These will need some doc comments, plus we'll need to add something to the `Code_block in ast.mli

jonludlam (approved):

src/parser/odoc_parser.mli
```diff
@@ -43,3 +43,29 @@ val position_of_point : t -> Loc.point -> Lexing.position
     the usual representation in the Lexing module. Note that this relies on the
     information passed in {!parse_comment}, and hence requires the result of
     that call in addition to the {!Loc.point} being converted. *)
+
+val codeblock_content : Loc.span -> string -> string * Warning.t list
+(** Process the content of a code block, following the rules described
+    {{!/odoc/odoc_for_authors.indentation_code_blocks}here}. To achieve this, it
```
panglesd, line 49: Isn't that cross-package link cool?

dbuenzli: It is.

PR #1326, opened 2025-03-04
Extract code blocks by name

panglesd:
This is an early prototype (on top of #1325) of an `extract-code` command, which extracts named code blocks into a file. Supersedes the venerable #303.

This demonstrates that the location is right:

`test.mld`:
```
{0 hello}

  {[

  let f x = x +. 2.

]}


  {[ let x = f 2

]}
```

```
$ dune exec -- odoc extract-code test.mld > ocaml.ml
$ ocaml ocaml.ml
File "test.mld", line 10, characters 16-17:
Error: This expression has type int but an expression was expected of type
         float
  Hint: Did you mean 2.?
```

### DIFF
--- a/src/parser/lexer.mll
+++ b/src/parser/lexer.mll
@@ -83,7 +83,8 @@ let warning_loc =
 let warning =
   with_location_adjustments warning_loc
 
-let unescape_tag s start_loc input =
+let unescape_tag s offset input =
+  let start_loc = Lexing.lexeme_start input.lexbuf + String.length offset in
   let warn n c =
     let start = input.offset_to_location @@ start_loc + n in
     let end_ = input.offset_to_location @@ start_loc + n + 2 in
@@ -389,7 +390,7 @@ and token input = parse
       in
       let emit_truncated_code_block () =
         let empty_content = with_location_adjustments (fun _ -> Loc.at) input "" in
-        emit ~start_offset input (`Code_block (Some (lang_tag, None), delim, empty_content, false))
+        emit ~start_offset input (`Code_block (Some (lang_tag, []), delim, empty_content, false))
       in
       (* Disallow result block sections for code blocks without a delimiter.
          This avoids the surprising parsing of '][' ending the code block. *)
@@ -401,7 +402,7 @@ and token input = parse
         code_block allow_result_block start_offset content_offset metadata
           prefix delim input lexbuf
       in
-      match code_block_metadata_tail None input [] lexbuf with
+      match code_block_metadata_tail input [] lexbuf with
       | `Ok metadata -> code_block_with_metadata metadata
       | `Eof ->
           warning input ~start_offset Parse_error.truncated_code_block_meta;
@@ -409,7 +410,7 @@ and token input = parse
       | `Invalid_char c ->
           warning input ~start_offset
             (Parse_error.language_tag_invalid_char lang_tag_ c);
-          code_block_with_metadata None
+          code_block_with_metadata []
     }
 
   | "{@" horizontal_space* '['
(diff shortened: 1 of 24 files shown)


panglesd:
So, this PR currently implements the `extract-code` command.

This commands takes a mld file, and `--line-directives` flag, and a list of `--name <name>` argument, and an option `-o`/`--output`.

I then outputs (to the `-o` or stdout) the concatenation of "matching" code blocks, with line directives if the flag is set.

The rules for matching code blocks is:
- If no `--name` are provided, all _ocaml_ code block matches. An ocaml code block is a code block without specified language or with "`ocaml`" specified as language.
- If some `--name` are provided, a code block match if it has a name that was passed. A code block has a name if its "tags" (the string after the language) contain `name=<name>`.

See the [test](https://github.com/ocaml/odoc/pull/1326/files#diff-e316ceff4966142b29d01e8c1ea7a774a104dab8b27e3ddffae9d4b725efa07f) for some examples.

Now, I wonder two things:
- Should we uniformize how tags are parsed? In this PR I use space as separators to look for `name=<...>`, but MDX seems to separate "labels" with `,`: see this [function](https://github.com/realworldocaml/mdx/blob/main/lib/label.ml#L207) called on [odoc tags](https://github.com/realworldocaml/mdx/blob/main/lib/mli_parser.ml#L182).
- What should we take as input in addition to mld files? Mli? cmti/cmi/cmt? (I don't think `.odoc` files should be taken as input as they move things around.)

dbuenzli:
The `mdx` docs I'm pointed to looks like a hack of the style, oh we are just going to use `String.split_on_char ','` and `String.split_on_char '='` to get our tags and variable bindings.

Please just don't. These things end up being infuriating at some point.

Cater for the fact that the tags/bindings may want to contain the separators, including and especially *spaces*. 

I think it's best to ignore whatever `mdx` has been doing and devise a proper grammar for tags and key values pairs. 

In general the basics of sexps are *always* sound to use. If I assume we want both "tags" (`atom`) and "bindings" (`atom=atom`). Define a notion of atom as being either an unquoted atom or quoted atom. A tag is an atom and binding two atoms separated by `=` without intervening whitespace (for simplicity). Tags and bindings are separated by whitespace.

*Very quickly*

```ocaml
type codeblock = {
  content: string;
  language: string;
  info_string: [ `Tag of string | `Binding of string * string ] list
}
```

With `info_string` parsed by:

```
ws = *(%x0020 / %x0009)  # Not sure if newlines are allowed here 

unquoted-char = … # Char.Ascii.is_graphic without `=` or `"`
unquoted-atom = 1*unquoted-char

escape = %005C %x0022 # More can be added, e.g. if we want newlines
quoted-char = … # Anything except `"` (U+0022)
quoted-atom = %x0022 (quoted-char / escape) %x0022 

atom = unquoted-atom / quoted-atom

tag = atom 
binding = atom %003D atom 
info-string = *(ws (tag / binding)) 
```

jonludlam:
> In general the basics of sexps are _always_ sound to use. If I assume we want both "tags" (`atom`) and "bindings" (`atom=atom`). Define a notion of atom as being either an unquoted atom or quoted atom. A tag is an atom and binding two atoms separated by `=` without intervening whitespace (for simplicity). Tags and bindings are separated by whitespace.
> 

This sounds good to me. Mdx is for 'live' files - those that can be updated - and so we can update the syntax of the tags. They don't affect rendering so we've got no worries about existing mld files in released packages. We might want to add a warning into mdx if it notices the comma-separated syntax maybe? We can probably come up with some heuristic.

panglesd:
## Tags

Ok, I implemented the parsing of the tags.
I tried to record all locations that could be useful:
- The location of tags (including potential `"` and `=`) and bindings
- The location of each binding part (including potential `"`)
- The full location of the tags (including initial and trailing spaces)

There is one thing. Mdx uses `<=` and other operators, which are not available anymore with our parsing. For instance, `version<=3` will be parsed as `("version<","3")`. I suggest mdx hack is way (eg by checking that the key ends with a `<` in the example above.)

## Cmti loading and compatibility

I first implemented cmti loading with `Tast_iterator`. Using this, I can just iter through the attributes, and I have basically no boilerplate traversal code to write.
However, the attribute traversal was only implemented in 5.2 or 5.1. And in fact, the typedtree traversal module was only implemented in 4.10 I think.
I still updated the traversal to be compatible with 4.10, as it is less annoying to do that write a compatible loading.

Could we say that odoc extract-code is a functionality which is only available in 4.10?

dbuenzli:
Ooops. Pushed wrong button :–)

jonludlam:
Personally I don't think we need to support extract-code on older versions of OCaml. It's way less important than rendering the docs for older versions. 

Also I'm not particularly convinced we need to go out of our way to try to support the syntax mdx currently uses. `version<=4.14` and the like seem a little dodgy (version of what?) and we could have a more generic syntax, e.g. `constraints="(ocaml <= 4.14)"` or similar

src/parser/test/test.ml
```diff
@@ -3401,6 +3468,148 @@ let%expect_test _ =
...
+        {|{@ocaml env=f1 version=4.06 "tag with several words" "binding with"=singleword also="other case" "everything has"="multiple words" [foo]}|};
+      [%expect
+        {|
+        ((output
+          (((f.ml (1 0) (1 137))
+            (code_block
+             (((f.ml (1 2) (1 7)) ocaml)
+              ((binding ((f.ml (1 8) (1 11)) env) ((f.ml (1 12) (1 14)) f1))
+               (binding ((f.ml (1 15) (1 22)) version) ((f.ml (1 23) (1 27)) 4.06))
+               (tag ((f.ml (1 29) (1 51)) "tag with several words"))
```
panglesd, line 3487: I think the location of quoted tags/bindings should include the quotes. Otherwise it's hard for programs that want to change the tags by modifying what is at this location...

src/parser/lexer.mll
```diff
@@ -358,15 +395,15 @@ and token input = parse
         code_block allow_result_block start_offset content_offset metadata
           prefix delim input lexbuf
       in
-      match code_block_metadata_tail input lexbuf with
+      match code_block_metadata_tail input None [] lexbuf with
       | `Ok metadata -> code_block_with_metadata metadata
       | `Eof ->
           warning input ~start_offset Parse_error.truncated_code_block_meta;
           emit_truncated_code_block ()
       | `Invalid_char c ->
```
panglesd, line 403: I think this `` `Invalid_char `` case cannot happen anymore.

src/parser/lexer.mll
```diff
@@ -692,24 +727,77 @@ and bad_markup_recovery start_offset input = parse
...
+ | '='
+   { match tag with
+     | Some t ->
+       let value = code_block_metadata_atom input lexbuf in
+       code_block_metadata_tail input None (`Binding (t, value) :: acc) lexbuf
+     | None ->
+       warning input (Parse_error.code_block_tag_invalid_char '=');
+       code_block_metadata_tail input None acc lexbuf }
+ | (_ # space_char # '[' # '=' as c) [^ '[']* '['
+   { warning input (Parse_error.code_block_tag_invalid_char c);
```
panglesd, line 798: In this case we are omitting all the rest of the tags. Might be worth mentioning in the error message, as one might be surprised that:
```
{@ocaml x=y \" z=t[
foo
]}
```
has binding `x=y` but not `z=t`.

panglesd: (or maybe the invalid character could be ignored, just like invalid `=` are ignored?

src/parser/lexer.mll
```diff
@@ -692,24 +727,77 @@ and bad_markup_recovery start_offset input = parse
...
+   {
+    let start_offset = Lexing.lexeme_start input.lexbuf + 1 in
+    Buffer.clear string_buffer;
+    let s = string input lexbuf in
+    with_location_adjustments ~start_offset ~adjust_end_by:"\"" (fun _ -> Loc.at) input s }
+ | (tag_unquoted_atom as value)
+   { with_location_adjustments (fun _ -> Loc.at) input value }
+ | (_ as c) 
+   { warning input (Parse_error.code_block_tag_invalid_char c);
+     with_location_adjustments (fun _ -> Loc.at) input "" }
```
panglesd, line 778: Nitpick: It's quite hard to read what values `c` can take, but in fact it can only be "=". I don't know if that is more unreadable to match on "=" or on "_"... but I think at least a comment would help!

```suggestion
 | (tag_unquoted_atom as value)
   { with_location_adjustments (fun _ -> Loc.at) input value }
 | ('=' as c) 
   { warning input (Parse_error.code_block_tag_invalid_char c);
     with_location_adjustments (fun _ -> Loc.at) input "" }
```

```suggestion
 | (tag_unquoted_atom as value)
   { with_location_adjustments (fun _ -> Loc.at) input value }
 | (_ as c)   (* At this point, c can only be "=" *) 
   { warning input (Parse_error.code_block_tag_invalid_char c);
     with_location_adjustments (fun _ -> Loc.at) input "" }
```

panglesd (commented):
Thanks, the ast simplification is welcome, and the escaping is better.

The only important comment is that I think locations should include the quotes!

src/parser/test/test.ml
jonludlam: Yes, that's fair.

src/parser/lexer.mll
jonludlam: True, good spot!

src/parser/lexer.mll
jonludlam: I think I prefer the explicit '=' version.

src/parser/lexer.mll
jonludlam: It's a little tricky because this bit of the lexer is required to start with whitespace, which makes the errors a bit odd. If we just skip the char, then for example, the test `{@ocaml =foo [ code ]}` if would complain about the '=' and then it will also complain about the 'f', the 'o' and the 'o'. It's currently trying to skip the rest of the metadata until the '[', but I could try to make it just skip until some whitespace.

PR #1300, opened 2025-02-03
Remove support for OCaml < 4.08

jonludlam:

PR #1328, opened 2025-03-10
Remove compatcmdliner and depend on cmdliner >= 1.3.0.

dbuenzli:
Now that < 4.08 is out since https://github.com/ocaml/odoc/pull/1300 that's possible. 

It will not be possible to make `compatcmdliner` compatible with cmdliner 2.0. 

This was a quick port, I did not try to improve along more idiomatic cmdliner usage.

Note that this make `odoc` compatible with cmdliner 2.0.


jonludlam:
Thanks @dbuenzli !

dbuenzli:
Thank you!

dbuenzli:
That didn't make it into 3.0.0 😢 ?

PR #1332, opened 2025-03-18
latex backend: full path in heading labels

Octachron:
This PR fixes the label used for headings in the latex backend.

Previously, the latex backend used only the heading label, for instance,

```ocaml

(** {1:s_name Title } *)
```
generated
```latex
\subsection{Title\label{s_name}}
```
whereas the references used the full path
```latex
\hyperref[Module_name-s_name]{Title[p\pageref*{Module_name-s-name}]}
```

This PR fixes this discrepancy (observed on the odoc version of the OCaml manual) by using the full path in the heading translation.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 - Drop support for OCaml < 4.08 (@jonludlam, #1300)
 - Allow referencing libraries from package added in `odoc-config.sexp`
   (@panglesd, #1343)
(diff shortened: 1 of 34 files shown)


Octachron:
There seems to be multiple changes required to make the reference manual works with odoc 3, I am planning to switch this PR to non-draft mode once I have compiled all the changes.

PR #1333, opened 2025-03-20
Minor doc fix!

jonludlam:


panglesd (approved):
I made a very long and careful review, and that looks good to me.

PR #1337, opened 2025-03-25
latex backends: add an option to shorten submodules

Octachron:
In the latex backend, inlining all submodule and functor definitions in their long form is not always the optimal choice.
In particular, it can be better to be more concise and use only the short description for nested submodules

This PR adds a flag `--shorten-beyond-depth` to the latex backend  to make it choose the abbreviated form
for submodules beyond a given depth.

Along the way, this PR fixes a collision between page url and anchor and adds missing label definitions for pages that have been inlined in the current page in order to redirect references to those page toward the inlined contents.

Combined with an option to remove link to functor arguments, this seems enough to remove all missing labels warning from the build of the reference manual.


Octachron:
@jonludlam , would you prefer to me to add a Changes entry with the initial PR?

By default, I tend to add it once the PR has been reviewed, but switching to an optimist mode for the Changes entry is fine with me.

PR #1338, opened 2025-04-04
Use `(format-dune-file)` instead of `dune format-dune-file`

nojb:
Hello,

You are currently calling `dune format-dune-file` from one of your tests. This command does not honour the Dune lang version used in the project which means that your test will break whenever we change the formatting style in Dune.

In Dune 3.18 we introduced a built-in action `(format-dune-file <src> <dst>)` precisely for this purpose. This action honours the current Dune lang version and does not require shelling out to an external process.

Note that `dune format-dune-file` may evolve in ways that may not be backwards compatible in the near future (as the CLI interface does not have any backwards compatibility or versioning guarantees), so we are encouraging users of this command to switch to the action instead.

See https://github.com/ocaml/dune/pull/10892#issuecomment-2503530595 for some of the context.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,4 +1,4 @@
-(lang dune 3.7)
+(lang dune 3.18)
 
 (name odoc)
 
--- a/odoc-bench.opam
+++ b/odoc-bench.opam
@@ -31,7 +31,7 @@ depends: [
   "astring"
   "cmdliner" {>= "1.3.0"}
   "cppo" {build & >= "1.1.0"}
-  "dune" {>= "3.7.0"}
+  "dune" {>= "3.18.0"}
   "fpath"
   "ocaml" {>= "4.02.0"}
   "result"
--- a/odoc-driver.opam
+++ b/odoc-driver.opam
@@ -36,7 +36,7 @@ documentation for installed packages.
 depends: [
   "ocaml" {>= "5.1.0"}
   "odoc" {= version}
-  "dune" {>= "3.7.0"}
+  "dune" {>= "3.18.0"}
   "odoc-md"
   "bos"
   "fpath"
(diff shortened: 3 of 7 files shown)


nojb:
> Thanks @nojb! I was about to do it today but you were clearly faster 😄
> 
> One question I have regarding this: shouldn't it add a dependency on `dune >= 3.18` to make sure it supports the functionality?

Yes, indeed. I pushed a fix (I think ; I am not very familiar with OPAM).

panglesd (approved):
Thanks @nojb! That looks good.

The benchmark CI failure seems to be related to https://github.com/ocaml/dune/issues/11045.
It does not happen on ocaml-ci, which install deps and then call dune. On the contrary, the benchmark ci seems to pin everything and install them with opam, which calls `dune subst` and that is what is breaking.

Maybe that would be better to wait for the https://github.com/ocaml/dune/issues/11045 fix to be released before raising `dune lang` if that is what triggers the bug?

panglesd:
I think it does include it. The old error message was:
```
File "odoc-bench.opam", line 1, characters 10-25:
1 | # This file is generated by dune, edit dune-project instead
              ^^^^^^^^^^^^^^^
Error: repository does not contain any version information
```
and it is now:
```
File "odoc-bench.opam", line 1, characters 0-0:
Error: repository does not contain any version information
```

I guess the error is either on the benchmark CI or on dune's side, so it would be a pity to block this PR on it. I don't know how much the benchmark CI is used, at least I struggle with its UI...

jonludlam:
OK, we'll assume a fix is on its way and merge this anyway.

shonfeder:
Thanks very much!

PR #1350, opened 2025-06-04
Remove use of /tmp from a test

jonludlam:


panglesd:
I'm not sure how this test is useful. It was introduced in https://github.com/ocaml/odoc/pull/1081, so I guess it's about the computed content of the signature?

PR #1355, opened 2025-06-24
OCaml 5.4 support

Octachron:
This PR updates the odoc model to support labelled type and bivariant type parameters (in commit 6928546a5952474e3ac2361b2770d076958379ba) and update the loader to be compatible with the OCaml 5.4 AST in dcd8b8bfb725b6a0e977ec6fca0d7b4cd0e9da3a .

There is also one test modified in eb5c6e5877ab77d262e5a33ad866df98fa2d06a7 to avoid a problematic case that is now rejected in OCaml 5.4 (a module type being abstracted due to a module shadowing).

Similarly, 4fe1af921b06fd17cccbdfa9017ca3e8fe7a86e5 updates two tests using unicode character to be more stable due to the improved handling of non-ascii character in Format in OCaml 5.4 .

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,8 @@
 
 ### Fixed
 - Fix bug in parsing META files when there are no dependencies (@jonludlam, #1352)
+- OCaml 5.4.0 support
+  (@Octachron, #???)
 
 # 3.0.0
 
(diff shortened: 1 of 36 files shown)


jonludlam:
Thanks @Octachron ! I'll take a look at this ASAP.

jonludlam:
Looks good - thanks @Octachron !

PR #1358, opened 2025-07-01
Add missing parentheses in type expr `(_ as 'a) t`

Julow:
This bug is found for example in Eio: https://ocaml.org/p/eio/latest/doc/Eio/Domain_manager/index.html#type-t
Type `t` renders as invalid syntax:
```
type 'a t = [> ty ] as 'a Resource.t
```
In the source code, it's:
```
type 'a t = ([> ty] as 'a) Resource.t
```


jonludlam:
Oops, forgot the CHANGES entry!

PR #1364, opened 2025-07-11
Fix license info for cmarkit code

jonludlam:


dbuenzli:
Yes thank you! (note I don't mind not having the text in `LICENSE`, I think it becomes a maintenance burden). The attribution in `renderer.ml` is enough for me.

jonludlam:
OK, in that case I will take it out - as you say, one less thing to think about :-)

PR #1375, opened 2025-08-11
Improve file magic handling

jonludlam:
Before this change, the magic string was read by reading the number of bytes we're expecting to read based on our magic string. If ours happens to be a prefix of the real one (e.g. ours is 'odoc-3.1.0' and the real one is 'odoc-3.1.0-2-g12345789') then the check was succeeding and then the subsequent unmarshalling was failing.

This fixes that by writing the length too, so we always unmarshal the right length.

### DIFF
--- a/src/odoc/odoc_file.ml
+++ b/src/odoc/odoc_file.ml
@@ -29,13 +29,24 @@ type content =
 type t = { content : content; warnings : Odoc_model.Error.t list }
 
 (** Written at the top of the files. Checked when loading. *)
-let magic = "odoc-%%VERSION%%"
+let magic = "ODOC"
+
+let magic_version = "%%VERSION%%"
 
 (** Exceptions while saving are allowed to leak. *)
 let save_ file f =
+  let len = String.length magic_version in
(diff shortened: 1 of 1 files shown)


panglesd (approved):
What if the version starts with a number? 😄 
(We could switch `ODOC` and the length, or have a "v" in between length and `%%VERSION%%`, but I'm not sure it is worth it, as `%%VERSION%%` is likely to never start with a number, if you say it is currently `odoc-3.1.0`!)

jonludlam:
I'm not sure I follow. The new code writes "ODOC" then a binary representation of the length of the version string, then it writes exactly that many bytes of version string.

jonludlam:
I noticed a silly mistake in the error message. Fixed!

panglesd:
> I'm not sure I follow. The new code writes "ODOC" then a binary representation of the length of the version string, then it writes exactly that many bytes of version string.

Oh you are right, for some reason I though it was a _string_ representation of the length of the version string. I deal too often with string serialization.

Then, it's all good!

PR #1383, opened 2025-08-28
Add support for opam 2.5

kit-ty-kate:
See https://github.com/ocaml/opam/pull/6442

String.starts_with has been available since OCaml 4.13, so this change is fully backward compatible given that odoc-driver requires at least 5.1


jonludlam:
Thanks!

PR #1388, opened 2025-10-15
Support multiple inputs in markdown-generate (and html-generate)

davesnx:
Since integrating markdown in dune (https://github.com/ocaml/dune/pull/12477), I found (idea by @jonludlam) would be helpful to generate markdown files from a list of odocl files.

This change is propagated into html-generate (which I belive is a nice addition and it doesn't break any previous usage of the cli). If you prefer to not change this, I could implement a new `Main.generate`


jonludlam:
This seems a totally reasonable change to me. Thanks, @davesnx !

PR #1392, opened 2025-11-20
Remove obsolete symlinks

Sudha247:
This patch removes symlinks that are not required anymore.

I was building some projects with `dune pkg` with `odoc-parser` as a dependency, and found these symlinks. Dune package management doesn't support symlinks yet. In an offline discussion with @jonludlam, he pointed out that a few of these symlinks are obsolete now. Removing these would make compiling with `dune pkg` easier.


jonludlam:
odoc.mld needs to be renamed to index.mld too, but otherwise this looks fine!

PR #1395, opened 2025-12-10
Fix URL remapping for page references

jonludlam:
Page links were not being remapped because the remap prefix check only matched against the directory path. For modules, the module name is part of the directory (e.g., prefix/otherpkg/Otherlib/index.html), so it matched. But for pages, the page name is in the filename (e.g., prefix/otherpkg/otherpage.html), leaving only prefix/otherpkg in the directory - which didn't match prefix/otherpkg/ due to the trailing slash.

Fix by including the filename in the path used for prefix matching. Also normalize remap entries at config creation time (ensure trailing slashes) rather than on each lookup.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ### Fixed
 - Markdown backend assumes ocaml as langId on declarations
 - Markdown backend collects all code snippets and render a single code block with comments, rather than split the variants/records per constructor/field
+- Fix URL remapping for page references (@jonludlam, #1395)
 
 # 3.1.0
 


src/html/config.ml
```diff
@@ -15,9 +15,18 @@ type t = {
   home_breadcrumb : string option;
 }
 
+let ensure_trailing_slash s =
+  if Astring.String.is_suffix ~affix:"/" s then s else s ^ "/"
```
Julow: I don't think a possibly long list of string with requirements in them is the right data structure.
I think a Trie structure would be better.

src/html/link.ml
```diff
@@ -26,9 +26,11 @@ module Path = struct
     match remaps with
     | [] -> None
     | (prefix, replacement) :: _ ->
-        let len = String.length prefix in
-        let l = String.sub l len (String.length l - len) in
-        Some (replacement ^ l)
+        let remainder =
+          String.sub path (String.length prefix)
+            (String.length path - String.length prefix)
```
Julow: This is also `Astring.String.with_range ~first:(String.length prefix) path`.

src/html/config.ml
jonludlam: True, though this PR doesn't change that. It'll be a good future improvement.

Issue #22, opened 2016-09-20
Non-uniform treatment of toplevel modules and submodules


dbuenzli:
[`Asetmap`](http://docs.mirage.io/odoc/asetmap/Asetmap/index.html) has no toplevel `module Asetmap : sig ... end` definition. 

[`Asetmap.Set`](http://docs.mirage.io/odoc/asetmap/Asetmap/Set/index.html) has a `module Set : sig end` definition. 

This should be unified, I would be in favor of dropping the `module Set : sig end` as it is kind of redundant.

Issue #21, opened 2016-09-20
odoc doesn't respect stop comments


dbuenzli:
For example in uuidm we have this stop comment:

https://github.com/dbuenzli/uuidm/blob/master/src/uuidm.mli#L38-L41

`create` still shows up in `odoc`'s API doc with dubious C multiline start comments.

```
opam install uuidm && opkg ocamldoc uuidm && opkg odoc uuidm && opkg doc -c uuidm
```

trefis:
The problem in this case is that doc-ock expects stop comments to be floating comments whereas in your example the compiler has attached the comments to `create`.

I feel like we should keep the implementation of odoc as is and you should add some blank lines in your code so they are not attached to the node.

dbuenzli:
I think this is going to be a problem in practice. The [specification](http://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#sec338) really says that everything between two `(**/**)` is dropped. It feels a little brittle to ask people to add whitespace for this to work.

trefis:
> It feels brittle

Really? Properly handling these comments when they are attached to a node involves making some relatively ugly `Location`-based decisions.
Just asking people to add a newline here and there doesn't feel "brittle" to me and we already asked them to add such newlines in the past, e.g. to disambiguate things like:
```ocaml
val foo : 'a -> 'b
(** where do I attach? *)
val bar : 'a -> 'b
```
so it doesn't seem too crazy to make the same request in this case.

@lpw25 any opinion?

dbuenzli:
> Really?

Yes. Your example is quite different. As you can read in the specification I linked to there's no ambiguity in a `(**/**)` comment. 

My take on this is that the comment lifting is wrongly done upstream. `(**/**)` should never get attached to a value specification.

trefis:
I agree on both points. It's just that I don't feel like the compiler is going to be "fixed" on that point (but @lpw25 might contradict me on this?) and I'm not really keen on introducing shitty "detachment" decisions based on locations in doc-ock.

That being said I don't mind submitting a PR on the compiler just to see how it goes.

dbuenzli:
> I'm not really keen on introducing shitty "detachment" decisions based on locations in doc-ock.

Sure it doesn't seem to be the right place to do this.

dbuenzli:
Excellent @trefis and thanks to @damiendoligez for merging this into 4.04

Issue #14, opened 2016-09-19
Enhance toplevel navigation.


dbuenzli:
I'd like to give direct links in that [nav](https://www.w3.org/TR/html5/sections.html#the-nav-element) tag to the package list and to the package page. That is
rather than simply have : `Up`, you'd get at any level `packages - pkg vX.Y.Z - Up`. 

The markup you want to add here seems sufficiently small to be specified on the cli rather than read from a file.

dbuenzli:
As @trefis mentions this is not trivial if we want relative links. The idea is to simply have: 

`<nav><a>Up</a> <a>package <span class="package">$PKG</span></a></nav>`

which for UI gives a consistantly located `Up` someone can bang on it until it gets to the package list or a direct click on the package page.

PR #1399, opened 2026-01-28
Upstream OxCaml

art-w:
(builds on top of #1398 which provides `ocaml-config:ox` to detect OxCaml)

This PR is a rebase of @lukemaurer 's patches from [oxcaml/odoc](https://github.com/oxcaml/odoc) (+ [small fixes from #22](https://github.com/oxcaml/odoc/pull/22)). I hope this will make it easier to keep everything uptodate, and that it isn't too much of a stretch as Odoc already supports multiple versions of the OCaml AST using `cppo`. (Note that the preprocessor was originally using `OCAML_VERSION=(5,2,0)` as a heuristic to check for OxCaml, but since dune 3.21 we can be exact.)

### DIFF
--- a/src/.ocamlformat-ignore
+++ b/src/.ocamlformat-ignore
@@ -1,14 +1,15 @@
 document/*.cppo.ml
-loader/*.cppo.ml
 loader/cmi.ml
 loader/cmi.mli
 loader/cmt.ml
 loader/cmti.ml
 loader/doc_attr.ml
+loader/ident_env.ml
 loader/implementation.ml
-loader/typedtree_traverse.ml
 loader/lookup_def.ml
 loader/lookup_def.mli
+loader/odoc_loader.ml
+loader/typedtree_traverse.ml
 syntax_highlighter/syntax_highlighter.ml
 model/*.cppo.ml
 odoc/*.cppo.ml


.github/workflows/oxcaml.yml
```diff
@@ -0,0 +1,59 @@
...
+          ocaml-compiler: ${{ matrix.ocaml-compiler }}
+          # Do not pin odoc to not break Mdx installation
+          opam-pin: ${{ matrix.run-mdx != true }}
+          opam-repositories: |
+            oxcaml: "git+https://github.com/oxcaml/opam-repository.git"
+            default: "git+https://github.com/ocaml/opam-repository.git"
+
+      - name: Install dependencies
+        run: |
+          sed -i '/bisect_ppx/d' odoc.opam
```
art-w, line 41: `bisect_ppx` is currently incompatible with OxCaml, bringing the coverage down

art-w:
@jonludlam If you have time, I think this PR is in good shape for a review!

This is only the first step to integrate oxcaml/odoc in this repo, so currently I've left out a [small patch "Support OxCaml polymorphic arguments (#4)"](https://github.com/art-w/odoc/commit/5b6066bbf079f64e057513ca2369bb39fd8d5ff8) from @lukemaurer as it was breaking the OCaml 4.08 regression tests. The polymorphic arguments require a bit more work and testing so I'll open another PR for them.

The other useful patches from @lukemaurer in oxcaml/odoc are not directly related to OxCaml support, so I've left them out to ease the review as we can cherry-pick them later:
* 77706571c - `CSS fix for layout when there is no global ToC (#12)` has "FIXME: This will probably get a better fix upstream"
* 8213d44b6 - `Add source locations to more places (#14)` uses an alternative `source_loc_jane` to point at file locations
* 4b8f98867 - `Print source information in implementations (#13)` is also useful but not OxCaml specific
* 328ae6f12 - `Support compiling against 'result' library (#19)` conflicts with 8478ff40c `Remove dependency on the 'result' package` by @Julow
* 0e57a5b83 - `Update integration test (#21)` is a fix of a test output, not required for this PR

src/xref2/ref_tools.ml
```diff
@@ -519,17 +521,28 @@ module CS = struct
...
+  let not_a_constructor name =
+    (* Let's pretend we didn't see the field/unboxed field and say we didn't find anything. *)
     Error (`Find_by_name (`Cons, name))
 
   let in_parent _env (parent : fragment_type_parent_lookup_result) name =
     let name_s = ConstructorName.to_string name in
     match parent with
     | `S (parent', parent_cp, sg) -> (
         let sg = Tools.prefix_signature (parent_cp, sg) in
+        let find_ambiguous =
```
jonludlam: This is a bit odd - what's going on here?

jonludlam:
Other than that minor concern about the 'find_ambiguous' thing, this looks good to me

src/xref2/ref_tools.ml
art-w: Oh good catch, I'll remove it since the type annotation doesn't seem useful

art-w:
Thanks for the review! I've addressed your comment :)

jonludlam:
Great, thanks. In it goes!

PR #1404, opened 2026-03-20
OxCaml: Fix missing parens for polymorphic parameters

art-w:
Follow up on #1399: OxCaml polymorphic parameters were missing parentheses in Odoc rendering (note that polymorphic parameters were upstreamed https://github.com/ocaml/ocaml/pull/13806 and are expected in OCaml 5.5).

(Only the last two commits are new: the first adds a test for OxCaml to check the support, the second patch fixes the missing parens.)

edit: marked as draft because we need to merge #1399 first, but otherwise the PR is ready to review!


art-w:
@jonludlam : Thanks for merging #1399! Hopefully, this PR should be an easy review now :)

jonludlam:
Thanks, yes, this looks good, and we've finally got a green tick for passing tests too :-)

PR #1406, opened 2026-03-25
Ocaml 5.5 compat

panglesd:
Support for 5.5 with:
- Support for external types, displayed as abstract types
- Support for module arrows.

The second one is trickier. In this PR, at least:
- The type is written
- There is no warning about missing elements from the environment.

However, the path referencing the module argument appear as unresolved links: underlined in red, suggesting an error...

The state of the PR is draft because it needs to be cleaned up and improved, but I already open it in case someone wants to try odoc on OCaml 5.5.

:robot: Generated with @xvw


odoc.opam
```diff
@@ -45,7 +45,7 @@ depends: [
   "cppo" {build & >= "1.1.0"}
   "dune" {>= "3.21.0"}
   "fpath" {>= "0.7.3"}
-  "ocaml" {>= "4.08.0" & < "5.5"}
+  "ocaml" {>= "4.08.0" & <= "5.5.1"}
```
kit-ty-kate:
```suggestion
  "ocaml" {>= "4.08.0" & < "5.6"}
```

odoc-parser.opam
```diff
@@ -14,7 +14,7 @@ dev-repo: "git+https://github.com/ocaml/odoc.git"
 doc: "https://ocaml.github.io/odoc/odoc_parser"
 depends: [
   "dune" {>= "3.21"}
-  "ocaml" {>= "4.08.0" & < "5.5"}
+  "ocaml" {>= "4.08.0" & <= "5.5.1"}
```
jonludlam: This one too!

jonludlam: Actually thinking about it, why do we have an upper bound on odoc-parser? I can understand with odoc where every ocaml release breaks it, but the parser ought to be way less brittle.

panglesd: I agree, there is no reason to have this "minor release" upper bound!

panglesd:
I'd still like to clean the history, but the PR is ready for review.

test/generators/cases/ocaml_55.mli
```diff
@@ -0,0 +1,19 @@
...
+
+val f'' : (module X with type t = int) -> int
+
+module type Y = sig type 'a t val return : 'a -> 'a t end
+
+val g : (module M : Y) -> int M.t
+
+val g' : (module M : Y) -> int
+
+val g'' : (module Y) -> int
```
Octachron, line 19: You should probably also add a test for polymorphic parameters:
```ocaml
val map2: ('a. 'a -> 'a) -> 'a * 'b -> 'a * 'b
```
after a quick look, I have the impression that the current document generation code is missing parentheses.

panglesd: Thanks! I'll add that today.

panglesd: I cherry-picked @art-w's 0752cc5e19bc759a3129f58dce0deab50dd1d4e3 which fixed exactly that (thanks!).

@art-w I did not include the Oxcaml test since it is not part of "5.5 support". Whichever is merged last, #1404 or this one, will have to rebase!

panglesd:
> I'd still like to clean the history

I rebased and cleaned the history. (And addressed ochtachron's comment)

panglesd:
Rebased since the merge of #1404

xvw (approved):
For what it's worth, and judging by the output results!

Octachron:
On the subject of the module argument, I was wondering if it would not better to link `M.t` in `(module M:T) -> M.t`
 to the `t` component of the module type since at least for now the module type of a module argument is always a named module.
 
 Nevertheless, I agree that this can be improved later.

panglesd:
> On the subject of the module argument, I was wondering if it would not better to link M.t in (module M:T) -> M.t
to the t component of the module type since at least for now the module type of a module argument is always a named module.

Would you link the same way when the type is replaced?

```ocaml
module type S =  sig type t = float end

val f : (module M : S with type t = int) -> M.t
```

I would find it misleading! The way to improve I can think of are too involved for this initial support...

Octachron:
Your example is a type error, thus I am not sure how illustrative it is.
 The constraints can only refine an abstract type:
 ```ocaml
 module type S = sig type t end
 val f : (module M : S with type t = int) -> M.t
 ```
 in which case is not that clear what is the best option between linking to `M.t` or `int`, but following the intent
 of the author that wrote `M.t` rather than `t` does not sound that bad.
 
 Anyway, I agree that this is probably best left for further refinement.

jonludlam:
Looks good! I came across some test failures when I was looking through it though, so I've fixed them (very minor). As soon as CI shows green we'll merge!

PR #1407, opened 2026-03-31
OxCaml: Support for unboxed named types

art-w:
Follow-up on #1404, only the last two commits are new:
- The first one introduces a test for OxCaml unboxed types `foo#` to show that the trailing `#` is missing in generated documentation
- The second one propagates the unboxing from the compiler representation `Pextra_ty (_, Punboxed_ty)` in odoc internals, to fix the test


panglesd (approved):
Looks good to me!

Issue #564, opened 2021-01-19
Support including submodules in their parent's page


lpw25:
Sometimes people have small sub-modules and they would prefer to get something like:

```ocaml
  module Foo : sig
  
    type t
    
    val x : t
    
  end
```
in their docs instead of:
```ocaml
  module Foo : sig ... end
```
and a separate page for `Foo`.

It would be nice to allow users to annotate modules with some kind of tag to indicate that they would like this behaviour. It would require support in the model since links that used to go to the separate `Foo` page would now need to go to its parent page.

lpw25:
I thought we already had an issue for this, but I couldn't find it. I also think this feature request is pretty low priority at the moment.

ulrikstrid:
Semi-related: https://github.com/ocaml/odoc/issues/323

Also, I really want this 😄

Drup:
I was slightly confused so I checked: It's already supported everywhere .... except there are no annotations for this in the syntax, unlike `include`. The models and everything down already does the right thing. In fact, both the latex and man output automatically unfold submodules, and the links are still correct in latex.

lpw25:
I don't see how the links can be correct. You need to add a new form to the path syntax to represent "project submodule 'Foo' that is in the same page as its parent" and then resolving has to detect an annotation on the definitions and decide to use that projection rather than the ordinary one. Otherwise it will try to link to a non-existent page for the submodule.

lpw25:
To be clear, this is fine in latex because there is only one possible projection in latex. In the html there will be two possible projections and resolution needs to produce the correct one.

ulrikstrid:
Related to this I would like it if we can expand `module type` in the file. My usecase is that I have a "API" `module type` that then different modules implement. So I would like to showcase the API in the main module so that it's immediately obvious instead of clicking into one of the implementations.

jonludlam:
Right - we'll also need to be careful with anchors as doing it naively will cause name clashes.

jonludlam:
I'll rename this so it applies to all expansions rather than just modules.

dmbaturin:
I also would like to see this implemented.

dbuenzli:
I think this is a pretty bad idea and should not be implemented. I'll just repeat here what I said [in that discussion](https://discuss.ocaml.org/t/inline-sub-module-documentation-with-odoc/8293) on discuss.

>  I'm not saying you won't be able to find examples where the feature makes sense. 
> 
[...]

PR #1408, opened 2026-04-06
Add @inline tag support for module declarations

mjambon:
This PR was generated by Claude Code to give a sense of how much work is needed to implement #564. It may or may not be great.

## Summary

Support `@inline` on module declarations, as requested in #564.

When a module declaration's doc comment includes `@inline`, its contents
are rendered directly on the parent page rather than on a separate sub-page:

```ocaml
(** An inlined module — contents appear directly on this page.
    @inline *)
module Inlined : sig
  type t
  val create : unit -> t
end
```

This mirrors the existing behaviour of `include … (** @inline *)`.

Modules without an inline signature (aliases) are unaffected, and nested
sub-modules within an inlined module still get their own pages unless they
are also marked `@inline`.

## Implementation notes

- `Lang.Module.t` gains an `inline : bool` field.
- A new `Expect_module_tags` variant in `Semantics` processes both `@canonical`
  and `@inline` in a single pass over a module's doc-comment tags, avoiding
  spurious warnings that would arise from two separate `find_tag` calls.
- The `inline` field is threaded through `Component.Module.t` and the xref2
  round-trip so it survives link resolution.
- The document generator emits `Item.Include { status = \`Inline }` for
  inline modules, reusing the rendering path already used by inlined includes.

**Known limitation:** cross-references inside an inline module's items (e.g.
`val create : unit -> t`) currently point to the ungenerated subpage
(`Module-Inlined.html#type-t`). This is the same situation as with inlined
`include` statements and can be addressed in a follow-up.

## Test plan

- [ ] New test case `test/generators/cases/module_inline.mli` covering:
  normal module, `@inline` module, `@inline` on an alias (no-op), and
  nested inlined module with a non-inlined inner module.
- [ ] Expected outputs checked in for HTML, LaTeX, man, and markdown generators.
- [ ] `dune runtest test/generators` passes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 # Unreleased
 
 ### Added
+- Support `[@inline]` tag on module declarations. When a module declaration's
+  doc comment includes `@inline`, its contents are rendered directly on the
+  parent page rather than on a separate sub-page — matching the behaviour of
+  `include … (** @inline *)`. (#564)
 - Allow persistent latex macros in HTML/KaTeX backend (@dlesbre, #1391)
 - `markdown-generate` command now accepts multiple `.odocl` files in a single
   invocation, eliminating the need for shell scripting (@davesnx, #1387)
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -1443,40 +1443,59 @@ module Make (Syntax : SYNTAX) = struct
         | ModuleType e -> expansion_of_module_type_expr e
       in
       let source_anchor = source_anchor t.source_loc in
-      let modname, status, expansion, expansion_doc =
-        match expansion with
-        | None -> (O.txt modname, `Default, None, None)
-        | Some (expansion_doc, items) ->
-            let status =
-              match t.type_ with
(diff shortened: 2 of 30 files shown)


jonludlam:
> Known limitation: cross-references inside an inline module's items (e.g.
val create : unit -> t) currently point to the ungenerated subpage
(Module-Inlined.html#type-t). This is the same situation as with inlined
include statements and can be addressed in a follow-up.

It isn't true that this is the same situation as inlined include statements. Includes are always on the same HTML page whether they're inlined or not, so the links are unaffected. This is in fact the big missing piece of this PR. The problem is that when you're resolving a path/reference we need to know whether the module will end up inlined or not, so we probably need to add some new constructors/fields to the paths, references and identifiers. We've got something a little similar with the 'OpaqueModule' constructor - where if we know that a module has no expansion then a link to it will resolve to the declaration point in the parent module rather than the expansion of the module. 

It's a good start, but there's quite a lot more work to do.

jonludlam:
I'll close this for now.

PR #1410, opened 2026-04-14
OxCaml: Support for kind annotations

art-w:
Builds on top of https://github.com/ocaml/odoc/pull/1404 (which adds the oxcaml test file)

This PR adds support for [OxCaml kind annotations](https://oxcaml.org/documentation/kinds/syntax/) :)

### DIFF
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -309,8 +309,7 @@ module Make (Syntax : SYNTAX) = struct
 
     val kind_annotation : Odoc_model.Lang.KindAnnotation.t -> text
 
-    val with_kind_annotation :
-      Odoc_model.Lang.KindAnnotation.t -> text -> text
+    val with_kind_annotation : Odoc_model.Lang.KindAnnotation.t -> text -> text
   end = struct
     let rec te_variant (t : Odoc_model.Lang.TypeExpr.Polymorphic_variant.t) =
       let style_arguments ~constant arguments =
@@ -453,8 +452,7 @@ module Make (Syntax : SYNTAX) = struct
     and with_kind_annotation kind base =
       match kind with
       | Odoc_model.Lang.KindAnnotation.Default -> base
-      | k ->
-          O.txt "(" ++ base ++ O.txt " : " ++ kind_annotation k ++ O.txt ")"
+      | k -> O.txt "(" ++ base ++ O.txt " : " ++ kind_annotation k ++ O.txt ")"
 
     and type_expr ?(needs_parentheses = false) (t : Odoc_model.Lang.TypeExpr.t)
         =
@@ -880,9 +878,12 @@ module Make (Syntax : SYNTAX) = struct
           let base = format_param_str x |> Syntax.Type.handle_format_params in
           Type_expression.with_kind_annotation x.kind (O.txt base)
       | lst ->
-          let l = match delim with `parens -> "(" | `brackets -> "[" in
-          let r = match delim with `parens -> ")" | `brackets -> "]" in
-          O.txt l ++ O.list lst ~sep:(O.txt ", ") ~f:format_param ++ O.txt r
+          let left, right =
+            match delim with `parens -> ("(", ")") | `brackets -> ("[", "]")
+          in
+          O.txt left
+          ++ O.list lst ~sep:(O.txt ", ") ~f:format_param
+          ++ O.txt right
 
     let format_constraints constraints =
       O.list constraints ~f:(fun (t1, t2) ->
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -306,6 +306,11 @@ module Make (Syntax : SYNTAX) = struct
 
     val format_type_path :
       delim:[ `parens | `brackets ] -> Lang.TypeExpr.t list -> text -> text
+
+    val kind_annotation : Odoc_model.Lang.KindAnnotation.t -> text
+
+    val with_kind_annotation :
+      Odoc_model.Lang.KindAnnotation.t -> text -> text
   end = struct
     let rec te_variant (t : Odoc_model.Lang.TypeExpr.Polymorphic_variant.t) =
       let style_arguments ~constant arguments =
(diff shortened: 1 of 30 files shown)


test/generators/gen_rules/gen_rules.ml
```diff
@@ -62,11 +62,12 @@ let constraints =
     ("ocamlary.mli", Min "4.14");
     ("recent.mli", Min "5.4");
     ("labels.mli", Min "4.09");
-    ("recent_impl.ml", Min "4.09");
+    ("recent_impl.ml", MinNotOxCaml "4.09");
```
art-w, line 65: `recent_impl.ml` is the only `.ml` file currently tested, and it uses simple `type t = A` which in OxCaml now renders as `type t : immediate = A` (which is correct, but not compatible with the OCaml output). I tried to "complexify" the type definition, but then you either get `immutable_data` or `mutable_data` instead so that doesn't help! :P

(This doesn't seem to be an issue for `.mli` testing, because OxCaml doesn't try to infer more precise kinds than the ones specified)

panglesd: That seems fine to me. OxCaml not being run on this specific test seems fine. Adding the variant `MinNotOxCaml` is fine.

Just to confirm I understand correctly:
```ocaml
type t = A
```
will be loaded as different values in odoc, depending on whether they are in an `ml` or `mli` file:
- In an ml file, some layout is inferred for `t`, from the definition, and stored in the `cmt`
- In an mli file, nothing is inferred and the `Default` annotation is loaded from the `cmi`.

This seems strange, and will discourage the use of `ml` without `mli`, since it will be impossible to have a type without an annotation!
But if it comes from the cmti, I'm not sure we should do anything about that.

panglesd (approved):
Looks good to me!

The only improvements I see is to have test the kind products (eg, `bit32 & float64`). In particular, do we need to use parentheses for thing like `(bit32 mod portable) & (float64 mod contended)`?  


I am also not sure what `Pjk_kind_of` represents, and whether it is tested or not...

src/document/generator.ml
```diff
@@ -426,6 +430,30 @@ module Make (Syntax : SYNTAX) = struct
...
+          | mods ->
+              O.txt " " ++ O.keyword "mod"
+              ++ O.txt (" " ^ String.concat ~sep:" " mods))
+      | Kind_of ty -> O.keyword "kind_of_" ++ O.txt " " ++ type_expr ty
+      | Product ks -> O.list ks ~sep:(O.txt " & ") ~f:kind_annotation
+
+    and with_kind_annotation kind base =
+      match kind with
+      | Odoc_model.Lang.KindAnnotation.Default -> base
+      | k -> O.txt "(" ++ base ++ O.txt " : " ++ kind_annotation k ++ O.txt ")"
```
panglesd: Probably it would make sense here to add some boxes and break hints. I don't know how pervasive are the multiplicity of kinds.

For comparison, `oxcamlformat` turns

```ocaml
type t : global aliased many contended portable forkable unyielding immutable stateless external_
```

into

```ocaml
type t :
  value
  mod
     aliased
     contended
     external_
     forkable
     global
     immutable
     many
     portable
     stateless
     unyielding
```

and it turns

```ocaml
type t : value mod global aliased many contended portable forkable unyielding immutable stateless external_ & value mod global aliased many contended portable forkable unyielding immutable stateless external_
```

into

```ocaml
type t :
  ((value
     mod
        aliased
        contended
        external_
        forkable
        global
        immutable
        many
        portable
        stateless
        unyielding)
   & value)
  mod
     aliased
     contended
     external_
     forkable
     global
     immutable
     many
     portable
     stateless
     unyielding
```

(which shows also that parenthesis are needed, as this is different from:

```ocaml
type t : (value mod global aliased many contended portable forkable unyielding immutable stateless external_) & (value mod global aliased many contended portable forkable unyielding immutable stateless external_)
```

which is formatted as

```ocaml
type t :
  (value
    mod
       aliased
       contended
       external_
       forkable
       global
       immutable
       many
       portable
       stateless
       unyielding)
  & (value
      mod
         aliased
         contended
         external_
         forkable
         global
         immutable
         many
         portable
         stateless
         unyielding)
```
)

Could you include those tests first? And then we decide if we need to fix the formatting in this PR.

src/document/generator.ml
```diff
@@ -426,6 +430,30 @@ module Make (Syntax : SYNTAX) = struct
...
+          kind_annotation base ++ O.txt " " ++ O.keyword "mod"
+          ++ O.txt (" " ^ String.concat ~sep:" " modes)
+      | With (base, ty, modalities) -> (
+          kind_annotation base ++ O.txt " " ++ O.keyword "with" ++ O.txt " "
+          ++ type_expr ty
+          ++
+          match modalities with
+          | [] -> O.noop
+          | mods ->
+              O.txt " " ++ O.keyword "mod"
```
lpw25: Drive by comment: I think this should be `@@` rather than `mod`. See [this page](https://oxcaml.org/documentation/kinds/syntax/) for a grammar for the syntax.

src/document/generator.ml
art-w: Thanks! I've fixed and added tests for the parens :) (not quite the examples you suggested, to avoid the compiler automatic simplifications)

src/document/generator.ml
art-w: Ha thanks! Fixed and added a test

art-w:
Thanks everyone! I've rebased and updated for OxCaml minus31, with some new tests and fixes following the reviews :)

@panglesd > The `Pjk_kind_of` is a new syntax which isn't supported atm (example at https://github.com/oxcaml/oxcaml/blob/main/testsuite/tests/parsetree/source_jane_street.ml#L1375 ), so I can't easily test it. While upgrading to minus31, I also noticed a new syntax to declare a kind abbreviation `kind_ my_abbrev = value mod etc` to be used later `type t : my_abbrev` (I added a test which shows the type declaration renders correctly, but that the `kind_` declaration is missing and that there's no link from the abbreviation to its declaration => I would prefer to follow-up on this in a separate PR if there's no objection)

jonludlam:
mdx tests failed. I'm very doubtful they provide any value any more - they're very old and have largely been superseded by newer tests. I'd be happy for them to be removed.

src/document/generator.ml
```diff
@@ -427,6 +434,45 @@ module Make (Syntax : SYNTAX) = struct
...
+          let res =
+            kind_annotation ~needs_parentheses:true base
+            ++ O.txt " " ++ O.keyword "with" ++ O.txt " " ++ type_expr ty
+            ++
+            match modalities with
+            | [] -> O.noop
+            | mods -> O.txt " @@ " ++ O.txt (String.concat ~sep:" " mods)
+          in
+          enclose_parens_if_needed res
+      | Kind_of ty -> O.keyword "kind_of_" ++ O.txt " " ++ type_expr ty
```
panglesd: Even though `kind_of` is not supported, this kinds of requires to be wrapped in `enclose_parens_if_needed`

panglesd (commented):
Looks good to me!

I'm happy to have the kind abbrev in another PR.

Kinds do not have any "format breaks", but this can also be fixed in another PR (especially, I don't know if it is a big deal, how often are there so many kinds it's better to break, but I hope not too often)

I left a nitpick on an unimplemented feature.

panglesd:
And I agree with the removal of mdx tests. I've never had a bug found by an mdx test not covered by a cram test. It could be in this PR or another.

panglesd (approved):

src/document/generator.ml
art-w: Done!

art-w:
I've fixed the mdx tests here, and submitted https://github.com/ocaml/odoc/pull/1437 to get rid of them in the future :)

jonludlam:
Thanks!

PR #1420, opened 2026-04-17
OxCaml: Support for modalities

### LINKED ISSUE #1416: Support OxCaml modalities
The modalities are not shown at all:

```ocaml
type 'a portended = Portend of 'a @@ portable contended

type 'a t = { shared : 'a @@ shared }
```

Specified by @art-w

art-w:
Builds on top of https://github.com/ocaml/odoc/pull/1410, only the last two commits are new (one commit to test, one to fix)

This PR adds support for [OxCaml modalities](https://oxcaml.org/documentation/modes/syntax/#:~:text=Modalities), fixes https://github.com/ocaml/odoc/issues/1416

### DIFF
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -309,8 +309,7 @@ module Make (Syntax : SYNTAX) = struct
 
     val kind_annotation : Odoc_model.Lang.KindAnnotation.t -> text
 
-    val with_kind_annotation :
-      Odoc_model.Lang.KindAnnotation.t -> text -> text
+    val with_kind_annotation : Odoc_model.Lang.KindAnnotation.t -> text -> text
   end = struct
     let rec te_variant (t : Odoc_model.Lang.TypeExpr.Polymorphic_variant.t) =
       let style_arguments ~constant arguments =
@@ -453,8 +452,7 @@ module Make (Syntax : SYNTAX) = struct
     and with_kind_annotation kind base =
       match kind with
       | Odoc_model.Lang.KindAnnotation.Default -> base
-      | k ->
-          O.txt "(" ++ base ++ O.txt " : " ++ kind_annotation k ++ O.txt ")"
+      | k -> O.txt "(" ++ base ++ O.txt " : " ++ kind_annotation k ++ O.txt ")"
 
     and type_expr ?(needs_parentheses = false) (t : Odoc_model.Lang.TypeExpr.t)
         =
@@ -880,9 +878,12 @@ module Make (Syntax : SYNTAX) = struct
           let base = format_param_str x |> Syntax.Type.handle_format_params in
           Type_expression.with_kind_annotation x.kind (O.txt base)
       | lst ->
-          let l = match delim with `parens -> "(" | `brackets -> "[" in
-          let r = match delim with `parens -> ")" | `brackets -> "]" in
-          O.txt l ++ O.list lst ~sep:(O.txt ", ") ~f:format_param ++ O.txt r
+          let left, right =
+            match delim with `parens -> ("(", ")") | `brackets -> ("[", "]")
+          in
+          O.txt left
+          ++ O.list lst ~sep:(O.txt ", ") ~f:format_param
+          ++ O.txt right
 
     let format_constraints constraints =
       O.list constraints ~f:(fun (t1, t2) ->
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -306,6 +306,11 @@ module Make (Syntax : SYNTAX) = struct
 
     val format_type_path :
       delim:[ `parens | `brackets ] -> Lang.TypeExpr.t list -> text -> text
+
+    val kind_annotation : Odoc_model.Lang.KindAnnotation.t -> text
+
+    val with_kind_annotation :
+      Odoc_model.Lang.KindAnnotation.t -> text -> text
   end = struct
     let rec te_variant (t : Odoc_model.Lang.TypeExpr.Polymorphic_variant.t) =
       let style_arguments ~constant arguments =
@@ -426,6 +431,31 @@ module Make (Syntax : SYNTAX) = struct
       then enclose ~l:lparen res ~r:")"
       else res
 
+    and kind_annotation (k : Odoc_model.Lang.KindAnnotation.t) =
+      match k with
+      | Default -> O.noop
+      | Abbreviation s -> O.txt s
+      | Mod (base, modes) ->
+          kind_annotation base ++ O.txt " " ++ O.keyword "mod"
+          ++ O.txt (" " ^ String.concat ~sep:" " modes)
+      | With (base, ty, modalities) -> (
+          kind_annotation base ++ O.txt " " ++ O.keyword "with" ++ O.txt " "
+          ++ type_expr ty
+          ++
+          match modalities with
+          | [] -> O.noop
+          | mods ->
+              O.txt " " ++ O.keyword "mod"
+              ++ O.txt (" " ^ String.concat ~sep:" " mods))
+      | Kind_of ty -> O.keyword "kind_of_" ++ O.txt " " ++ type_expr ty
+      | Product ks -> O.list ks ~sep:(O.txt " & ") ~f:kind_annotation
+
+    and with_kind_annotation kind base =
+      match kind with
+      | Odoc_model.Lang.KindAnnotation.Default -> base
+      | k ->
+          O.txt "(" ++ base ++ O.txt " : " ++ kind_annotation k ++ O.txt ")"
+
     and type_expr ?(needs_parentheses = false) (t : Odoc_model.Lang.TypeExpr.t)
         =
       let enclose_parens_if_needed res =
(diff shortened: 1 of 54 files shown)


src/document/generator.ml
```diff
@@ -567,23 +567,29 @@ module Make (Syntax : SYNTAX) = struct
       text * bool
 
     val format_constraints : (Lang.TypeExpr.t * Lang.TypeExpr.t) list -> text
+
+    val format_modalities : Odoc_model.Lang.Modalities.t -> text
   end = struct
+    let format_modalities modalities =
```
panglesd: As mentioned in https://github.com/ocaml/odoc/pull/1410/changes/BASE..d0a2d4e0dc3abd4bccec3cb8ac8da0e87da8c6fe#r3131327149, the display of modalities in `kind_annotation` should use this function!

art-w: Thanks, refactored :)

src/loader/cmt.ml
```diff
@@ -53,7 +53,7 @@ let rec read_pattern env parent doc pat =
           Cmi.mark_type_expr pat.pat_type;
           let type_ = Cmi.read_type_expr env pat.pat_type in
           let value = Abstract in
-          [Value {id; source_loc; doc; type_; value}]
+          [Value {id; source_loc; doc; type_; value; modalities = []}]
```
panglesd: Are modalities ignored on cmt loading on purpose?

src/loader/cmti.ml
```diff
@@ -645,7 +666,17 @@ let rec read_with_constraint env global_parent parent (_, frag, constr) =
 and read_module_type env parent label_parent mty =
   let open ModuleType in
     match mty.mty_desc with
-    | Tmty_ident(p, _) -> Path { p_path = Env.Path.read_module_type env.ident_env p; p_expansion = None }
+    | Tmty_ident(p, _) ->
+        (match mty.mty_type with
+#if defined OXCAML
+         | Mty_signature sg ->
+             (* For modules with modalities (e.g. [module M : S @@ portable]),
```
panglesd, line 670: Is it what the docs describes as (they use the "modes" word but it is also in the "modalities" section):

> Support for modules with modes is being worked on and not ready for wide adoption. More documentation will come as it becomes ready.

If so, I'm fine with the current code being merged, but I would need to update/re-review when there are more documentation (eg, can we add modalities on functor module types?)

src/loader/cmi.mli
```diff
@@ -101,4 +101,13 @@ val read_exception : env ->
...
   Odoc_model.Lang.Kind.t
+
+val read_modalities :
+  Types.mutability ->
+  Mode.Modality.Const.t ->
+  Odoc_model.Lang.Modalities.t
+
+val read_value_modalities :
+  Mode.Modality.t ->
+  Odoc_model.Lang.Modalities.t
```
panglesd, line 112: I think we would avoid quite a lot of `#if defined OXCAML` if we had `read_value_modalities`, `read_field_modalities`, etc., taking not `Mode.Modality.t` but the ast nodes above.

The "non-oxcaml" version of these function would exist and return `[]`.

(This is nitpicky! But I tend to try to avoid having cppo in too many places, I find it more readable and I can ocamlformat individual structure items. Feel free to ignore if you want, it is really good-enough!)

test/generators/cases/oxcaml.mli
```diff
@@ -119,3 +119,107 @@ kind_ my_abbrev = value_or_null mod non_null global
...
+
+type modalities_cstr =
+  | A of string @@ global
+      (** Constructor argument with [global] modality. *)
+  | B of (int -> int) @@ portable
+      (** Function constructor argument with modality. *)
+  | C of int * string @@ portable
+      (** Tuple constructor argument with modality. *)
+  | D of int @@ portable * string @@ global
+      (** Per-element modalities in a constructor tuple. *)
```
panglesd, line 227: For completeness it could be nice to add an inline record `D' of {a : int @@ portable; b : string @@ global }`

test/generators/cases/oxcaml.mli
```diff
@@ -119,3 +119,107 @@ kind_ my_abbrev = value_or_null mod non_null global
...
+
+type modalities_all = {
+  f_global : opaque @@ global;
+      (** Locality modality. *)
+  f_local : opaque @@ local;
+      (** Locality modality (identity, not rendered). *)
+  f_unique : opaque @@ unique;
+      (** Uniqueness modality. *)
+  f_aliased : opaque @@ aliased;
+      (** Uniqueness modality (identity, not rendered). *)
```
panglesd: In my test, that is the other way around: `unique` is not rendered and `aliased` is rendered. What is to blame? The comment or the output of the test?

test/generators/cases/oxcaml.mli
```diff
@@ -119,3 +119,107 @@ kind_ my_abbrev = value_or_null mod non_null global
...
+}
+
+type modalities_fn = {
+  g : int -> int @@ portable;
+      (** Function field with modality. *)
+}
+
+(** {1 Modalities on constructor arguments} *)
+
+type modalities_cstr =
```
panglesd, line 219: I think it would be worth adding a GADT to the test!

test/generators/cases/oxcaml.mli
```diff
@@ -119,3 +119,107 @@ kind_ my_abbrev = value_or_null mod non_null global
...
+      (** Statefulness modality (identity, not rendered). *)
+  f_immutable : opaque @@ immutable;
+      (** Visibility modality. *)
+  f_read_write : opaque @@ read_write;
+      (** Visibility modality (identity, not rendered). *)
+  f_no_modality : opaque;
+      (** No modality, for reference. *)
+}
+
+(** {1 Multiple modalities on a field} *)
```
panglesd, line 198: It would be nice if the subheadings were of level 2 (this file keeps growing and it would improve navigation in the browser!)

test/generators/cases/oxcaml.mli
```diff
@@ -119,3 +119,107 @@ kind_ my_abbrev = value_or_null mod non_null global
...
+
+module M2 : S @@ portable
+(** Module with [portable] modality. The modality is applied to
+    all value members of [M2]. *)
+
+module M3 : sig @@ contended
+  val f : string -> bool
+  type s
+end
+(** [contended] modality applied to all definitions in the module. *)
```
panglesd: Maybe worth adding `val` that also sets some modalities (on the contendedness axis and/or on another axis)

panglesd (commented):
Looks good! I just want confirmation on:
- not loading modalities on cmt files
- the discrepancy between the input of the test and the output (see the comment there)

The other comments are less important.

src/loader/cmt.ml
art-w: As far as I understand, here we don't have modalities but modes for these values (there's another PR coming soon for modes). See [this documentation comment in `types.mli`](https://github.com/oxcaml/oxcaml/blob/main/typing/types.mli#L1176) and [an example where modalities are dropped for pattern values](https://github.com/oxcaml/oxcaml/blob/main/typing/typecore.ml#L1460) in the oxcaml compiler :)

src/loader/cmti.ml
art-w: Agree, I ran into this syntax accidentally when searching for examples (and it seemed useful, but my understanding is limited to what I was able to test)

test/generators/cases/oxcaml.mli
art-w: Sure, done!

src/loader/cmi.mli
art-w: Agree, it's a bit more readable now :)

test/generators/cases/oxcaml.mli
art-w: Oh good catch, thanks!

test/generators/cases/oxcaml.mli
art-w: Done!

test/generators/cases/oxcaml.mli
art-w: Agree :)

test/generators/cases/oxcaml.mli
art-w: I've updated the test, thanks :)

art-w:
Thanks for the review! The PR is now rebased and I've updated the tests and code as advised :) (I left some comments open in case there are follow-up questions)

test/xref2/classes.t/run.t
```diff
@@ -51,7 +51,8 @@ resolve correctly. All of the 'Class' json objects should contain
         []
       ]
     },
-    "value": "Abstract"
+    "value": "Abstract",
```
jonludlam, line 54: It's a bit annoying that these tests had to be changed - we should fix the tests to be a big more targetted in their output (in a separate PR!)

src/model/lang.ml
```diff
@@ -212,13 +212,19 @@ end =
...
+  Modalities
+
 and TypeDecl : sig
   module Field : sig
     type t = {
       id : Identifier.Field.t;
       doc : Comment.docs;
       mutable_ : bool;
       type_ : TypeExpr.t;
+      modalities : Modalities.t;
```
jonludlam, line 227: We could just use Lang.Modalities here - there's no reason to redefine it. We use that pattern in a few other places, like `TypeExpr.Polymorphic_variant.kind`, `TypeExpr.label` and so on.

jonludlam: Oops ignore me :-)

jonludlam:
Looks good! Thanks @art-w

PR #1422, opened 2026-04-28
Zero alloc annotation for OxCaml

Leonidas-from-XIV:
This PR aims to take the [zero alloc](https://oxcaml.org/documentation/miscellaneous-extensions/zero_alloc_check/) annotation from the source and retain it in the parsed output.

Currently this PR supports the `[@@zero_alloc]` annotation on signatures. Next step would be to support the `[@zero_alloc]` annotation on bindings; not sure if it would make sense to add it in this PR or another one.

This is my first PR to Odoc so excuse me for being a bit lost, I'll happily fix up/change stuff.

cc @art-w


src/document/generator.ml
```diff
@@ -988,13 +988,19 @@ module Make (Syntax : SYNTAX) = struct
...
+        | false -> O.noop
+      in
       let content =
         O.documentedSrc
           (O.box_hv
           @@ O.keyword Syntax.Value.variable_keyword
              ++ O.txt " " ++ O.txt name
              ++ O.txt Syntax.Type.annotation_separator
              ++ O.cut ++ type_expr t.type_
+             ++ O.cut ++ zero_alloc
```
art-w: nitpick: Should the `O.cut` be on line 993 instead, to avoid a cut before the following semicolon?

src/loader/doc_attr.ml
```diff
@@ -74,6 +74,12 @@ let attribute_unpack = function
   | { Location.txt = name; loc }, attr_payload -> (name, attr_payload, loc)
 #endif
 
+let known_attribute attr =
+  let name, _, _ = attribute_unpack attr in
+  match String.equal name "zero_alloc" with
+  | true -> Some Lang.Value.Zero_alloc
+  | false -> None
```
art-w: The documentation mentions that it's possible to add some option to the zero_alloc, e.g. `[@@zero_alloc strict]` or `assume` or `opt` etc. This doesn't look critical since strict/assume shouldn't be relevant for end-users consuming the documentation, but maybe `opt` is important?

art-w (approved):
It looks good! I have a minor nitpick, mostly to be safe because I don't know if the additional `O.cut` could break existing docs. Do you prefer addressing the remaining TODOs for the cmti in a separate PR?

src/loader/doc_attr.ml
Leonidas-from-XIV: Good idea! I've looked into how the OxCaml toplevel prints it and it looks like this:

| Implementation | Inferred interface        |
|------------------------|-------------------------|
| `[@zero_alloc assume]` | `[@@zero_alloc]`        |
| `[@zero_alloc strict]` | `[@@zero_alloc strict]` |
| `[@zero_alloc opt]`    | `[@@zero_alloc opt]`    |

I will add these.

src/model/lang.ml
```diff
@@ -321,12 +321,18 @@ end =
 and Value : sig
   type value = Abstract | External of string list
 
+  module Zero_alloc : sig
+    type t = Assume | Opt | Strict
+  end
+  type attr = Zero_alloc of Zero_alloc.t
```
panglesd, line 334: According to the [docs](https://oxcaml.org/documentation/miscellaneous-extensions/zero_alloc_check/), the only allowed payloads are `opt`, `strict` and `arity <n>`.

If I understand correctly "`assume`" does not correspond to the default. (Assume in a module type would be: do not check that the implementation do not allocate, but use that information in uses outside.)

Moreover, it seems possible to use multiple (valid) payloads.

So I think we have two options in terms of type representation: the "more correct" one

```suggestion
  module Zero_alloc : sig
    type t = {opt : unit option ; strict: unit option; arity: int option }
  end
  type attr = Zero_alloc of Zero_alloc.t
```
or the one that uses the fact that it has already been validated by the oxcaml compiler:
```suggestion
  module Zero_alloc : sig
    type t = Arity of n | Opt | Strict
  end
  type attr = Zero_alloc of Zero_alloc.t
```
Both are fine to me! And the second one is probably less work.

Here are some examples of zero_alloc uses in module types to "validate" the oxcaml doc:
```ocaml
# module type T = sig val[@zero_alloc] f : int -> int -> int end ;;
module type T = sig val f : int -> int -> int [@@zero_alloc] end

(** The error says the accepted attributes {e for implementation}, not interface, unfortunately *)

# module type T = sig val[@zero_alloc gloubli-boulga] f : int -> int -> int end ;;
Warning 47 [attribute-payload]: illegal payload for attribute 'zero_alloc'.
It must be either 'assume', 'assume_unless_opt', 'strict', 'opt', 'opt strict', 'assume strict', 'assume never_returns_normally', 'assume never_returns_normally strict', 'assume error', 'ignore', 'arity <int_constant>', 'custom_error_message <string_constant>' or empty

(** assume is not allowed *)

# module type T = sig val[@zero_alloc assume] f : int -> int -> int end ;;
Error: zero_alloc assume attributes are not supported in signatures

(** strict, opt, and arity <n> (and combinations) are allowed *)

# module type T = sig val[@zero_alloc strict] f : int -> int -> int end ;;
module type T = sig val f : int -> int -> int [@@zero_alloc strict] end
# module type T = sig val[@zero_alloc opt] f : int -> int -> int end ;;
module type T = sig val f : int -> int -> int [@@zero_alloc opt] end
# module type T = sig val[@zero_alloc arity 1] f : int -> int -> int end ;;
module type T = sig val f : int -> int -> int [@@zero_alloc arity 1] end
# module type T = sig val[@zero_alloc strict opt] f : int -> int -> int end ;;
module type T = sig val f : int -> int -> int [@@zero_alloc strict opt] end
# module type T = sig val[@zero_alloc strict arity 2] f : int -> int -> int end ;;
module type T = sig val f : int -> int -> int [@@zero_alloc strict] end
# module type T = sig val[@zero_alloc arity 1 opt strict] f : int -> int -> int end ;;
module type T =
  sig val f : int -> int -> int [@@zero_alloc strict opt arity 1] end
# module type T = sig val[@zero_alloc] f : int -> int -> int end ;;
module type T = sig val f : int -> int -> int [@@zero_alloc] end
# module type T = sig val[@zero_alloc] f : int -> int -> int end ;;
module type T = sig val f : int -> int -> int [@@zero_alloc] end
```

src/loader/doc_attr.ml
```diff
@@ -74,6 +74,42 @@ let attribute_unpack = function
   | { Location.txt = name; loc }, attr_payload -> (name, attr_payload, loc)
 #endif
 
+let ident (pexp_desc : Parsetree.expression_desc) =
```
panglesd: Extreme nitpick: I find the name slightly too generic. Maybe "`ident_of_expression`"?

src/loader/doc_attr.ml
```diff
@@ -74,6 +74,42 @@ let attribute_unpack = function
...
+let zero_alloc_argument (payload : Parsetree.payload) =
+  match payload with
+  | PStr structure_items -> (
+    match structure_items with
+    | [] -> Some Lang.Value.Zero_alloc.Assume
+    | [{pstr_desc; _}] -> zero_alloc_structure_arg pstr_desc
+    | _ -> None)
+  | _ -> None
+
+let known_attribute attr =
```
panglesd: Could this function always return `None` outside of oxcaml? If the annotation has no effect in OCaml, let's not print it!

panglesd (commented):
Thanks!

I've left a nitpick, a comment about payload support and one about not loading attributes in O~x~Caml.

Some small comments that probably require no change in this PR:
- In the oxcaml doc, they use `val[@zero_alloc] f : t` rather than `val f : t [@@zero_alloc]` as in this PR (they are different syntax for the same thing). So maybe oxcaml users are more familiar with the former? But keeping as is is fine to me!
- There is already some logic to extract special attributes (eg `@alert` and `@deprecated`). However, they are rendered differently (using odoc `@`-tags). It's a pity to have two ways to handle attributes, but for `zero_alloc` (and other oxcaml-specific attributes to come), I think I prefer your approach.

Leonidas-from-XIV:
> In the oxcaml doc, they use `val[@zero_alloc] f : t` rather than `val f : t [@@zero_alloc]` as in this PR (they are different syntax for the same thing). So maybe oxcaml users are more familiar with the former? But keeping as is is fine to me!

Very good point. I mostly did it this way as this is how the oxcaml toplevel prints it. So I assumed this is the way to go, but I will at least add a test-case to make sure that the `val[@zero_alloc]` variant is supported correctly.

art-w:
> In the oxcaml doc, they use val[@zero_alloc] f : t rather than val f : t [@@zero_alloc] as in this PR (they are different syntax for the same thing).

We received feedback that the later form looks nicer for documentation :)

src/loader/doc_attr.ml
Leonidas-from-XIV: Yes! I also moved it to a `Zero_alloc` module as it's not particularly generic and only really useful in the context of parsing `@zero_alloc`, so now it doesn't spam up the namespace. It could be made more generic in case we will need something similar in the future but for now I think it's ok to keep it limited in what it can do to what we need for `@zero_alloc`.

Leonidas-from-XIV:
I've rewritten the code to parse the arguments in `[@@zero_alloc]`, with better names and better self-contained. My testing has revealed that `ignore`, `opt`, `strict`, `arity` and `custom_error_message` are valid in signatures.

However there are cases where the parser silently ignores errors (e.g. "arity" used but no argument given), in that case I think it would be nice to do something. Probably exiting is too extreme so I didn't want to throw but it would be nice to emit a warning that something went wrong?

Or maybe its ok to silently continue because the OxCaml compiler will complain anyway? WDYT?

panglesd:
> Or maybe its ok to silently continue because the OxCaml compiler will complain anyway? WDYT?

I would say it is ok to silently continue, as I would say those warnings are more the responsibility of the compiler.

But that made me think... Isn't this information already available in the compiler artefact? I have quickly checked in `types.mli`, and I have found something that [looks like it](https://github.com/oxcaml/oxcaml/blob/main/typing/types.mli#L1159), although I'm not sure (it is in a `Wrapped` functor that's instantiated and inlined :face_with_diagonal_mouth: ).

If we can avoid the parsing ourself, that would be much better!

panglesd:
After setting up an OxCaml environment :sweat_smile: I was able to find that yes, the information is already there!

In cmi files, the "zero alloc" info is stored at `vd.val_zero_alloc`, and in cmti files, it is stored in `vd.val_val.val_zero_alloc`. For instance:

```ocaml
    let za = vd.val_val.val_zero_alloc |> Zero_alloc.get in
    let za = match za with
      | Default_zero_alloc -> _
      | Ignore_assert_all -> _
      | Check _ -> _
      | Assume _ -> _
    in
    Value.Zero_alloc za
```

using this info to create a lang, rather than reparse it from the attribute, seems better! And would remove all the parsing code.

Leonidas-from-XIV:
@panglesd Thanks for the pointers, they were very useful. I've implemented your changes, adapting the value type to be more aligned with the oxcaml compiler-libs version of it. However there's two concerns:

1. The compiler-libs return the "computed" arity, not the specified one. So I have no way (bar checking the annotation manually) to distinguish whether an arity value was passed or not. So for now we always print the arity. This doesn't seem to be a problem per-se, but it is a bit messy. Maybe I could read the type and determine whether the arities match and omit it in such case?
2. In the CMT case I don't always have a `value_description`. Maybe also not a problem, because the CMT loader calls `Cmti.read_value_description` which does the right thing.

Leonidas-from-XIV:
@panglesd Could you take another look and share your thoughts? I've rebased the branch on `master` which works on a minus31 OxCaml switch.

src/document/generator.ml
```diff
@@ -989,13 +989,36 @@ module Make (Syntax : SYNTAX) = struct
...
+            (function Odoc_model.Lang.Value.Zero_alloc _ -> true)
+            t.ext_attr
+        with
+        | exception Not_found -> O.noop
+        | Zero_alloc { opt; strict; arity; custom_error_msg } ->
+            let ext_arg =
+              match (opt, strict) with
+              | true, false -> " opt"
+              | false, true -> " strict"
+              | _, _ -> ""
```
jonludlam: Is `true, true` invalid, and `false, false` the only thing that should actually result in `""`?

src/xref2/lang_of.ml
```diff
@@ -694,6 +694,8 @@ and value_ map parent id v =
     doc = docs (parent :> Identifier.LabelParent.t) v.doc;
     type_ = type_expr map (parent :> Identifier.LabelParent.t) v.type_;
     value = v.value;
+    (* TODO this needs to be fixed *)
```
jonludlam: Without mapping between lang types and component types we'll lose all these attributes in the wrapped library (ie, default!) case.

panglesd:
> In the CMT case I don't always have a value_description. Maybe also not a problem, because the CMT loader calls Cmti.read_value_description which does the right thing.

Hmm, I think it is a problem, as I don't see value cmt loading calling into `Cmti.read_value_description`. However, after a ~~quick~~ slow look at oxcaml's typedtree, I think you can use [`Typedtree.let_bound_idents_with_modes_sorts_and_checks`](https://github.com/oxcaml/oxcaml/blob/main/typing/typedtree.mli#L1505) to get the `ident -> Zero_alloc.t` map.

src/xref2/lang_of.ml
panglesd: Indeed!

What you need to do is:
- Add `extr_attrs` field in the `Value.t` type in (`xref2/components.ml{i}`),
- Handle the conversion from "model" to component (same file, `Of_lang` module): just propagate the value.
- Handle the conversion from component to model (the TODO that was left).

Component are used as an intermediate types when manipulating modules (eg when doing `include M`). Without this, the zero alloc info would be lost whenever such conversion happen, eg in wrapped modules.

test/generators/html/Oxcaml.html
```diff
@@ -14,26 +14,7 @@
   <header class="odoc-preamble">
    <h1>Module <code><span>Oxcaml</span></code></h1>
   </header>
-  <div class="odoc-content">
-   <div class="odoc-spec">
-    <div class="spec value anchored" id="val-f">
-     <a href="#val-f" class="anchor"></a>
-     <code>
-      <span><span class="keyword">val</span> f :
```
panglesd, line 59: Why is this removed from the output of the test? `val f` is still present in the input!

Leonidas-from-XIV: This looks an issue when rebasing where `master` added `f` and I also added `f` later in the file and upon rebasing my `f` shadowed the former. Thanks for pointing out, I've renamed my identifier so its not shadowing anymore.

src/loader/cmt.ml
```diff
@@ -70,49 +75,52 @@ let rec read_pattern env parent doc pat =
           Cmi.mark_type_expr pat.pat_type;
           let type_ = Cmi.read_type_expr env pat.pat_type in
           let value = Abstract in
-          let ext_attr = [] in
-          Value {id; source_loc; doc; type_; value; ext_attr} :: read_pattern env parent doc pat
+          let ext_attr = [vb]
+            |> let_bound_idents_with_modes_sorts_and_checks
```
panglesd: I think this is not how `let_bound_idents_with_modes_sorts_and_checks` is supposed to be used!

It creates an associative list between idents, and zero_alloc values (among other data). Here you use the whole associative list for every ident, ignoring the keys!

(A single value binding can define multiple ident, as for instance in:

```ocaml
let x,y = 5, "five"
```
)

I think, it should be done this way:
- The associative list is computed in `read_value_bindings` from the value binding list, and passed until `read_pattern`
- In `read_pattern`, the zero_alloc value is got by finding the zero_alloc value corresponding to the ident in the associative list.

Leonidas-from-XIV: Ah I see, thanks. I wasn't sure what exactly a `value_binding`, I should probably add a test for this as well to make sure it works when binding multiple values and only one of them is `zero_alloc`. Will update the code!

src/document/generator.ml
Leonidas-from-XIV: I checked in the Toplevel and accepts `opt` and `strict` at the same time in any order and normalizes it to `strict opt`. I've done the same here now.

src/xref2/lang_of.ml
Leonidas-from-XIV: I have added a test to show the issue and implemented your suggested fix which indeed also fixed the tests. Thanks!

src/loader/doc_attr.ml
```diff
@@ -74,6 +74,29 @@ let attribute_unpack = function
...
+
+let attrs_of_value_description (vd : Types.value_description) =
+#if defined OXCAML
+  let zero_alloc = lang_value_attr_of_zero_alloc vd.val_zero_alloc in
+#else
+  let zero_alloc = None in
+#endif
+  match zero_alloc with
+  | Some za -> [za]
+  | None -> []
```
panglesd: I tend to dislike cppo in the middle of a structure item. I find it harder to read and to format. In this case, I think it should be split! (and have a single "`#if defined OXCAML`" block.)

```suggestion
#if defined OXCAML
let attrs_of_value_description (vd : Types.value_description) =
  let zero_alloc = lang_value_attr_of_zero_alloc vd.val_zero_alloc in
  match zero_alloc with
  | Some za -> [za]
  | None -> []
#else
let attrs_of_value_description (vd : Types.value_description) = []
#endif
```

src/loader/cmt.ml
```diff
@@ -119,21 +121,36 @@ let rec read_pattern env parent doc pat =
...
   let container = (parent : Identifier.Signature.t :> Identifier.LabelParent.t) in
+  let id_attrs =
+#if defined OXCAML
+    vbs |> let_bound_idents_with_modes_sorts_and_checks |> List.fold_left (fun tbl (ident, _, zero_alloc) ->
+      match Doc_attr.lang_value_attr_of_zero_alloc zero_alloc with
+      | None -> tbl
+      | Some attr -> Ident.add ident [attr] tbl) Ident.empty
+#else
+    Ident.empty
+#endif
```
panglesd: This could be in `doc_attr`.

panglesd (approved):
Thanks a lot for the work!

I think the only thing that would need feedback from an actual user of oxcaml, is the added arity: Should we, as you suggested, omit it when it corresponds to the visible arity of the function?

Probably, but I would be more confident if I had experience using the `zero_alloc` tag myself!

Anyway, that can be in another PR: this one is already good as is. I left some small comment if you have time to do them, and then we can merge.

src/loader/cmt.ml
Leonidas-from-XIV: It's a good point; it also obsoletes the need to expose `lang_value_attr_of_zero_alloc` thus I can remove CPPO from `doc_attr.mli`

src/loader/doc_attr.ml
Leonidas-from-XIV: I see your point. I was somewhat thinking that this function could potentially add other attributes in the future, but I think we can revisit that if and when we get there. For now there's just zero alloc and your code is indeed nicer to read.

jonludlam:
Thanks @Leonidas-from-XIV !

PR #1423, opened 2026-04-29
Compatibility with OxCaml-minus31

art-w:
The latest release of OxCaml introduced some breaking changes to the ASTs, which @jonludlam fixed in the following commits :) Note that this PR only intend to fix the compilation of Odoc with OxCaml, as the real support for the new features is being done in separate PRs (so in the mean time we just ignore the new AST nodes).

Requires https://github.com/oxcaml/opam-repository/pull/39 and https://github.com/oxcaml/opam-repository/pull/42 to fix Odoc's dependencies on dune and opam-format

### DIFF
--- a/sherlodoc/jsoo/tyxml.ml
+++ b/sherlodoc/jsoo/tyxml.ml
@@ -47,7 +47,8 @@ end = struct
       let next = i + 1 in
       loop next next
     in
-    loop 0 0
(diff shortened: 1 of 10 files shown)


jonludlam:
Let's get this in - it certainly improves the situtation from not compiling at all to compiling!

PR #1424, opened 2026-04-29
Add OxCaml implementation file test

Leonidas-from-XIV:
This adds a file that contains just an implementation, without an interface, to show how OxCaml code is rendered by Odoc. Thus it is not `oxcaml.ml` to avoid being overridden by `oxcaml.mli`.

The sample is directly from [the oxcaml documentation](https://oxcaml.org/documentation/miscellaneous-extensions/zero_alloc_check/), to include a very simple OxCaml feature.

It is a separate PR from #1422 as having a test for implementation files might be useful outside of #1422 as well, as discussed with @art-w.

### DIFF
--- a/test/generators/cases/oxcaml_impl.ml
+++ b/test/generators/cases/oxcaml_impl.ml
@@ -1,0 +1,1 @@
+let[@zero_alloc] add b x y = if b then x + y else x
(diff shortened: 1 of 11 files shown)


art-w (approved):
Thanks!

panglesd (approved):
Thanks! And sorry for the delay. I'll merge as soon as the CI is green (I think you just rebased).

Issue #1426, opened 2026-05-02
Error when building docs for `base` with OCaml 5.5


jonludlam:
Repro: Install OCaml 5.5 beta1 and Odoc 3.2.0, then run `dune build @doc -p base` in base.v0.17.3.  I get:

```
(cd _build/default/src/.base.objs/byte && /Users/jon/.opam/5.5.0~beta1/bin/odoc compile -I . -I ../../../_doc/_odoc/pkg/base -I ../../../hash_types/src/.base_internalhash_types.objs/byte -I ../../../shadow-stdlib/src/.shadow_stdlib.objs/byte --pkg base -o base__Hashable_intf.odoc base__Hashable_intf.cmt)
Exception Not_found handling type_expr: ((module_arg)) -> [resolved((param root(module_arg_1_) Key).t)] resolved(t/386)
backtrace:
Raised at Stdlib__List.assoc in file "list.ml", line 201, characters 10-25
Called from Odoc_xref2__Lang_of.type_expr_module_arg in file "src/xref2/lang_of.ml", line 1022, characters 11-48
Called from Odoc_xref2__Lang_of.type_expr in file "src/xref2/lang_of.ml", line 1053, characters 16-53

odoc: internal error, uncaught exception:
      Not_found
      Raised at Stdlib__List.assoc in file "list.ml", line 201, characters 10-25
      Called from Odoc_xref2__Lang_of.type_expr_module_arg in file "src/xref2/lang_of.ml", line 1022, characters 11-48
      Called from Odoc_xref2__Lang_of.type_expr in file "src/xref2/lang_of.ml", line 1053, characters 16-53
      Re-raised at Odoc_xref2__Lang_of.type_expr in file "src/xref2/lang_of.ml", line 1061, characters 4-11
      Called from Odoc_xref2__Lang_of.value_ in file "src/xref2/lang_of.ml", line 693, characters 12-70
      Called from Odoc_xref2__Lang_of.signature_items.inner in file "src/xref2/lang_of.ml", line 432, characters 26-50
      Called from Odoc_xref2__Lang_of.signature in file "src/xref2/lang_of.ml", line 460, characters 12-43
      Called from Odoc_xref2__Lang_of.simple_expansion in file "src/xref2/lang_of.ml", line 605, characters 30-51
      Called from Odoc_xref2__Compile.include_.get_expansion in file "src/xref2/compile.ml", line 431, characters 25-71
[...]

PR #1427, opened 2026-05-02
Fix 1426

jonludlam:
#1426  is caused by not keeping track of identifier intoduced in the new `Arrow_functor` type expression. This patch fixes that.


panglesd (approved):
Looks good! Thanks (and sorry for introducing the bug!)

jonludlam:
No problem! I hadn't realised that this new constructor wasn't just for new syntax, so I hadn't bothered checking it on existing libraries. I only noticed because of my new version of ocaml-docs-ci has a profile that tracks odoc master!

Issue #930, opened 2023-02-17
Crash when substituting for the same name at different arities


lukemaurer:
The MWE is highly suggestive of what's going on:

```ocaml
module type S1 = sig
  type t0
  type 'a t := unit

  val x : t0 t
end

module type S2 = sig
  type t (* must be the same name as [S1.t] *)

  include S1 with type t0 := t
end

module type S3 = sig
  type t1

  include S2 with type t := t1
end
```

The result is an `Invalid_argument` out of `List.fold_left2`, caused by a length mismatch between the parameter list and the argument list.

```
$ ocamlc -c test.mli -bin-annot
$ _build/install/default/bin/odoc compile test.cmti
  odoc: internal error, uncaught exception:
      Invalid_argument("List.fold_left2")
      Raised at Stdlib.invalid_arg in file "stdlib.ml", line 33, characters 20-45
      Called from Odoc_xref2__Subst.type_expr in file "subst.ml", line 598, characters 21-59
      Called from Odoc_xref2__Subst.value in file "subst.ml" (inlined), line 842, characters 19-38
      Called from Odoc_xref2__Subst.apply_sig_map.inner.(fun) in file "subst.ml", line 1089, characters 19-52
      Called from Odoc_xref2__Component.Delayed.get in file "component.ml" (inlined), line 55, characters 16-22
      Called from Odoc_xref2__Lang_of.signature_items.inner in file "lang_of.ml", line 438, characters 16-39
      Called from Odoc_xref2__Lang_of.signature in file "lang_of.ml" (inlined), line 466, characters 12-43
      Called from Odoc_xref2__Lang_of.include_ in file "lang_of.ml", line 641, characters 18-69
      ...
```

gpetiot:
I cannot reproduce this with the `master` version, do you still observe this bug with the last odoc build?

lukemaurer:
> ```ocaml
> module type S1 = sig
>   type t0
>   type 'a t := unit
> 
>   val x : t0 t
> end
> 
> module type S2 = sig
>   type t (* must be the same name as [S1.t] *)
> 
>   include S1 with type t0 := t
> end
> 
> module type S3 = sig
>   type t1
> 
>   include S2 with type t := t1
> end
> ```

I do. Notably, I only get it when going through a `.cmti`, though. I'll update with more specific instructions.

lukemaurer:
> I cannot reproduce this with the `master` version, do you still observe this bug with the last odoc build?

[...]

PR #1430, opened 2026-05-06
Fix 1429

jonludlam:
This is a regression, but due to code that's been there a while. We're just exercising it more with the fix for #930.


panglesd (approved):
I haven't investigated enough the issue to say this is the correct fix, or to provide a better one. But I still can say that it is better than before, and should go in!

PR #1432, opened 2026-05-07
Conflict OxCaml < 5.2.0minus31

dra27:


jonludlam:
Shouldn't this only be on the odoc package? I don't think the other packages depend on any particular version of oxcaml.

dra27:
Oh yes, it's a bit more belt and braces than ideal!

Issue #1431, opened 2026-05-07
Failure when building core_unix


jonludlam:
Running `dune build @doc` in core_unix fails with odoc master:

```
File "time_float_unix/time_unix/.time_unix.objs/byte/_unknown_", line 1, characters 0-0:
odoc: internal error, uncaught exception:
      Failure("error")
      Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
      Called from Odoc_xref2__Tools.apply_inner_substs.inner in file "src/xref2/tools.ml", line 2414, characters 24-34
      Called from Odoc_xref2__Tools.apply_inner_substs.inner in file "src/xref2/tools.ml", line 2414, characters 24-34
      Called from Odoc_xref2__Tools.apply_inner_substs in file "src/xref2/tools.ml", line 2417, characters 20-34
      Called from Odoc_xref2__Compile.include_.get_expansion in file "src/xref2/compile.ml", line 430, characters 19-51
      Called from Odoc_xref2__Compile.signature_items.loop in file "src/xref2/compile.ml", line 348, characters 27-41
      Called from Odoc_xref2__Compile.signature in file "src/xref2/compile.ml", line 365, characters 19-49
      Called from Odoc_xref2__Compile.content.(fun) in file "src/xref2/compile.ml", line 100, characters 15-54
      Called from Odoc_xref2__Compile.unit in file "src/xref2/compile.ml", line 68, characters 21-47
      Called from Odoc_xref2__Lookup_failures.with_ref in file "src/xref2/lookup_failures.ml", line 13, characters 10-14
      Called from Odoc_xref2__Lookup_failures.catch_failures in file "src/xref2/lookup_failures.ml", line 60, characters 20-37
      Called from Odoc_odoc__Compile.resolve_and_substitute in file "src/odoc/compile.ml", line 154, characters 4-49
      Called from Odoc_model__Error.catch in file "src/model/error.ml", line 52, characters 21-27
      Called from Odoc_model__Error.catch_warnings.(fun) in file "src/model/error.ml", line 87, characters 18-22
      Called from Odoc_model__Error.with_ref in file "src/model/error.ml", line 65, characters 12-16
      Re-raised at Odoc_model__Error.with_ref in file "src/model/error.ml", line 70, characters 4-11
      Called from Odoc_odoc__Compile.compile.(fun) in file "src/odoc/compile.ml", lines 407-410, characters 6-23
[...]

PR #1433, opened 2026-05-11
Fix 1431

jonludlam:
The fix for #930 introduced a problem where a local module substitution fails to resolve, which
led to an unrecoverable exception being raised. This commit detects the problem and essentially
leaves the signature as it was found. This means that the fix for #930 is _partial_ (and
always was). The behaviour that this commit introduces is better now, but not perfect.

The original #930 problem is due to a clash of identifier caused by an include with a module
substitution. The fix was to turn the module substitution into a `sig ... end with module Foo :=`
expression, calculate the result and use that as the body of the signature, hence no clash.
This was applied universally, so any time an include had a substitution in it, we used this code.

However, if the right-hand side of the module substitution was a local module, as in the test in
this PR, odoc cannot resolve this, and hence raises an error. This commit detects that problem
and reverts the signature to its original form.

Thus the new behaviour means that there's still potential for #930, but now the condition for
a problem is that there's a module substitution being included into a signature that has a module
with the same name (the #930 problem), _and_ that the substitution was for a local module (the
cause of #1431)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
   (@jonludlam, #1427)
 - Fix #1429, which broke docs in packages including `merlin-lib` (@jonludlam,
   #1430)
+- Fix #1431, a regression introduced in v3.2.0 (#1433, @jonludlam)
 
 # 3.2.0
 
--- a/src/xref2/tools.ml
+++ b/src/xref2/tools.ml
@@ -2356,25 +2356,32 @@ let apply_inner_substs env (sg : Component.Signature.t) : Component.Signature.t
   let rec inner (items : Component.Signature.item list) :
       Component.Signature.item list =
     match items with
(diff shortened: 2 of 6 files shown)


panglesd (approved):
Even though this is not a full fix, it is at least strictly better than what we had before, in terms of behaviour.

Slight summary of the description above, since I initially misunderstood it.
We had issue #930. Then, #1400 fixed it, but introduced the other issue, #1431.
This PR restores the pre-#1400 behavior, only whenever #1431 happens. This way, we only have a problem in the "#930 + #1431 at the same time" situation, which is hopefully quite rare.

jonludlam:
gah I was going to say this was 'no changelog' but I realised there _is_ a need as 3.2.0 is affected by this.

PR #1434, opened 2026-05-11
voodoo: transitive-close lib_deps before per-compile -I emission

jonludlam:
We've been missing some include paths when building docs for some packages. This doesn't affect other modes of odoc_driver as they go through `Packages.fix_missing_deps`.

### DIFF
--- a/src/driver/voodoo.ml
+++ b/src/driver/voodoo.ml
@@ -104,6 +104,29 @@ let of_voodoo pkg =
       (Util.StringMap.empty, []) metas
   in
 
+  (* Transitively close [all_lib_deps] over itself so each lib's deps
+     reflect every lib reachable through META requires. *)
+  let all_lib_deps =
+    let cache : (string, Util.StringSet.t) Hashtbl.t =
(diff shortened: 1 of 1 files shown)


src/driver/voodoo.ml
```diff
@@ -104,6 +104,29 @@ let of_voodoo pkg =
...
+  (* Transitively close [all_lib_deps] over itself so each lib's deps
+     reflect every lib reachable through META requires. *)
+  let all_lib_deps =
+    let cache : (string, Util.StringSet.t) Hashtbl.t =
+      Hashtbl.create (Util.StringMap.cardinal all_lib_deps) in
+    let rec close lib =
+      match Hashtbl.find_opt cache lib with
+      | Some r -> r
+      | None ->
+        Hashtbl.add cache lib Util.StringSet.empty;
```
panglesd, line 116: Why do you need this?

panglesd (commented):
Looks good to me!

src/driver/voodoo.ml
jonludlam: just being defensive against circular deps!

PR #1436, opened 2026-05-14
Fix syntax of changelog

Leonidas-from-XIV:
Looking at the rendering of https://github.com/ocaml/odoc/blob/master/CHANGES.md there is something broken with the rendering.

Turns out there's some invalid whitespace. This PR fixes it.


jonludlam:
Thanks!

Issue #865, opened 2022-06-09
Support for breaking long URLs in URL reference syntax


dbuenzli:
In my docsets I have hundreds of long links into the www that break my 80 source columns limit. While I may never get to break those in the future, I would really like to be able to do so for the next hundreds one.

More precisely be able to turn: 
```
{{:https://github.com/ocaml/odoc/issues/865}this issue}
```
into
```
{{:https://github.com/ocaml/\
   odoc/issues/865}this issue}
```

PR #1439, opened 2026-05-22
Fix issue #865

Tim-ats-d:
First try to fix the issue #865. The patch is dead simple, I've only added a `multilines_link_to_link` which properly reformats multi lines link into normal ones.

As I'm not yet really familiar with the Odoc codebase, don't hesitate to tell where can I define the `multilines_link_to_link` function or if the test need to reside in its own file!

Cc @panglesd

### DIFF
--- a/src/parser/syntax.ml
+++ b/src/parser/syntax.ml
@@ -167,6 +167,8 @@ type token_that_always_begins_an_inline_element =
   | `Begin_link_with_replacement_text of string
   | `Math_span of string ]
 
+(** When a backslash is followed by space, remove both the backslash and the
+    space. *)
 let escape_link link =
   let link = String.trim link in
   let buf = Buffer.create (String.length link) in
--- a/src/parser/syntax.ml
+++ b/src/parser/syntax.ml
@@ -170,6 +170,10 @@ type token_that_always_begins_an_inline_element =
 let escape_link link =
   let link = String.trim link in
   let buf = Buffer.create (String.length link) in
+  let add_backslash = function
+    | `Backslash -> Buffer.add_char buf '\\'
+    | `Escaping | `Char -> ()
+  in
   let last_state =
     String.fold_left
       (fun state chr ->
@@ -179,16 +183,13 @@ let escape_link link =
             Buffer.add_char buf chr;
             `Char
         | (`Backslash | `Escaping), _ when Char.Ascii.is_white chr -> `Escaping
-        | (`Backslash | `Escaping), _ ->
+        | ((`Backslash | `Escaping) as state), _ ->
+            add_backslash state;
             Buffer.add_char buf chr;
             `Char)
       `Char link
   in
-  let () =
-    match last_state with
-    | `Backslash -> Buffer.add_char buf '\\'
-    | `Escaping | `Char -> ()
-  in
+  add_backslash last_state;
   Buffer.contents buf
 
 (* Check that the token constructors above actually are all in [Token.t]. *)
--- a/src/parser/syntax.ml
+++ b/src/parser/syntax.ml
@@ -1,4 +1,3 @@
-
 (* This module is a recursive descent parser for the ocamldoc syntax. The parser
    consumes a token stream of type [Token.t Stream.t], provided by the lexer,
    and produces a comment AST of the type defined in [Parser_.Ast].
@@ -173,17 +172,17 @@ let escape_link link =
   let buf = Buffer.create (String.length link) in
   let last_state =
     String.fold_left
-         (fun state chr ->
-           match (state, chr) with
-           | `Char, '\\' -> `Backslash
-           | `Char, _ ->
-               Buffer.add_char buf chr;
-               `Char
-           | (`Backslash | `Escaping), (' ' | '\t' | '\n') -> `Escaping
-           | (`Backslash | `Escaping), _ ->
-               Buffer.add_char buf chr;
-               `Char)
-         `Char link
+      (fun state chr ->
+        match (state, chr) with
+        | `Char, '\\' -> `Backslash
+        | `Char, _ ->
+            Buffer.add_char buf chr;
+            `Char
+        | (`Backslash | `Escaping), _ when Char.Ascii.is_white chr -> `Escaping
+        | (`Backslash | `Escaping), _ ->
+            Buffer.add_char buf chr;
+            `Char)
+      `Char link
   in
   let () =
     match last_state with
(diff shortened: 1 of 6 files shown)


src/parser/syntax.ml
```diff
@@ -277,7 +287,7 @@ let rec inline_element :
           location
         |> add_warning input;
 
-      Loc.at location (`Link (u, []))
```
panglesd, line 280: I think `trim` should be called on the singlelinified input, not the other way around!

src/parser/syntax.ml
```diff
@@ -167,6 +168,15 @@ type token_that_always_begins_an_inline_element =
...
   | `Math_span of string ]
 
+let multilines_link_to_link link =
+  String.split_on_char '\n' link
+  |> List.map (fun line ->
+      String.trim @@
+      if String.ends_with ~suffix:{|\|} line then
+        String.(sub line 0 (length line - 1))
+      else line)
+  |> String.concat ""
```
panglesd: This does not really correspond to what we expect (escaping newlines):
```
a
b\      <--- there are whitespace after the \
c\      <--- no newlines after the \
```
will be interpreted as `abc`. The biggest problem probably being that the trailing `\` is removed (?)

src/parser/syntax.ml
```diff
@@ -1,3 +1,4 @@
+
```
panglesd:
```suggestion
```

src/parser/syntax.ml
```diff
@@ -295,7 +305,7 @@ let rec inline_element :
           input
       in
 
-      `Link (u, content) |> Loc.at (Loc.span [ location; brace_location ])
+      `Link ((multilines_link_to_link u), content) |> Loc.at (Loc.span [ location; brace_location ])
```
panglesd: same as https://github.com/ocaml/odoc/pull/1439/changes#r3304357614

panglesd (commented):
I left some comments. This also needs a change entry!

src/parser/syntax.ml
panglesd: Oh the location of the github comment is weird. What I mean is that (omitting the warning here, just to get the idea) it should not be 

```
      Loc.at location (`Link (u |> String.trim |> multilines_link_to_link, []))
```

but 

```
      Loc.at location (`Link (u |> multilines_link_to_link |> String.trim, []))
```

Tim-ats-d:
I've changed the behavior to be more fine-grained. Ending backslash should now be preserved in the extracted link.
Here are some examples:

```
{{:https://github.com/ocaml/\
   odoc/\
issues/\
                865\  }this issue}
``` 
=> `https://github.com/ocaml/odoc/issues/865\`
```
{{:https://github.com/ocaml/\
   odoc/issues/\

865

}this issue}
```
=> `https://github.com/ocaml/odoc/issues/865`

```
{{:https://github.com/ocaml/odoc/issues/\
865/   /\
ending worlds}this issue} 
```
=> `https://github.com/ocaml/odoc/issues/865/   /ending worlds"`

src/parser/syntax.ml
```diff
@@ -167,6 +168,30 @@ type token_that_always_begins_an_inline_element =
...
+  let buf = Buffer.create (String.length link) in
+  let last_state =
+    String.fold_left
+         (fun state chr ->
+           match (state, chr) with
+           | `Char, '\\' -> `Backslash
+           | `Char, _ ->
+               Buffer.add_char buf chr;
+               `Char
+           | (`Backslash | `Escaping), (' ' | '\t' | '\n') -> `Escaping
```
panglesd: Probably `'\r'` is needed for windows. The best is to use `Char.is_white`:
```suggestion
           | (`Backslash | `Escaping), c when Char.is_white c -> `Escaping
```

src/parser/syntax.ml
panglesd:
```suggestion
```

panglesd (commented):
(I've checked in the [definition of a URI](https://datatracker.ietf.org/doc/html/rfc3986#section-2), whitespace and `\` are not allowed in URI. Which does not mean they should be removed, it would be surprising as browsers handle their percent-encoding.)

One sentence should be added to the docs. Without it, it is difficult to say whether there is a bug in the implementation, or if it had the behaviour you wanted. My problem with the current implementation is that they cannot satisfyingly be expressed in English. ("When a backslash is followed by whitespace, remove both except at the end where if there is no whitespace the backslash is kept").

I suggest to make it either:
> When a backslash is followed by whitespace, both are removed

which implies a small modification to your code

or:
> Removes all backlashes and the space that follows them

which has a simpler implementation but is more invasive. I think I prefer the first one.

Could you add a sentence in the docs that says what is done to backslash and whitespace, and have the implementation correspond?

src/parser/syntax.ml
```diff
@@ -167,6 +167,30 @@ type token_that_always_begins_an_inline_element =
...
+      (fun state chr ->
+        match (state, chr) with
+        | `Char, '\\' -> `Backslash
+        | `Char, _ ->
+            Buffer.add_char buf chr;
+            `Char
+        | (`Backslash | `Escaping), _ when Char.Ascii.is_white chr -> `Escaping
+        | (`Backslash | `Escaping), _ ->
+            Buffer.add_char buf chr;
+            `Char)
```
panglesd: The docs says:
> When a backslash is followed by whitespace, both are removed.

but here, even without whitespace, the backslash is removed...

panglesd:
```suggestion
        | (`Backslash | `Escaping), _ ->
            if state = `Backslash then Buffer.add_char buf '\\';
            Buffer.add_char buf chr;
            `Char)
```
~~or a more elegant version if you have one.~~

Actually see https://github.com/ocaml/odoc/pull/1439/changes#r3623267227

src/parser/syntax.ml
```diff
@@ -167,6 +167,30 @@ type token_that_always_begins_an_inline_element =
...
+      `Char link
+  in
+  let () =
+    match last_state with
+    | `Backslash -> Buffer.add_char buf '\\'
+    | `Escaping | `Char -> ()
+  in
```
panglesd: Actually, this could be turned into a function to factor it with the same above.

Tim-ats-d:
Should be fine now :smile:

panglesd (approved):
Looks good to me! Thanks!

Leonidas-from-XIV:
I've kicked the tests to run but a lot of them fail because `String.fold_left` is missing (only added in OCaml 4.13). It probably needs a polyfill, like there exist for `Odoc_util.List`.

Likewise, `Char.Ascii` only exists since OCaml 5.4.

Issue #1198, opened 2024-09-26
UI: Fix rendering of tags


sabine:
Styling of tags is by default such that OCaml.org uses their own styles for tags. It would be good to fix up the styles on the odoc side to make the default sensible.

Related:
* https://github.com/ocaml/odoc/issues/1138
* https://github.com/ocaml/odoc/issues/1136

Issue #1136, opened 2024-05-26
Weird newline in rendering of tags


dlesbre:
The `@raise` tag is described as `@raise Error text` which seems intuitive enough, but the HTML render puts the text in a separate paragraph:
```
  raise Error

  text
```
This looks weird and can often break the flow of text (I tend to write `@raise Exn if bla bla bla`, which looks fine in the mli but weird in the docs since the paragraph break also breaks a sentence in two).

I haven't checked if this also occurs with other tags (`@param`).

This can be fixed by the following CSS:
```css
li.raises>p:first-of-type {
  display: inline;
}
```
However, a cleaner fix would involve not generating this first `<p>` block, or moving the `<span>raises<\span> <a>Exn<\a>` inside the first `<p>`.

PR #1440, opened 2026-05-26
Inline raise tag content in HTML rendering.

Tim-ats-d:
The fix is a bit ugly (`l.attr = [ "raises"]`) but I can easily remove this part by adding a field `should_inline` in the record definition of the type of `Odoc_document.Types.Description.t` to avoid relying on a fragile string comparison.

Should fix theses two issues:
- https://github.com/ocaml/odoc/issues/1136
- https://github.com/ocaml/odoc/issues/1198

cc @panglesd

### DIFF
--- a/src/html/generator.ml
+++ b/src/html/generator.ml
@@ -209,7 +209,7 @@ let text_align = function
 
 let cell_kind = function `Header -> Html.th | `Data -> Html.td
 
-let rec block ~config ~resolve (l : Block.t) : flow Html.elt list =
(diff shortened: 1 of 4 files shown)


panglesd:
The current fix has several issues:
- It only deals with the html backend: the other backend still have the issue.
- (If I understand correctly your code, did not check) It basically turns all paragraphs inside a raise tag into span, removing all the paragraph styling. So multi-paragraph raise tags are broken!
- Yes, the string comparison is not something we want.

I think it would be much cleaner to fix that in "Document", by putting the tag inside the first paragraph (if there is one, creating a paragraph otherwise). I think that would fix the problems I mentioned above!

PR #1442, opened 2026-06-03
Make cram tests portable to OpenBSD

Leonidas-from-XIV:
Since we got the OCaml-CI testing on OpenBSD and fixed the `jq` depext we are now running the tests mostly successfully on OpenBSD. The failures are mostly due to small discrepancies between the tools, e.g. they fail because the `-C <NUM>` syntax of GNU `grep` is unsupported. What is supported is `-C<NUM>` and `--context=<NUM>` so for clarity I've opted for the latter.

Also apparently OpenBSD `sed` does not interpret `\n` in sed patterns so I've swapped in `awk` instead.

### DIFF
--- a/test/xref2/github_issue_857.t/run.t
+++ b/test/xref2/github_issue_857.t/run.t
@@ -9,7 +9,11 @@ A quick test to repro the issue found in #857
   $ odoc latex-generate -o latex/ a.odocl
 
 In latex, labels in subpages should be disambiguated since the subpage is inlined inside the generated latex source.
-  $ cat latex/A.tex | sed 's/\\/\n\\/g' | grep label
+
+(Replace \\ with \n\ for readability)
+
+  $ awk '{ gsub("\\\\","\n\\",$0); print $0 }' < latex/A.tex > newlines.tex
+  $ grep label < newlines.tex
   \label{A}%
   \label{A--module-type-A}
   \label{A-module-type-A}
--- a/test/integration/depends.t/run.t
+++ b/test/integration/depends.t/run.t
@@ -4,7 +4,7 @@ Testing the depends command.
   $ ocamlc -c -bin-annot -I . -o lib_a.cmti a.mli
   $ ocamlc -c -bin-annot -I . -o lib_b.cmti b.mli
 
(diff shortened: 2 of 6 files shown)


test/xref2/github_issue_857.t/run.t
```diff
@@ -9,7 +9,7 @@ A quick test to repro the issue found in #857
   $ odoc latex-generate -o latex/ a.odocl
 
 In latex, labels in subpages should be disambiguated since the subpage is inlined inside the generated latex source.
-  $ cat latex/A.tex | sed 's/\\/\n\\/g' | grep label
+  $ awk '{ gsub("\\\\","\n\\",$0); print $0 }' < latex/A.tex | grep label
```
panglesd: The original line seems to replace `\\` with `\n\\`.
The new line seems to replace `\\` with `\n\`. It does not seem a big deal, but let's keep the original behavior (which makes sense as it does not modify the latex, just adds new lines).

Also, maybe a small comment would help future reader who are more comfortable with sed than awk.

> Add new lines before `\\` before grepping lines.

panglesd (approved):
Thanks

I just think one `\` is missing.

test/xref2/github_issue_857.t/run.t
Leonidas-from-XIV: The original also replaced it with `\n\` (which to be fair does not speak well of the readability of `sed`; in the Dune cram tests we removed `sed` because it lead to portability and readability issues all the time):

```
$ awk '{ gsub("\\\\","\n\\",$0); print $0 }' < latex/A.tex > awk-sol
$ cat latex/A.tex | sed 's/\\/\n\\/g' > sed-sol
$ diff -u sed-sol awk-sol
$ # no diff
```

If I add the missing `\` then the `\label` matches become matches on `\\label`.

I did add the comment for clarity and split it into two commands.

Leonidas-from-XIV:
@panglesd I've switched to `-C<NUM>` as Busybox does not support `--context`, but it looks like that's the compromise that works between BSD, GNU and Busybox.

jonludlam:
Thanks!

PR #1444, opened 2026-06-04
Omit displaying zero alloc arity when it matches the function arity

Leonidas-from-XIV:
When `[@@zero_alloc arity N]` is specified, this means that it applies to a function of arity `N`. This can be the arity of the function onto which it is annotated but it can also be less, e.g. in the case when a function returns a function that is `zero_alloc`.

Odoc currently displays the arity always, however omitting the arity means that the arity is the same as the arity of the function, thus it doesn't add any additional information. In OxCaml `arity N` is optional, leaving it out is perfectly acceptable. This PR changes odoc to avoid printing the `arity N` argument, when the type expression is of the same arity. Therefore the `arity` annotation becomes more valuable when it is printed as it means that the arity of the zero allocation part is actually different from the value binding.

The test case only modifies `oxcaml.mli` as specifying the arity explicitely is only valid in signatures, not in implementations. Trying otherwise triggers warning 47 about invalid payloads of attributes. 

Discussed this with @panglesd before and we agreed that this makes sense as a separate PR.


panglesd (approved):
I changed my mind and I think you are right that `document/generator.ml` is a good place to do the check.

So looks good to me!

CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 - Support for OxCaml unboxed named types (@art-w, #1407)
 - Support for OxCaml zero alloc definitions (@Leonidas-from-XIV, #1422)
+- Support for OxCaml zero alloc definitions (@Leonidas-from-XIV, #1422, #1444)
```
panglesd: There was a hiccup during the rebase!

Leonidas-from-XIV: Oh, true! I was looking at this diff and at first I didn't see the issue but you're right; the line was duplicated by accident! I've fixed it now.

jonludlam:
Thanks!

PR #1445, opened 2026-06-05
No ppx expect

jonludlam:
Remove ppx_expect, but keep the tests!


jonludlam:
This should be reviewed commit-by-commit.

test/model/semantics/test.ml
```diff
@@ -33,16 +33,19 @@ let test ?(tags_allowed = true) ?(location = { Location_.line = 1; column = 0 })
...
-    Format.fprintf Format.std_formatter "%s" (Yojson.Basic.to_string yojson)
-  in
-  print_json_desc parser_output_desc parser_output;
-  Format.pp_print_flush Format.std_formatter ()
+  let buf = Buffer.create 256 in
+  let f = Format.formatter_of_buffer buf in
+  let yojson = Type_desc_to_yojson.to_yojson parser_output_desc parser_output in
+  Format.fprintf f "%s" (Yojson.Basic.to_string yojson);
+  Format.pp_print_flush f ();
+  let body = Buffer.contents buf in
```
panglesd: This seems a very convoluted way of doing:

```suggestion
  let yojson = Type_desc_to_yojson.to_yojson parser_output_desc parser_output in
  let body = Yojson.Basic.to_string yojson in
```
or maybe I am missing something

src/parser/test/test.ml
```diff
@@ -275,11 +275,18 @@ let test ?(location = { Loc.line = 1; column = 0 }) str =
     }
   in
   let ast = Odoc_parser.parse_comment ~location ~text:str in
-  Format.printf "%a" parser_output ast
+  let buf = Buffer.create 256 in
+  let f = Format.formatter_of_buffer buf in
+  Format.fprintf f "%a" parser_output ast;
+  Format.pp_print_flush f ();
+  let body = Buffer.contents buf in
```
panglesd: This seems a very convoluted way of doing:

```suggestion
  let yojson = Type_desc_to_yojson.to_yojson parser_output_desc parser_output in
  let body = Yojson.Basic.to_string yojson in
```

(similar to the other comment

test/model/semantics/dune
```diff
@@ -1,13 +1,25 @@
 (rule
  (copy ../../odoc_print/type_desc_to_yojson.ml type_desc_to_yojson.ml))
 
-(library
- (name odoc_model_semantics_test)
- (package odoc)
- (optional)
- (inline_tests)
+(executable
+ (name test)
```
panglesd: We now have `test.exe` and `test_main.exe`. Those names are confusing, they should be fully qualified: `odoc_parser_test.exe` and `odoc_semantic_test.exe` or something like that.

panglesd (commented):
Removing this dependency is really good, but:
- With the current organization, we lose to have all three things at the same place: the description of the test (eg whether it is expected behaviour or a FIXME), the input and the output. The doc and input is in one file, the input and output in another.
- It is slightly more complex but the setup is well explained so I guess it is ok (and expected since we remove a dependency, we have to redo things).

So we could add a `~doc` argument to the `print` and `expect` functions to be able to read the `.expected` files with all the context.

And I left a comment on an intermediate commit (so not important), but also about name of test executables (more important).

test/model/semantics/test.ml
jonludlam: oops yes, at least this is fixed by the subsequent commits. I think there was a slightly more complex setup at some point.

panglesd (approved):

jonludlam:
Tests renamed, and docs updated.

PR #1447, opened 2026-06-09
Remove the odoc-bench.opam file

jonludlam:
This installed a bunch of libraries for the purposes of benchmarking. However, it doesn't seem to have run for a long time, and it ends up pulling lots of deps in to the normal test runs.


panglesd (approved):
CI run for nothing is wasted resources!

Issue #960, opened 2023-05-02
Raised at Odoc_xref2__Tools.assert_not_functor in colibri2.0.4


sabine:
Originally reported in: https://github.com/ocaml-doc/voodoo/issues/61

> From this build https://docs.ci.ocaml.org/job/2023-04-26/140729-voodoo-do-7c9b58:

```
odoc: internal error, uncaught exception:
      File "src/xref2/tools.ml", line 1600, characters 9-15: Assertion failed
      Raised at Odoc_xref2__Tools.assert_not_functor in file "src/xref2/tools.ml", line 1600, characters 9-21
      Called from Odoc_xref2__Utils.ResultMonad.bind in file "src/xref2/utils.ml" (inlined), line 10, characters 38-41
      Called from Odoc_xref2__Tools.resolve_module.resolve.(fun) in file "src/xref2/tools.ml", line 933, characters 8-129
      Called from Odoc_xref2__Env.with_recorded_lookups in file "src/xref2/env.ml", line 186, characters 17-23
      Re-raised at Odoc_xref2__Env.with_recorded_lookups in file "src/xref2/env.ml", line 191, characters 4-11
      Called from Odoc_xref2__Tools.MakeMemo.memoize.no_memo in file "src/xref2/tools.ml", line 291, characters 10-64
      Called from Odoc_xref2__Tools.handle_canonical_module_real.resolve in file "src/xref2/tools.ml", line 1324, characters 4-69
      Called from Odoc_xref2__Tools.canonical_helper.resolve in file "src/xref2/tools.ml", line 81, characters 10-23
      Called from Odoc_xref2__Tools.filter_map.(fun) in file "src/xref2/tools.ml", line 16, characters 27-30
      Called from Stdlib__List.fold_left in file "list.ml", line 121, characters 24-34
      Called from Odoc_xref2__Tools.filter_map in file "src/xref2/tools.ml", line 15, characters 5-101
      Called from Odoc_xref2__Tools.canonical_helper in file "src/xref2/tools.ml", line 90, characters 21-62
      Called from Odoc_xref2__Tools.handle_canonical_module_real in file "src/xref2/tools.ml", line 1334, characters 8-59
      Called from Odoc_xref2__Env.with_recorded_lookups in file "src/xref2/env.ml", line 186, characters 17-23
      Re-raised at Odoc_xref2__Env.with_recorded_lookups in file "src/xref2/env.ml", line 191, characters 4-11
[...]

PR #1452, opened 2026-06-30
`include functor` support for OxCaml

Leonidas-from-XIV:
This PR is a follow up to #1368 and adds support for `include functor` where it now displays that the items were included via the functor.


src/xref2/tools.ml
```diff
@@ -1621,22 +1627,28 @@ and signature_of_module_type_of :
...
   match m with
   | Component.ModuleType.U.Path p -> (
       match resolve_module_type env p with
-      | Ok (_, mt) -> expansion_of_module_type env mt >>= assert_not_functor
+      | Ok (_, mt) -> (
+          expansion_of_module_type env mt
+          >>=
+          match allow_functor with
+          | true -> assert_functor
+          | false -> assert_not_functor)
```
art-w: Can we avoid the `~allow_functor` is we match on `| Signature s | Functor (_ Signature s) -> s` or would that hide some logic errors? (since both `assert_.._functor` fail with `assert false` anyway, I assume the error reporting isn't a concern)

src/document/generator.ml
```diff
@@ -1916,23 +1916,27 @@ module Make (Syntax : SYNTAX) = struct
...
             match t.decl with
             | Odoc_model.Lang.Include.Alias mod_path ->
-                Link.from_path (mod_path :> Paths.Path.t)
-            | ModuleType mt -> umty mt
+                (O.keyword "include", Link.from_path (mod_path :> Paths.Path.t))
+            | Functor (Path mod_path) ->
+                ( O.keyword "include functor",
+                  Link.from_path (mod_path :> Paths.Path.t) )
+            | Functor (ModuleType mt) -> (O.keyword "include functor", umty mt)
+            | ModuleType mt -> (O.keyword "include", umty mt)
```
art-w: nitpick: there's some minor redundancy which could be refactored here by pattern matching twice on `t.decl` instead of once

test/generators/cases/oxcaml.mli
```diff
@@ -280,3 +280,31 @@ module M3 : sig @@ contended
...
+  module Make (T : sig type t end) : sig type included end
+  type t
+  include functor module type of Make
+end
+
+module Include_functor_named_type : sig
+(** This is a Module where the type is named and then included. *)
+  module type Make = (_ : sig type t end) -> sig type included end
+  type t
+  include functor Make
```
art-w, line 660: Out of curiosity, can we test the behavior with `(** @inline *)`?

art-w (approved):
Thanks a lot, this looks great! I only have minor feedback but nothing blocking :)

src/xref2/tools.ml
jonludlam: Yes - at the very least we ought to rename the argument. I'd quite like to double check the semantics here though, as we've got at least one instance of this bit going wrong: https://github.com/ocaml/odoc/issues/960

Leonidas-from-XIV: Truth be told I am a bit confused about the purpose of `assert_not_functor`:

1. Why do we need to make sure the module expression is not a functor?
2. The function doesn't actually just assert (I thought I could just remove it to allow for functors, but then the types don't match), it also unpacks the signature.

Hence to answer @art-w's question, I don't know - it would change semantics as then it would always allow functors; in which case we could rewrite it into something like

```ocaml
and assert_signature : type err. expansion -> (Component.Signature.t, err) result
    = function
  | Signature sg -> Ok sg
  | Functor (_, mty) -> (
      match mty with Signature sg -> Ok sg | _ -> assert false)
  | _ -> assert false
```

jonludlam:
I think there's a chunk of logic missing from this PR. At the moment, all it does AFAICT is get the expansion of the functor and splice it into the expansion of the include. While this works for simple examples, it's not sufficient for more involved cases.

As a simple example, consider this:

```ocaml
module F ( I : sig type t end ) = struct
  type myt = I.t
end

module M = struct
  type t = float
  include functor F
end
```
Running this through `ocamlc -c -i` gives what we ought to end up with (roughly):

```ocaml
module F : functor (I : sig type t end) -> sig type myt = I.t end
module M : sig type t = float type myt = float end
```

If we just simply get the signature of the functor body and splice it in, we end up something more like:

```ocaml
...
module M : sig
  type t = float
  type myt = I.t
end
```

and obviously that `I.t` need something done to it. If you run it through odoc as of this branch, it chokes on exactly this:

```
Exception Failure("Not_found: I/3") handling type_expr: resolved(I/3.t)
backtrace:
Raised at Stdlib.failwith in file "stdlib.ml" (inlined), line 39, characters 17-33
...
```

I suspect the most straightforward implementation of this would be to do exactly what the docs suggest it does internally. Treat it more like:

```ocaml
module M = struct
  module __DUMMY__ = struct
    type t = float
  end

  include __DUMMY__
  include F(__DUMMY__)
end
```

We can ensure that the dummy module is hidden, meaning that the include will just inline the contents,
so it should end up rendering in a sensible way. I'm not sure whether it would be best to do this as
a preprocessing step or whether we might be able to rewrite the loaders with an accumulator so we can
do this sort of mangling on loading of the cmt/cmtis.

jonludlam (changes requested):

src/xref2/tools.ml
Leonidas-from-XIV: Ok, the whole code disappeared now with @jonludlam's proposal to use a different compilation scheme so I am marking this conversation as resolved since I removed the whole `allow_functor` machinery.

Leonidas-from-XIV:
I've now implemented @jonludlam's proposal of compiling down the include functor to basically a `BODY` module and a functor application. This has some extra steps to deal with odd cases like anonymous functors but seems to work fine. I've also incorporated the example and it compiles now properly.

The way it works is by parsing the `include functor` as `Include.Functor` expression (with a path or module type) and then post-processing the structure items if there were include functors and compiling them to `Include.Functor` with paths (to be able to use `` `Apply`` thus avoiding having to infer types) in the shape of `<GENERATED_FUNCTOR_MODULE>(GENERATED_WRAPPER_MODULE>)`. I'm also retaining the original expression, so the `include functor <X>` will be rendered as `include functor <X>` instead of `include functor APPLY_23(BODY_12)`.

However there's one outstanding issue: while the items coming from the functor expansions contain the right path (e.g. `val functor_output = APPLY_23(BODY_12).functor_output`) this is is quite ugly and messy. If I hide the generated modules the cross-reference disappears and just the value remains. This might be fine but also somewhat less informative.

What I could imagine would be best if I could hide the generated modules but also tell the cross-linker that I want the item to be cross-referenced to the ("include") functor, not the generated module. Is there a way to tell odoc to "redirect" paths?

Leonidas-from-XIV:
Odoc has a mechanism to specify a substitution via `@canonical` which is represented by the `` `Canonical`` constructor, however that can only be constructed with resolved modules, whereas the Typedtree transformation is happening before. Maybe it is possible to resolve these modules early.

test/generators/cases/oxcaml.mli
```diff
@@ -280,3 +280,79 @@ module M3 : sig @@ contended
 end
 (** [contended] modality applied to all definitions in the module, except the
     ones which have already specified this axis. *)
+
+(** {1 Include functor on signatures} *)
+
+module No_include_functor : sig
+(** Module without any [include functor] features, this is how things are done
+    in plain OCaml at the moment. *)
+  module Make (T : sig type t end) : sig type included end
```
panglesd: I think it would be nice to put this outside of the modules, so that the output is more readable

test/generators/cases/oxcaml.mli
```diff
@@ -280,3 +280,79 @@ module M3 : sig @@ contended
...
+
+  include module type of T
+  include module type of Make(T)
+end
+
+module Include_functor : sig
+(** Module which defines a functor and includes it via [module type of] *)
+  module Make (T : sig type t end) : sig type included end
+  type t
+  include functor module type of Make
```
panglesd: We should have at least one test where some items are after the `include functor`.

panglesd (commented):
I'll continue the review, but here is already some comments. I think the `Make` functor should be made a little bit more complex, and in particular use its argument.

```ocaml
module Make (T : sig type t end) : sig type included = T.t end
```
Now suppose it is used in the following way:

```ocaml
type t
include functor Make
```

From my understanding of the oxcaml feature, the rendered signature should be:

```ocaml
type t
> include functor Make
  type included = t
```
(where `>` is for the opened include accordion 🪗 )
In my testing, that is not the case!

I wonder if it is possible to fix that using the current "sugar". I believe the "desugared" form of the example above (extended with a value and a module) could use aliases, for types, modules and module types, instead of relying on the `DUMMY__` include :

```ocaml
type t
val f : t
module X : S
module DUMMY__ : sig
    type nonrec t = t
    val f : t
    module X = X
  end

  module STRUCT__Make : Make
  include module type of STRUCT__Make(DUMMY__)
end
```
What do you think?

Also, definitely we want those generated modules to be hidden.  As @jonludlam mentioned, that would turn the "inlining" invisible if we keep it, and odoc would use links to a visible alias.

Also note that while I'd like to investigate the alternate sugar, the current version is better than no support, and as such (I still need more review of course) I would not block the merge on the problem noted above!

panglesd:
I had a go at the "alias desugaring" idea: https://github.com/panglesd/odoc/tree/include-functor-alias-desugaring

It looks a reasonable first solution, I think. Maybe implementing the "include functor expansion" mechanism in odoc would be better, but slightly more invasive...

Leonidas-from-XIV:
I've ended up cherry-picking the commits from @jonludlam's branch, which among an alternative desugaring also solves another real problem that I didn't think of: the `BODY` module needs to be re-created for every include functor application as one `include functor` application might depend on items that were generated by a preceding `include functor` application (e.g. imagine a `Identifier` include functor that generates (amongst others) a `type key` and a then a `Map` include functor which consumes as a `type key`, the latter needs to see the output of the former).

Also added an additional example of when the body of the module contains an `include` (a potential issue noticed by @panglesd) and rebased on master (thus running on Oxcaml minus39).

src/loader/cmi.ml
```diff
@@ -1638,3 +1638,90 @@ let read_interface root name ~warnings_tag intf =
...
+      in
+      Signature.ModuleType { mt with ModuleType.id; expr; canonical = None; source_loc = None } :: acc
+    | Include incl ->
+      (* The items an [include] brings in are visible to the functor too. *)
+      List.fold_left (fun acc item -> wrapper_item item acc) acc incl.Include.expansion.content.items
+    | Value _ | ModuleSubstitution _ | ModuleTypeSubstitution _ | Open _
+    | TypeSubstitution _ | TypExt _ | Exception _ | Comment _ ->
+      (* Nothing in the expansion of [F(BODY__n)] can refer to these. *)
+      acc
+    | Class _ | ClassType _ ->
```
panglesd: Nothing can refer to them, but without including them, we apply `F(BODY_n)` without `BODY_n` having module type what `F` is expecting.

Maybe that is fine, I don't think odoc will check that as it assumes the compiler has already enforced this, however I think it is worth adding tests (as for instance if later odoc requires `BODY_n` to have the right signature.
But maybe that is not fine! What happens to the values, if the input module is stored in the expansion?

I think it would be good to add a test with values, ..., classes in the input to the `include functor`. Moreover, the input functor should alias its input, to test the expansion of the situation described above.

src/loader/cmi.ml
panglesd, line 1: It is slightly strange to have the function defined here, and only used in another file. Maybe they should be defined in the other file, probably `cmt.ml`?

panglesd (commented):
I have some small comments on the included items in the wrapper module, which needs some testing and (I would bet but not a lot) some implementation.

Otherwise, very good work! Thanks!

src/loader/cmi.ml
Leonidas-from-XIV: I put it there since both `Cmti` and `Cmt` use it, to avoid creating dependencies between these. But `Cmt` already depends on `Cmti` so I moved the code to `Cmti` :+1:

src/loader/cmi.ml
jonludlam: Oops, yes, you're right. Nothing stopping the functor from just making an alias of its argument, and then we'd lose all of those other items. My mistake, sorry!

Leonidas-from-XIV: I've added a test for class, class types and aliases which exposes this issue:

* The classes and class types refer to the generated `BODY__n` module. It's somewhat surprising given `BODY__n` should be hidden, I think there's some `is_hidden` check missing in some place.
* The `Aliased` module is constructed properly but indeed missing the class types.

jonludlam:
Great, thanks for this!

PR #1453, opened 2026-07-01
Use `grep` CLI in a portable way

Leonidas-from-XIV:
OpenBSD keeps failing, because BSD grep can't take `grep <term> <options>` but needs `grep <options> term`. This PR fixes the failures. Also the `-A` `-B`, `-C` options don't accept spaces.

Followup to #1442.


jonludlam:
Nice thanks!

PR #1454, opened 2026-07-01
Oxcaml: Support for modes

### LINKED ISSUE #1417: Support OxCaml modes
Odoc doesn't show any mode at all:

```ocaml
val globalize : 'a @ local -> 'a
val localize : 'a -> 'a @ local
val restrict : 'a @ unique -> 'a @ once
```

art-w:
This PR adds support for rendering OxCaml modes. Most of the complexity comes from the heuristics copied from the OxCaml compiler to omit modes which are either legacy, implied by other axes, or inferred from the context (for arrow types, e.g. `a @ local -> b -> c -> d` is internally represented as `a @ local -> (b -> (c -> d @ local) @ local) @ local` because a partial application of the first argument `a @ local` must return a `local` closure (otherwise the local value could escape its scope)... but since the `@ local` expansion is unreadable, we display the short form as it implies the long one)

Fixes https://github.com/ocaml/odoc/issues/1417


sherlodoc/index/load_doc.ml
```diff
@@ -82,21 +82,22 @@ let searchable_type_of_constructor args res =
   | TypeDecl.Constructor.Tuple args -> begin
       match args with
       | _ :: _ :: _ ->
-          TypeExpr.Arrow (None, Tuple (List.map (fun (x, _mods) -> None, x) args), res)
-      | [ (arg, _) ] -> TypeExpr.Arrow (None, arg, res)
+          TypeExpr.Arrow
+            (None, (Tuple (List.map (fun (x, _mods) -> None, x) args), []), (res, []))
+      | [ (arg, _) ] -> TypeExpr.Arrow (None, (arg, []), (res, []))
```
Leonidas-from-XIV, line 87: I wonder whether it wouldn't potentially make sense to make this into a record type (with some usefully named helper constructor functions) because the empty lists everywhere make it a bit verbose.

src/loader/cmti.ml
```diff
@@ -41,6 +41,9 @@ let opt_map f = function
 let read_label = Cmi.read_label
 
 let rec read_core_type env container ctyp =
+  read_core_type_modal env Cmi.legacy_modes container ctyp
```
Leonidas-from-XIV, line 44: Why does this always read `legacy_modes`?

src/model/lang.ml
```diff
@@ -496,7 +501,7 @@ and TypeExpr : sig
     | Var of string
     | Any
     | Alias of t * string
-    | Arrow of label option * t * t
+    | Arrow of label option * (t * Modes.t) * (t * Modes.t)
```
Leonidas-from-XIV: As mentioned above maybe it would be better to have a `with_mode` record type of sorts here.

test/generators/cases/oxcaml.mli
```diff
@@ -280,3 +280,189 @@ module M3 : sig @@ contended
...
+val mode_arg : int @ local -> int
+(** Mode on a function argument. *)
+
+val mode_ret : int -> int @ local
+(** Mode on a function return. *)
+
+val mode_both : int @ local -> int @ local
+(** Modes on both argument and return. *)
+
+val mode_multi : string @ local once -> string @ local unique
```
Leonidas-from-XIV, line 295: What if the modes are flipped?

Leonidas-from-XIV (commented):
Looks nice, the logic determining what to leave out is actually more readable than the description makes it sound IMHO :)

test/generators/cases/oxcaml.mli
art-w: Flipped as in `string @ once local -> string @ unique local`? I'll add a test to show that similarly to the OxCaml compiler printer for the toplevel or inferred interfaces, the modes order is normalized by `Cmi.tree_of_modes` predefined order :)

src/loader/cmti.ml
art-w: This argument is the base value for the accumulator for the "currently implicit modes" (such that we don't need to display them). It gets updated when traversing an arrow type such that the argument' modes carry over implicit modes for the output (otherwise we would display the redundant `@ local` described in the PR). In OCaml `legacy_modes` is just `():unit` since we don't make use of it :)

src/loader/cmt.ml
```diff
@@ -297,7 +297,7 @@ let rec read_class_field env parent cf =
                must keep only the type after the first arrow. *)
             let type_ =
               match Cmi.read_type_expr env expr.exp_type with
-              | Arrow (_, _, t) -> t
+              | Arrow (_, _, (t, _)) -> t
```
jonludlam: Is there anything subtle here? Can the first argument have any mode that we're throwing away?

test/generators/cases/oxcaml.mli
Leonidas-from-XIV: Yep, that's exactly what I meant :)

src/loader/cmi.ml
```diff
@@ -561,16 +561,102 @@ let read_label_modalities ld =
...
+  in
+  let portability =
+    match modes.statefulness, modes.portability with
+    | Stateless, Portable
+    | Observing, Shareable
+    | Stateful, Nonportable -> None
+    | _, _ -> Some modes.portability
+  in
+  let diff = Mode.Alloc.Const.diff modes Mode.Alloc.Const.legacy in
+  let diff = { diff with forkable; yielding; contention; portability } in
```
panglesd: Maybe it would be nice to fully destruct `diff` here: It would make a compiler error if a new axis is introduced as a field in the record, warning us that there is a line to add in the list below.

test/generators/html/Oxcaml-module-type-Mode_sig.html
panglesd, line 1: I'm not sure to see where this module comes from. Maybe a leftover from a previous organization where things were grouped into modules?

src/loader/cmi.ml
```diff
@@ -561,16 +561,102 @@ let read_label_modalities ld =
 let read_constructor_argument arg =
   arg.ca_type, read_modalities Immutable arg.ca_modalities
 
+let tree_of_modes (modes : Mode.Alloc.Const.t) : string list =
+  (* Same as the OxCaml's [Printtyp.tree_of_modes]: axes whose value is legacy
```
panglesd: In today's OxCaml's main, `tree_of_modes` has been moved to `Out_type`. Maybe we could avoid naming the module containing the function?

```suggestion
  (* Same as the OxCaml's [tree_of_modes]: axes whose value is legacy
```

src/loader/cmi.ml
panglesd: I realize only now this is a copy of an oxcaml function, but I still think we could modify it to be warned when a new field is added.

src/loader/cmi.ml
```diff
@@ -561,16 +561,102 @@ let read_label_modalities ld =
...
+let read_alloc_modes m = tree_of_modes (Mode.Alloc.zap_to_legacy m)
+
+type modes = Mode.Alloc.Const.t
+
+let legacy_modes = Mode.Alloc.Const.legacy
+
+let curried_acc modes arg_mode =
+  Ctype.curry_mode modes (Mode.Alloc.zap_to_legacy arg_mode)
+
+let mode_is_implied modes res_mode =
```
panglesd: `modes` could be named `implied_mode`, if I understand correctly. And I think the same in other functions would make them easier to read!

panglesd (commented):
I will need a bit more time to review the loading, since it is touching the internals of oxcaml (eg I'm not sure what `zap_to_legacy` is) but the rest is solid, and very well tested! Thanks!

test/generators/html/Oxcaml-module-type-Mode_sig.html
art-w: Thanks, I removed it, it was indeed a residual from a previous test

test/generators/cases/oxcaml.mli
art-w: Done!

src/model/lang.ml
art-w: Thanks, I introduced a type alias! But I kept a product as the record was even more verbose :/

src/loader/cmi.ml
art-w: Done!

src/loader/cmi.ml
art-w: Good idea, done!

src/loader/cmi.ml
art-w: Sure, done :)

sherlodoc/index/load_doc.ml
art-w: I tried but it was even more verbose with a record :/ (there's a lot of pattern matching on those arrows)

src/loader/cmt.ml
art-w: This part of the code is destructuring the `self -> t` from the method type, and as far as I know there's no syntax to annotate methods with modes/modalities. The modes within the arrow (e.g. `method foo: string @ local -> ...` are in `t = string @ local -> ...` so there are not lost. Nevertheless I added an explicit check to detect if that ever changes :)

panglesd (approved):
Let's merge after rebasing after #1469 is merged and green CI

jonludlam:
I've rebased on top of the minus39 fixes.

jonludlam:
Thanks!

PR #1457, opened 2026-07-07
Driver: Transitive deps in voodoo mode

jonludlam:
When in voodoo mode, we only have access to one packages META files. Finding dependencies from this meant that we only had direct deps, not transitive deps. This fixes that problem by recording the module hashes in the stamp files we drop per library, then using the pre-existing function that adds missing dependencies via the hash.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,14 @@
 # Unreleased
 
+### Added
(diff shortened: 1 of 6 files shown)


jonludlam:
I've got a better idea.

PR #1458, opened 2026-07-07
Remove code for OCaml < 4.08

fantazio:
The minimal supported version according to .opam files is OCaml 4.08. There is no need to conserve and maintain support for prior versions.


Leonidas-from-XIV (approved):
I've added the "no changelog" tag since technically the compatibility with 4.08 was dropped in 3.1.0 (#1300) but feel free to add to the entry or add your own.

I've looked though the changes and they look very good, I don't have any notes on that. Merging this is a good idea as it reduces the mental gymnastics of understanding what the code that is run looks like.

Thanks for your contribution!

jonludlam:
Agreed. Thanks for this, in it goes!

PR #1460, opened 2026-07-10
Eliminate duplicate branch

Leonidas-from-XIV:
While thinking on how to best implement `include functor` I was looking at the code that loads modules to see how to best create dummy modules. Which led me to this piece of code.

Both sides of the branch seemed have the same code (in the end I verified with `diff` to make sure I am not missing something subtle), thus might as well get rid of the branch itself.

I assume this might have been part of a bigger branching code for different versions but it seems all supported versions can use the same code these days.


jonludlam:
Thanks @Leonidas-from-XIV !

PR #1464, opened 2026-07-23
OxCaml: support for parameterized libraries

### LINKED ISSUE #1390: Add support for representing parameterized libraries
Dune has been extended to support parametric libraries (currently supported by the oxcaml compiler), in the work tracked by the ['Support for Parameterized Libraries' milestone](https://github.com/ocaml/dune/milestone/64).

Our evaluation (documented on https://github.com/ocaml/dune/issues/12565) has shown that odoc can build documentation for projects using the feature without encountering problems. But odoc does not currently render library parameters or parametric libraries why any distinction. We would like to add support for the latter.

@art-w can offer any needed elaboration

art-w:
This PR integrates OxCaml parameterized libs in Odoc, by adding the missing informations (is that library a description of a parameter? is that library parameterized and by what? does that library implement a parameter?) and fixing the broken references to instantiated parameterized libraries. The later is similar to a functor application, with the difference that arguments are nominal instead of positional: The OxCaml compiler internally uses the syntax `Lib[Param:Impl]`, which we replicate (with individual links on `Lib`, `Param` and `Impl`).

See the cram test or [dune's documentation on parameterized libs](https://dune.readthedocs.io/en/latest/tutorials/oxcaml-parameterized-library/index.html)

Fixes #1390

### DIFF
--- a/src/document/doctree.ml
+++ b/src/document/doctree.ml
@@ -322,14 +322,15 @@ module PageTitle : sig
 
   val render_src_title : Source_page.t -> Item.t list
 end = struct
+  let prefixed_title ~source_anchor name s =
+    let level = 0 and label = None in
+    let title =
+      Types.inline (Text (s ^ " ")) :: Codefmt.code (Codefmt.txt name)
+    in
+    [ Types.Item.Heading { level; label; title; source_anchor } ]
+
   let format_title ~source_anchor kind name preamble =
-    let mk title =
-      let level = 0 and label = None in
-      [ Types.Item.Heading { level; label; title; source_anchor } ]
-    in
-    let prefix s =
-      mk (Types.inline (Text (s ^ " ")) :: Codefmt.code (Codefmt.txt name))
-    in
+    let prefix = prefixed_title ~source_anchor name in
     match kind with
     | `Module -> (prefix "Module", preamble)
     | `Parameter _ -> (prefix "Parameter", preamble)
@@ -349,9 +350,10 @@ end = struct
     | Some p -> Printf.sprintf "%s.%s" p.name name
 
   let render_title ?source_anchor (p : Page.t) =
-    format_title ~source_anchor p.url.kind
-      (make_name_from_path p.url)
-      p.preamble
+    let name = make_name_from_path p.url in
+    if p.library_parameter then
+      (prefixed_title ~source_anchor name "Library parameter", p.preamble)
+    else format_title ~source_anchor p.url.kind name p.preamble
 
   let render_src_title (p : Source_page.t) =
     format_title ~source_anchor:None p.url.kind (make_name_from_path p.url) []
--- a/src/document/doctree.ml
+++ b/src/document/doctree.ml
@@ -332,6 +332,7 @@ end = struct
     in
     match kind with
     | `Module -> (prefix "Module", preamble)
+    | `LibraryParameter -> (prefix "Library parameter", preamble)
     | `Parameter _ -> (prefix "Parameter", preamble)
     | `ModuleType -> (prefix "Module type", preamble)
     | `ClassType -> (prefix "Class type", preamble)
--- a/src/document/generator.ml
+++ b/src/document/generator.ml
@@ -84,10 +84,11 @@ let prepare_preamble comment items =
   in
   (Comment.standalone preamble, Comment.standalone first_comment @ items)
 
-let make_expansion_page ~source_anchor url comments items =
+let make_expansion_page ?(library_parameter = false) ~source_anchor url comments
+    items =
   let comment = List.concat comments in
   let preamble, items = prepare_preamble comment items in
-  { Page.preamble; items; url; source_anchor }
+  { Page.preamble; items; url; source_anchor; library_parameter }
 
 include Generator_signatures
 
@@ -95,11 +96,38 @@ module Make (Syntax : SYNTAX) = struct
   module Link : sig
     val from_path : Paths.Path.t -> text
 
+    val href_of_path : Paths.Path.t -> Url.t option
+
     val from_fragment : Paths.Fragment.leaf -> text
 
     val render_fragment_any : Paths.Fragment.t -> string
   end = struct
     open Paths
+
+    let href_of_resolved rp =
+      (* If the path is pointing to an opaque module or module type
+         there won't be a page generated - so we stop before; at
+         the parent page, and link instead to the anchor representing
+         the declaration of the opaque module(_type) *)
+      let stop_before =
+        match rp with
+        | `OpaqueModule _ | `OpaqueModuleType _ -> true
+        | _ -> false
+      in
+      match Paths.Path.Resolved.identifier rp with
+      | Some id -> Some (Url.from_identifier ~stop_before id)
+      | None -> None
+
+    let rec href_of_path : Path.t -> Url.t option =
+     fun path ->
+      match path with
+      | `Substituted m -> href_of_path (m :> Path.t)
+      | `SubstitutedMT m -> href_of_path (m :> Path.t)
+      | `SubstitutedT m -> href_of_path (m :> Path.t)
+      | `SubstitutedCT m -> href_of_path (m :> Path.t)
+      | `Resolved _ when Paths.Path.is_hidden path -> None
+      | `Resolved rp -> href_of_resolved rp
+      | _ -> None
 
     let rec from_path : Path.t -> text =
      fun path ->
@@ -129,24 +157,18 @@ module Make (Syntax : SYNTAX) = struct
           let link1 = from_path (p1 :> Path.t) in
           let link2 = from_path (p2 :> Path.t) in
           link1 ++ O.txt "(" ++ link2 ++ O.txt ")"
+      | `ApplyParam (p1, p2, p3) ->
+          let link1 = from_path (p1 :> Path.t) in
+          let link2 = from_path (p2 :> Path.t) in
+          let link3 = from_path (p3 :> Path.t) in
+          link1 ++ O.txt "[" ++ link2 ++ O.txt ":" ++ link3 ++ O.txt "]"
       | `Resolved _ when Paths.Path.is_hidden path ->
           let txt = Url.render_path path in
           unresolved [ inline @@ Text txt ]
       | `Resolved rp -> (
-          (* If the path is pointing to an opaque module or module type
-             there won't be a page generated - so we stop before; at
-             the parent page, and link instead to the anchor representing
-             the declaration of the opaque module(_type) *)
-          let stop_before =
-            match rp with
-            | `OpaqueModule _ | `OpaqueModuleType _ -> true
-            | _ -> false
-          in
           let txt = [ inline @@ Text (Url.render_path path) ] in
-          match Paths.Path.Resolved.identifier rp with
-          | Some id ->
-              let href = Url.from_identifier ~stop_before id in
-              resolved href txt
+          match href_of_resolved rp with
+          | Some href -> resolved href txt
           | None -> O.elt txt)
 
     let dot prefix suffix = prefix ^ "." ^ suffix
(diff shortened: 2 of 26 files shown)


test/integration/parameterized.t/run.t
```diff
@@ -0,0 +1,468 @@
...
+  > EOF
+  $ cat > final/final.ml <<EOF
+  > let x = Both_a1_b1.make 1
+  > let y = Both_a2_b1.make 2
+  > let combos : Both_a1_b1.combined * Both_a2_b1.combined = (x, y)
+  > EOF
+
+Everything builds and odoc generates documentation for all the libraries:
+
+  $ dune build @doc-private 2>&1
```
art-w, line 221: Using `@doc-private` because otherwise dune skips the documentation of libraries with an `implements` (fixed in the next release of dune by https://github.com/ocaml/dune/pull/16200 )

art-w:
Rebased for OxCaml minus39

test/integration/parameterized.t/run.t
Leonidas-from-XIV: I think this is a good hint, maybe worth writing that rationale in the test.

Also, come to think of it, it might be worth adding a constraint on `dune` versions in the OPAM file in the case of `with-test` (and maybe if we can filter on the oxcaml compiler, so we don't bump the minimal dune version for tests on OCaml)

test/integration/parameterized.t/run.t
```diff
@@ -0,0 +1,516 @@
...
+
+Render to markdown for inspection:
+
+  $ for f in $(find _build/default/_doc/_odocls -name '*.odocl' | sort); do
```
Leonidas-from-XIV, line 225: I wonder if there's a way to get this without having to inspect the internals of `_build`? Just to make sure it's not unnecessarily fragile.

test/integration/parameterized.t/run.t
```diff
@@ -0,0 +1,516 @@
...
+  >   odoc markdown-generate "$f" -o markdown 2>&1
+  > done
+
+The library names below are suffixed with an opaque hash by dune; we normalise
+it away so the output is stable.
+
+  $ md() { cat markdown/$1@*/$2.md | sed 's/@[0-9a-f]*/@HASH/g'; }
```
Leonidas-from-XIV, line 232: Useless use of `cat` potentially. But I see the improved readability aspect of it.

test/integration/parameterized.t/run.t
```diff
@@ -0,0 +1,516 @@
...
+the sidebar marks it as the current unit and keeps its children, exactly like a
+plain module:
+
+  $ odoc compile-index --root _build/default/_doc/_odocls/
+  $ odoc sidebar-generate index.odoc-index
+  $ for lib in a_param a1; do
+  >   odoc html-generate --indent --sidebar sidebar.odoc-sidebar -o html \
+  >     $(find _build/default/_doc/_odocls -iname "$lib.odocl")
+  > done
+  $ grep -rl 'current_unit' html/ | sed 's/@[0-9a-f]*/@HASH/g' | sort
```
Leonidas-from-XIV, line 510: OpenBSD `grep` discourages `-r` and encourages usage of `-R`, however that has slightly different semantics in GNU `grep`. Potentially worth switching to `-R`, however its not super important.

src/xref2/link.ml
```diff
@@ -228,9 +270,39 @@ and module_path : Env.t -> Paths.Path.Module.t -> Paths.Path.Module.t =
...
+  | `ApplyParam (inst, param, arg) ->
+      `ApplyParam
+        ( resolve_module_path_parts env inst,
+          resolve_module_path_parts env param,
+          resolve_module_path_parts env arg )
+  | _ -> module_path env m
+
+and instance_prefix env (m : Paths.Path.Module.t) : Paths.Path.Module.t option =
+  let unfolded = unfold_alias env m in
+  if unfolded = m || not (is_instance unfolded) then None
```
Leonidas-from-XIV, line 296: Is `=` safe to use on `Path.Path.Module.t` values?

Leonidas-from-XIV, line 296: Also, the logic is not apparent to me. If the path is not aliased, then there is not instance prefix (`None`)? Maybe worth adding a comment on why that would be the case.

src/xref2/errors.ml
```diff
@@ -245,6 +250,10 @@ let rec kind_of_module_cpath = function
       match kind_of_module_cpath a with
       | Some _ as a -> a
       | None -> kind_of_module_cpath b)
+  | `ApplyParam (a, _param, b) -> (
+      match kind_of_module_cpath a with
+      | Some _ as a -> a
+      | None -> kind_of_module_cpath b)
```
Leonidas-from-XIV, line 256: Maybe worth joining with the case above to avoid the same code diverging in the future.

src/loader/odoc_loader.ml
```diff
@@ -42,6 +42,45 @@ exception Not_an_interface
...
+        parameterisation_of_cmi ~cmi_kind:cmi.Cmi_format.cmi_kind
+          ~cmi_params:cmi.Cmi_format.cmi_params
+    | None -> no_parameterisation
+  in
+  match cmt with
+  | Some cmt -> (parameterisation, cmt)
+  | None -> raise (Cmt_format.Error (Cmt_format.Not_a_typedtree filename))
+#else
+let read_cmt_and_parameterisation filename =
+  (no_parameterisation, Cmt_format.read_cmt filename)
```
Leonidas-from-XIV, line 81: Small nitpick but the order of items in the name of the function and in the tuple are swapped.

src/loader/odoc_loader.ml
```diff
@@ -42,6 +42,45 @@ exception Not_an_interface
...
+  let parameterisation =
+    match cmi with
+    | Some cmi ->
+        parameterisation_of_cmi ~cmi_kind:cmi.Cmi_format.cmi_kind
+          ~cmi_params:cmi.Cmi_format.cmi_params
+    | None -> no_parameterisation
+  in
+  match cmt with
+  | Some cmt -> (parameterisation, cmt)
+  | None -> raise (Cmt_format.Error (Cmt_format.Not_a_typedtree filename))
```
Leonidas-from-XIV, line 78: Maybe have this at the top, since no point of figuring out a `parameterisation` if `cmt` is `None` from the get-go.

Also, why is the error `Not_a_typedtree`?

src/loader/ident_env.ml
```diff
@@ -709,8 +709,31 @@ let is_shadowed
     List.mem id env.shadowed
 module Path = struct
 
+let module_of_id id = `Root (ModuleName.of_ident id)
```
Leonidas-from-XIV, line 712: Maybe it would be more readable to have this in the `#else` block, to avoid shadowing it in the `OXCAML` case.

src/loader/ident_env.ml
```diff
@@ -709,8 +709,31 @@ let is_shadowed
...
+
+#if defined OXCAML
+  let rec read_global_name (n : Global_module.Name.t) : Paths.Path.Module.t =
+    (* OxCaml parameterized library application "Lib[Param:Impl][P2:I2]" *)
+    let base = `Root (ModuleName.make_std n.head) in
```
Leonidas-from-XIV, line 717: That would be `Lib` in this example, correct?

src/document/generator.ml
```diff
@@ -84,23 +84,51 @@ let prepare_preamble comment items =
   in
   (Comment.standalone preamble, Comment.standalone first_comment @ items)
 
-let make_expansion_page ~source_anchor url comments items =
+let make_expansion_page ?(library_parameter = false) ~source_anchor url comments
```
Leonidas-from-XIV, line 87: I wonder whether it wouldn't be better to make this a mandatory named param, to avoid accidentally setting this to `false` by leaving it out. But maybe there's really just one place where it should be set to true.

Leonidas-from-XIV (commented):
Looks overall very nice, I only have a few nitpicks.

I am a bit thrown by the spelling of "parameterisation" since I would've used "parametrization". I think it makes sense to adopt the spelling that OxCaml uses, however the docs never seem to mention it so I don't know what their preferred spelling is.

src/document/types.ml
```diff
@@ -185,6 +185,7 @@ and Page : sig
     source_anchor : Url.t option;
         (** Url to the corresponding source code. Might be a whole source file
             or a sub part. *)
+    library_parameter : bool;
```
jonludlam: I'm not keen on this boolean appearing here - that doesn't seem right.

src/document/doctree.ml
```diff
@@ -349,9 +350,10 @@ end = struct
     | Some p -> Printf.sprintf "%s.%s" p.name name
 
   let render_title ?source_anchor (p : Page.t) =
-    format_title ~source_anchor p.url.kind
-      (make_name_from_path p.url)
-      p.preamble
+    let name = make_name_from_path p.url in
```
jonludlam: I think we should just have a new 'kind' rather than this new mechanism purely for parameters.

jonludlam:
Thanks @art-w ! sorry it's taken me some time to get to this. I think the loader parts look good, and the rendering looks mostly fine (modulo the boolean tacked onto the Page.t).

I'm more hesitant about the resolution parts. I was expecting there to be more changes in tools.ml where we resolve paths, but you're essentially throwing an error there and trying to handle it in link.ml instead. Aside from being in a surprisingly different place, it also means that we don't resolve subpaths, like `Lib[Param:Impl].Inner` as when we reresolve in link.ml they don't match `` `ApplyParam ``. I'd like to see how it looks to handle everything in tools.ml - I've had a very rough attempt [here](https://github.com/jonludlam/odoc/commit/16417e310b165f2c3ce9299db7a05f026a12bbbe) though I'd suggest using that as a suggested direction rather than something to cherry pick.

For the boolean in Page.t, I think it might be better if we get a new kind variant in document/url.ml and then we can just amend `make_name_from_path` rather than prefixing.

test/integration/parameterized.t/run.t
art-w: Removed the internal dune path since `_build` is enough

test/integration/parameterized.t/run.t
art-w: Removed cat, thanks :)

test/integration/parameterized.t/run.t
art-w: I wasn't aware of that, thanks! Switched to `-R`, it seems fine :)

src/xref2/link.ml
art-w: This got removed by Jon's resolution :)

art-w: (it's been removed)

src/xref2/errors.ml
art-w: Done!

src/loader/odoc_loader.ml
art-w: Ha thanks, fixed :)

src/loader/odoc_loader.ml
art-w: Refactored a bit the code... otherwise the `Not_a_typedtree` error predates my changes so I just kept it as is

src/loader/ident_env.ml
art-w: Sure, done

src/loader/ident_env.ml
art-w: Yes :)

src/document/generator.ml
art-w: That boolean and parameter are now gone :)

art-w:
Thanks everyone for the reviews :) The PR has been rebased for minus39 and I believe most issues have been addressed. I cherry-picked your resolution commit @jonludlam but reverted the resolution of parameters to preserve the names of the interfaces (which would otherwise be replaced by their implementation, an information that is already displayed).

PR #1467, opened 2026-07-23
OxCaml: support for kind abbreviations

### LINKED ISSUE #1443: Support OxCaml `kind_` definition
Kind definition enables declaring an abbreviation for complex kinds:

https://github.com/ocaml/odoc/blob/af26831b2a3f28ae2eccdfd145b4f04131b6fd3d/test/generators/cases/oxcaml.mli#L117-L121

Currently the use of abbreviations is rendered like any builtin kinds (by https://github.com/ocaml/odoc/pull/1410 ), but this isn't enough for users to understand what the abbreviation stands for:
- The `kind_` definition is missing in the rendered documentation
- Links from an abbreviation to its definition are missing too

art-w:
This PR adds support for rendering and linking to kind abbreviations (this introduces a new namespace, similar to types, so a lot of mechanical changes are required!).

As we are currently testing with OxCaml-minus31, there's a small inconvenience that one has to write an explicit `[@@ocaml.doc "..."]` to document `kind_` abbreviation as the parser was missing an `add_docs_attrs` ( https://github.com/oxcaml/oxcaml/blob/5.2.0minus-31/parsing/parser.mly#L4123-L4134 ). The oxcaml parser has since been [fixed](https://github.com/oxcaml/oxcaml/blob/main/parsing/parser.mly#L4197-L4209) so we'll be able to clean up the test in the near future :)

Fix https://github.com/ocaml/odoc/issues/1443

### DIFF
--- a/src/loader/cmi.ml
+++ b/src/loader/cmi.ml
@@ -495,36 +495,58 @@ let read_parsetree_core_type (ct : Parsetree.core_type) =
   | Ptyp_any _ -> Any
   | _ -> failwith "invalid core type"
 
-let rec read_jkind_annotation (jk : Parsetree.jkind_annotation) =
+let rec kind_signature_reference : Longident.t -> Paths.Reference.Signature.t =
+  function
+  | Longident.Lident m -> `Root (m, `TUnknown)
+  | Longident.Ldot (p, m) -> `Dot (kind_label_parent_reference p, m)
+  | Longident.Lapply _ -> `Root ("_", `TUnknown)
+
+and kind_label_parent_reference x =
+  (kind_signature_reference x :>  Paths.Reference.LabelParent.t)
+
+let read_kind_abbreviation env (lid : Longident.t) =
+  let name = Format.asprintf "%a" Pprintast.longident lid in
+  let reference : Paths.Reference.t option =
+    match lid with
+    | Longident.Lident n ->
+        Option.map
+          (fun id -> `Resolved (`Identifier (id :> Identifier.t)))
+          (Env.find_kind_abbreviation env n)
+    | Longident.Ldot (prefix, n) ->
+        Some (`KindAbbreviation (kind_signature_reference prefix, TypeName.make_std n))
+    | Longident.Lapply _ -> None
+  in
+  Kind.Abbreviation (name, reference)
+
+let rec read_jkind_annotation env (jk : Parsetree.jkind_annotation) =
   let open Kind in
   match jk.pjka_desc with
   | Pjk_default -> Default
-  | Pjk_abbreviation s -> Abbreviation (Env.Fragment.read_type s.txt)
+  | Pjk_abbreviation s -> read_kind_abbreviation env s.txt
   | Pjk_mod (jk', modes) ->
     let modes = List.map (fun (m : Parsetree.mode Location.loc) ->
       let (Parsetree.Mode s) = m.txt in s) modes in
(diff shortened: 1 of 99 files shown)


panglesd (changes requested):
Odoc references have this quite particular semantic that they can be "forward reference". So for instance:

```ocaml
type t
(** Abstract but can be created with {!v} *)

val v : unit -> t
```
The reference to `v` is made before `v` is defined.

As a result, in the following example:

```ocaml
kind_ to_be_shadowed = value mod portable

module X : sig
  type t : to_be_shadowed

  kind_ to_be_shadowed = value_or_null mod non_null global
end
```

When clicking on `to_be_shadowed` associated to the type `t`, odoc redirects to the second abbrev, `value_or_null mod non_null global`. I believe the first one, `value mod portable`, is the one oxcaml would use.

I'll think about what other path you could use. (But tell me if this example is wrong!)

Leonidas-from-XIV:
Just wanted to confirm that the `add_doc_attrs` is part of [5.2.0minus-39](https://github.com/oxcaml/oxcaml/blob/dd4e8507373d22fb295422eb6dd3d997c76c47cb/parsing/parser.mly#L4140-L4152), the release that we are currently targetting on `master` :)

src/model/paths.ml
```diff
@@ -70,6 +71,7 @@ module Identifier = struct
     | `Result x -> is_hidden (x :> t)
     | `ModuleType (_, name) -> ModuleTypeName.is_hidden name
     | `Type (_, name) -> TypeName.is_hidden name
+    | `KindAbbreviation (_, name) -> TypeName.is_hidden name
```
Leonidas-from-XIV, line 74: Can/should kind abbreviations be hidden?

src/loader/cmi.ml
```diff
@@ -1492,8 +1516,21 @@ and read_module_rec_status rec_status =
   | Trec_first -> Rec
   | Trec_next -> And
 
+#if defined OXCAML
+and read_kind_abbreviation_from_types env parent id (jkd : Types.jkind_declaration) =
+  let open KindAbbreviation in
+  let identifier = Env.find_kind_abbreviation_identifier env.ident_env id in
+  let source_loc = None in
```
Leonidas-from-XIV, line 1523: Not sure if this is ever set, so I'd suggest either removing it for now or determining the location.

src/loader/cmi.ml
```diff
@@ -581,70 +604,71 @@ let rec read_parsetree_core_type (ct : Parsetree.core_type) =
...
 
-let read_jkind_annotation = function
+let read_jkind_annotation env = function
   | None -> Kind.Default
-  | Some jk ->
-    match read_jkind_annotation jk with
-    | Abbreviation (`Dot (`Root, "value")) -> Default
-    | k -> k
+  | Some { Parsetree.pjka_desc = Pjk_abbreviation ({ txt = Longident.Lident "value"; _ }, _); _ } ->
+    Kind.Default
```
Leonidas-from-XIV, line 665: I wonder, is this correct? OxCaml defines the kind abbreviation `value` but in my testing in the topleve it seems like its possible to redefine it:

```ocaml
kind_ value = value mod portable
```

Leonidas-from-XIV (commented):
I think the approach with a specific namespaces for kinds makes a lot of sense.

Besides the few things I've mentioned in the code I think the `{!kind:xyz}` syntax should be documented somewhere.

PR #1469, opened 2026-07-30
OxCaml minus39 fixes.

jonludlam:
Note that we'll need to do some work to handle module types with `with kind` expressions. For now, I've changed it to return an option and switched to `filter_map`. The other changes are mostly mechanical.


test/xref2/lib/common.cppo.ml
```diff
@@ -620,11 +620,17 @@ let mkresolver () =
...
+   List.filter (fun s -> s <> "") (visible @ paths.hidden))
+#elif OCAML_VERSION >= (5,2,0)
   (let paths = Load_path.get_paths () in
    List.filter (fun s -> s <> "") (paths.visible @ paths.hidden))
-#else
+#elif OCAML_VERSION >= (4,8,0)
     (Load_path.get_paths () |> List.filter (fun s -> s <> ""))
+#else
+    !Config.load_path
 #endif
```
art-w, line 630: This is starting to get hard to read ^^' Maybe consider extracting?

art-w (approved):
Thanks! I'll rebase my other PRs on top once this is merged :)

odoc.opam
```diff
@@ -58,7 +58,7 @@ depends: [
 
 conflicts: [
   "ocaml-option-bytecode-only"
-  "oxcaml-compiler" {< "5.2.0minus31"}
+  "oxcaml-compiler" {< "5.2.0minus39"}
```
panglesd, line 61: Shouldn't it be 
```suggestion
  "oxcaml-compiler" {!= "5.2.0minus39"}
```
?
I just tried to install odoc on oxcaml minus39, and instead of failing as not compatible, it tried (and failed without this PR).

panglesd (approved):
Let's merge so that the many oxcaml PRs can be rebased!

odoc.opam
jonludlam: I think the plan was to be 'optimistic' - what we should do is modify the opam file for odoc.3.2.1 in opam-repository to have an upper bound, but to leave this as is, just in case by some miracle minus47 comes out next week and it doesn't require changes :-)

test/xref2/lib/common.cppo.ml
jonludlam: Good call - I've tidied it a bit

PR #1472, opened 2026-08-05
OCaml 5.6 support

jonludlam:


Leonidas-from-XIV:
Rebased it to get the CI fix from #1490 in.

jonludlam:
thanks!

Leonidas-from-XIV:
We should probably add a runner for OCaml 5.6 in the CI. No idea whether `setup-ocaml` supports that yet (maybe worth trying out).

Leonidas-from-XIV:
Ok so turns out you can get a preview of OCaml 5.6 in opam-repository if you ask for `ocaml-compiler.5.6`.

I've also updated the expected output of Markdown to remove the seemingly left-over backslash. It has shown me that the OCaml 5.6+ tests do run successfully on CI, so there's that upside :)

Leonidas-from-XIV:
The CI in 5.6 reveals a missing case in the lexer.

`ocamllex` warnings are new in OCaml 5.6: I've filed an issue in Dune to https://github.com/ocaml/dune/issues/16429 to improve the handling but on the odoc side it should be easy to add the empty-word case. Possibly as a separate PR even as it has nothing to do with OCaml 5.6

PR #1474, opened 2026-08-07
fix: odoc driver linking to stdlib in `$prefix/lib64`

### LINKED ISSUE #1477: odoc driver cannot link to stdlib in `$prefix/lib64`
It seems that in some cases, opam will install the stdlib into `$prefix/lib64` rather than `$prefix/lib/ocaml`. This causes problems for odoc_driver's detection of the ocaml-compiler package and its nested stdlib packages.

It's not clear to me when this happens or whether it depends on the ocaml version or OS. I have opam 2.5.2, installed by distro package manager, on openSUSE Tumbleweed. I can observe it for Ocaml 5.3, 5.4, and 5.1.

This can be checked by:
```console
$ opam switch create NAME --packages=ocaml-base-compiler.5.3.0 
$ grep /stdlib.cmi ~/.opam/NAME/.opam-switch/install/ocaml-compiler.changes 
  "lib64/stdlib.cmi" {"F:S13047T1786672613.32"}
```

Then, we can try to odoc a package which links to stdlib and observe:
```console
$ opam install odoc-driver fmt
$ odoc_driver fmt             
/tmp/odoc-6b5ba9/fmt/fmt/fmt.odoc
---------------------------------
File "/tmp/odoc-6b5ba9/fmt/fmt/fmt.odoc":
Warning: Couldn't find the following modules:
  Stdlib
```

This is the ocamlfind config:
```console
$ ocamlfind printconf
Effective configuration:
Configuration file:
    /home/rina/.opam/NAME/lib/findlib.conf
Search path:
    /home/rina/.opam/NAME/lib64
    /home/rina/.opam/NAME/lib
Packages will be installed in/removed from:
    /home/rina/.opam/NAME/lib
META files will be installed in/removed from:
    the corresponding package directories
The standard library is assumed to reside in:
    /home/rina/.opam/NAME/lib64
The ld.conf file can be found here:
    /home/rina/.opam/NAME/lib64/ld.conf
```

katrinafyi:
This does two related changes to fix https://github.com/ocaml/odoc/issues/1477:

- Use Findlib's own logic for inferring stdlib directory
- Adjust odoc_driver's classify_libs logic to also accept `lib64/` and also accept files appearing outside of a package subdirectory, as the ocaml stdlib files do.

I can test that it works with `dune exec odoc_driver -- fmt`.

--------

Removing the explicitly provided arguments to findlib  will be more reliable than hardcoding a subpath of the opam prefix.

As of about 2 years ago, ocamlfind is able to automatically infer the config path from opam environment variables: https://github.com/ocaml/ocamlfind/blob/be2335f14cc45b5551dea4278affdfb2718eb72e/src/findlib/findlib_config.mlp#L71-L77

Also, hardcoding the `env_camllib` can be risky because it's not portable or overridable. On openSUSE Tumbleweed, my stdlib was actually in `{prefix}/lib64` rather than the `lib/ocaml` which was hardcoded.

Findlib also has logic to let the user override the paths using environment variables, if needed: https://github.com/ocaml/ocamlfind/blob/be2335f14cc45b5551dea4278affdfb2718eb72e/src/findlib/findlib.ml#L281-L289

### DIFF
--- a/src/driver/opam.ml
+++ b/src/driver/opam.ml
@@ -196,18 +196,19 @@ let classify_libs prefix only_package contents =
     match only_package with None -> true | Some p -> p = pkg
   in
 
-  let libs =
-    List.fold_left
-      (fun set fpath ->
-        match Fpath.segs fpath with
-        | "lib" :: "stublibs" :: _ -> set
-        | "lib" :: pkg :: _ :: _
-          when Fpath.has_ext ".cmi" fpath && pkg_match pkg ->
-            Fpath.Set.add Fpath.(prefix // fpath |> split_base |> fst) set
-        | _ -> set)
-      Fpath.Set.empty contents
-  in
-  libs
+  contents
+  |> List.filter (Fpath.has_ext ".cmi")
+  |> List.filter_map (fun fpath ->
+         match Fpath.segs fpath with
+         | ("lib" | "lib64") :: "stublibs" :: _ -> None
+         (* e.g., [lib/cohttp/**/*.cmi] *)
+         | ("lib" | "lib64") :: pkg :: _ :: _ when pkg_match pkg ->
+             Some Fpath.(prefix // fpath |> split_base |> fst)
+         (* e.g., [lib64/stdlib.cmi] *)
+         | [ ("lib" | "lib64"); _ ] ->
+             Some Fpath.(prefix // fpath |> split_base |> fst)
+         | _ -> None)
+  |> Fpath.Set.of_list
 
 let find_odoc_config prefix only_package contents =
   let pkg_match pkg =
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 - Support for OxCaml zero alloc definitions (@Leonidas-from-XIV, #1422, #1444)
 - Remove requirement for ppx_expect in tests (@jonludlam, #1445)
 - Support for OxCaml modalities (@art-w, #1420)
+- Use Findlib to detect library directories in odoc_driver, avoiding hard-coding (@katrinafyi, #1474)
 
 # 3.2.1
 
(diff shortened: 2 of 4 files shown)


jonludlam:
Thanks! I just put 'no changelog required' on this PR, but thinking about it, it's fixing a real bug that you found and that others might too - in which case it'd be useful to create an issue then have a changelog referencing it. Could you do that please?

Leonidas-from-XIV:
Looking at the [commit that added it](https://github.com/ocaml/ocamlfind/commit/9e07b8d07132aafd09f4a938f40bcc29115cd208) the code is only available in ocamlfind >= 1.9.8, so I think it requires some updates to the lower bounds in odoc.

katrinafyi:
Thanks! Added and updated the changelog, let me know if that looks okay. 

I do also still intend to file the issue, as Jon suggested. I just wanted to double-check some things first and I'm away from the relevant computer right now.

katrinafyi:
@jonludlam I've filed the issue and incorporated another small change which was needed to fully fix the issue :) Let me know how it looks.

src/driver/opam.ml
```diff
@@ -196,18 +196,19 @@ let classify_libs prefix only_package contents =
...
-      Fpath.Set.empty contents
-  in
-  libs
+  contents
+  |> List.filter (Fpath.has_ext ".cmi")
+  |> List.filter_map (fun fpath ->
+         match Fpath.segs fpath with
+         | ("lib" | "lib64") :: "stublibs" :: _ -> None
+         (* e.g., [lib/cohttp/**/*.cmi] *)
+         | ("lib" | "lib64") :: pkg :: _ :: _ when pkg_match pkg ->
```
jonludlam, line 205: This is pretty grim stuff that we should remove! (not in this PR though)

jonludlam:
I've merged this manually. Normally I'd just have fixed the conflict and pushed to your branch and merged through the PR, but I can't seem to do that today!

PR #1476, opened 2026-08-12
Use implementations from `Stdlib` if available

Leonidas-from-XIV:
In my `include functor` PR I've used `List.partition_map`, which made the CI fail on OCaml 4.11, so I looked for the place where Odoc (like every project) would have its own Stdlib extensions.

However the way it is currently implemented the Odoc implementations take preference over the OCaml stdlib. This is mostly fine, however it would be somewhat nicer if Odoc could profit from potential future improvements. In Dune we have arrived at a pattern which allows us to polyfill missing functions while also using the compilers implementation of them - without having to resort to conditional compilation.

This PR implements that pattern.

### DIFF
--- a/src/utils/odoc_list.ml
+++ b/src/utils/odoc_list.ml
@@ -6,7 +6,6 @@ let rec concat_map_sep ~sep ~f = function
       let tl = concat_map_sep ~sep ~f xs in
       hd @ (sep :: tl)
 
-(** @raise Failure if the list is empty. *)
 let rec last = function
   | [] -> failwith "Odoc_utils.List.last"
   | [ x ] -> x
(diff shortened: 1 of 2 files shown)


dbuenzli:
Note it's not against your PR, but I note that for myself I would likely not do that.

When additions are made to the stdlib, your own functions get automatically overriden without you noticing. I think I'd rather first check that we are on page that the semantics matches. Stuff like differences in iteration order in the implementation of functions can make it for the odd and hard to track bug that depends on which version of OCaml you installed.

art-w (approved):
No objection on my end, I trust that the stdlib will not change the semantics of those functions :) (but I also wouldn't mind if the `include List` came first like before, since that's less tricky and avoids any uncertainties as to which code will be running, e.g. for "helper functions not yet in the stdlib")

dbuenzli:
> No objection on my end, I trust that the stdlib will not change the semantics of those functions

That was not my point. My point is if you have your own implementation of functions that subsequently become part of the stdlib. There can be subtle differences and it's better for you check before you switch to them.

> (but I also wouldn't mind if the `include List` came first like before, since that's less tricky and avoids any uncertainties as to which code will be running, e.g. for "helper functions not yet in the stdlib")

Which basically implements what I said :–) but negates the objective of this PR.

art-w:
Alright sorry @Leonidas-from-XIV, I'm closing this PR because I botched the review and somehow thought another PR was blocked by missing stdlib functions and then I forgot what I was reviewing >_> ... which goes to say that I don't care at all about this refactoring (this pattern seems more useful for abstract types where the overlays can't be implemented as efficiently as the stdlib, but for lists it's simpler not to).
Thanks @dbuenzli for pointing that out!

PR #1478, opened 2026-08-14
Remove `Forward constructor

Leonidas-from-XIV:
I was trying to understand what `` `Forward`` does. To do so figure out where a value of `` `Forward`` is constructed (from scratch, not as part of a `map` operation) and see what context this is useful. But it seems that such a value is never constructed. To test my hypothesis I removed the `` `Forward`` constructor from the codebase and it seems to compile fine and the tests also pass.

Thus my investigation seems to point to the fact that `` `Forward`` is unnecessary and can be removed. I'm opening a draft PR and invite maintainers to take a look - maybe I am missing some context in which `` `Forward`` is used, but if anything then at least it points at the fact that this functionality is not covered by any tests.

### DIFF
--- a/src/xref2/cpath.ml
+++ b/src/xref2/cpath.ml
@@ -176,14 +176,6 @@ let is_class_type_substituted : class_type -> bool = function
   | `DotT (a, _) -> is_module_substituted a
   | `Class (a, _) | `ClassType (a, _) -> is_resolved_parent_substituted a
 
-let rec is_module_forward : module_ -> bool = function
-  | `Resolved _ -> false
-  | `Root _ -> false
-  | `Identifier _ -> false
-  | `Local _ -> false
-  | `Substituted p | `Dot (p, _) | `Apply (p, _) -> is_module_forward p
-  | `Module (_, _) -> false
-
 let rec is_module_hidden : module_ -> bool = function
   | `Resolved r -> is_resolved_module_hidden ~weak_canonical_test:false r
   | `Substituted p | `Dot (p, _) | `Apply (p, _) -> is_module_hidden p
--- a/src/xref2/cpath.ml
+++ b/src/xref2/cpath.ml
@@ -57,7 +57,6 @@ and Cpath : sig
     | `Local of Ident.module_ * bool
     | `Identifier of Identifier.Path.Module.t * bool
     | `Root of ModuleName.t
-    | `Forward of string
     | `Dot of module_ * ModuleName.t
     | `Module of Resolved.parent * ModuleName.t (* Like dot, but typed *)
     | `Apply of module_ * module_ ]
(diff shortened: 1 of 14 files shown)


src/xref2/cpath.ml
art-w: The whole `is_module_forward` function could be removed as it always returns `false` :)

art-w (changes requested):
I agree it looks like pure dead code!

Probably dead code too (?):
https://github.com/ocaml/odoc/blob/1c3102cabfffc2ca1bacebec5d95a77af88f28e1/src/xref2/tools.ml#L956-L958
https://github.com/ocaml/odoc/blob/1c3102cabfffc2ca1bacebec5d95a77af88f28e1/src/xref2/tools.mli#L33-L37

(`env.ml` also have some `Forward` which might be removable, and the `UnresolvedForwardPath` error seems guarded by `is_module_forward = const false`)

Leonidas-from-XIV:
Good points, I removed some more unreachable code.

I hoped I could remove the `UnresolvedForwardPath` error, but it is emitted by https://github.com/ocaml/odoc/blob/1c3102cabfffc2ca1bacebec5d95a77af88f28e1/src/xref2/tools.ml#L953-L954. I looked up `Forward` and through some layers it seems this can only happen if https://github.com/ocaml/odoc/blob/1c3102cabfffc2ca1bacebec5d95a77af88f28e1/src/odoc/resolver.ml#L328 evaluates to a a `Forward_reference` which is dependent on the function called with `important_digests = true`. This seems to be set when the odoc binary is launched with the `-r`/`--resolve-fwd-refs` option. Unfortunately this path seems to be untested, so I am a bit cautious removing things that might have a user-visible effect.

jonludlam:
> Good points, I removed some more unreachable code.
> 
> I hoped I could remove the `UnresolvedForwardPath` error, but it is emitted by
> 
> https://github.com/ocaml/odoc/blob/1c3102cabfffc2ca1bacebec5d95a77af88f28e1/src/xref2/tools.ml#L953-L954
> . I looked up `Forward` and through some layers it seems this can only happen if
> 
> https://github.com/ocaml/odoc/blob/1c3102cabfffc2ca1bacebec5d95a77af88f28e1/src/odoc/resolver.ml#L328
> evaluates to a a `Forward_reference` which is dependent on the function called with `important_digests = true`. This seems to be set when the odoc binary is launched with the `-r`/`--resolve-fwd-refs` option. Unfortunately this path seems to be untested, so I am a bit cautious removing things that might have a user-visible effect.

Yeah this bit is still necessary. It's used when the code is compiled with `--no-alias-deps` - there will be paths that refer to modules that haven't yet been compiled by odoc, and that's handled by the `` `UnresolvedForwardPath `` handler. I think this PR currently removes all the code we _should_ be removing, so I think I'll merge as is. Thanks @Leonidas-from-XIV !

PR #1479, opened 2026-08-27
Drop id wrapper

jonludlam:
This simplification is the removal of an optimisation that's measurably unnecessary now. The end result is something that runs just as fast as before, but with a large decrease in memory and disk usage. The reduction in disk space is around 30% overall. The memory usage decreases by a lot less - around 5%.

### DIFF
--- a/src/model/paths.ml
+++ b/src/model/paths.ml
@@ -20,17 +20,13 @@ module Ocaml_env = Env
 open Names
 
 module Identifier = struct
-  type 'a id = 'a Paths_types.id = { iv : 'a }
-
   module Id = Paths_types.Identifier
 
   type t = Id.any
 
-  type t_pv = Id.any_pv
-
   let rec name_aux : t -> string =
    fun x ->
-    match x.iv with
+    match x with
     | `Root (_, name) -> ModuleName.to_string name
     | `Page (_, name) -> PageName.to_string name
     | `LeafPage (_, name) -> PageName.to_string name
@@ -61,7 +57,7 @@ module Identifier = struct
 
   let rec is_hidden : t -> bool =
    fun x ->
-    match x.iv with
+    match x with
     | `Root (_, name) -> ModuleName.is_hidden name
     | `Page (_, _) -> false
     | `LeafPage (_, _) -> false
@@ -86,11 +82,11 @@ module Identifier = struct
     | `SourceLocationInternal _ | `AssetFile _ ->
         false
 
-  let name : [< t_pv ] id -> string = fun n -> name_aux (n :> t)
+  let name : [< t ] -> string = fun n -> name_aux (n :> t)
 
   let rec full_name_aux : t -> string list =
    fun x ->
-    match x.iv with
+    match x with
     | `Root (_, name) -> [ ModuleName.to_string name ]
     | `Page (None, name) -> [ PageName.to_string name ]
     | `Page (Some parent, name) ->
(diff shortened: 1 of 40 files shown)


src/model/paths.mli
```diff
@@ -379,6 +333,12 @@ module rec Path : sig
 
       val is_hidden : t -> weak_canonical_test:bool -> bool
 
+      val equal : t -> t -> bool
+
+      val hash : t -> int
+
+      module Hashtbl : Hashtbl.S with type key = t
+
```
panglesd: Why are these added?

src/model/paths.ml
```diff
@@ -140,182 +136,172 @@ module Identifier = struct
...
+     breadth-first walk that stops after [count] meaningful nodes (ints,
+     strings) or [limit] enqueued ones, and each identifier level costs several
+     of each, so a budget of 10 is exhausted a few levels down and never reaches
+     the root.
 
-  let compare x y = compare x.ikey y.ikey
+     256/256 is the most reach obtainable: the runtime clamps [limit] to its
+     fixed-size traversal queue (HASH_QUEUE_SIZE, runtime/hash.c), so larger
+     values are silently ignored and cannot help. *)
+  let hash x = Hashtbl.hash_param 256 256 x
```
panglesd, line 185: That is frightening!

src/model/paths.ml
```diff
@@ -325,135 +311,116 @@ module Identifier = struct
...
     let hash = hash
   end
 
   module NonSrc = struct
     type t = Paths_types.Identifier.non_src
-    type t_pv = Paths_types.Identifier.non_src_pv
 
-    let equal x y = x.ihash = y.ihash && x.ikey = y.ikey
+    (* [==] guard as in [equal] above: [( = )] alone never checks pointers. *)
+    let equal x y = x == y || x = y
```
panglesd: Why not use `equal` here, just like in other places?

src/model/paths.ml
```diff
@@ -325,135 +311,116 @@ module Identifier = struct
...
 
-    let hash x = x.ihash
+    let hash x = Hashtbl.hash_param 256 256 x
```
panglesd: Same! We only need one frightening place.

src/model/paths.ml
```diff
@@ -807,6 +728,18 @@ module Path = struct
 
       let is_hidden m =
         is_resolved_hidden (m : t :> Paths_types.Resolved_path.any)
+
+      (* Same policy as [Identifier.equal]/[Identifier.hash] - see the comment
+         there for why [Hashtbl.hash] is too shallow for these spines. *)
+      let equal x y = x == y || x = y
+      let hash x = Hashtbl.hash_param 256 256 x
```
panglesd: Why not alias `Identifier.*`?

src/xref2/lang_of.ml
```diff
@@ -52,6 +52,20 @@ let lookup_module map : Ident.module_ -> _ = function
 
 module Opt = Component.Opt
 
+(* Hash-cons converted resolved module paths so that structurally equal paths
+   become physically shared. Marshal preserves physical sharing, so this is what
+   keeps repetitive alias chains small both on disk and in memory. *)
```
panglesd, line 57: Nice!

panglesd (approved):
Apart from adding one frightening bit, getting rid of this annoying wrapping is very nice!

src/model/paths.ml
jonludlam: In what way frightening?

src/xref2/lang_of.ml
jonludlam: yeah, we used to have this, then it fell off! See the change in the test case :-)

jonludlam:
Thanks for pointing out the duplication. Fixed!

src/model/paths.ml
panglesd: Just because it touches on how OCaml represents values, which is very often just a black box. And, there is a hardcoded value from a _c_ file.

(But that was more a comment than something to address)

jonludlam: it's not as bad as it seems :-) The code is fairly readable - or at least, it's straightforward to understand what the two non-default values mean.

PR #1480, opened 2026-09-01
Memoize docstring parsing

jonludlam:
OxCaml's libraries tend to have a lot of repeated comments due to ppx_template. This PR memoizes the parsing and skips resolving if they don't need it. On `Container_intf` in `base`:
- ~155,000 doc-comment instances
- only 33 unique doc-comment texts (a 99.98% duplicate rate!)

Impact (on `odoc_driver core`):
- odoc user CPU: **−13.0%**; total CPU −5.2%.
- Peak single-process RSS **−16%**; aggregate peak −14%.
- **`.odoc` files −18%, `.odocl` files −23% on disk**: memoized doc ASTs are physically shared, and Marshal serialises shared structure once. This compounds downstream (less unmarshalling, less memory).

On stock OCaml, there's no cpu usage change, but we do save some space in the odoc and odocl files (~4%).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,9 @@
 - Support for OxCaml modes (@art-w, #1454)
 - Fix OxCaml with-bounds for arbitrary types (@art-w, #1466)
 
+### Performance
+- Memoize doc-comment parsing and skip doc rebuilding during link when nothing needs resolution (@jonludlam, #1480)
+
 ### Fixed
 - Fix optional arguments rendering as `?arg:???` in modules without an mli on
   OCaml 5.5 (@jonludlam, #1489)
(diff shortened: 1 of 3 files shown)


panglesd (commented):
I have mixed feelings about this...

It will be strange when someone adds one reference to a big text and suddenly the doc generation is longer!

I'm also a bit worried about the maintainability. Suppose at some point we add a warning on "malformed code blocks" (for some definition of malformed). Then, it is very easy to miss that codeblocks need to be added to `*_needs_own_location`!

jonludlam:
I think that on "normal" OCaml code this won't do anything at all. The real use-case of this is for heavily templated code that ends up putting the same (identical) doc comments on many items. AIUI when there's more polymorphism in oxcaml we can safely GC this code.

PR #1481, opened 2026-09-02
More careful handling of strengthened signatures from includes

jonludlam:
Odoc was a little inconsistent with what happened with signatures that contain an `include`, sometimes incorrectly dropping type equalities (see the test in the first commit). This PR fixes that, making the handling consistent, which then enables an optimisation where we can drop the body of `include sig ... end`, which turns out to be a huge win on ppx_template-heavy oxcaml code.


src/xref2/tools.ml
```diff
@@ -1860,7 +1860,13 @@ and fragmap :
...
                 Ok
                   (Component.Signature.Include
-                     { i with decl; expansion_; strengthened = None })
+                     {
+                       i with
+                       decl;
+                       expansion_;
+                       expanded = i.expanded;
+                       strengthened = None;
+                     })
```
panglesd: I think `expanded = i.expanded` is redundant with `with`:

```suggestion
                     {
                       i with
                       decl;
                       expansion_;
                       strengthened = None;
                     })
```

src/model/lang.ml
```diff
@@ -213,6 +213,7 @@ and Include : sig
     status : [ `Inline | `Closed | `Open | `Default ];
     decl : decl;
     expansion : expansion;
+    expanded : bool;
```
panglesd: It is slightly slightly strange to have an expansion, that can be not expanded... Perhaps a better naming could be found, or at least a docstring explaining in what sense an expansion is or is not expanded!

src/xref2/compile.ml
```diff
@@ -450,17 +450,26 @@ and include_ : Env.t -> Include.t -> Include.t * Env.t =
...
+     from the decl in dependent modules, so an inline Signature decl is
+     redundant; strip it. *)
+  let decl =
+    match decl with
+    | Include.ModuleType (Signature _) ->
+        Include.ModuleType
+          (Signature { items = []; compiled = true; removed = []; doc = i.doc })
+    | _ -> decl
+  in
+  ({ i with decl; expansion; expanded = true }, env')
```
panglesd: Is the synthetic decl to save space?

It would probably be more readable to have the `decl` field have type `Unexpanded of decl | Expanded` to enforce that this synthetic decl is never used? (Or it seems, you still keep the `doc` field, is that on purpose?

src/xref2/compile.ml
```diff
@@ -450,17 +450,26 @@ and include_ : Env.t -> Include.t -> Include.t * Env.t =
...
     {
       expansion with
       content = { expansion.content with items; compiled = true };
     }
   in
-  ({ i with decl = include_decl env i.parent i.decl; expansion }, env')
+  let decl = include_decl env i.parent i.decl in
+  (* expanded=true: the expansion is authoritative and is not re-derived
+     from the decl in dependent modules, so an inline Signature decl is
+     redundant; strip it. *)
```
panglesd:
```suggestion
     redundant; strip it to save space on Marshal. *)
```

src/xref2/compile.ml
```diff
@@ -450,17 +450,26 @@ and include_ : Env.t -> Include.t -> Include.t * Env.t =
...
+  let decl =
+    match decl with
+    | Include.ModuleType (Signature _) ->
+        Include.ModuleType
+          (Signature { items = []; compiled = true; removed = []; doc = i.doc })
```
panglesd: Why do we keep the `doc` field? It seems that the generator only use that decl to render `sig ... end`, sot the docs can be removed too.

src/xref2/compile.ml
```diff
@@ -450,17 +450,26 @@ and include_ : Env.t -> Include.t -> Include.t * Env.t =
...
+    | _ -> decl
```
panglesd: Reading `generator.ml`, it seems few more cases could be stripped, but only if it makes any improvement to OxCaml I guess. Can be done later...

test/xref2/stripped_include_reexpansion.t/run.t
```diff
@@ -0,0 +1,61 @@
...
+    val internal : int
+  end
+
+
+
+
+
+
+
+
```
panglesd: I don't think the `lib.mli` has those trailing lines

```suggestion
```

test/xref2/stripped_include_subst.t/run.t
```diff
@@ -0,0 +1,46 @@
...
+  module type S = sig
+    include sig
+      type t
+  
+      val x : t
+      val y : t -> t
+    end
+  end
+
+
```
panglesd:
```suggestion
```

test/xref2/stripped_include_subst.t/run.t
```diff
@@ -0,0 +1,46 @@
...
+
+  $ cat b.mli
+  (** Module B includes [A.S] with a type substitution. *)
+  
+  module M : A.S with type t := int
+
+
```
panglesd:
```suggestion
```

CHANGES.md
```diff
@@ -37,6 +37,8 @@
 
 ### Performance
 - Cache Shape_reduce functor instantiation per environment during source resolution (@jonludlam, #1487)
+- Strip expanded include decls from .odoc files, reducing link memory and file size,
+  particularly on ppx_template-heavy code in OxCaml (@jonludlam, #1481)
```
panglesd: I think this PR also has effects on correctness, changing rendering in the tests. So probably worth a change entry in the "Fix" section.

panglesd (commented):
I left a few nitpicks and a request for better name choice!

The code looks good and (if the suggested renaming is done) the `if i.expansion.origin = `Odoc ...` is easier to read and reason about than the `if i.expansion.content.compiled` check. The tests are quite complex and hard to read but they seem to be improvements! 😵

PR #1482, opened 2026-09-02
Fix extended open support

jonludlam:
2 related fixes for the "extended opens", with tests. Tests are committed failing then updated in the commit that fixes the behaviour.


CHANGES.md
```diff
@@ -12,6 +12,7 @@
 - Remove requirement for ppx_expect in tests (@jonludlam, #1445)
 - Fix resolving functor through `module type of` (@Leonidas-from-XIV, #1471)
 - Fix odoc_driver's detection of `stdlib` when it is in `$prefix/lib64`, requires ocamlfind >= 1.9.8 (@katrinafyi, #1477, #1474)
+- Fix resolution of items bound by extended opens (`open struct ... end`) (@jonludlam, #1482)
```
panglesd: This seems to be only the second fix.

panglesd (approved):
Two simple and easy to review fixes. Nice!

CHANGES.md
jonludlam: Fixed!

PR #1483, opened 2026-09-02
Fix canonical chain search

jonludlam:
This is an enabling PR - it allows people to fix the problems that happen when you write your own wrapper module that doesn't contain _all_ of the library's modules. You should be able to put @canonical tags in the appropriate places after this.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,8 @@
 - Fix odoc_driver's detection of `stdlib` when it is in `$prefix/lib64`, requires ocamlfind >= 1.9.8 (@katrinafyi, #1477, #1474)
 - Fix loss of items bound by extended opens (`open struct ... end`) following a top-of-file docstring (@jonludlam, #1482)
 - Fix resolution of paths through the bindings of an extended open (@jonludlam, #1482)
+- Don't let an unrelated `@canonical` tag halt the search for a self-canonical
+  module, which lost the expansion of the module it named (@jonludlam, #1483)
 
 ### Performance
 - Cache Shape_reduce functor instantiation per environment during source resolution (@jonludlam, #1487)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,8 +20,7 @@
 - Fix odoc_driver's detection of `stdlib` when it is in `$prefix/lib64`, requires ocamlfind >= 1.9.8 (@katrinafyi, #1477, #1474)
 - Fix loss of items bound by extended opens (`open struct ... end`) following a top-of-file docstring (@jonludlam, #1482)
 - Fix resolution of paths through the bindings of an extended open (@jonludlam, #1482)
-- Don't let an unrelated `@canonical` tag halt the search for a self-canonical
-  module, which lost the expansion of the module it named (@jonludlam, #1483)
+- Unresolved @canonical tags no longer abort the search for canonical paths (@jonludlam, #1483)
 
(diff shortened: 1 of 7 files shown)


test/xref2/canonical_missing_alias.t/run.t
```diff
@@ -0,0 +1,135 @@
...
+  $ ocamlc -c -bin-annot -no-alias-deps main__Zone.mli
+  $ ocamlc -c -bin-annot -no-alias-deps -open Main__ main__Foo.mli
+  $ ocamlc -c -bin-annot -no-alias-deps -open Main__ main.ml
+
+  $ odoc compile -I . main__Zone.cmti
+  $ odoc compile -I . main__.cmt
+  $ odoc compile -I . main__Foo.cmti
+  $ odoc compile -I . main.cmt
+  $ odoc link -I . main.odoc
+  $ odoc html-generate --indent -o html main.odocl
```
Leonidas-from-XIV: Maybe worth declaring that as a shell function to show that this is doing the exact same steps as above.

test/xref2/canonical_missing_alias.t/run.t
```diff
@@ -0,0 +1,135 @@
...
+  html/Main/Foo/index.html
+  html/Main/Private/Zone_alias/index.html
+  html/Main/Private/index.html
+  html/Main/index.html
+
+But the reference in `Main.Foo` is unresolved: the canonical path `Main.Zone`
+can't be resolved, and odoc has no way of knowing that
+`Main.Private.Zone_alias` names the same module.
+
+  $ grep -A3 'val</span> f' html/Main/Foo/index.html
```
Leonidas-from-XIV, line 88: I wonder could we get the `xref-unresolved` information from another output format? Matching and outputting HTML here seems a bit fragile.

CHANGES.md
```diff
@@ -12,6 +12,8 @@
 - Remove requirement for ppx_expect in tests (@jonludlam, #1445)
 - Fix resolving functor through `module type of` (@Leonidas-from-XIV, #1471)
 - Fix odoc_driver's detection of `stdlib` when it is in `$prefix/lib64`, requires ocamlfind >= 1.9.8 (@katrinafyi, #1477, #1474)
+- Don't let an unrelated `@canonical` tag halt the search for a self-canonical
+  module, which lost the expansion of the module it named (@jonludlam, #1483)
```
Leonidas-from-XIV: I find this a bit hard to read and understand what it means. How about something like:

> Unresolved `@canonical` tags no longer abort the search for an existing canonical path

(Correct me if I understand the purpose of this PR wrong)

Leonidas-from-XIV (commented):
It's a surprisingly simple change with very nice test coverage documenting what the issue is/was.

CHANGES.md
jonludlam: Done, thanks!

test/xref2/canonical_missing_alias.t/run.t
jonludlam: Done!

test/xref2/canonical_missing_alias.t/run.t
jonludlam: Yeah, would be nice to do this better. I think I'd quite like to extend my PR #1448 and have more tests use this printing library. For a later PR though!

PR #1484, opened 2026-09-02
Fix broken test for hidden paths

jonludlam:
We were only checking if a type's _parent_ was hidden, not the type's name itself.


panglesd (commented):
I agree with the fix.

However, I find the test quite hard to get in my head, with 5 includes... Would there be a simpler test (even if not inspired from real world example) that demonstrate the bug?

So probably LGTM :)

jonludlam:
I've simplified the test a bit, but it still required a few modules to actually hit it.

PR #1486, opened 2026-09-03
Odoc compress payloads

jonludlam:
Note: this branch includes the changes in both #1479 and #1375

The last commit in this PR enables compression on odoc/odocl files if the compiler supports it. This results in a big drop in file sizes (around 4-5 times smaller).


panglesd (commented):
I think it would be nicer to introduce another compatibility module so that `odoc_file.ml` is not infected by cppo...

Otherwise, I'm fine with the current change (as long as no `%%VERSION%%` ends with `-zstd`)

PR #1487, opened 2026-09-03
Build the shape reducer once rather than many times

jonludlam:
We can share its caches as a result, which massively reduces allocation.


CHANGES.md
```diff
@@ -13,6 +13,9 @@
 - Fix resolving functor through `module type of` (@Leonidas-from-XIV, #1471)
 - Fix odoc_driver's detection of `stdlib` when it is in `$prefix/lib64`, requires ocamlfind >= 1.9.8 (@katrinafyi, #1477, #1474)
 
+### Performance
+- Cache Shape_reduce functor instantiation per environment during source resolution (@jonludlam, #XXXX)
```
panglesd:
```suggestion
- Cache Shape_reduce functor instantiation per environment during source resolution (@jonludlam, #1487)
```

panglesd (approved):
LGTM!

PR #1488, opened 2026-09-03
Odoc merge word runs

jonludlam:
Rather than storing comments as `` [`Word "rather"; `Space; `Word "than"; `Space; `Word "storing";...] `` each of which actually also carries a location, store them as just one: `` [`Word "rather than storing ..."] `` 

This is built on top of #1480 so we can be sure it doesn't clobber the memoizing.

### DIFF
--- a/src/document/comment.ml
+++ b/src/document/comment.ml
@@ -148,9 +148,26 @@ module Reference = struct
             [ inline @@ Inline.Link link ])
 end
 
+(* Merged word runs keep the source whitespace, which isn't significant
+   inline, so collapse each run to a single space for rendering. *)
+let collapse_whitespace s =
+  let b = Buffer.create (Stdlib.String.length s) in
+  let in_ws = ref false in
+  Stdlib.String.iter
+    (fun c ->
+      match c with
+      | ' ' | '\t' | '\n' | '\r' ->
+          if not !in_ws then Buffer.add_char b ' ';
+          in_ws := true
+      | c ->
+          Buffer.add_char b c;
+          in_ws := false)
+    s;
+  Buffer.contents b
+
 let leaf_inline_element : Comment.leaf_inline_element -> Inline.one = function
-  | `Space -> inline @@ Text " "
-  | `Word s -> inline @@ Text s
+  | `Space s -> inline @@ Text (collapse_whitespace s)
+  | `Word s -> inline @@ Text (collapse_whitespace s)
   | `Code_span s -> inline @@ Source (source_of_code s)
   | `Math_span s -> inline @@ Math s
   | `Raw_markup (target, s) -> inline @@ Raw_markup (target, s)
--- a/src/model/comment.ml
+++ b/src/model/comment.ml
@@ -15,7 +15,7 @@ type media = [ `Image | `Audio | `Video ]
 type raw_markup_target = string
 
 type leaf_inline_element =
-  [ `Space
+  [ `Space of string
   | `Word of string
   | `Code_span of string
   | `Math_span of string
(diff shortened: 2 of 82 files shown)


panglesd:
Does that affect ocamlformat @Julow?

jonludlam:
It won't affect ocamlformat - deliberately - as we're only changing the internal AST not the AST of odoc-parser. In any case, ocamlformat vendors odoc-parser :-)

src/document/comment.ml
```diff
@@ -148,9 +148,26 @@ module Reference = struct
             [ inline @@ Inline.Link link ])
 end
 
+(* Merged word runs keep the source whitespace, which isn't significant
+   inline, so collapse each run to a single space for rendering. *)
+let collapse_whitespace s =
+  let b = Buffer.create (Stdlib.String.length s) in
+  let in_ws = ref false in
+  Stdlib.String.iter
```
panglesd, line 156: `String` probably does not need `Stdlib`?

src/model/comment.ml
```diff
@@ -15,7 +15,7 @@ type media = [ `Image | `Audio | `Video ]
 type raw_markup_target = string
 
 type leaf_inline_element =
-  [ `Space
+  [ `Space of string
   | `Word of string
```
panglesd, line 19: Is there anywhere where `` `Space `` and `` `Word `` are handled differently? It seems to me they have the same meaning, and could be merged into `` `String ``.

test/generators/html/Markup.html
```diff
@@ -70,99 +70,105 @@ <h1>Module <code><span>Markup</span></code></h1>
...
-    <a href="#val-foo" title="foo"><code>code</code></a>. It's also possible
-     to surround a reference in a style: 
-    <b><a href="#val-foo"><code>foo</code></a></b>. References can't 
-    be nested inside references, and links and references can't be nested
-     inside each other.
+    <a href="#val-foo" title="foo"><code>code</code></a>
+    . It's also possible to surround a reference in a style: 
+    <b><a href="#val-foo"><code>foo</code></a></b>
+    . References can't be nested inside references, and links and references
+    can't be nested inside each other.
```
panglesd, line 171: There is an issue here, where whitespace is added before a `.`...

panglesd (changes requested):
> It won't affect ocamlformat - deliberately - as we're only changing the internal AST not the AST of odoc-parser. In any case, ocamlformat vendors odoc-parser :-)

Oh, right, thanks!

There is one blocking issue where space is added before a dot. I don't know exactly what caused this... (the rest looks good, although I'd like to have a second look when the issue is fixed!)

test/generators/html/Markup.html
jonludlam: I believe this is because we're asking tyxml to output the string `". It's also possible to surround a reference in a style:"` and it just breaks the line before writing it. It's only an issue in `--indent` mode, which is broken anyway, so I don't think this is a problem.

src/model/comment.ml
jonludlam: yeah I think you're probably right! I'll have a go.

test/generators/html/Markup.html
panglesd: Yes, `--indent` is (sadly) broken, and I now understand that this could also happen before but just less likely, eg with `"._It's_also_possible_to_surround_a_reference_in_a_style:"`.

Thanks for the explanation!

src/model/comment.ml
jonludlam: Oh my mistake - we don't merge the runs in the frontmatter tags, so they retain their `Space` tags. We could just have a guard checking a `` `Word`` for whitespace but I'm not sure it's worth it.

Julow:
Hi! Thanks for pinging about OCamlformat. In general, changes to odoc-parser don't affect OCamlformat, unless it becomes sensitive to the changes that OCamlformat makes. Don't hesitate to ping me again if you add new syntaxes or break it.

PR #1489, opened 2026-09-03
Fix optional argument rendering in OCaml 5.5

jonludlam:
In OCaml 5.5, optional arguments get wrapped in a `TPoly` constructor, confusing odoc. This results in rendering values such as:

```ocaml
val foo: ?arg:??? -> unit
```

This PR fixes this by unwrapping them from the TPoly.

### DIFF
--- a/test/xref2/dune
+++ b/test/xref2/dune
@@ -72,7 +72,7 @@
 ; 5.5.0 and above
 
 (cram
- (applies_to github_issue_1426)
+ (applies_to github_issue_1426 optional_arg_inferred)
  (enabled_if
   (>= %{ocaml_version} 5.5.0)))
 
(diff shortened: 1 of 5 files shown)


test/xref2/dune
```diff
@@ -72,7 +72,7 @@
 ; 5.5.0 and above
 
 (cram
- (applies_to github_issue_1426)
+ (applies_to github_issue_1426 optional_arg_inferred)
```
panglesd, line 75: Why only test on 5.5? We want the behavior to be the same on <5.5!

panglesd (approved):
I could not find in ocaml/ocaml the PR that changed the representation. I only found https://github.com/ocaml/ocaml/pull/14985 which is not merged (at the time of posting).

By curiosity, do you have a link easily? (otherwise, that's ok! the CI tests it already)

test/xref2/dune
jonludlam: heh true!

jonludlam:
I think it's this one: https://github.com/ocaml/ocaml/pull/13806

PR #1490, opened 2026-09-11
Update Sherlodoc tests for tyxml 5.0.0

jonludlam:
Let's fix that sea of red crosses.

### DIFF
--- a/sherlodoc.opam
+++ b/sherlodoc.opam
@@ -22,6 +22,8 @@ depends: [
   "menhir" {>= "20230608"}
   "ppx_blob" {>= "0.9.0"}
   "tyxml" {>= "4.6.0"}
+  # The cram tests depend on the precise version of tyxml
+  "tyxml" {with-test & >= "5.0.0"}
   "result" {>= "1.5"}
   "odig" {with-test}
   "alcotest" {with-test}
--- a/sherlodoc/test/cram/base_cli.t
+++ b/sherlodoc/test/cram/base_cli.t
@@ -11,6 +11,7 @@
   152 type Html_types.b_attrib = [ common ]
   152 type Html_types.i_attrib = [ common ]
   152 type Html_types.p_attrib = [ common ]
+  152 type Html_types.s_attrib = [ common ]
   152 type Html_types.u_attrib = [ common ]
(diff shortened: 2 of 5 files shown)


Leonidas-from-XIV:
I was just looking why my PR seems to be failing and rebasing on this PR will probably fix it.

However if it is due to tyxml 5, we probably need to restrict (at least the tests) to use tyxml 5, otherwise users with with older versions of Tyxml will have failing tests.

jonludlam:
Agreed!

jonludlam:
Huh, that dune check for the version of tyxml has broken `dune build @fmt` - @Leonidas-from-XIV any idea how to fix this?

Leonidas-from-XIV:
Hmm, I think the issue here is the combination of OCaml-CI's `lint-fmt` and Dune's rules, as it runs `dune build @fmt` without installing the dependencies, so Dune is (somewhat) right in pointing out that there's no `tyxml` package, thus it can't load the rule.

Dune probably shouldn't need to load cram rules for `@fmt` however.

I would suggest a different solution: run the tyxml cram tests always but add a second line

```
"tyxml" {>= "5.0" & :with-test}"
```

to the `sherlodoc.opam` file. That way a potential contributor is not surprised that the tests run successfully on their machine (with tyxml 4.6) but their PR fails on CI (with tyxml 5+).

Leonidas-from-XIV:
There's a similar issue in Dune already: https://github.com/ocaml/dune/issues/7454, however reading it it seems like this is a distinct issue.

jonludlam:
Good plan. I've added the constraint. Thanks!

Leonidas-from-XIV (approved):

dbuenzli:
Could I please ask you not to run these tests in the public CIs (that is the opam-repository). 

It's annoying for other people that add to the repository which have to check what these failures are about (e.g. [here](https://github.com/ocaml/opam-repository/pull/30683)) and they can potentially break on very new `tyxml` release.

jonludlam:
Hmm yes, that _is_ annoying that it's caused breakage on the opam-repository tests too. OK, let's fix our CI by merging this one once the tests are green, and figure out before the next release a better testing strategy for this.

